PR #793, opened 2021-02-25
Clicking timestamp can search for surrounding lines

darksheik:
One feature I have been deeply missing from our previous logging service is the ability to click the timestamp of a log line and have it show me the log lines immediately surrounding it.

I think I have the right idea here but didn't have a proper local setup to test it.  Might need some timezone adjustment in the `querystring` because I think the query expects local timezones.

### DIFF
--- a/lib/logflare_web/templates/log/log_event_body.html.leex
+++ b/lib/logflare_web/templates/log/log_event_body.html.leex
@@ -6,7 +6,8 @@
         <div><%= Timex.to_datetime(@timestamp) %></div>
         <div>
           <%= if @user_local_timezone do %>
-            <%= Timex.to_datetime(@timestamp, @user_local_timezone) %>
+            <%= link Timex.to_datetime(@timestamp, @user_local_timezone), to: Routes.live_path(@socket, LogflareWeb.Source.SearchLV, @source,
+            querystring: "t:>#{Timex.format!(Timex.shift(@timestamp, seconds: -30), "{ISO:Extended:Z}")} t:<#{Timex.format!(Timex.shift(@timestamp, seconds: 30), "{ISO:Extended:Z}")} c:count(*) c:group_by(t::second)", tailing?: false) %>
           <% else %>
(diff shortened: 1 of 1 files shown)


chasers:
https://github.com/Logflare/logflare/pull/795 should be on prod in about an hour. Play with it and let me know. The only issue I see is that we limit the number of log events listed so if more than I think 100 are returned within the timeframe, that might be confusing. Ideally, I'd like to enhance our LQL so you could pass an event ID with an `around` count and it would pull in N log events before and after said log event. Or maybe an `around` param which lets you pass in a duration in `ms` with the timestamp. This way at least if there is a lot returned you could easily limit it to a few seconds vs 30.

chasers:
Ah found a bug ... working on it.

darksheik:
Yes, that would be ideal.  30 seconds is a guess, usually if an event is happening, I really only need 10 seconds or so.  But agree that getting `n` events around the specific log is a much better solution.

chasers:
Bug is fixed ... and Logflare is stable after some super weird Google Cloud issues.

chasers:
@darksheik you can now step forwards and backwards. I would think this is useful for you along with this feature!

darksheik:
Great!  I'm need seeing anything in my front end yet - should I be?  Or is there a date you plan to roll it out?

chasers:
You should see it: https://twitter.com/logflare_logs/status/1366465124714774528

darksheik:
Ah, thanks!  I wasn't sure where to look for the timestamp link.  Found in one of your previous tweets.

chasers:
Yeah I put it where you put it in this pull request :)

darksheik:
I ended up being completely wrong about which template I was changing - I thought it was the index view. 😬

chasers:
All good ... we should have a quicker link in the logs list view but I thought it was a good spot to put it so we can play with it a bit before making it more prominent.

PR #971, opened 2021-08-12
Problem: Logflare.SQL crashing when reporting errors

yrashk:
It's an out of bounds exception

Solution: ensure we respect strings boundaries
when inserting content for error explanation.


yrashk:
@chasers looks like this one was never merged in?

chasers:
There you go.

PR #1036, opened 2021-12-03
Problem: it's hard to understand the schema of sandboxed endpoints

yrashk:
Solution: expose an API call for extracting sandboxed endpoint query schema

The columns defined in CTEs have to be explicitly named and if the type information
is important, it has to be casted using `CAST(value AS type) AS column_name` in
order to derive type information.


Ziinc:
Closing for now, as CTE sandboxing improvements are not being focused on atm. Also, will likely not extend kotlin codebase any more.

PR #1043, opened 2021-12-08
Problem: under memory pressure, caching endpoints can hog resources

yrashk:
This can lead to memory exhaustion on a given node and a subsequent restart,
which is not great.

Solution: start endpoint caches on other nodes in the cluster
when memory is higher than a set watermark (90%)

What is even more important, upon reaching this condition, the local node
will start pruning (or forcefully evicting) local endpoint caches that
haven't been used for over 5 minutes.

Note: this change also fixes a bug in endpoint caches where DynamicSupervisor
restarts a child that is normally terminated. They are now properly accounted
as transient children.

### DIFF
--- a/lib/logflare/endpoint/cache.ex
+++ b/lib/logflare/endpoint/cache.ex
@@ -11,6 +11,14 @@ defmodule Logflare.Endpoint.Cache do
     |> Enum.map(&:global.whereis_name/1)
   end
 
+  @high_watermark 0.9
+  @forced_eviction_after_inactivity_minutes 5
+
+  defp child_params(query, params) do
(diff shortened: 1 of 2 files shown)


Ziinc:
closing for now, will consider merging in the pruning of processes approach.

Not a fan of redistributing processes across nodes manually, can trigger a looping effect where processes get moved to another node, then that node gets near and exceeds the watermark and then triggers the processes moving again.

Ziinc:
moved to internal tracker [here](https://www.notion.so/supabase/Endpoints-v2-88abadc47a15494c890c3c32d05bc83a#fd073ca8fc4d49629a72e891a5fa1769)

PR #1046, opened 2021-12-15
Update README.md with dev setup instructions

Ziinc:
@chasers let me know if i missed anything


chasers:
@Ziinc looks good, but can you pull down staging make changes from there and merge into that?

PR #1105, opened 2022-05-09
chore: Add github actions unit testing

Ziinc:
Need to remove and mock 3rd party deps, for easier isolated unit testing.

- [x] add in postgres db service in gh action.

Reorganized the `application.ex:start` callback to make the children clearer for each env. Also remove `:goth` child in `test` children, and prevented app early startup when in test, so that we can avoid setting distribution settings. 

Removed coveralls, which is unnecessary, and added in aliases for various checks for code quality. They are to be enabled on GH actions once the existing issues for each are cleaned up.


chasers (approved):
Much cleaner application.ex!

Ziinc:
Trying to get GH actions to properly run the unit tests before i merge it in. Lots of trial and error...

PR #1116, opened 2022-05-26
feat: add in /api/endpoints/query route

Ziinc:
Adds in the `/api/endpoints/query/:token` route, with the `/endpoints/query/:token` route to be considered deprecated.


chasers (changes requested):
Can you get rid of the request_path pattern matching here?  We don't need it for this and should use the router pipelines to scope which routes this plug module runs.

chasers (approved):

PR #1117, opened 2022-05-27
chore: refactor Logflare.Contact context

Ziinc:
@chasers i notice that the contact form has been disabled completely in the html template. Since we aren't using the contact form as a support/sales channel, perhaps we can direct them to the supabase contact page instead, then I can rip out all the code relating to the contact form + emailing


chasers:
Sure ... keep the page up but you can do something like "Logflare is now a part of Supabase! Click here for Supabase support."

chasers (approved):

Ziinc:
Ripped out all logic, new copy is as follows

Screenshot 2022-05-27 at 2 24 46 PM

prod.secrets.exs does not need the following configs anymore:
```
config :logflare,
  recaptcha_site_key: ...
  recaptcha_secret:  ...
```

PR #1133, opened 2022-06-10
feat: tentative deploy scripts

Ziinc:
Deployment scripts for incremental deployment automation, to bypass gcloud ui and make things less error prone.


Ziinc:
Closing in favour of using pulumi directly

PR #1137, opened 2022-06-10
chore: endpoints controller tests

Ziinc:
Adds endpoint controller tests for better pre-release testing.

Initial attempts to mock `Logflare.SQL`  with Mimic proves too involved. Other possible solutions:
- Mimic and stubbing with a module, running `Logflare.SQL` server in application children. Stubs only public functions, and only need to stub external api, not genserver callbacks
- Mock server, init with `Logflare.SQL` in conn_case setup. Would be able to simulate communication with parser better. Will have to redefine callbacks, as well as keep the callbacks in sync.


chasers:
Oh I did not see this PR!  I'll take a look at what you've got here.  I started some tests in my last PR which put up a lot of Endpoints changes.

Ziinc:
I've rebased your changes in, this PR focuses on the `query` action. for ui-related we'll convert it to a liveview eventually so not as important

Ziinc:
@chasers I've been able to get the endpoints query tests to pass, the main file to look at is [this](https://github.com/Logflare/logflare/blob/4dbaa92f17f38cb70a76288ac9b109d1929027aa/test/logflare_web/controllers/endpoints_controller_test.exs).

I've also done a bit of module renaming: 
- `Logflare.Endpoint` -> `Logflare.Endpoints`
- `LogflareWeb.EndpointsController` -> `Logflare.EndpointsController` 

This is to cut down ambiguity when referring to the module, which clashes with the Phoenix.Endpoint module. It is mainly isolated to [this commit](https://github.com/Logflare/logflare/pull/1137/commits/908701ff5f08fefbc5eb323384936314edbf09e8), so we can revert it easily if needed.

chasers (approved):

PR #1143, opened 2022-06-21
Timeout endpoint queries

chasers:
Previously Endpoints would block when queries were running even when they were cached. 

Now It doesn't block except for the first response, and if queries are not cached at all.

Need some feedback with the tests I created here too, but I got one to work!


chasers:
@Ziinc I merged this because I want to get it out there with your stuff and don't want to wait until 24 hours or whatever it would be for our days to cycle.  But please review, with my tests ... need some help there!

Ziinc:
ok gotcha, i'll merge it with the endpoint tests that i'm also working on

PR #1153, opened 2022-06-29
feat: Multi Backends

Ziinc:
- [x] Webhooks backend
   - [x] adaptor module
   - [x] use tesla as a http client
   - [x] memory buffer
   - [x] broadway pipeline(?)
- [x] backend behaviour
- Migration for backends table
  - [x] create `source_backends` table
  - [x] store config as map/embedded schema
    - [x] v2 pipeline toggle column in `sources` table
- [x] adaptor-specific config validation 
  - [x] adaptor behaviour function
- [x] able to toggle source v2 pipeline in ui if admin
  - [x] able to create backend
  - [x] able to delete backend
- [x] `architecture.md` + diagram
- [x] per-source ingestion buffer + pipeline
- [x] recent logs cache
  - [x] naive implementatino
  - [x] global cache syncing
  - [x] only retain latest 100 items
- [x] api ingestion tests

Separate PRs:
- [ ] able to update backend config through source settings
- [ ] Search query executor for v2 pipeline
- bigquery adaptor
  - [ ] ingest callback
  - [ ] transformers
  - [ ] Migration to copy over BQ config
  - [ ] Config handling for big query (users table and backends table)
- pipeline telemetry metrics
- [ ] web: crud backend config
stretch:
- [ ] recent logs fetch latest n values on boot
- [ ] rate limiting logic for each source
  - [ ] global genserver, one per node, using deltacrdt to sync.
- [ ] consolidated Notifier server, one per source


Ziinc:
@chasers I've added some more implementation files as well as a behaviour. Thinking of the following:
- `source_backends` table, with `Logflare.Backends.SourceBackend` struct to represent the stored backends for each source.
- `Logflare.Backends.Adaptor` for defining any macros or behaviours that each adaptor needs to provide and implement. I have created a stub module for the webhook implementation. Each backend is likely to have its own supervision tree if needed.
- `Logflare.Backends` as the top level interface for interacting with the backends and sending log events/queries to the respective backends. Still fleshing out the interface.

Ziinc:
@chasers I've got the backend management stuff in, very rough around the edges but we can tidy it up later. I decided to cut the scope down further, so bigquery backend would come in a separate PR. With this PR in, we can start testing throwing logs at the source and seeing the webhook request throughput. Will also need to get in telemetry metrics for the pipeline for benchmarking. I haven't done the v2 handling for search page and recent logs page yet, so those are likely to error out.

https://user-images.githubusercontent.com/22714384/182143661-f5129c3f-c75f-4fcc-9a7c-5fbb3a878d72.mov

chasers (approved):

PR #1156, opened 2022-08-25
Change source table to replica identity full

chasers:


chasers:
@Ziinc also provides a mix alias for `start.orange` and `start.pink` for testing distributed stuff locally

PR #1165, opened 2022-09-16
fix: Adds handling for unrecognized oauth handling for OauthController

Ziinc:
This PR adds handling and logging for unrecognized Oauth providers, which should let us better track these random auth errors that keep popping up.

Reference ticket: [link](https://www.notion.so/supabase/bug-OauthController-undefined-function-858073f81a1c410ea9b5123dc2014956)


chasers (changes requested):
Can you address these logging metadata changes?

chasers (approved):

Ziinc:
rebased on staging.

PR #1167, opened 2022-09-21
chore: add initial test coverage for ingest transformers

Ziinc:
Adds tests, docs, and specs for ingest transformers and related utility functions.

A few areas that would need some clarifications @chasers :
- the `:field_length` transformer is set to max 128 characters for bigquery, however, the field length quoting transformer adds the `_` before the column name, resulting in the length of the field name always being `max + 1`, just want to check if this is expected behaviour, i'm not sure if the prefixed underscore is counted as part of field length for bigquery
  - The closest google documentation i can find that mentions table column name length limitation is [here](https://cloud.google.com/bigquery/quotas#table_functions), but it is for table functions.
- The `:strip_bq_prefixes` transformer does not actually strip the bigquery prefixes, but adds an additional underscore to the field. This results in a double underscore like `__TABLE_something`. Is this expected behaviour?

Pre-merge todo:
- [x] fix bug relating to ingest transformers `:strip_bq_prefixes`, [ref ticket](https://www.notion.so/supabase/bug-IngestTransformers-do_transform-strip_bq_prefixes-55ff87b4295445aeae108b43dabc550f)
  - Some users are sending non-map values on the metadata key, resulting in this error. To fix, i've done a passthrough for non-map data types. 

I've also added top-level key transformation, in preparation for allowing top-level log params.


chasers:
> The :strip_bq_prefixes transformer does not actually strip the bigquery prefixes, but adds an additional underscore to the field. This results in a double underscore like __TABLE_something. Is this expected behaviour?

I'm not even sure why we're striping the `TABLE` like prefixes. I've never seen those in the field names anywhere. Ah maybe in the schema responses?  I could see why we want to strip those there but I don't remember them being there either.

chasers:
`max + 1` is probably not expected behavior.  It probably should be `max` including the added underscore.

chasers:
> Some users are sending non-map values on the metadata key, resulting in this error. To fix, i've done a passthrough for non-map data types.

This either needs to be a map or a list of maps. Other types won't work at all at the moment I don't think.

lib/logflare/logs/ingest_transformers.ex
```diff
@@ -73,4 +80,7 @@ defmodule Logflare.Logs.IngestTransformers do
         key
     end)
   end
+
+  # passthrough for non-matching data types
+  defp do_key_transform(log_params, _rule), do: log_params
```
chasers, line 85: Yeah for now it's okay for this to blow up, or we should handle the error and respond appropriately to the client.  If this goes through it's just going to blow up down the line.

chasers (changes requested):
See comment

lib/logflare/logs/ingest_transformers.ex
Ziinc: But if we're moving towards allowing users to send in any value that they want for each top level key, it would not make sense to force the `metadata` key to be special anymore.
The only critical fields are `timestamp` and `event_message`, after all. All other fields can be treated any normal metadata field.

Ziinc:
For the different key prefix transformations (`_TABLE`, `_FILE`, and `_PARTITION`), I've added some logging so that we can monitor if these keys are getting sent in and transformed. I'll check again later on down the line and if they aren't used we can get rid of them.

max key length transformation logic has been updated to include the underscore `_`.

For the non-map value passthrough, see comment reply

Ziinc:
- [ ] Add in ticket to check back in few months on prefix key transformation

lib/logflare/logs/ingest_transformers.ex
chasers: Yeah I know I said I was okay with rewriting the BigQuery pipeline in place the last time we spoke but I think I was in an abnormally good mood or something.

I still very much think it should be rewritten as a "backend". 100% bad things will happen with an in place rewrite, even if done in small chunks.

I know the requirements and edge cases. Maybe it'll be approachable if we sit down and enumerate a v1 in more detail.

Ziinc: hmm ok, i understand where you come from. lets focus on the v2 backend then. Just a bit worried that we'll miss edge cases from the original v1 code and spend extra time hunting them down.

PR #1192, opened 2022-11-03
Add Cloudflare log push Plug parser with test

chasers:
Adds support for the [Cloudflare HTTP destination ](https://developers.cloudflare.com/logs/get-started/enable-destinations/http/)available to Cloudflare Enterprise accounts.

- `EdgeStartTimestamp` is the `timestamp`
- The `event_message` is the stringy JSON log line
- The `metadata` is the decoded JSON

For these parsers we can do a v2 which will put this stuff at the top level, and have a different Router pipeline for all the v2 endpoints and pasers.

API call to enable an HTTP destination:

```
curl -s https://api.cloudflare.com/client/v4/zones/ZONE_ID/logpush/jobs -X POST -d '{  "name": "supabase.co-https",  "logpull_options": "fields=ClientRequestProtocol,ClientSSLCipher,ClientRequestURI,WAFProfile,WAFRuleID,EdgePathingOp,ClientRequestReferer,FirewallMatchesRuleIDs,SecurityLevel,EdgeResponseBytes,ClientASN,EdgePathingStatus,EdgeResponseContentType,ParentRayID,WorkerStatus,EdgeRequestHost,EdgeServerIP,ClientDeviceType,WorkerCPUTime,EdgeStartTimestamp,ClientSrcPort,ClientXRequestedWith,OriginResponseBytes,OriginResponseHTTPLastModified,EdgePathingSrc,ClientRequestHost,ClientRequestPath,EdgeColoID,CacheResponseStatus,ClientRequestBytes,CacheCacheStatus,ZoneID,WorkerSubrequestCount,ClientIPClass,CacheResponseBytes,EdgeColoCode,WAFFlags,WAFAction,ClientRequestUserAgent,FirewallMatchesActions,Cookies,EdgeRateLimitID,OriginResponseHTTPExpires,ClientSSLProtocol,RayID,EdgeEndTimestamp,RequestHeaders,WAFMatchedVar,ClientIP,EdgeRateLimitAction,OriginResponseTime,ClientCountry,OriginIP,CacheTieredFill,WAFRuleMessage,ClientRequestMethod,WorkerSubrequest,FirewallMatchesSources,OriginSSLProtocol,OriginResponseStatus,EdgeResponseStatus,ResponseHeaders",  "destination_conf": "https://api.logflare.app/logs/cloudflare?source=SOURCE_ID&header_X-API-KEY=API_KEY",  "max_upload_bytes": 5000000,  "max_upload_records": 1000,  "dataset": "http_requests",  "enabled": true}' \
    -H "Authorization: Bearer BEARER" | jq
```


test/logflare_web/plugs/logpush_parser_test.exs
```diff
@@ -0,0 +1,217 @@
...
+          34, 235, 187, 213, 51, 146, 70, 124, 203, 240, 77, 131, 248, 174, 97, 109, 71, 255, 50,
+          125, 123, 70, 220, 136, 23, 120, 193, 19, 226, 246, 117, 8, 219, 11, 218, 245, 183, 194,
+          246, 8, 228, 247, 100, 206, 33, 246, 107, 101, 206, 117, 2, 242, 197, 100, 206, 124,
+          181, 204, 145, 192, 114, 255, 223, 58, 199, 177, 187, 7, 34, 103, 91, 158, 147, 38, 45,
+          35, 159, 22, 57, 226, 133, 158, 255, 47, 68, 206, 176, 108, 215, 32, 150, 105, 56, 175,
+          149, 184, 208, 218, 165, 218, 211, 74, 71, 108, 195, 65, 198, 187, 30, 6, 11, 191, 136,
+          236, 125, 24, 205, 30, 21, 189, 106, 215, 191, 94, 243, 182, 255, 51, 30, 215, 188, 154,
+          57, 175, 82, 189, 191, 0, 29, 159, 90, 32, 167, 13, 0, 0>>
+
+      data = %{
```
Ziinc, line 68: quite verbose to do a full equality check on the result. Only need to test that the parser is converting, don't need to check full payload.

Can have some match assertions on a few keys and assume the rest will be parsed correctly.


```elixir
assert {:ok, [ %{"batch" => [%{
  "message" => "{\" <> _,
  "metadata" => %{"ClientRequestProtocol"=> "HTTP/1.1"},
  "timestamp" => _
},_] }, %_{}
```

Ziinc: should be able to cut this file loc to <50, better for readability.

lib/logflare_web/controllers/plugs/cf_log_push_parser.ex
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def decode({:ok, <<>>, conn}) do
+    {:ok, %{}, conn}
+  end
+
+  def decode({:ok, body, conn}) do
+    body =
+      if @gzip_header in conn.req_headers do
+        body |> :zlib.gunzip() |> String.split("\n", trim: true)
```
Ziinc, line 32: i don't think the test case covers both gzip and non-gzip.

lib/logflare_web/controllers/plugs/cf_log_push_parser.ex
```diff
@@ -0,0 +1,71 @@
...
+        end
+      end
+
+    {:ok, %{"batch" => batch}, conn}
+  rescue
+    e ->
+      reraise Plug.Parsers.ParseError, [exception: e], __STACKTRACE__
+  end
+
+  def decode({:more, _, conn}) do
```
Ziinc, line 60: these other decode errors are not covered by test case as well

test/logflare_web/plugs/logpush_parser_test.exs
Ziinc: might actually be better to shift this test over to the `LogsControllerTest` to cover whole ingestion for logpush integration.

Maybe can change base to https://github.com/Logflare/logflare/pull/1184 and [extend the cloudflare ingestion tests](https://github.com/Logflare/logflare/pull/1184/files#diff-0acb00fed743ea98b799247a798c5e66ad98bb2d90ef3aa24c0edffcfdf456c4R80) ?

lib/logflare_web/controllers/plugs/cf_log_push_parser.ex
```diff
@@ -0,0 +1,71 @@
...
+        case Jason.decode(line) do
+          {:ok, log_event} ->
+            %{
+              "timestamp" => log_event["EdgeStartTimestamp"],
+              "message" => line,
+              "metadata" => log_event
+            }
+
+          {:error, error} ->
+            Logger.warn(inspect(error))
```
Ziinc, line 48: should use `Logger.error`?Would need more descriptive error here

e.g. 
```
Logger.error("[LOGPUSH] Decode error for input batch..." <> inspect(error))
```

Ziinc (changes requested):
Lets avoid copy-pasting the raw integration structs into the test cases directly and instead keep them in the fixtures utility functions.

Added a few more comments on the testing portion, implementation looks good.

Ziinc:
Was looking over implementation and how logpush works, perhaps we can use headers to insert top level keys? We can start off with merging headers with `X-LOGFLARE-MERGE` header directive and merge in the k-v list into the batch.

lib/logflare_web/controllers/plugs/cf_log_push_parser.ex
chasers: No but the only difference is the gunzip so if effectively tests both.  But I can add a test for it.

chasers:
> Was looking over implementation and how logpush works, perhaps we can use headers to insert top level keys? We can start off with merging headers with `X-LOGFLARE-MERGE-INTO` header directive and merge in the k-v list into the batch.

It would be nice to have some sort of generic way to manipulate payloads as they come in.

There is this [custom event message stuff](https://github.com/Logflare/logflare/blob/206437f82d7e28b0305d8d76a15f3908f3556b83/lib/logflare/logs/log_event.ex#L255) we do.  

Maybe it can be something similar. A key value form where we put keys, and the value could be from a JSON query via `Warpath`.

chasers:
You can have Cloudflare include custom headers in your logs as a separate field.  We could do this with the `host` header for now.

test/logflare_web/plugs/logpush_parser_test.exs
chasers: We could additionally do this, but there really is no logic for this in the controller so having the plug tested directly I think is a good idea regardless.

chasers:
Changed a couple things:
 - Parser is now NDJSON so if any api requests come in as NDJSON it'll get handled correctly
 - Added a function plug to swap the content-type and created a pipeline only for Cloudflare
 - Cleaned up the tests per your suggestions @Ziinc 

I think this makes the routing of this content type more correct. And if Cloudflare changes their content type to the correct one `application/x-ndjson` then it'll all still work and we can simply pull out that special pipeline, and it'll all still just work.

lib/logflare_web/controllers/plugs/ndjson_parser.ex
```diff
@@ -0,0 +1,71 @@
...
+            %{
+              "timestamp" => log_event["EdgeStartTimestamp"],
+              "message" => line,
+              "metadata" => log_event
+            }
+
+          {:error, error} ->
+            Logger.warn(inspect(error))
+
+            nil
```
Ziinc, line 50: should we filter out nils? will likely lead to ingestion error for the entire batch

Ziinc (approved):
👍 just one last thing i noticed for json decode error handling, if any.

test/logflare_web/plugs/logpush_parser_test.exs
Ziinc: it'll be fine to test plug directly

Ziinc:
@chasers could you fix up the compilation checks as well before merging
Screenshot 2022-11-15 at 5 15 26 PM

lib/logflare_web/controllers/plugs/ndjson_parser.ex
chasers: I saw that in one of the other parsers but was trying not to have to Enum over the list if I didn't need to.

I will do that though just in case.

PR #1196, opened 2022-11-10
Index Sources for Account

filipecabaco:
New API endpoints to index sources for logged-in accounts.

* GET `/api/sources`  - Lists all sources for logged in account

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -104,7 +104,7 @@ config :logflare, Logflare.Vercel.Client,
   install_vercel_uri: "https://vercel.com/integrations/logflare-v2-dev/new"
 
 config :goth,
-  json: "config/secrets/logflare-dev-238720-63d50e3c9cc8.json" |> File.read!()
+  json: "config/secrets/logflare-dev-238720-344624644919.json" |> File.read!()
 
 config :logflare, Logflare.Cluster.Utils, min_cluster_size: 1
 
--- a/lib/logflare/sources.ex
+++ b/lib/logflare/sources.ex
@@ -178,8 +178,7 @@ defmodule Logflare.Sources do
 
   def preload_defaults(source) do
     source
-    |> Repo.preload(:user)
-    |> Repo.preload(:rules)
+    |> Repo.preload([:user, :rules])
     |> refresh_source_metrics()
     |> maybe_compile_rule_regexes()
     |> put_bq_table_id()
@@ -198,7 +197,7 @@ defmodule Logflare.Sources do
 
     Repo.preload(
       source,
-      saved_searches: from(SavedSearch) |> where([s], s.saved_by_user)
+      saved_searches: from(s in SavedSearch, where: s.saved_by_user)
     )
   end
 
@@ -354,5 +353,15 @@ defmodule Logflare.Sources do
       nil -> :pseudo
       x -> x
     end
+  end
+
+  @spec preload_for_dashboard(list(Source.t())) :: list(Source.t())
+  def preload_for_dashboard(sources) do
+    sources
+    |> Enum.map(&preload_defaults/1)
+    |> Enum.map(&preload_saved_searches/1)
+    |> Enum.map(&put_schema_field_count/1)
+    |> Enum.sort_by(& &1.name, &<=/2)
+    |> Enum.sort_by(& &1.favorite, &>=/2)
   end
 end
--- a/lib/logflare_web/controllers/api/source_controller.ex
+++ b/lib/logflare_web/controllers/api/source_controller.ex
@@ -1,0 +1,11 @@
+defmodule LogflareWeb.Api.SourceController do
+  use LogflareWeb, :controller
+  alias Logflare.Sources
+  alias Logflare.Users
+
+  def index(%{assigns: %{user: user}} = conn, _) do
+    user = Users.preload_sources(user)
+    sources = Sources.preload_for_dashboard(user.sources)
+    json(conn, sources)
+  end
+end
--- a/test/logflare/source/billing_writer_test.exs
+++ b/test/logflare/source/billing_writer_test.exs
@@ -1,9 +1,11 @@
 defmodule Logflare.Source.BillingWriterTest do
   @moduledoc false
   use Logflare.DataCase
-  alias Logflare.Source.{BillingWriter, RecentLogsServer}
   alias Logflare.Billing.BillingCount
   alias Logflare.Repo
+  alias Logflare.Source.BillingWriter
+  alias Logflare.Source.RecentLogsServer
+
   setup :set_mimic_global
 
   setup do
@@ -13,14 +15,14 @@ defmodule Logflare.Source.BillingWriterTest do
     user = user |> Logflare.Repo.preload(:billing_account)
     plan = insert(:plan, type: "metered")
 
-    pid =
-      start_supervised!(
+    {:ok, pid} =
+      start_supervised(
         {BillingWriter,
          %RecentLogsServer{source_id: source.token, source: source, user: user, plan: plan}}
       )
 
     # increase log count
-    start_supervised!(Logflare.Sources.Counters)
+    start_supervised(Logflare.Sources.Counters)
     Logflare.Sources.Counters.incriment(source.token)
 
     # Stripe mocks
--- a/test/logflare/sources_test.exs
+++ b/test/logflare/sources_test.exs
@@ -8,22 +8,21 @@ defmodule Logflare.SourcesTest do
   alias Logflare.Google.BigQuery.GenUtils
   alias Logflare.Source.RecentLogsServer, as: RLS
   alias Logflare.Source
-  @moduletag :this
 
-  setup do
-    u = Users.get_by(email: System.get_env("LOGFLARE_TEST_USER_WITH_SET_IAM"))
-    s = insert(:source, token: Faker.UUID.v4(), rules: [], user_id: u.id)
-
-    Source.BigQuery.Schema.start_link(%RLS{
-      source_id: s.token,
-      plan: %{limit_source_fields_limit: 500}
-    })
-
-    {:ok, sources: [s], users: [u]}
-  end
-
+  @tag :failing
   describe "Sources" do
-    @tag :failing
+    setup do
+      u = Users.get_by(email: System.get_env("LOGFLARE_TEST_USER_WITH_SET_IAM"))
+      s = insert(:source, token: Faker.UUID.v4(), rules: [], user_id: u.id)
+
+      Source.BigQuery.Schema.start_link(%RLS{
+        source_id: s.token,
+        plan: %{limit_source_fields_limit: 500}
+      })
+
+      {:ok, sources: [s], users: [u]}
+    end
+
     test "get_bq_schema/1", %{sources: [s | _], users: [u | _]} do
       source_id = s.token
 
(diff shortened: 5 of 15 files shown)


chasers (approved):

config/dev.exs
```diff
@@ -104,7 +104,7 @@ config :logflare, Logflare.Vercel.Client,
   install_vercel_uri: "https://vercel.com/integrations/logflare-v2-dev/new"
 
 config :goth,
-  json: "config/secrets/logflare-dev-238720-63d50e3c9cc8.json" |> File.read!()
+  json: "config/secrets/logflare-dev-238720-344624644919.json" |> File.read!()
```
Ziinc, line 107: @chasers if we're using different dev jwt secrets now, we shld move this to `dev.secrets.exs`. I'm still using `63d50e3c9cc8`

lib/logflare/sources.ex
```diff
@@ -355,4 +354,14 @@ defmodule Logflare.Sources do
...
+
+  @spec preload_for_dashboard(list(Source.t())) :: list(Source.t())
+  def preload_for_dashboard(sources) do
+    sources
+    |> Enum.map(&preload_defaults/1)
+    |> Enum.map(&preload_saved_searches/1)
+    |> Enum.map(&put_schema_field_count/1)
+    |> Enum.sort_by(& &1.name, &<=/2)
+    |> Enum.sort_by(& &1.favorite, &>=/2)
+  end
```
Ziinc, line 366: we can split these into 2 different functions, one that iterates over each item and another that does the sort.

This one isn't very efficient  because it iterates over the list 5 time, one for each map and sort_by.

Would be more efficient to do a singular map (where each op is performed element-wise), then a singular sort.

also, function name doesn't match the actual operations, would be more accurate to say `prepare_for_dashboard/1` instead. In this case, the function name is not self-documenting and would be misleading.

lib/logflare_web/controllers/api/source_controller.ex
```diff
@@ -0,0 +1,11 @@
+defmodule LogflareWeb.Api.SourceController do
+  use LogflareWeb, :controller
+  alias Logflare.Sources
+  alias Logflare.Users
+
+  def index(%{assigns: %{user: user}} = conn, _) do
```
Ziinc, line 6: missing fallback controller, but i think you added it in the other PR.

test/logflare/source/billing_writer_test.exs
```diff
@@ -13,14 +15,14 @@ defmodule Logflare.Source.BillingWriterTest do
...
-      start_supervised!(
+    {:ok, pid} =
+      start_supervised(
         {BillingWriter,
          %RecentLogsServer{source_id: source.token, source: source, user: user, plan: plan}}
       )
 
     # increase log count
-    start_supervised!(Logflare.Sources.Counters)
+    start_supervised(Logflare.Sources.Counters)
```
Ziinc, line 25: this would silently fail

test/logflare/source/billing_writer_test.exs
```diff
@@ -13,14 +15,14 @@ defmodule Logflare.Source.BillingWriterTest do
     user = user |> Logflare.Repo.preload(:billing_account)
     plan = insert(:plan, type: "metered")
 
-    pid =
-      start_supervised!(
+    {:ok, pid} =
+      start_supervised(
```
Ziinc, line 19: no meaningful difference, because we would want the test to fail if the genserver can't start.

test/logflare/sources_test.exs
```diff
@@ -8,22 +8,21 @@ defmodule Logflare.SourcesTest do
...
-  @moduletag :this
 
-  setup do
-    u = Users.get_by(email: System.get_env("LOGFLARE_TEST_USER_WITH_SET_IAM"))
-    s = insert(:source, token: Faker.UUID.v4(), rules: [], user_id: u.id)
+  @tag :failing
+  describe "Sources" do
+    setup do
+      u = Users.get_by(email: System.get_env("LOGFLARE_TEST_USER_WITH_SET_IAM"))
+      s = insert(:source, token: Faker.UUID.v4(), rules: [], user_id: u.id)
```
Ziinc, line 16: can simplify this to:

```
user = insert(:user)
source = insert(:source, user: user)
```

also, would defer towards using fully qualified variable  names where possible.

test/logflare/sources_test.exs
```diff
@@ -82,4 +81,28 @@ defmodule Logflare.SourcesTest do
...
+      sources = insert_list(3, :source, %{user: user})
+      sources = Sources.preload_for_dashboard(sources)
+
+      assert Enum.all?(sources, &Ecto.assoc_loaded?(&1.user))
+      assert Enum.all?(sources, &Ecto.assoc_loaded?(&1.rules))
+      assert Enum.all?(sources, &Ecto.assoc_loaded?(&1.saved_searches))
+    end
+
+    test "sorts data by name and favorite flag", %{user: user} do
+      source_1 = insert(:source, %{user: user, name: "C"})
```
Ziinc, line 100: lets use lists for the factory attrs, less verbose

test/logflare/sources_test.exs
```diff
@@ -82,4 +81,28 @@ defmodule Logflare.SourcesTest do
...
+      source_2 = insert(:source, %{user: user, name: "B", favorite: true})
+      source_3 = insert(:source, %{user: user, name: "A"})
+      sources = Sources.preload_for_dashboard([source_1, source_2, source_3])
+
+      assert Enum.map(sources, & &1.id) == Enum.map([source_2, source_3, source_1], & &1.id)
```
Ziinc, line 105: can simplify this to 

```
ids = [source_1, source_2, source_3] | > Sources.preload_for_dashboard() | >  Enum.map(& &1.id)
assert [source_2.id,source_3.id,source_1.id] == ids
```

test/logflare_web/controllers/api/source_controller_test.exs
```diff
@@ -0,0 +1,45 @@
...
+    insert(:plan, name: "Free")
+    u1 = insert(:user)
+    u2 = insert(:user)
+    Teams.create_team(u1, %{name: "u1 team"})
+    Teams.create_team(u2, %{name: "u2 team"})
+
+    s1 = insert(:source, public_token: Faker.String.base64(16), user_id: u1.id)
+    s2 = insert(:source, user_id: u1.id)
+    s3 = insert(:source, user_id: u2.id)
+
```
Ziinc, line 21: we can cut this down to 
```
user = insert(:user)
team = insert(:team, user: user)
source = insert(:source, user: user)
```
Might already have a team factory, not sure.

Also, we don't need to start the Counters, we can mock its behaviour instead (we're going to refactor it heavily)

Ziinc: can also just pass  at the end of setup

```
[user: user, team: team, source: source]
```
since the api test only needs to test that the api works, don't need to test logic for fetching of multiple sources

test/logflare_web/controllers/api/source_controller_test.exs
```diff
@@ -0,0 +1,45 @@
...
+    {:ok, users: [u1, u2], sources: [s1, s2, s3]}
+  end
+
+  describe "index/2" do
+    test "returns list of sources for given user", %{
+      conn: conn,
+      users: [u1, _u2],
+      sources: [s1, s2 | _]
+    } do
```
Ziinc, line 32: can cut to `user: user, source: source`

test/logflare_web/controllers/api/source_controller_test.exs
```diff
@@ -0,0 +1,45 @@
...
+      response =
+        conn
+        |> login_user(u1)
+        |> get("/api/sources")
+        |> json_response(200)
+
+      response = Enum.map(response, & &1["id"])
+      expected = Enum.map([s1, s2], & &1.id)
+
+      assert_lists_equal(response, expected)
```
Ziinc, line 42: seems quite unnecessary to add `:assertions` for list comparison, esp when lists are so small. 
actually id matching would suffice

```
assert [%{"id"=> ^source_id}] = json_response(conn, 200)
```

test/logflare_web/controllers/source_controller_test.exs
```diff
@@ -374,13 +372,7 @@ defmodule LogflareWeb.SourceControllerTest do
...
 
-  def login_user(conn, u) do
-    conn
-    |> Plug.Test.init_test_session(%{user_id: u.id})
-    |> assign(:user, u)
-  end
-
-  def expect_user_plan(_ctx) do
+  defp expect_user_plan(_ctx) do
     expect(Logflare.Billing, :get_plan_by_user, fn _ ->
```
Ziinc, line 376: can move this up to the common setup, i think. don't need separate mock setup.

Ziinc (commented):
refactor logic looks ok

config/dev.exs
filipecabaco: Agreed, I can tackle that in a fix-it-up PR.

We should explore ways of making this a bit easier for development (e.g. mock authentications for development?)

lib/logflare/sources.ex
filipecabaco: I thought about it and makes a lot of sense 👍 

we can also make an intermediary step where:
1.  Use Streams to compose transformations 
2. `sort_by` we should actually return a tuple of {name, favorite} so it can sort directly with multiple params but transverse only once ( https://hexdocs.pm/elixir/1.13/Enum.html#sort_by/3-examples )

lib/logflare_web/controllers/api/source_controller.ex
filipecabaco: yes, tackled in another PR. 

The possible errors are related with lack of auth which would be caught by the plugs, the controller itself can only return lists (either empty or not)

test/logflare/source/billing_writer_test.exs
filipecabaco: It's ok to leave it as is. This was failing with `:already_started` errors, but if it fails to startup, the test will catch it.

Other than this approach, we could check if the process exists and only start at that point but that felt bloated for a test.

test/logflare/source/billing_writer_test.exs
filipecabaco: similar to https://github.com/Logflare/logflare/pull/1196#discussion_r1020190243
this was also throwing the `:already_started` error.

we could use the same approach

test/logflare/sources_test.exs
filipecabaco: can start to tackle it in a cleanup PR 👍

test/logflare/sources_test.exs
filipecabaco: 👍

test/logflare/sources_test.exs
filipecabaco: 👍

test/logflare_web/controllers/api/source_controller_test.exs
filipecabaco: will change, this was a big copy-paste from the previous code and I wanted to avoid messing too much with it but you are right it needs a cleanup to standardize test approaches

As for Counters the issue is actually the lack of an ETS table that is initialized with the process itself. We do not have an easy way now to mock it from what I can see (no way to define the implementation to be used).

test/logflare_web/controllers/api/source_controller_test.exs
filipecabaco: this was a bit of future-proofing since it's a lightweight dependency but can refactor to remove it. 👍

test/logflare_web/controllers/source_controller_test.exs
```diff
@@ -1,24 +1,28 @@
...
+
   setup_all do
     Sources.Counters.start_link()
     :ok
   end
 
   setup do
+    insert(:plan, name: "Free")
     u1 = insert(:user)
     u2 = insert(:user)
```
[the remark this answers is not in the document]
filipecabaco: 👍 will do

test/logflare_web/controllers/source_controller_test.exs
filipecabaco: do you think we should remove this check?

test/support/conn_case.ex
```diff
@@ -39,6 +39,12 @@ defmodule LogflareWeb.ConnCase do
       setup context do
         Mimic.verify_on_exit!(context)
       end
+
+      def login_user(conn, u) do
+        conn
+        |> Plug.Test.init_test_session(%{user_id: u.id})
```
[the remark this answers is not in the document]
filipecabaco: here we're only assigning a user to a connection so we already went a step ahead of validating access tokens from my point of view. But we should add some tests to `lib/logflare_web/controllers/plugs/set_verify_user.ex` to tackle the authentication pipeline for API

config/dev.exs
chasers: Yeah this is fine.  This new one is in Bitwarden btw.

filipecabaco: will move to dev.secrets.exs

PR #1197, opened 2022-11-10
Account operations source api

filipecabaco:
Adds multiple operations to an accounts source API 

* GET `/api/sources/:token` - Creates a new source to a given user
* POST `/api/sources` - Creates a new source to a given user
* PATCH `/api/sources/:token` - Updates a given source
* DELETE `/api/sources/:token` - Deletes a given source

### DIFF
--- a/lib/logflare_web/controllers/api/fallback_controller.ex
+++ b/lib/logflare_web/controllers/api/fallback_controller.ex
@@ -9,4 +9,10 @@ defmodule LogflareWeb.Api.FallbackController do
     |> put_status(:unprocessable_entity)
     |> json(%{errors: errors})
   end
+
+  def call(conn, nil) do
+    conn
+    |> put_status(:not_found)
+    |> text("")
+  end
 end
--- a/lib/logflare_web/controllers/api/fallback_controller.ex
+++ b/lib/logflare_web/controllers/api/fallback_controller.ex
@@ -3,7 +3,7 @@ defmodule LogflareWeb.Api.FallbackController do
   alias Ecto.Changeset
 
   def call(conn, {:error, %Changeset{} = changeset}) do
-    errors = Changeset.traverse_errors(changeset, fn _, _, {message, _} -> "#{message}" end)
+    errors = Changeset.traverse_errors(changeset, fn _, _, {message, _} -> message end)
 
     conn
     |> put_status(:unprocessable_entity)
--- a/lib/logflare_web/controllers/api/fallback_controller.ex
+++ b/lib/logflare_web/controllers/api/fallback_controller.ex
@@ -1,0 +1,12 @@
+defmodule LogflareWeb.Api.FallbackController do
+  use Phoenix.Controller
+  alias Ecto.Changeset
+
+  def call(conn, {:error, %Changeset{} = changeset}) do
+    errors = Changeset.traverse_errors(changeset, fn _, _, {message, _} -> "#{message}" end)
+
+    conn
+    |> put_status(:unprocessable_entity)
+    |> json(%{errors: errors})
+  end
+end
--- a/lib/logflare_web/controllers/api/source_controller.ex
+++ b/lib/logflare_web/controllers/api/source_controller.ex
@@ -11,26 +11,32 @@ defmodule LogflareWeb.Api.SourceController do
     json(conn, sources)
   end
 
+  def show(%{assigns: %{user: user}} = conn, %{"token" => token}) do
+    with source when not is_nil(source) <- Sources.get_by(token: token, user_id: user.id),
+         {:ok, source} <- Sources.preload_for_dashboard([source]) do
+      json(conn, source)
+    end
+  end
+
   def create(%{assigns: %{user: user}} = conn, params) do
-    with params <- Map.put(params, "token", Ecto.UUID.generate()),
-         {:ok, _} <- Sources.create_source(params, user) do
+    with {:ok, source} <- Sources.create_source(params, user) do
       conn
-      |> Plug.Conn.send_resp(204, [])
-      |> Plug.Conn.halt()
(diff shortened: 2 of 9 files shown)


lib/logflare_web/controllers/api/fallback_controller.ex
```diff
@@ -0,0 +1,18 @@
+defmodule LogflareWeb.Api.FallbackController do
+  use Phoenix.Controller
+  alias Ecto.Changeset
+
+  def call(conn, {:error, %Changeset{} = changeset}) do
+    errors = Changeset.traverse_errors(changeset, fn _, _, {message, _} -> "#{message}" end)
```
Ziinc, line 6: the `message` will already be a string, no need to type convert.

lib/logflare_web/controllers/api/source_controller.ex
```diff
@@ -3,9 +3,38 @@ defmodule LogflareWeb.Api.SourceController do
...
+  action_fallback LogflareWeb.Api.FallbackController
+
   def index(%{assigns: %{user: user}} = conn, _) do
     user = Users.preload_sources(user)
     sources = Sources.preload_for_dashboard(user.sources)
     json(conn, sources)
   end
+
+  def create(%{assigns: %{user: user}} = conn, params) do
+    with params <- Map.put(params, "token", Ecto.UUID.generate()),
```
Ziinc, line 15: we should shift down uuid generation to changeset level, actually. or even better, autogenerate in schema.

test/logflare_web/controllers/api/source_controller_test.exs
```diff
@@ -42,4 +45,107 @@ defmodule LogflareWeb.Api.SourceControllerTest do
...
+    test "creates a new source for an authenticated user", %{
+      conn: conn,
+      users: [u1, _u2]
+    } do
+      name = TestUtils.random_string()
+
+      conn
+      |> login_user(u1)
+      |> post("/api/sources", %{name: name})
+      |> response(204)
```
Ziinc, line 59: shouldnt this be 201?

test/logflare_web/controllers/api/source_controller_test.exs
```diff
@@ -42,4 +45,107 @@ defmodule LogflareWeb.Api.SourceControllerTest do
...
+
+      assert Repo.one(
```
Ziinc, line 61: should assert on response instead, controller layer shouldn't have repo knowledge.

test/logflare_web/controllers/api/source_controller_test.exs
```diff
@@ -42,4 +45,107 @@ defmodule LogflareWeb.Api.SourceControllerTest do
...
+    } do
+      name = TestUtils.random_string()
+
+      conn
+      |> login_user(u1)
+      |> patch("/api/sources/#{s1.id}", %{name: name})
+      |> response(204)
+
+      assert name ==
+               Repo.one(
```
Ziinc, line 85: as above, can pattern match on response instead.

Ziinc: usual created resonse would be the return the created object

test/logflare_web/controllers/api/source_controller_test.exs
```diff
@@ -42,4 +45,107 @@ defmodule LogflareWeb.Api.SourceControllerTest do
...
+      sources: [_s1, _s2, s3]
+    } do
+      conn
+      |> login_user(u1)
+      |> delete("/api/sources/#{s3.id}", %{name: TestUtils.random_string()})
+      |> response(404)
+    end
+  end
+
+  test "changeset errors handled gracefully", %{
```
Ziinc, line 139: we can merge this test together with update/2 and create/2 blocks, just adds 1 extra line of assertion for each test case.

```
# in setup
conn = login_user(user)


assert %{"error"=> %{"name"=> [ "can't" <>_]}} = conn 
        |> post("/api/sources" , @invalid) 
        |> json_response(422)


assert %{"error"=> %{"name"=> [ "can't" <>_]}} = conn 
        |> put("/api/sources/${s.id}" , @invalid) 
        |> json_response(422)

Ziinc (commented):
another point to note is that throughout the app, `source.id` is used for routing and internal source identification, however, from an external user POV, they identify sources using the `source.token` uuid, as it is easier to remember the char strings start/endings.
This pattern is in line with the Logflare Endpoints api path params, where the `endpoint_query.token` is used in the path.
we should keep the api consistent and use uuids for external user params.

lib/logflare_web/controllers/api/fallback_controller.ex
filipecabaco: nice call! I had a different message being built 🤦

lib/logflare_web/controllers/api/source_controller.ex
filipecabaco: will change it 👍

test/logflare_web/controllers/api/source_controller_test.exs
filipecabaco: oh yes! way better

test/logflare_web/controllers/api/source_controller_test.exs
filipecabaco: the idea here was to just create the entity and reply an empty body. I can change it to have the entity in the response body and assert on it.

test/logflare_web/controllers/api/source_controller_test.exs
filipecabaco: will change 👍

filipecabaco:
Made requested changes. We're also using token instead of id to identify sources in the URL

chasers (approved):

PR #1202, opened 2022-11-11
Fix Show endpoint ; Apply more refactoring suggestions

filipecabaco:
Show endpoint had bad test cases and bad implementation. This PR fixes both and also simplifies tests and applies other refactoring suggestions


Ziinc (approved):
👍

PR #1204, opened 2022-11-14
Feat: Replace closed source SQL parser

Ziinc:
This PR replaces the closed source sql parser with an open-source implementation, enabled through a rust NIF.
This parser integration relies on [pg-parse](https://github.com/paupino/pg_parse), but it is likely that we would need to switch the parser again to a different parser due to certain postgres-specific limitations that causes valid bigquery statements to be converted with extra symbols.

For example, table names with backticks are wrapped with double quotes unnecessarily
```
-- closed source parser
from `my-table` ...

-- pg-parse
from \"`my-table`\" ...

```


chasers:
BOOM

Screen Shot 2022-12-02 at 6 24 54 AM

chasers:
Assuming your tests pass this looks great @Ziinc!!!

chasers (approved):

PR #1209, opened 2022-11-16
chore: fix compilation check for docs in test utils

Ziinc:


chasers (approved):
What were you getting?  

Or is it because I had two `@doc`s on the same function name maybe?

test/support/test_utils.ex
```diff
@@ -99,10 +99,7 @@ defmodule Logflare.TestUtils do
     end)
   end
 
-  @doc """
-  gzipped request body from the Cloudflare Log Push HTTP service.
-  """
-
+  # gzipped request body from the Cloudflare Log Push HTTP service.
```
chasers, line 102: Ah yeah saw your message in Slack. You should just be able to keep the first `@doc`.

Ziinc: 🆗 updated

PR #1217, opened 2022-11-17
Remove editor roles

chasers:
Sets the correct perms for the compute engine service account.


chasers:
Merging so I can start the staging build to test there.

PR #1220, opened 2022-11-17
Fix/UI fixes

chasers:
- Adds a color background to the dropdown expanded nav on mobile (not sure why this changed)
- Makes live tailing more reliable (really needs a re-write but this should work for now)


chasers:
Merging to get on staging.

PR #1221, opened 2022-11-17
Add new local environment for single tenant mode

filipecabaco:
To support new single-tenant efforts, we're adding a new Mix env that setups a minimal version of Logflare that can be used by a new Dockerfile to boot up the full stack locally via docker compose.


config/releases.exs
```diff
@@ -1 +1,98 @@
 import Config
+
```
Ziinc, line 2: we can have a portion here that sets the`:single_tenant` config key, something like
```
config :logflare, single_tenant: System.get_env("LOGFLARE_SINGLE_TENANT")
```
which will get merged with the compile-time keys, so the final config on `:logflare` is
```
# for prod single tenant
config :logflare, env: :prod, single_tenant: true


# for dev single tenant
config :logflare, env: :dev, single_tenant: true

# for dev cluster
config :logflare, env: :dev, single_tenant: false

# for prod cluster
config :logflare, env: :prod, single_tenant: false
```
we should also think of phasing out the `staging` env, because we should see it as a "prod" environment with its own env var file, so then our staging will be as close to prod as possible.

.local.env
```diff
@@ -0,0 +1,62 @@
...
+LIBCLUSTER_TOPOLOGY_GOOGLE_ZONES=
+
+VERCEL_AUTH_CLIENT_ID=
+VERCEL_AUTH_HOST=
+
+VERCEL_CLIENT_CLIENT_ID=
+VERCEL_CLIENT_INSTALL_URI=
+VERCEL_CLIENT_REDIRECT_URI=
+
+GOOGLE_DATASET_ID_APPEND=
```
Ziinc, line 29: see below, we should be able of these configs for the local env

docker/local.app.Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+WORKDIR /logflare
+RUN mix local.rebar --force
+RUN mix local.hex --force
+RUN mix do deps.get, deps.compile
+
+RUN mix phx.digest
+RUN mix release --force --overwrite
+
+FROM elixir:1.12-alpine as app
```
Ziinc, line 27: the final docker file does not need elixir,  just needs to be the same alpine version as the build image.

Ziinc: it should cut down the image size even more 😍

config/releases.exs
chasers: This sounds nice. 

Let's move away from "single tenant" because frankly I'd rather not feature flag a bunch of stuff just to feature flag it. The features are there so let them use them if they want to figure it out. If they're motivated enough they'll fork it and figure it out anyways. 

Maybe "self host".

config/releases.exs
```diff
@@ -1 +1,98 @@
...
+  config :logflare, LogflareWeb.Auth.VercelAuth,
+    vercel_app_host: System.get_env("VERCEL_AUTH_HOST"),
+    client_id: System.get_env("VERCEL_AUTH_CLIENT_ID")
+
+  config :logflare, Logflare.Vercel.Client,
+    client_id: System.get_env("VERCEL_CLIENT_CLIENT_ID"),
+    redirect_uri: System.get_env("VERCEL_CLIENT_REDIRECT_URI"),
+    install_vercel_uri: System.get_env("VERCEL_CLIENT_INSTALL_URI")
+
+  config :logflare, Logflare.Google,
```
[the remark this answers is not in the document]
chasers: Yeah I'll double check these too.

docker/local.app.Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+ENV MIX_ENV prod
+ENV MAGIC_COOKIE $magic_cookie
+
+COPY --from=builder ./logflare/_build/prod ./app
+COPY --from=builder ./logflare/priv/static ./app/priv/static
+
+ADD .google.secret.json app/.google.secret.json
+
+CMD ["./app/rel/logflare/bin/logflare", "start", "--name" , "logflare"]
```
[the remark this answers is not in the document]
filipecabaco: For the life of me, I can't understand why I still get the `Can't set long node name` ... need some help with that

docker-compose.yml
```diff
@@ -9,6 +9,15 @@ services:
...
+  logflare:
+    build:
+      dockerfile: ./docker/local.app.Dockerfile
+    ports:
+      - "4000:4000"
+    links:
+      - "db:db"
+    env_file:
+      - .local.env
+    hostname: logflare
```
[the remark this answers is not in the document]
filipecabaco: will remove

lib/logflare/application.ex
```diff
@@ -41,7 +41,7 @@ defmodule Logflare.Application do
     env = Application.get_env(:logflare, :env)
     # Setup Goth for GCP connections
     credentials =
-      if env in [:dev, :test],
+      if env in [:dev, :test, :local],
```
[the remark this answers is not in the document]
filipecabaco: will remove

chasers (approved):

chasers:
Whenever you think this is good to go feel free to merge.

.gitignore
```diff
@@ -50,6 +50,12 @@ gcloud_prod.json
...
+.prod.env
+.local.env
+.staging.env
 
 # rustler nif binaries
-priv/native/
\ No newline at end of file
+/priv/native
+/native/sqlparser_ex/target
+/native/sqlparser_ex/.config
```
Ziinc, line 61: don't need the last two under `/native/*`, see above.

.gitignore
```diff
@@ -50,6 +50,12 @@ gcloud_prod.json
 /cover/
 /node_modules/
 .vscode/
+.google.secret.json
+.prod.env
+.local.env
+.staging.env
```
Ziinc, line 56: we can do `.*.env` to be more concise.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+WORKDIR /logflare
+RUN mix do local.rebar --force, local.hex --force
+RUN mix do deps.get, deps.compile
+
+RUN mix phx.digest
+RUN mix release
+
+FROM alpine:3.16.0 as app
+WORKDIR /root/
+# erlexec requires SHELL to be set
```
Ziinc, line 24: erlexec has been removed, can remove

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ENV SHELL /bin/sh
+ENV MAGIC_COOKIE=$magic_cookie
+
+RUN apk update && apk add -f openssl libgcc libstdc++ ncurses-libs
```
Ziinc, line 28: could you add a comment for what each dep is needed for?

README.md
```diff
@@ -69,7 +69,9 @@ Logflare is using a SQL parser from sqlparser.com. To set this up on your dev ma
 ```elixir
 iex> LogflareLogger.info("testing log message")
 ```
-
+### Using Docker
+1. Build images with `docker compose build --build-arg=magic_cookie=$(cat /dev/urandom | LC_ALL=C tr -dc 'a-zA-Z0-9' | fold -w 50 | head -n 1)`
```
Ziinc, line 73: as above, we don't need to set the magic cookie at build time.

Also, we can use a simple value like `mycookie` for the cookie as it is for illustration purposes.

the dcker compose file would also serve as an example.

config/runtime.exs
```diff
@@ -0,0 +1,106 @@
+import Config
+
+if config_env() != :test do
```
Ziinc, line 3: ideally we should test the env var loading by having an `.test.env` checked into the repo. that would ensure that the env var passing to config is correct.

docker/db.Dockerfile
```diff
@@ -0,0 +1,9 @@
+from postgres:13.4
+
+RUN echo "\connect postgres" >> /docker-entrypoint-initdb.d/init.sql
```
Ziinc, line 3: it is better to add the wal commands to a `priv/wal.sql` and mount it to the postgres image's initdb volume. That way we don't have to build the db image and have one file to handle.

Something like:
```
db:
    volumes:
      - /priv/wal.sql:/docker-entrypoint-initdb.d/wal.sql
      - pg-data:/var/lib/postgresql/data
```

run.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/sh
+echo "Migrating..."
```
Ziinc, line 2: lets move logging over to the Logflare.Release module

run.sh
```diff
@@ -0,0 +1,5 @@
...
+./logflare eval Logflare.Release.migrate
+echo "Migration done; Starting app"
+./logflare start --sname logflare
```
Ziinc, line 5: would prefer to have not have a separate run script, seems unnecessary when we can put the migration and start command inside of the dockerfile image RUN command and chain them with `&&`.

Ziinc: Also, more files = more maintenance

Ziinc (changes requested):

config/runtime.exs
Ziinc: we can do this in another PR, not as important.

.github/workflows/tests.yml
```diff
@@ -39,14 +39,23 @@ jobs:
...
+        with:
+          rust-version: stable
       - name: Restore dependencies cache
         uses: actions/cache@v3
         with:
           path: |
             deps
             _build
+            native/sqlparser_ex/.cargo
+            native/sqlparser_ex/target
```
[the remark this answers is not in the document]
filipecabaco: will change

.github/workflows/tests.yml
```diff
@@ -39,14 +39,23 @@ jobs:
...
           key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
           restore-keys: ${{ runner.os }}-mix-
+      - name: Compile native
+        working-directory: ./native/sqlparser_ex
+        run: cargo build
```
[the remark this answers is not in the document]
filipecabaco: I think I tried it and it failed 🤔 will try again probably something was missing

.github/workflows/tests.yml
```diff
@@ -61,3 +70,26 @@ jobs:
...
+    #     password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build
+        uses: docker/build-push-action@v3
+        with: 
+          push: false
+          tags: logflare/app:${{ env.LOGFLARE_VERSION }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          build-args: | 
+            MAGIC_COOKIE=${{ env.GITHUB_SHA }}
```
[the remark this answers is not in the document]
filipecabaco: ok will change

.gitignore
filipecabaco: what do you mean? and this is for local development purposes

.gitignore
filipecabaco: will change

Dockerfile
filipecabaco: will remove

Dockerfile
filipecabaco: honestly, they are just there because the Beam needs it... can leave a comment saying that 😅

config/runtime.exs
filipecabaco: No need, it works without it and this works for now, we can refine as we go for now

run.sh
filipecabaco: I would rather keep it like this for now. 

It's one more small file that enable us to add more tasks if need be (potentially seeding?). 

Plus the RUN command would be run during the build process of docker so we would be running migrations during the building process. We could have a CMD that does it but not a big win in terms of readability.

README.md
filipecabaco: will change

docker/db.Dockerfile
Ziinc: Mounting sql files to the init directory is in the first line of the section you're referring to:
Screenshot 2022-12-16 at 1 10 54 PM

run.sh
Ziinc: ah yes i was referring to the `CMD` directive, sorry for confusion.

Sure, no issue then. Seeding would be done as an elixir module, not through shell scripting.

docker/db.Dockerfile
filipecabaco: ok that was dumb on my part, will tackle it the way you mentioned

.gitignore
filipecabaco: got it will remove, I did cargo build inside `/native/sqlparser_ex/` hence creating things there and ignoring it but will change it

Ziinc (approved):

PR #1223, opened 2022-11-18
Use IntersectionObserver instead

chasers:
Holy crap I could never get this to work before! Just needed to go in a different spot.

### DIFF
--- a/assets/js/source.js
+++ b/assets/js/source.js
@@ -177,7 +177,7 @@ function resetScrollTracker() {
         })
       })
     
-    const target = document.querySelector("#search-tip")
+    const target = document.querySelector("#observer-target")
     observer.observe(target)
 }
 
--- a/assets/js/source.js
+++ b/assets/js/source.js
@@ -103,6 +103,7 @@ async function renderLog(event) {
 }
 
 export function scrollBottom() {
+
   const y = document.body.clientHeight
 
   window.scrollTo(0, y)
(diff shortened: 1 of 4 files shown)


assets/js/source.js
```diff
@@ -161,20 +162,23 @@ export async function switchDateFormat() {
...
+            
+            window.scrollTracker = true
+          } else {
+            // don't stick to bottom
+            window.scrollTracker = false
+          }
+        })
+      })
+    
+    const target = document.querySelector("#search-tip")
```
Ziinc, line 180: could we not use the search tip as the target? I'm thinking of removing it and moving all the tips to the docs so that there is better visibility for them.

Ziinc (approved):
Minor comment on target

assets/js/source_lv_hooks.js
```diff
@@ -23,30 +23,22 @@ hooks.SourceLogsSearchList = {
...
-        // We don't reset window.playing here because we don't want it to start playing again when you scroll back down
-        window.removeEventListener("scroll", _.throttle(resetScrollTracker, 250))
-        hook.pushEvent("soft_pause", {})
-      }
-    };
+    const observer =
+      new IntersectionObserver((entries, observer) => {
+        entries.forEach((entry) => {
+          let searchInView = entry.isIntersecting
+          if (searchInView) {
```
Ziinc, line 30: we can cut out the commented part and use
```
if (!searchInView){
  hook.pushEvent("soft_pause", {})
}
```

assets/js/source.js
chasers: What's more visible than actually in the UI?

assets/js/source_lv_hooks.js
chasers: I think the comment is useful here because I've thought about auto-playing the search when you scroll to the bottom.

Ultimately, it's not what we want to do because you may be looking at some logs which you don't want to lose view of and that could happen if we auto-play.  This comment reminds me that we've tried this before, but it's also useful to uncomment and play with it if you're working on it.

assets/js/source.js
chasers: I changed the observer target tho.

PR #1235, opened 2022-11-29
Fix/clustering key fix

Ziinc:
This PR adds in a guard for the top level `project` field for specific sources, as a fix for phasing the workaround for the top level project key rewrite.

Full removal of the clustering key logic can only be done once CF and pg log pipelines are updated, hence the addition of the guard ensure that we only do the rewrite when the key is not provided (should not be the case anymore once the api gateway fix is rolled to prod).


Ziinc:
@chasers just checked, for the CF logs the value gets overwritten unless host header is empty, so will need to merge this still

chasers (approved):

PR #1238, opened 2022-11-30
chore: remove clustering keys hack

Ziinc:
Once the prod api gateway has been updated to send top level project, we can safely remove the cloudflare hack.

Postgres logs is already sending top level project key, as can be [seen here](https://github.com/supabase/infrastructure/blob/d5a08873610ad0755aa0ec5e981608bdbd0da154/init-scripts/postgres/vector.toml#L79)

### DIFF
--- a/lib/logflare/logs/log_event.ex
+++ b/lib/logflare/logs/log_event.ex
@@ -125,7 +125,6 @@ defmodule Logflare.LogEvent do
         other ->
           other
       end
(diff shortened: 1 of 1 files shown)


chasers (approved):

chasers:
@Ziinc this should be [good to go](https://github.com/supabase/api-gateway/pull/96)

PR #1244, opened 2022-12-10
Add Teams / Teams User management API

filipecabaco:
* Add token field to Team
* Add List Teams endpoint
* Add Show Team endpoint
* Add Create team endpoint
* Add Update team endpoint
* Add Delete endpoint
* Fix assertions on other endpoint tests

When listing or showing teams, we will consider the team the user is owner of and the ones he is part of via Team User.

Create, Update and Delete can only be done to the team they are owner of.

### DIFF
--- a/lib/logflare/teams.ex
+++ b/lib/logflare/teams.ex
@@ -1,13 +1,15 @@
 defmodule Logflare.Teams do
   @moduledoc false
   import Ecto.Query, warn: false
-  alias Logflare.{Repo, Teams.Team, TeamUsers.TeamUser, Users, User}
+  alias Logflare.Repo
+  alias Logflare.Teams.Team
+  alias Logflare.TeamUsers.TeamUser
+  alias Logflare.Users
+  alias Logflare.User
 
   @doc "Returns a list of teams. Unfiltered."
   @spec list_teams() :: [Team.t()]
-  def list_teams do
-    Repo.all(Team)
-  end
+  def list_teams, do: Repo.all(Team)
 
   @doc "Gets a single team. Raises `Ecto.NoResultsError` if the Team does not exist."
   @spec get_team!(String.t() | number()) :: Team.t()
@@ -21,28 +23,29 @@ defmodule Logflare.Teams do
   @spec get_home_team(TeamUser.t()) :: Team.t() | nil
   def get_home_team(%TeamUser{email: email}) do
     case Users.get_by(email: email) |> Users.preload_team() do
-      nil ->
-        nil
-
-      user ->
-        user.team
+      nil -> nil
+      %{team: team} -> team
     end
   end
 
   @doc "Preloads the `:user` assoc"
-  @spec preload_user(nil | Team.t()) :: Team.t()
+  @spec preload_user(nil | Team.t()) :: Team.t() | nil
   def preload_user(nil), do: nil
   def preload_user(team), do: Repo.preload(team, :user)
 
-  @doc "preloads the `:team_users` assoc"
-  @spec preload_team_users(nil | Team.t()) :: Team.t()
+  @doc "Preloads the `:team_users` assoc"
+  @spec preload_team_users(nil | Team.t()) :: Team.t() | nil
   def preload_team_users(nil), do: nil
   def preload_team_users(team), do: Repo.preload(team, :team_users)
 
+  @doc "Preloads given fields of a Team"
+  @spec preload_fields(nil | Team.t(), list(keyword() | atom())) :: Team.t() | nil
+  def preload_fields(nil, _fields), do: nil
+  def preload_fields(team, fields), do: Repo.preload(team, fields)
+
   @doc "Creates a team from a user."
-  @spec create_team(User.t(), %{name: String.t()}) ::
-          {:ok, Team.t()} | {:error, Ecto.Changeset.t()}
-  def create_team(user, %{name: _name} = attrs) do
+  @spec create_team(User.t(), map()) :: {:ok, Team.t()} | {:error, Ecto.Changeset.t()}
+  def create_team(user, attrs) do
     user
     |> Ecto.build_assoc(:team)
     |> Team.changeset(attrs)
@@ -64,4 +67,44 @@ defmodule Logflare.Teams do
   @doc "Returns an `%Ecto.Changeset{}` for tracking team changes"
   @spec change_team(Team.t()) :: Ecto.Changeset.t()
   def change_team(%Team{} = team), do: Team.changeset(team, %{})
+
+  @doc """
+  Lists all the teams a given user is part of.
+
+  ## Examples
+  ### User is owner of a team and belongs another user team
+  iex(1) > Logflare.Teams.list_teams_by_user_access(user, "token")
+  [%Logflare.Team{}, %Logflare.Team{}]
+  """
+  @spec list_teams_by_user_access(User.t()) :: [Team.t()]
+  def list_teams_by_user_access(user) do
+    user
+    |> query_teams_by_user_access()
+    |> Repo.all()
+  end
+
+  @doc """
+  Fetchesa single team with the given token from the list of teams he's part of
+
+  ## Examples
+  ### User is owner of a team and belongs another user team
+  iex(1) > Logflare.Teams.get_team_by_user_access(user, "token")
+  %Logflare.Team{}
+  """
+  @spec get_team_by_user_access(User.t(), binary()) :: Team.t() | nil
+  def get_team_by_user_access(user, token) do
+    user
+    |> query_teams_by_user_access()
+    |> where([t, tu], t.token == ^token or tu.token == ^token)
+    |> Repo.one()
+  end
+
+  defp query_teams_by_user_access(%User{email: email, id: id}) do
+    from t in Team,
+      join: tu in TeamUser,
+      on: tu.email == ^email,
+      where: t.user_id == ^id or tu.email == ^email,
+      distinct: true,
+      preload: [:user, :team_users]
+  end
 end
--- a/lib/logflare_web/controllers/api/team_controller.ex
+++ b/lib/logflare_web/controllers/api/team_controller.ex
@@ -1,0 +1,45 @@
+defmodule LogflareWeb.Api.TeamController do
+  use LogflareWeb, :controller
+  alias Logflare.Teams
+  action_fallback LogflareWeb.Api.FallbackController
+
+  def index(%{assigns: %{user: user}} = conn, _) do
+    teams = Teams.list_teams_by_user_access(user)
+    json(conn, teams)
+  end
+
+  def show(%{assigns: %{user: user}} = conn, %{"token" => token}) do
+    with team when not is_nil(team) <- Teams.get_team_by_user_access(user, token),
+         team <- Teams.preload_fields(team, [:user, :team_users]) do
+      json(conn, team)
+    end
+  end
+
+  def create(%{assigns: %{user: user}} = conn, params) do
+    with {:ok, team} <- Teams.create_team(user, params),
+         team <- Teams.preload_fields(team, [:user, :team_users]) do
+      conn
+      |> put_status(201)
+      |> json(team)
+    end
+  end
+
+  def update(%{assigns: %{user: user}} = conn, %{"token" => token} = params) do
+    with team when not is_nil(team) <- Teams.get_team_by(token: token, user_id: user.id),
+         {:ok, team} <- Teams.update_team(team, params),
+         team <- Teams.preload_fields(team, [:user, :team_users]) do
+      conn
+      |> put_status(204)
+      |> json(team)
+    end
+  end
+
+  def delete(%{assigns: %{user: user}} = conn, %{"token" => token}) do
+    with team when not is_nil(team) <- Teams.get_team_by(token: token, user_id: user.id),
+         {:ok, _} <- Teams.delete_team(team) do
+      conn
+      |> Plug.Conn.send_resp(204, [])
+      |> Plug.Conn.halt()
+    end
+  end
+end
--- a/lib/logflare_web/router.ex
+++ b/lib/logflare_web/router.ex
@@ -359,6 +359,10 @@ defmodule LogflareWeb.Router do
     resources "/endpoints", Api.EndpointController,
       param: "token",
       only: [:index, :show, :create, :update, :delete]
+
+    resources "/teams", Api.TeamController,
+      param: "token",
+      only: [:index, :show, :create, :update, :delete]
   end
 
   # Old log ingest endpoint. Deprecate.
--- a/priv/repo/migrations/20221210011115_set_token_default_values.exs
+++ b/priv/repo/migrations/20221210011115_set_token_default_values.exs
@@ -1,0 +1,14 @@
+defmodule Logflare.Repo.Migrations.SetTokenDefaultValues do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  def change do
+    alter table(:teams) do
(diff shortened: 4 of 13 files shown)


lib/logflare/teams.ex
```diff
@@ -64,4 +68,40 @@ defmodule Logflare.Teams do
...
+      from t in Team,
+        join: tu in TeamUser,
+        on: tu.email == ^email
+
+    team_query =
+      from t in Team,
+        where: t.user_id == ^user_id,
+        union: ^team_user_query,
+        preload: [:user, :team_users]
+
```
Ziinc, line 85: Can consolidate the query into:
```
from t in Team,
left_join: tu in assoc(t, :team_users)
where: tu.email == ^email or t.user_id == ^user_id
```
will be more performant than the union.

Also, should not preload by default, only conditionally. Each preload is an additional db query.

Also, using email might not be a good idea, not sure if the TeamUser `:token` is the same as the User `:token` when copied over, would be better to filter on it if possible.

lib/logflare_web/controllers/api/team_controller.ex
```diff
@@ -0,0 +1,51 @@
...
+  alias Logflare.Teams
+  action_fallback LogflareWeb.Api.FallbackController
+
+  def index(%{assigns: %{user: user}} = conn, _) do
+    teams = Teams.list_for_user(user)
+    json(conn, teams)
+  end
+
+  def show(%{assigns: %{user: user}} = conn, %{"token" => token}) do
+    with team when not is_nil(team) <- Teams.get(user, token),
```
Ziinc, line 12: should handle situation when it is nil

lib/logflare_web/controllers/api/team_controller.ex
```diff
@@ -0,0 +1,51 @@
...
+    with team when not is_nil(team) <- Teams.get_team_by(token: token, user_id: user.id),
+         {:ok, _} <- Teams.delete_team(team) do
+      conn
+      |> Plug.Conn.send_resp(204, [])
+      |> Plug.Conn.halt()
+    end
+  end
+
+  defp params_keys_string_to_atom(params) do
+    params |> Enum.map(fn {k, v} -> {String.to_existing_atom(k), v} end) |> Enum.into(%{})
```
Ziinc, line 49: why is this needed? Not very safe. ecto changeset should be used for casting.

lib/logflare_web/controllers/api/team_controller.ex
```diff
@@ -0,0 +1,51 @@
...
+         {:ok, team} <- Teams.create_team(user, params),
+         team <- Teams.preload_fields(team, [:user, :team_users]) do
+      conn
+      |> put_status(201)
+      |> json(team)
+    end
+  end
+
+  def update(%{assigns: %{user: user}} = conn, %{"token" => token} = params) do
+    with params <- params_keys_string_to_atom(params),
```
Ziinc, line 29: see comment below.

lib/logflare_web/controllers/api/team_controller.ex
```diff
@@ -0,0 +1,51 @@
...
+
+  def show(%{assigns: %{user: user}} = conn, %{"token" => token}) do
+    with team when not is_nil(team) <- Teams.get(user, token),
+         team <- Teams.preload_fields(team, [:user, :team_users]) do
+      json(conn, team)
+    end
+  end
+
+  def create(%{assigns: %{user: user}} = conn, params) do
+    with params <- params_keys_string_to_atom(params),
```
Ziinc, line 19: see comment below.

lib/logflare_web/router.ex
```diff
@@ -359,6 +359,10 @@ defmodule LogflareWeb.Router do
     resources "/endpoints", Api.EndpointController,
       param: "token",
       only: [:index, :show, :create, :update, :delete]
+
+    resources "/teams", Api.TeamController,
+      param: "token",
+      only: [:index, :show, :create, :update, :delete]
```
Ziinc, line 365: can use `except: [:new, :edit]` instead

priv/repo/migrations/20221210011115_set_token_default_values.exs
```diff
@@ -0,0 +1,14 @@
...
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  def change do
+    alter table(:teams) do
+      modify :token, :string, default: fragment("gen_random_uuid()")
+    end
+
+    execute "UPDATE teams SET token=gen_random_uuid()"
```
Ziinc, line 12: this will replace all existing uuids already present. should filter to null.

test/logflare/teams_test.exs
```diff
@@ -1,44 +1,91 @@
...
+    user = insert(:user)
+    teams = insert_list(2, :team_user, email: user.email)
+
+    expected = Enum.map(teams, & &1.team.id) |> Enum.sort()
+    result = Enum.map(Teams.list_for_user(user), & &1.id) |> Enum.sort()
+
+    assert expected == result
+  end
+
+  test "get/2 gets a team for a given user and token" do
```
Ziinc, line 83: should combine this function with `Teams.get_team/1` and pattern match on 2 arity

test/logflare_web/controllers/api/endpoint_controller_test.exs
```diff
@@ -112,15 +112,15 @@ defmodule LogflareWeb.Api.EndpointControllerTest do
...
-      |> login_user(user)
-      |> delete("/api/endpoints/#{endpoint.token}", %{name: name})
-      |> response(204)
-
-      conn
-      |> login_user(user)
-      |> get("/api/endpoints/#{endpoint.token}")
-      |> response(404)
+      assert conn
+             |> login_user(user)
```
Ziinc, line 116: can move `|> login_user(user)` to setup block

Ziinc: or can use a setup function for :valid_user/:invalid_user logins

test/logflare_web/controllers/api/team_controller_test.exs
```diff
@@ -0,0 +1,171 @@
...
+      non_owner_team: non_owner_team
+    } do
+      response =
+        conn
+        |> login_user(user)
+        |> get("/api/teams")
+        |> json_response(200)
+
+      response = response |> Enum.map(& &1["name"]) |> Enum.sort()
+      expected = [main_team, non_owner_team] |> Enum.map(& &1.name) |> Enum.sort()
```
Ziinc, line 37: we can do this instead to make it more concise.
```
assert [_, _] = json_resonse(conn, 200)
```
don't need to check equality, since we know there will only be 2 teams.

test/logflare_web/controllers/api/team_controller_test.exs
```diff
@@ -0,0 +1,171 @@
...
+      user: user,
+      main_team: main_team
+    } do
+      response =
+        conn
+        |> login_user(user)
+        |> get("/api/teams/#{main_team.token}")
+        |> json_response(200)
+
+      assert response["name"] == main_team.name
```
Ziinc, line 54: can do instead to be more concise.
```
test " .... " %{..., main_team: %{id: team_id}}
...
assert %{"id"=> ^team_id} = json_respnse(conn, 200)
```

test/logflare_web/controllers/api/team_controller_test.exs
```diff
@@ -0,0 +1,171 @@
...
+      invalid_user = insert(:user)
+
+      assert conn
+             |> login_user(invalid_user)
+             |> delete("/api/teams/#{main_team.token}")
+             |> response(404)
+    end
+  end
+
+  test "changeset errors handled gracefully", %{conn: conn, user: user, main_team: main_team} do
```
Ziinc, line 154: can be moved to the post/patch test cases since this is expected behaviour, will be less verbose.

Ziinc (changes requested):
main issue is with the `String.to_existing_atom` usage, which looks unnecessary.

lib/logflare/teams.ex
```diff
@@ -64,4 +68,40 @@ defmodule Logflare.Teams do
   @doc "Returns an `%Ecto.Changeset{}` for tracking team changes"
   @spec change_team(Team.t()) :: Ecto.Changeset.t()
   def change_team(%Team{} = team), do: Team.changeset(team, %{})
+
+  @doc "List all teams for user"
+  @spec list_for_user(User.t()) :: list(Team.t())
```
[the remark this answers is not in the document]
Ziinc: could we also add in example usage for docs as well for the modified functions, so that we can incrementally improve the overall internal docs

lib/logflare/teams.ex
filipecabaco: won't work because `user_id` will be used to join `team_user`. Since the user isn't the owner of all `team_user` entities it won't find all the teams this user is a member of

`team_user` and `user` do not have matching tokens, we would need to add a new field to `team_user` with the `user` token to match with it which would require way more code changes

lib/logflare/teams.ex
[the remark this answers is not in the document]
filipecabaco: will add. 

[Team.t()] is the same as list(Team.t()) and more descriptive IMO

lib/logflare_web/controllers/api/team_controller.ex
filipecabaco: it is handled, it bubbles up to the action_fallback and returns 404

lib/logflare_web/controllers/api/team_controller.ex
filipecabaco: wasn't working for some reason 😞 unsure why

test/logflare/teams_test.exs
filipecabaco: at the moment they have different purposes. `get_team/1` fetches the Team entity, `get/2` fetches the Team or a Team of a Team User the given user might be part of (that was confusing to write...)

but in the end, it has different purposes ... it might need a different name but it's a tricky one to handle

test/logflare_web/controllers/api/endpoint_controller_test.exs
filipecabaco: honestly don't seem much benefit, this way the tests are more descriptive of the flow they are executing for a given test

test/logflare_web/controllers/api/team_controller_test.exs
filipecabaco: there are more teams created during setup ( the `_non_relevant_to_main_user`) to ensure we're fetching with authorization in mind hence the check of equality

test/logflare_web/controllers/api/team_controller_test.exs
filipecabaco: can change 👍

lib/logflare/teams.ex
filipecabaco: actually I will try to left join with email vs union 👁️

lib/logflare_web/controllers/api/team_controller.ex
filipecabaco: stupid me it was something else for sure because it works... will fix the code

lib/logflare/teams.ex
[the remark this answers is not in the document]
Ziinc: 👍 both are fine to me

test/logflare_web/controllers/api/team_controller_test.exs
Ziinc: i see. fetching a two len array would imply that it is working correctly as well.

if an equality check is really needed, would be simpler to check if name is in the response name array, as this is rather hard to read.

filipecabaco: oh good call! I can even get by token because I've added the token to the jason derive

lib/logflare/teams.ex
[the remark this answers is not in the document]
filipecabaco: I'm going to change, from the code I've seen we use more the version you mentioned

test/logflare/teams_test.exs
Ziinc: seems like you're trying to do too much in one function, and also the function name doesn't self document and convey what is being fetched or checked for.

would suggest having a `Teams.has_team_access?/1` which accepts a User/TeamUser/email and does the check before actually fetching the team with `Teams.get_team/1`

lib/logflare/teams.ex
Ziinc: ok on 2nd point.

for 1st point, can join on the user email instead of using the assoc macro.

test/logflare/teams_test.exs
filipecabaco: that would mean extra queries right? it's a weird balance because I want to fetch all the teams that user has in their dashboard which is the one they own and the ones they have been invited to... honestly I feel that the other methods are the ones wrongly named since they should be "owner" focused fetches 😅

lib/logflare/teams.ex
filipecabaco: joining on user email won't work either because the join will then apply the where clauses and fail:
```elixir
team_query =
      from t in Team,
        join: tu in TeamUser,
        on: tu.email == ^email,
        where: t.user_id == ^user_id,
        preload: [:user, :team_users]
```

lib/logflare/teams.ex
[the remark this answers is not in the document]
Ziinc: @filipecabaco could we update the name as well to be more in line with codegen conventions

test/logflare/teams_test.exs
Ziinc: if thats the case then we can reuse the `list_teams_by_user_access` query and add a where filter to a team it to to one input team, that way we reuse the access logic. 

I think that having `Team.get_team/2` where first arg is the User/TeamUser is quite self explanatory that it is checking for whether user has access, though it is also ok to separate it out to two functions `Teams.get_team/1` which just retrieves any team, and `Teams.get_team_by_user_access/2` which then extends the has-access query. 

In general we opt for smaller, faster, and more flexible functions and less code dupe, because we can always use cachex if necessary for such rarely-updated tables.

Ziinc: in general we want to keep context functions as flexible as possible and don't want to put duplicate business logic in too many functions, as it makes refactoring or logic adjustments more complicated.

test/logflare_web/controllers/api/endpoint_controller_test.exs
Ziinc: sure, just seems like it is being repeated for each test

lib/logflare/teams.ex
filipecabaco: Screenshot 2023-02-02 at 11 26 25
still need to do a distinct and do the check on the email but this is closer 😄

test/logflare/teams_test.exs
filipecabaco: I will extract the query from `list_teams_by_user_access` and then use composition to add the extra where clause on `get_team_by_user_access` 👀

lib/logflare/teams.ex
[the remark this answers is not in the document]
filipecabaco: changed the names to have that concept of `*_by_user_access`

priv/repo/migrations/20221210011115_set_token_default_values.exs
Ziinc: @filipecabaco i think should err on the side of not mutating existing data

Ziinc (approved):
changes look good, could you add in the migration where clause before merge? 💯 nice work

filipecabaco:
Changed migrations and added a unique index to the token with the corresponding constraint

PR #1248, opened 2022-12-19
Update Staging secret file

filipecabaco:
Added a new value to the .enc staging file


filipecabaco:
Going to merge due to the low impact of changing the .enc file

PR #1292, opened 2023-01-10
Aliases for easier setup

filipecabaco:
Added multiple alias to facilitate startup and setup of deployment:
* start commands now source the .dev.env file
* added command to decrypt and encrypt dev .env file
* addec command to encrypt and decrypt staging .env and gcloud.json files

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -199,10 +199,18 @@ defmodule Logflare.Mixfile do
   defp aliases do
     [
       setup: ["deps.get", "cmd elixir --sname orange --cookie monster -S mix ecto.setup"],
-      start: "cmd PORT=4000 iex --sname orange --cookie monster -S mix phx.server",
-      "start.orange":
-        "cmd PORT=4000 iex --name orange@127.0.0.1 --cookie monster -S mix phx.server",
-      "start.pink": "cmd PORT=4001 iex --name pink@127.0.0.1 --cookie monster -S mix phx.server",
+      start: [
+        "cmd source .dev.env",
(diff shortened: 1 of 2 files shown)


mix.exs
```diff
@@ -215,7 +223,19 @@ defmodule Logflare.Mixfile do
       "lint.diff": ["credo diff master"],
       "lint.all": ["credo --strict"],
       "ecto.setup": ["ecto.create", "ecto.migrate", "run priv/repo/seeds.exs"],
-      "ecto.reset": ["ecto.drop", "ecto.setup"]
+      "ecto.reset": ["ecto.drop", "ecto.setup"],
+      "decrypt.dev":
+        "cmd gcloud kms decrypt --ciphertext-file='./.dev.env.enc' --plaintext-file=./.dev.env --location=us-central1 --keyring=logflare-keyring-us-central1 --key=logflare-secrets-key --project=logflare-staging",
```
Ziinc, line 228: aliases looks good, seems like we missed out prod.
Could you backlog a task for moving these to a Mix.Task? we can probably reduce a lot of repetition.

Ziinc (approved):

mix.exs
filipecabaco: didn't add production yet since we don't have that env set

Ziinc:
@filipecabaco for CI, looks like docker push is getting triggered even though we haven't merged it yet into staging.

filipecabaco:
@Ziinc it triggers the build to check it actually builds but doesn't push the image

PR #1296, opened 2023-01-12
feat: Multi-backend google analytics adaptor

Ziinc:
Implements a source backend for forwarding/generating GA4 events from Logflare.

Extracts event `name` and `client_id` from a log event. Allows one to generate multiple GA4 events from a single log event.

Api reference used for implementation is [here](https://developers.google.com/analytics/devguides/collection/protocol/ga4/sending-events?client_type=gtag#required_parameters).

- [x] adaptor logic
- [x] UI


Ziinc:
closing for now until we expand v2 backends

PR #1305, opened 2023-01-13
fix: move dev seeding to setup alias, remove runtime config defaults

Ziinc:
This PR removes config defaults, which should already be handled by config.exs and friends at compiletime. The defaults interfere with the nil filtering at runtime.

Also move db seeding to `mix setup`, which would only apply to the dev database now.


Ziinc:
merging in first to fix dev workflow.

PR #1306, opened 2023-01-16
fix: dev setup, more runtime configs

Ziinc:
Fixes the following:
- moves logging settings back to compile time, to prevent logger setting getting overwritten at runtime. Also correctly set log levels for each env.
- changes dev db name back to`logflare_dev`, as per phoenix conventions.
- update dev db seeding command to include distribution info.
- fixes misanmed env var, should be `LOGFLARE_MIN_CLUSTER_SIZE`
- removes `dev.secret.exs` import
- fixes default value setting for check_origin, which is messing with compile-time disabling for dev/test env
- remove `PHX_LIVE_VIEW_SIGNING_SALT` and shortened it to `PHX_LV_SIGNING_SALT`, sets a different signing salt for staging, which is does not need to be in secrets. Can still be overridden by env var if desired by users.
- correctly set `PHX_SECRET_KEY_BASE` for `.dev.env`


filipecabaco:
>moves logging settings back to compile time, to prevent logger setting getting overwritten at runtime. Also correctly set log levels for each env.

we should not do this since most envs will be compiled with prod but run with different settings (e.g. Docker locally)

> PHX_LV_SIGNING_SALT
I rather have a more descriptive name, we don't spend that many characters 😅 non-blocking thing

Everything else makes sense but the one for logging I would rather have it in the runtime since, as mentioned, we can expect most of the times to run a version that was compiled with MIX_ENV production which will be an issue when running locally since it will try to run log to `LogflareLogger.HttpBackend`

Ziinc:
Comments have been resolved, switched name of signing salt back. Agree on the logger, missed that. Moved http backend config to only be there when logflare backend url is set.

filipecabaco (approved):

PR #1307, opened 2023-01-16
Get Production pipeline changes ready

filipecabaco:
Replicate the changes made in staging onto our production env.


mix.exs
```diff
@@ -237,6 +237,14 @@ defmodule Logflare.Mixfile do
       "encrypt.staging": [
         "cmd gcloud kms encrypt --ciphertext-file='./.staging.env.enc' --plaintext-file=./.staging.env --location=us-central1 --keyring=logflare-keyring-us-central1 --key=logflare-secrets-key --project=logflare-staging",
         "cmd gcloud kms encrypt --ciphertext-file='./gcloud_staging.json.enc' --plaintext-file=./gcloud_staging.json --location=us-central1 --keyring=logflare-keyring-us-central1 --key=logflare-secrets-key --project=logflare-staging"
+      ],
+      "decrypt.production": [
```
Ziinc, line 241: lets do `decrypt.prod` and `encrypt.prod` to keep it consistent and short.

.github/workflows/pipeline.yml
```diff
@@ -148,3 +148,30 @@ jobs:
           project_id: 'logflare-staging'
       - name: 'Trigger Cloud Build'
         run: 'gcloud beta builds triggers run logflare-app-staging-trigger --branch=staging --format "value(name)"'
+  trigger_cloud_build_production:
+    name: Trigger Cloud Build in Production
+    if: github.ref == 'refs/heads/staging'
```
Ziinc, line 153: should be for `master`

.github/workflows/pipeline.yml
```diff
@@ -148,3 +148,30 @@ jobs:
...
+    needs:
+      - publish_latest
+      - publish_version
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
Ziinc, line 162: just checking, we need py for gcp sdk right?

.github/workflows/pipeline.yml
```diff
@@ -148,3 +148,30 @@ jobs:
...
+          export_environment_variables: true
+          cleanup_credentials: false
+      - name: 'Set up Cloud SDK'
+        uses: 'google-github-actions/setup-gcloud@v1'
+        with:
+          version: '413.0.0'
+          install_components: 'beta'
+          project_id: 'logflare-232118'
+      - name: 'Trigger Cloud Build'
+        run: 'gcloud beta builds triggers run logflare-master --branch=staging --format "value(name)"'
```
Ziinc, line 177: should be `master` branch

Ziinc: also, the cloudbuild trigger names between prod and staging don't seem to be consistent.

on another note, we can also remove the old base build triggers which are no longer used anymore.

Ziinc (changes requested):
spotted some misnamed branch reference

.github/workflows/pipeline.yml
filipecabaco: yes 👍 it's an annoyance 😓

.github/workflows/pipeline.yml
filipecabaco: will do that cleanup afterwards (aka after first deploy to prod)

Ziinc:
@filipecabaco Re the pushing of the versioned image to docker hub, it is OK to use semver to tag the image. During the sync we were referring to automating the version tag updating of the prod instance template as not worth automating.
Will do a proper review in a bit.

Ziinc (approved):

filipecabaco:
Had to do a change to `cloudbuild/prod/cloudbuild.yaml` to set build-arg so production only uses the versioned docker image

PR #1321, opened 2023-01-18
feat: config cat integration

Ziinc:
adds config cat for feature flagging, replaces multi-backend admin flag.

Updates secrets for all envs.


lib/logflare_web/templates/source/edit.html.eex
```diff
@@ -244,8 +244,8 @@
   <%= link "Setup custom backend", to: Routes.user_path(@conn, :edit) <> "#bigquery-backend", class: "btn btn-primary form-button", role: "button" %>
   <% end %>
 
-  <%= if @conn.assigns.user.admin do %>
-    <%= section_header("V2 Pipeline (Admin only)") %>
+  <%= if LogflareWeb.Utils.flag("multibackend") |> IO.inspect() == true do %>
```
filipecabaco, line 247:
```suggestion
  <%= if LogflareWeb.Utils.flag("multibackend") == true do %>
```

filipecabaco (approved):

PR #1334, opened 2023-01-24
Cloudbuild in prod only uses versioned docker image

filipecabaco:
To avoid issues of pushing unwanted revisions, production will only used the versioned image. The version is set using the VERSION file in the root of the project.

### DIFF
--- a/cloudbuild/prod/cloudbuild.yaml
+++ b/cloudbuild/prod/cloudbuild.yaml
@@ -18,22 +18,11 @@ steps:
       - --keyring=logflare-prod-keyring-us-central1
       - --key=logflare-prod-secrets-key
   - name: "gcr.io/cloud-builders/docker"
+    entrypoint: 'sh'
     args:
       [
-        "build",
-        "--build-arg",
-        "MAGIC_COOKIE",
-        "--build-arg",
-        "TAG_VERSION=1.0.0",
-        "-t",
-        "gcr.io/$PROJECT_ID/logflare_app:$COMMIT_SHA",
-        "-t",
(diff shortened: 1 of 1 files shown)


cloudbuild/prod/cloudbuild.yaml
```diff
@@ -18,20 +18,10 @@ steps:
       - --keyring=logflare-prod-keyring-us-central1
       - --key=logflare-prod-secrets-key
   - name: "gcr.io/cloud-builders/docker"
+    entrypoint: 'sh'
     args:
       [
-        "build",
-        "--build-arg",
-        "MAGIC_COOKIE",
```
Ziinc, line 25: We can pull out all of the MAGIC_COOKiE stuff as well. See [here](https://hexdocs.pm/mix/1.14/Mix.Tasks.Release.html)
We just need to set the `RELEASE_COOKIE` in `.env.*` and they will connect.

Ziinc: right now the MAGIC_COOKIE build arg is not even being used so we can remove it safely, the cluster is likely using the `release/COOKIE` file.

Ziinc: I had a look at the staging cloudbuild.yaml we can remove MAGIC_COOKIE_PREFIX as well (and the MAGIC_COOKIE in the staging env)

Ziinc (approved):

cloudbuild/prod/cloudbuild.yaml
filipecabaco: will remove

Ziinc:
@filipecabaco looks good, feel free to merge.

PR #1351, opened 2023-02-02
Add editor role to cloudbuild service account

filipecabaco:
We required this so we avoid an override of the IAM policies on our GCP Cloudbuild service account


lib/logflare/google/resource_manager.ex
```diff
@@ -123,6 +110,11 @@ defmodule Logflare.Google.CloudResourceManager do
         members: ["serviceAccount:#{env_service_account()}"],
         role: "roles/resourcemanager.projectIamAdmin"
       },
+      %Model.Binding{
+        condition: nil,
+        members: ["serviceAccount:#{env_cloud_build_sa()}"],
+        role: "roles/cloudbuild.builds.editor"
+      },
```
filipecabaco, line 117: I think this is the relevant change 😅 but had to clean up a bit for readability sake

Ziinc (approved):
changes look good 👍 , could you add a quick sanity check for the refactored parts? I dont think that there is test cov for ResourceManager yet

Ziinc:
i think let chase have a once over as well before merge

filipecabaco:
Added the requested tests after much fighting 😆

lib/logflare/google/resource_manager.ex
```diff
@@ -12,85 +12,75 @@ defmodule Logflare.Google.CloudResourceManager do
...
+      [%Model.Binding{members: members, role: "roles/bigquery.jobUser"}] ++
+        get_service_accounts()
+
+    policy = %Model.Policy{bindings: bindings}
+    body = %Model.SetIamPolicyRequest{policy: policy}
+
+    {function, arity} = __ENV__.function
+    caller = "#{function}" <> "_" <> "#{arity}"
+
+    case Api.Projects.cloudresourcemanager_projects_set_iam_policy(conn, env_project_number(),
```
Ziinc, line 52: it would be simpler to mock this function and assert on the received arguments in the mock

For example:
```
# in test
Api.Projects
|> expect(:cloudresourcemanager_projects_set_iam_policy, fn conn, proj_num-> 
  assert [%_{...} | _ ] = body.policy.bindings 
  {:ok, %{}}
end)
```
likewise for all the other `Api.Projects.*` apis.
Mocking the dep directly makes it easier to check that:
1. args are passed correctly
2. 3rd party func is called correctly
3. Tesla's not-so-nice mocking interface is not used

test/logflare/google/resource_manager_test.exs
```diff
@@ -0,0 +1,193 @@
...
+      end)
+
+      CloudResourceManager.set_iam_policy(async: false)
+    end
+  end
+
+  defp expected_url(%{project_number: project_number}),
+    do: "https://cloudresourcemanager.googleapis.com/v1/projects/#{project_number}:setIamPolicy"
+
+  defp expected_service_accounts(google_configs) do
```
Ziinc, line 45: since this is a sanity check, we don't need to assert on all the members (because we'd be essentially testing implementation in this case).

The main point of the test is to check that external iam policy is called.

it would suffice to check that service accounts length is `>1`, and maybe just assert on one or two special accounts are present.

The length of the test file is almost exceeding the length of actual implementation, which is a maintenance concern

test/logflare/google/resource_manager_test.exs
```diff
@@ -0,0 +1,193 @@
...
+  end
+
+  defp setup_test_state() do
+    stripe_id = "testing"
+    insert(:plan, name: "Free")
+    insert(:plan, name: "Legacy")
+    insert(:plan, name: "Lifetime")
+    insert(:plan, name: "Stripe", stripe_id: stripe_id)
+
+    _user_without_billing_account =
```
Ziinc, line 107: i think we can move all these to the Factory module for reuse

test/logflare/google/resource_manager_test.exs
filipecabaco: this tests that we're setting the correct service account roles hence this assertion because it is actually the thing we're changing with this added editor role 😅

lib/logflare/google/resource_manager.ex
filipecabaco: Plus the argument that this is testing the Google Library... damn it yeah this is way more correct with mimic... will change 👍

filipecabaco: I had to do the assertions on the stub otherwise it would call Tesla 🤷

chasers (approved):

chasers:
Looks good, let's just make sure to triple check the IAM list on staging when this is there.

PR #1356, opened 2023-02-09
(task) Add OAS3.0 to Endpoints Management API

filipecabaco:
Added Open API Specs to the Endpoints Management API as a first attempt on using open_api_spex as our doc generator.

Screenshot 2023-02-09 at 23 15 36


lib/logflare_web/schemas.ex
```diff
@@ -0,0 +1,60 @@
+defmodule LogflareWeb.Schemas do
```
Ziinc, line 1: Let's name this to OpenApiSchemas, using Schemas by itself is too ambiguous and can be mistaken for table schemas.

Ziinc (approved):
Looks good, just a minor module rename needed. 👍

PR #1357, opened 2023-02-11
(chore) Bump up asset dependencies

filipecabaco:
* Removed polyfill since it's deprecated
* Bumped up date-fns
* Remove node-fetch
* Fix run button on endpoints show
* Bump luxon version
* Update docs with `npm audit fix`


Ziinc (approved):
👍  will be dropping yarn usage soon to cut down our tooling deps.

PR #1368, opened 2023-02-16
(chore) Add Team OAS documentation

filipecabaco:
Added Open API Specs for Team Management endpoint.

### DIFF
--- a/lib/logflare_web/open_api_schemas.ex
+++ b/lib/logflare_web/open_api_schemas.ex
@@ -76,11 +76,63 @@ defmodule LogflareWeb.OpenApiSchemas do
     def params(), do: {"Source Params", "application/json", __MODULE__}
   end
 
-  defmodule SourceList do
+  defmodule User do
     require OpenApiSpex
 
-    OpenApiSpex.schema(%{type: :array, items: Source})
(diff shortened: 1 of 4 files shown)


lib/logflare_web/open_api_schemas.ex
```diff
@@ -76,11 +76,63 @@ defmodule LogflareWeb.OpenApiSchemas do
...
+        bigquery_dataset_location: %Schema{type: :string},
+        bigquery_dataset_id: %Schema{type: :string},
+        api_quota: %Schema{type: :integer},
+        company: %Schema{type: :string},
+        token: %Schema{type: :string}
+      },
+      required: [:name]
+    })
+
+    def response(), do: {"Team Response", "application/json", __MODULE__}
```
Ziinc, line 103: shouldn't this be "User Response"?

Ziinc (approved):
seems like a lot of boilerplate code, can consider writing a macro all the schema builderplate in the future.

lib/logflare_web/open_api_schemas.ex
filipecabaco: yes it should! thank you

filipecabaco:
I was thinking of doing that 👀 will tackle it when adding the next doc

PR #1370, opened 2023-02-16
(add) Stripe Mock server for dev environment

filipecabaco:
Adds Stripe Mock server for dev environment so we can develop billing features easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Sign up at https://logflare.app.
 1. Copy over secrets to two locations
    1. Dev secrets - `.dev.env`
    2. Google JWT key - `.gcloud.json`
-1. Start database `docker-compose up -d db`
+1. Start database and stripe mock `docker compose up -d db stripe-mock`
 1. Run `mix setup` for deps, migrations, and seed data.
 1. Run `(cd assets; yarn)` from project root, to install js dependencies
 1. Install `sqlparser` by following the steps in **Closed Source Usage** section.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,10 @@ services:
     volumes:
       - ./priv/wal.sql:/docker-entrypoint-initdb.d/wal.sql
       - pg-data:/var/lib/postgresql/data
+  stripe-mock:
+    image: stripe/stripe-mock:latest
+    ports:
+      - "12111:12111"
   logflare:
     build:
       dockerfile: ./Dockerfile
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -98,4 +98,8 @@ config :logflare, LogflareWeb.Plugs.EnsureSuperUserAuthentication,
 
 config :open_api_spex, :cache_adapter, OpenApiSpex.Plug.NoneCache
 
-import_config "telemetry.exs"
+config :stripity_stripe,
+  api_key: "sk_test_thisisaboguskey",
+  api_base_url: "http://localhost:12111/v1/"
+
+import_config("telemetry.exs")
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -9,22 +9,205 @@
 #
 # We recommend using the bang functions (`insert!`, `update!`
 # and so on) as they will fail if something goes wrong.
+plans = [
+  %{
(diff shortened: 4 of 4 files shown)


README.md
```diff
@@ -63,3 +63,12 @@ iex> LogflareLogger.info("testing log message")
 Use the `:error_string` metadata key when logging, which is for additional information that we want to log but don't necessarily want searchable or parsed for schema updating.
 
 For example, do `Logger.error("Some error", error_string: inspect(params) )`
+
+### Seed for development
+
+On your first run with `iex -S mix phx.server`:
+```
+Logflare.Repo.insert(Logflare.Billing.Plan.changeset(%Logflare.Billing.Plan{}, %{stripe_id: "1", price: 0, period: "month", name: "Free"}))
+Logflare.Repo.insert(Logflare.Billing.Plan.changeset(%Logflare.Billing.Plan{}, %{stripe_id: "2", price: 10, period: "month", name: "Metered BYOB"}))
```
Ziinc, line 72: Could we move this to the seed.exs? The script is already run on the setup alias, we can do this instead of directly inserting.

docker-compose.yml
```diff
@@ -10,6 +10,10 @@ services:
     volumes:
       - ./priv/wal.sql:/docker-entrypoint-initdb.d/wal.sql
       - pg-data:/var/lib/postgresql/data
+  stripe-mock:
```
Ziinc, line 13: Would relying on this affect tests?

An alternative approach is to have a mock cowboy server and add it to the application tree on test env.

Using this for dev env is fine as we want it as close to realistic as possible, but if this was the consideration then it makes more sense to just use the actual stripe api

Ziinc:
If the consideration is for full offline dev then it is a good addition, but then we rely on the image so you'd need to pull the image too beforehand.

README.md
filipecabaco: Will move

docker-compose.yml
filipecabaco: It won't affect tests since we're only setting the config for dev.

I wanted to avoid complex solutions and have something quick that is supported by Stripe.

Using it for dev only at the moment and it works for the main scenario which is "accessing the billing tab without errors". If we need more complex examples that require the real API then we can override with `runtime.exs` the base url.

filipecabaco:
We only need to pull it if doing work that requires calling Stripe so it won't affect day to day work. Plus, with `docker compose up -d db stripe` it becomes really easy to have everything up and running.

README.md
filipecabaco: Just checked and we're not running the seed.exs when running ecto.setup, in fact the seed.exs has price plans that do not get inserted

Ziinc: seeding is not automatically run on ecto.setup because we don't want to have the seeded plans for tests. the seeded plans are mostly for dev purposes only and are dumped from the prod db

Ziinc: the seed runs on `mix setup`, which is in the dev setup instructions

docker-compose.yml
Ziinc: ok sounds good 👍

mix.exs
```diff
@@ -224,7 +224,7 @@ defmodule Logflare.Mixfile do
       "lint.diff": ["credo diff master"],
       "lint.all": ["credo --strict"],
       "ecto.seed": ["run priv/repo/seeds.exs"],
-      "ecto.setup": ["ecto.create", "ecto.migrate"],
+      "ecto.setup": ["ecto.create", "ecto.migrate", "ecto.seed"],
```
Ziinc, line 227: this would seed twice when running `mix setup` for setting up the dev environment.

I was referring to moving the initial repo inserts that you had written to the seed.exs so that we can remove the raw inserts.

Ziinc (changes requested):

README.md
filipecabaco: Ok now I got it sorry. Will remove from the setup.

filipecabaco:
@Ziinc what is missing in this PR?

Ziinc (approved):

Ziinc:
looks good 👍

PR #1371, opened 2023-02-18
(add) Table to show estimate price and usage.

filipecabaco:
* Adds element to show usage and cost estimate
* Adds tests to BillingCounts
Screenshot 2023-02-20 at 12 06 18

### DIFF
--- a/test/logflare/billing/billing_counts_test.exs
+++ b/test/logflare/billing/billing_counts_test.exs
@@ -1,0 +1,71 @@
+defmodule Logflare.Billing.BillingCountsTest do
+  use Logflare.DataCase
+  alias Logflare.Billing.BillingCounts
+
+  setup do
+    user = insert(:user)
+
+    _non_matching_user = insert(:billing_counts, inserted_at: ~U[2023-02-01 00:00:00Z])
+
+    _outside_boundary = insert(:billing_counts, user: user, inserted_at: ~U[2023-01-31 00:00:00Z])
+
+    _outside_boundary = insert(:billing_counts, user: user, inserted_at: ~U[2023-03-01 00:00:00Z])
+
+    inside_boundary_0 = insert(:billing_counts, user: user, inserted_at: ~U[2023-02-17 00:00:00Z])
+
+    inside_boundary_1 = insert(:billing_counts, user: user, inserted_at: ~U[2023-02-17 00:00:00Z])
+
+    inside_boundary_2 = insert(:billing_counts, user: user, inserted_at: ~U[2023-02-10 00:00:00Z])
+
+    %{
+      user: user,
+      inside_boundary_0: inside_boundary_0,
+      inside_boundary_1: inside_boundary_1,
+      inside_boundary_2: inside_boundary_2
+    }
+  end
+
+  describe "timeseries/3" do
+    test "fetches billing counts for a certain period in a time series format", %{
+      user: user,
+      inside_boundary_0: inside_boundary_0,
+      inside_boundary_1: inside_boundary_1,
+      inside_boundary_2: inside_boundary_2
+    } do
+      result = BillingCounts.timeseries(user, ~U[2023-02-01 00:00:00Z], ~U[2023-02-28 00:00:00Z])
+
+      days =
+        Date.range(~D[2023-02-01], ~D[2023-02-28]) |> Enum.map(&DateTime.new!(&1, ~T[00:00:00]))
+
+      assert Enum.all?(result, fn [d | _] ->
+               Enum.find(days, &(DateTime.compare(&1, d) == :eq))
+             end)
+
+      expected_sum_day = inside_boundary_0.count + inside_boundary_1.count
+
+      assert result
+             |> Enum.find(fn [d | _] -> DateTime.compare(d, ~U[2023-02-17 00:00:00Z]) == :eq end)
+             |> then(fn [_, sum | _] -> sum end) == expected_sum_day
+
+      assert result
+             |> Enum.find(fn [d | _] -> DateTime.compare(d, ~U[2023-02-10 00:00:00Z]) == :eq end)
+             |> then(fn [_, sum | _] -> sum end) == inside_boundary_2.count
+    end
+  end
+
+  describe "cumulative_usage/3" do
+    test "returns cumulative total for a given period", %{
+      user: user,
+      inside_boundary_0: inside_boundary_0,
+      inside_boundary_1: inside_boundary_1,
+      inside_boundary_2: inside_boundary_2
+    } do
+      result =
+        BillingCounts.cumulative_usage(user, ~U[2023-02-01 00:00:00Z], ~U[2023-02-28 00:00:00Z])
+
+      expected = inside_boundary_0.count + inside_boundary_1.count + inside_boundary_2.count
+
+      assert result == expected
+    end
+  end
+end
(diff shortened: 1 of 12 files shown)


chasers (approved):

test/logflare/billing/billing_counts_test.exs
```diff
@@ -0,0 +1,71 @@
...
+    _outside_boundary = insert(:billing_counts, user: user, inserted_at: ~U[2023-01-31 00:00:00Z])
+
+    _outside_boundary = insert(:billing_counts, user: user, inserted_at: ~U[2023-03-01 00:00:00Z])
+
+    inside_boundary_0 = insert(:billing_counts, user: user, inserted_at: ~U[2023-02-17 00:00:00Z])
+
+    inside_boundary_1 = insert(:billing_counts, user: user, inserted_at: ~U[2023-02-17 00:00:00Z])
+
+    inside_boundary_2 = insert(:billing_counts, user: user, inserted_at: ~U[2023-02-10 00:00:00Z])
+
```
Ziinc, line 19: a little more verbose than necessary, 2 in boundary and 1 outside boundary would suffice.

test/logflare/billing/billing_counts_test.exs
```diff
@@ -0,0 +1,71 @@
...
+      inside_boundary_0: inside_boundary_0,
+      inside_boundary_1: inside_boundary_1,
+      inside_boundary_2: inside_boundary_2
+    } do
+      result = BillingCounts.timeseries(user, ~U[2023-02-01 00:00:00Z], ~U[2023-02-28 00:00:00Z])
+
+      days =
+        Date.range(~D[2023-02-01], ~D[2023-02-28]) |> Enum.map(&DateTime.new!(&1, ~T[00:00:00]))
+
+      assert Enum.all?(result, fn [d | _] ->
```
Ziinc, line 40: this block of assertions are a little too complex and hard to read.

if we have 1 out and 2 in, then we just need to assert that there is 2 dates returned with a sum of 3 counts.

test/logflare/billing/billing_counts_test.exs
```diff
@@ -0,0 +1,71 @@
...
+    test "returns cumulative total for a given period", %{
+      user: user,
+      inside_boundary_0: inside_boundary_0,
+      inside_boundary_1: inside_boundary_1,
+      inside_boundary_2: inside_boundary_2
+    } do
+      result =
+        BillingCounts.cumulative_usage(user, ~U[2023-02-01 00:00:00Z], ~U[2023-02-28 00:00:00Z])
+
+      expected = inside_boundary_0.count + inside_boundary_1.count + inside_boundary_2.count
```
Ziinc, line 66: it is fine to hardcode oure xpected to 3, for example.

test/logflare_web/live/billingaccount_live/estimate_usage_component_test.exs
```diff
@@ -0,0 +1,51 @@
...
+      expected_usage = billing_counts |> Enum.map(& &1.count) |> Enum.sum()
+      expected_cost = expected_usage * plan.price
+
+      result = render_component(EstimateUsageComponent, %{user: user, plan: plan})
+
+      assert result =~
+               "Estimate usage and cost between #{expected_presentation_start_date} and #{expected_presentation_end_date}:"
+
+      assert result =~ "Usage: #{expected_usage} inserts"
+      assert result =~ "Estimated Cost: #{expected_cost} USD"
```
Ziinc, line 48: don't need to test for exact text, can do a Regex match instead. main thing we're testing for is that estimated cost is presented, actual value has already been covered by other tests.
`~r/Estimated cost: [0-9]+ USD/` (and a similar one for inserts) would check that everything works with less setup needed.
What you're doing above is testing the implementation, when we'd want to instead test for end result (i.e. that something is rendered and the rendering doesn't break)

So for example, we can simplify the start/end date assertion to a regex test
```
assert Regex.match?(~r/Estimated usage and cost between .+ and .+/, result)
```
the actual calendar dates are irrelevant and would be testing for implementation.

Ziinc (approved):
👍 LGTM, just some minor comments on testing

mix.exs
```diff
@@ -227,7 +227,7 @@ defmodule Logflare.Mixfile do
       "lint.diff": ["credo diff master"],
       "lint.all": ["credo --strict"],
       "ecto.seed": ["run priv/repo/seeds.exs"],
-      "ecto.setup": ["ecto.create", "ecto.migrate"],
+      "ecto.setup": ["ecto.create", "ecto.migrate", "ecto.seed"],
```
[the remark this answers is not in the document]
filipecabaco: weird regression, probably rebased badly 🤦

test/logflare/billing/billing_counts_test.exs
filipecabaco: I want to test the cumulative nature of the time-series output

test/logflare/billing/billing_counts_test.exs
filipecabaco: I'm using random values for integers here

test/logflare_web/live/billingaccount_live/estimate_usage_component_test.exs
filipecabaco: We do have the cumulative usage tested elsewhere but the cost is not yet, that is done by getting the user's plan and multiplying on the view so that it isn't tested at the moment in any other spot except here and probably we should keep it that way for now to avoid more complexity

Ziinc: Then we should shift the cost calculation logic down from there view to the context instead, where it'll be easier to test and isolate.

The test logic is quite complex and hard to read in its current state, I'm mostly concerned about the test code maintenance. We should be testing for whether the view layer breaks, and not for biz logic which is already covered in the context. 

As this is a purely presentational liveview component, testing that text gets rendered and doesn't crash out is enough

test/logflare/billing/billing_counts_test.exs
Ziinc: Is it necessary to do so? Wouldn't 1 + 1 suffice for cumulation testing?

test/logflare_web/live/billingaccount_live/estimate_usage_component_test.exs
filipecabaco: honestly don't see this being used elsewhere but will change to be more inline with code expectations

test/logflare/billing/billing_counts_test.exs
filipecabaco: no since we need to test that we're accumulating same day counts and secondary day counts to see if the query is doing the sum as expected

test/logflare/billing/billing_counts_test.exs
Ziinc: 👌👌

test/logflare_web/live/billingaccount_live/estimate_usage_component_test.exs
Ziinc: It's fine, the table is a temp measure in any case, we can pick up billing ui improvements post LW

filipecabaco: ended up doing the change, didn't see this reply 🤦 😅

test/logflare/billing/billing_counts_test.exs
filipecabaco: I want to ensure that we're capturing the right date boundaries to avoid regressions

PR #1375, opened 2023-02-21
Reduce OAS generation boilerplate

filipecabaco:
* Create a common list schema from a given module with schema
* Create macro to reduce boilerplate for object schemas


Ziinc (approved):
very nice looks great 💯

PR #1392, opened 2023-03-03
Feat/self hosting supabase mode

Ziinc:
depends on #1382 

Adds in a `LOGFLARE_SUPABASE_MODE` flag, `LOGFLARE_API_KEY`, data seeding on startup.

Also makes config cat flagging return false by default if sdk key is not provided.

Adds auth adjustments to ensure that legacy `user.api_key` authentication is accepted by endpoints.

- [x] auth adjustments

Screenshot 2023-03-07 at 1 53 22 AM

Screenshot 2023-03-07 at 1 21 16 AM


lib/logflare_web/controllers/endpoints_controller.ex
```diff
@@ -16,16 +19,32 @@ defmodule LogflareWeb.EndpointsController do
...
-  def query(conn, %{"token" => token}) do
+  def query(%{assigns: assigns} = conn, %{"token" => token}) do
     endpoint_query = Endpoints.get_mapped_query_by_token(token)
+    user = Map.get(assigns, :user)
 
-    case Endpoints.run_cached_query(endpoint_query, conn.query_params) do
-      {:ok, result} ->
-        Logger.debug("Endpoint cache result, #{inspect(result, pretty: true)}")
-        render(conn, "query.json", result: result.rows)
+    with :ok <- check_auth(endpoint_query, user),
```
chasers, line 26: Is there a reason you didn't make a plug out of this? 

With a plug you don't have to worry about this logic in the controller, and it's what we do in all the other controllers unless we have to do something otherwise.

lib/logflare_web/controllers/plugs/verify_api_access.ex
```diff
@@ -4,43 +4,70 @@ defmodule LogflareWeb.Plugs.VerifyApiAccess do
...
     conn = fetch_query_params(conn)
     # fetch endpoint info
-    with endpoint_token <- conn.params["token"],
-         %Endpoints.Query{enable_auth: true, user_id: user_id} <-
-           Endpoints.get_query_by_token(endpoint_token),
-         {:ok, token} <- extract_token(conn),
-         {:ok, user} <- Auth.verify_access_token(token),
-         true <- user_id == user.id do
-      assign(conn, :user, user)
+    endpoint = Endpoints.get_query_by_token(conn.params["token"])
```
chasers, line 32: We should set the mapped query in the assigns once in the plug pipeline (probably in a different plug) and we'll have it in the controller. 

We're currently doing the db lookup twice because of this and then where we do `get_mapped_query_by_token` in the EndpointController.

lib/logflare_web/controllers/endpoints_controller.ex
Ziinc: reason why i moved down the resource ownership check to controller level was because the VerifyApiAccess was doing a token checks and didn't feel right to do ownership checks in there. I can move up the ownership check to a separate plug, and cover sources and controller as well.

lib/logflare_web/controllers/plugs/verify_api_access.ex
Ziinc: 👌 can do this in a resource fetching plug.

Ziinc:
Implemented a `FetchResource` plug and a `VerifyResourceOwnership` plug. Also enhanced the `VerifyApiAccess` plug to now implement the public/private scopes at the same time for mgmt api and the future user.api_key migration.

lib/logflare/single_tenant.ex
```diff
@@ -50,7 +93,11 @@ defmodule Logflare.SingleTenant do
   Creates an enterprise user
   """
   def create_default_user do
-    case Users.insert_user(@user_attrs) do
+    attrs =
+      @user_attrs
+      |> Map.put(:api_key, Application.get_env(:logflare, :api_key))
```
filipecabaco, line 98:
```suggestion
    attrs = Map.put(@user_attrs, :api_key, Application.get_env(:logflare, :api_key))
```
https://github.com/rrrene/credo/issues/221

lib/logflare_web/controllers/api/fallback_controller.ex
```diff
@@ -10,6 +10,12 @@ defmodule LogflareWeb.Api.FallbackController do
     |> json(%{errors: errors})
   end
 
+  def call(conn, {:error, :unauthorized}) do
+    conn
+    |> put_status(401)
+    |> halt()
```
filipecabaco, line 16: We should return an error message

lib/logflare_web/controllers/endpoints_controller.ex
```diff
@@ -16,14 +18,13 @@ defmodule LogflareWeb.EndpointsController do
...
-        render(conn, "query.json", result: result.rows)
+  def query(%{assigns: %{endpoint: endpoint}} = conn, _params) do
+    endpoint_query = Endpoints.map_query_sources(endpoint)
 
+    with {:ok, result} <- Endpoints.run_cached_query(endpoint_query, conn.query_params) do
+      Logger.debug("Endpoint cache result, #{inspect(result, pretty: true)}")
+      render(conn, "query.json", result: result.rows)
+    else
       {:error, err} ->
         render(conn, "query.json", error: err)
```
filipecabaco, line 29: This could be removed since the fallback controller will know how to handle this

lib/logflare_web/controllers/plugs/VerifyResourceOwnership.ex
```diff
@@ -0,0 +1,38 @@
...
+      when id == user_id,
+      do: conn
+
+  # halts all others
+  def call(%{assigns: assigns} = conn, _)
+      when is_map_key(assigns, :source) or is_map_key(assigns, :endpoint) do
+    # halt all unmatching
+    conn
+    |> put_status(401)
+    |> halt()
```
filipecabaco, line 33: We should return an error message

lib/logflare_web/controllers/plugs/verify_api_access.ex
```diff
@@ -4,43 +4,51 @@ defmodule LogflareWeb.Plugs.VerifyApiAccess do
...
-         {:ok, token} <- extract_token(conn),
-         {:ok, user} <- Auth.verify_access_token(token),
-         true <- user_id == user.id do
-      assign(conn, :user, user)
+  def call(conn, opts) do
+    opts = Enum.into(opts, %{scopes: []})
+    # generic access
+    with {:ok, user} <- identify_requestor(conn, opts.scopes) do
+      conn
+      |> assign(:user, user)
```
filipecabaco, line 26:
```suggestion
     assign(conn, :user, user)
```

lib/logflare_web/controllers/plugs/verify_api_access.ex
```diff
@@ -4,43 +4,51 @@ defmodule LogflareWeb.Plugs.VerifyApiAccess do
...
+
+      {:error, _} = err ->
+        # try to use legacy api_key
+        case Users.get_by(api_key: elem(extracted, 1)) do
+          %_{} = user when is_private_route? == false -> {:ok, user}
+          _ when is_private_route? == true -> {:error, :unauthorized}
+          _ -> err
+        end
+    end
   end
```
filipecabaco, line 53:
```suggestion
      {:error, _} = err -> handle_legacy_api_key(extracted, err)
    end
  end
  def handle_legacy_api_key(extracted, err) do
    case Users.get_by(api_key: elem(extracted, 1)) do
          %_{} = user when is_private_route? == false -> {:ok, user}
          _ when is_private_route? == true -> {:error, :unauthorized}
          _ -> err
    end
  end
```
This extracted to a function will make it easier to see that is an special case

filipecabaco (approved):
Minor readability stuff, overall looks good to me

lib/logflare_web/controllers/endpoints_controller.ex
Ziinc: lets do this change in a separate ticket & PR, currently we always return 200 for bigquery errors to bubble up, using fallback controller for this endpoint would result in status code change and api integration might go haywire.

Ziinc: ticket [created](https://www.notion.so/supabase/Endpoints-api-controller-error-response-consistency-9b4d4272a75c48dd838de72c12f60cde?pvs=4)

PR #1396, opened 2023-03-06
Feat/mgmt api verify api access

Ziinc:
Moves management api auth checking to common `VerifyApiAccess` plug, which will be the defacto api verification plug.

relies on #1392

### DIFF
--- a/lib/logflare_web/router.ex
+++ b/lib/logflare_web/router.ex
@@ -76,7 +76,7 @@ defmodule LogflareWeb.Router do
   end
 
   pipeline :require_mgmt_api_auth do
-    plug(LogflareWeb.Plugs.VerifyApiAccess)
+    plug(LogflareWeb.Plugs.VerifyApiAccess, scopes: ~w(private))
   end
 
   pipeline :require_auth do
--- a/test/support/conn_case.ex
+++ b/test/support/conn_case.ex
@@ -56,11 +56,11 @@ defmodule LogflareWeb.ConnCase do
       end
 
       # for api use
-      def add_access_token(conn, user) do
-        {:ok, access_token} = Logflare.Auth.create_access_token(user)
+      def add_access_token(conn, user, scopes \\ ~w(public)) do
+        scopes = if is_list(scopes), do: Enum.join(scopes, " "), else: scopes
+        {:ok, access_token} = Logflare.Auth.create_access_token(user, %{scopes: scopes})
 
-        conn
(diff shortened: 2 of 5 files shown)


lib/logflare_web/router.ex
```diff
@@ -74,7 +74,7 @@ defmodule LogflareWeb.Router do
   end
 
   pipeline :require_mgmt_api_auth do
-    plug(LogflareWeb.Plugs.SetVerifyUser)
+    plug(LogflareWeb.Plugs.VerifyApiAccess)
```
chasers, line 77: As I read it ... `VerifyApiAccess` will let ingest API keys in.

We should consider all `Logflare.User` `api_key`s exposed because they are used on in client side libs and potentially non-secure 3rd party forms (log drain forms).

Specific and obvious handling of an "ingest API key" vs a "management API key" is very important I think. 

I get the scent of RBAC here, which I like very much. But I'm not sure even with full RBAC we should ever let an API key have perms for ingest and management at the same time.

Ziinc: underestand, definitely would need to implement scope checking for access tokens. Had done the PR with the idea that we would do a db migration to copy all existing `user.api_key` values over to the access tokens table. 

Can add in scope checking as well as separate PR.

RE Scope checking logic, we can have a `public` scope for non-secure use only (ingest only), `private` scope for everything, and also fine grained scope checking `ingest|endpoint_query|management`  if no scope is set (which is the case for current existing access tokens) then we consider it as "public", making this compatible with the above migration path as well.

test/support/conn_case.ex
```diff
@@ -48,11 +48,20 @@ defmodule LogflareWeb.ConnCase do
...
         |> Plug.Test.init_test_session(%{user_id: user.id})
         |> Plug.Conn.assign(:user, user)
       end
+
+      # for api use
+      def add_access_token(conn, user) do
+        {:ok, access_token} = Logflare.Auth.create_access_token(user)
+
+        conn
+        |> put_req_header("authorization", "Bearer #{access_token.token}")
```
filipecabaco, line 63:
```suggestion
       put_req_header(conn, "authorization", "Bearer #{access_token.token}")
```

filipecabaco (approved):

Ziinc:
will  need to rebase on the self-hosted supabase mode before merge.

filipecabaco:
I know, the approval still stands 😄 just leave the PR ready to go

filipecabaco (approved):

PR #1397, opened 2023-03-06
Build Docker image for ARM platform

filipecabaco:
Create ARM image to be used on macos.

### DIFF
--- a/.github/workflows/arm-docker-latest-push.yml
+++ b/.github/workflows/arm-docker-latest-push.yml
@@ -1,0 +1,42 @@
+# This file will be removed when we have dedicated instances to reduce time to docker push
+# The platforms tag in the steps using docker/build-push-action@v3 will be moved to .github/workflows/docker-ci.yml
+
+name: Build Latest ARM Docker Image
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [staging]
+
+permissions:
+  contents: read
+
+jobs:
+  publish_latest:
+    name: Publish latest tagged image
+    runs-on: ubuntu-latest
+    env:
+      PUSH_TO_DOCKER: ${{github.ref == 'refs/heads/staging'}}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Get version
+        run: echo "LOGFLARE_VERSION=$(cat VERSION)" >> $GITHUB_ENV
+      - name: Get commit SHA
+        run: echo "GITHUB_SHA=$(git rev-parse HEAD)" >> $GITHUB_ENV
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/arm-docker-latest-push.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build Version
+        uses: docker/build-push-action@v3
+        with:
+          push: ${{env.PUSH_TO_DOCKER}}
+          tags: supabase/logflare:latest
```
Ziinc, line 36: usually `latest` is for the latest version image, lets use a different tag instead like `next` for both this workflow and the other workflow file for staging deploy.

Ziinc: also, should add the arm suffix, so `next-arm`

.github/workflows/arm-docker-versioned-push.yml
```diff
@@ -0,0 +1,40 @@
...
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build Version
+        uses: docker/build-push-action@v3
+        with:
+          push: ${{env.PUSH_TO_DOCKER}}
+          tags: supabase/logflare:${{ env.LOGFLARE_VERSION }}-arm
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          platforms: linux/arm64,linux/amd64
```
Ziinc, line 39: shouldn't we remove `linux/amd64` for both?

.github/workflows/arm-docker-latest-push.yml
```diff
@@ -0,0 +1,39 @@
...
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          platforms: linux/arm64,linux/amd64
```
Ziinc, line 39: see other comment

Ziinc (changes requested):
currently overwriting the latest tag without prefix

.github/workflows/arm-docker-latest-push.yml
filipecabaco: that is on purpose, if we have that we will need to have some docker compose weirdness depending on the platform of the user running it. 

Since we are not doing semver at the moment it's ok to override for now and later with dedicated machines and a more workable time to docker push we can move this into our main GH action flow.

Ziinc (approved):

Dockerfile
```diff
@@ -9,6 +9,7 @@ COPY . /logflare
 
 
 WORKDIR /logflare
+ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
```
Ziinc, line 12: @filipecabaco could you also add in an inline comment here to say why we need this in the image, so that we don't accidentally remove it when refactoring

PR #1400, opened 2023-03-15
(fix) Docker environment wasn't properly handlign docker hostname

filipecabaco:
The way we have our settings setup wouldn't work due to a bad hostname and a bad database name.

### DIFF
--- a/.docker.env
+++ b/.docker.env
@@ -1,6 +1,6 @@
 PHX_STATIC_PATH=/opt/app/rel/logflare/bin/priv/static
 DB_DATABASE=logflare_docker
-DB_HOSTNAME=localhost
+DB_HOSTNAME=db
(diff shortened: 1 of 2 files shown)


.docker.env
```diff
@@ -1,6 +1,6 @@
 PHX_STATIC_PATH=/opt/app/rel/logflare/bin/priv/static
 DB_DATABASE=logflare_docker
-DB_HOSTNAME=localhost
+DB_HOSTNAME=db
```
Ziinc, line 3: the db service in compose is exposed on host's `5432`, so both are fine.

Ziinc (approved):

.docker.env
filipecabaco: true but this way we are more explicit on what we're using

PR #1401, opened 2023-03-15
Feat/self hosting custom schema

Ziinc:
allows for creating a schema and setting a custom schema name for the search path, such that logflare will use that schema only. Allows for specifying a separate schema other than public.


docs/docs.logflare.com/docs/self-hosting.md
```diff
@@ -86,3 +86,9 @@ Allows you to pass in an API key that will used for authentication. This is inte
 > Boolean, defaults to false
 
 This is a special mode for Logflare which will seed additional resources for usage with Supabase self-hosted.
+
+### `DB_SCHEMA`
+
+> String, defaults to nil
+
+This ENV variable sets the search path to a custom database schema. THis allows you to customize the schema used on the database.
```
filipecabaco, line 94:
```suggestion
This ENV variable sets the search path to a custom database schema. This allows you to customize the schema used on the database.
```

PR #1404, opened 2023-03-17
feat: Logflare Alerts v0

Ziinc:
User can specify custom alert query to be triggered at set intervals (via cron). Any results returned will trigger an alert notification to be sent to specified notification channels.

- [x] crud 
- [x] one alert genserver per cluster, create/remove jobs on change
- [x] slack messenger
- [x] webhook messenger
- [x] restrict cron expressions (max 15 min intervals)
- [x] description
- [x] config cat user targeting

Screenshot 2023-09-21 at 6 07 08 AM


config/config.exs
```diff
@@ -116,4 +116,6 @@ config :logflare, Logflare.Cluster.Utils, min_cluster_size: 1
 
 config :grpc, start_server: true
 
+config :logflare, Logflare.AlertsScheduler, init_task: {Logflare.Alerting, :init_alert_jobs, []}
```
chasers, line 119: Citrine automatically warms with this `init_task` function when started?

lib/logflare/alerting.ex
```diff
@@ -0,0 +1,242 @@
...
+  @doc """
+  Performs the check lifecycle of an AlertQuery.
+
+  Send notifications if necessary configurations are set. If no results are returned from the query execution, no alert is sent.
+  """
+  @spec run_alert(AlertQuery.t()) :: :ok
+  def run_alert(%AlertQuery{} = alert_query) do
+    alert_query = alert_query |> Repo.preload([:user])
+
+    with {:ok, [_ | _] = results} <- execute_alert_query(alert_query) do
```
chasers, line 173: Ah nice so if it returns >0 results it'll alert with the results.

chasers: So what would my query be for say "send me an alert if my percentage of error logs was > 5% in the last 5 minutes?"

lib/logflare/alerting.ex
```diff
@@ -0,0 +1,242 @@
...
+      if alert_query.webhook_notification_url do
+        WebhookAdaptor.Client.send(alert_query.webhook_notification_url, %{
+          "result" => results
+        })
+      end
+
+      if alert_query.slack_hook_url do
+        SlackAdaptor.send_message(alert_query.slack_hook_url, results)
+      end
+
```
chasers, line 183: With the way this works we could actually replace the current counter based alerts with it as is long as we can query BigQuery enough. 

Eventually we could get a separate Postgres db (or Clickhouse) to store all logs only for the last ~48 hours which we could use to query more liberally.

chasers:
@Ziinc super clean as always!! As long as the tests pass let's it on prod and play with it.

lib/logflare/alerting.ex
```diff
@@ -0,0 +1,242 @@
...
+    with {:ok, transformed_query} <-
+           Logflare.Sql.transform(:bq_sql, alert_query.query, alert_query.user_id),
+         {:ok, %{rows: rows}} <-
+           Logflare.BqRepo.query_with_sql_and_params(
+             alert_query.user,
+             alert_query.user.bigquery_project_id || env_project_id(),
+             transformed_query,
+             [],
+             parameterMode: "NAMED",
+             maxResults: 1000,
```
chasers, line 233: Only thing is maybe limit maxResults to 5 because these will be displayed somewhere.  We currently limit the Slack alerts to the latest 5 log events because they're just too much usually.

lib/logflare/alerting.ex
```diff
@@ -0,0 +1,242 @@
...
+             location: alert_query.user.bigquery_dataset_location
+           ) do
+      {:ok, rows}
```
chasers, line 236: Oh ... not catching the error here. What happens when it crashes with Citrine? We likely don't want it retrying a bunch out of the gate.

Catch it and log it? Actually let's send the crash as an alert! 

"Your query crashed, it will get ran again next time. If it continues to crash you might need to change your alert query."

config/config.exs
chasers: What happens when multiple nodes start up and try to add the same job?

lib/logflare/backends/adaptor/slack_adaptor.ex
```diff
@@ -0,0 +1,72 @@
...
+
+  def to_markdown(%{} = row) do
+    bullet = "•"
+    indent = "    "
+
+    text =
+      for {key, value} <- Map.to_list(row) |> Enum.sort_by(fn {k, _v} -> k end) do
+        "#{key}: #{value}"
+      end
+      |> Enum.join("\r" <> indent)
```
filipecabaco, line 68:
```suggestion
    text =
      row
      |> Enum.sort_by({k,_} -> v end)
      |> Enum.map(fn {k, v} -> "#{k}:#{v}")
      |> Enum.join("\r" <> indent)
```

lib/logflare_web/live/alerts/components/alert_form.heex
```diff
@@ -0,0 +1,79 @@
+<div class="tw-flex tw-w-full tw-my-10 tw-gap-2">
```
filipecabaco, line 1: wrong file name, should be `lib/logflare_web/live/alerts/components/alert_form.html.heex`

lib/logflare_web/live/alerts/components/run_query_result.heex
```diff
@@ -0,0 +1,7 @@
+<div :if={@query_result_rows}>
```
filipecabaco, line 1: wrong file name, should be lib/logflare_web/live/alerts/components/run_query_result.html.heex

filipecabaco (commented):
🤯 this looks awesome!!!

I share some of the concerns Chase mentioned regarding job failure but overall this looks really good! 🚀

config/config.exs
chasers: All these jobs are distributed with mnesia is there a key to unique them or something?

lib/logflare/alerting.ex
Ziinc: Restricting it at the query level makes it much less flexible, as being able to send such events to a webhook can be quite and interesting use case.
Also, it's up to the user to fine tune their query to make the display optimal, and I would expect most queries to be doing aggregates.

We could add in more docs and best practices around it instead

config/config.exs
Ziinc: yes the `init_task` is triggered for each node startup. using `put_job/1` would update the job completely, including changing the task value.

https://hexdocs.pm/citrine/Citrine.Scheduler.html#c:put_job/1

The `:id` key is used for uniqueness.

Did some digging, and what happens when `put_job/1` is triggered is the distributed process' state will be updated. It would be good if we can avoid the update, as it is linear time complexity to do the update:

https://github.com/brndnmtthws/citrine/blob/master/lib/citrine/job_executor.ex#L158

A few possible strategies is:
1. storing last_init times on db, and having a staleness threshold. The `put_job/1` would only then be executed if it is stale.
2. upstreaming some changes to the library to allow separate of create/update as well as adding in last initialized timestamps to the stored Citrine.Job.

option 2 is more of a long run optimization, and option 1 is likely the simpler way to go. 

Another optimization is use Repo.stream() and startup the alerts in batches. Since this stream would be long running and has linear time complexity, we can add in some checks to ensure that only one node is performing the full table stream of the `init_task`. So that would involve:

- having a genserver that would store the state of whether this initializaiton is taking place
- on startup, the init_task checks across the cluster if an initialization is running. If already running, immediately exit the task. If not, update state and begin the Repo.stream transaction.

lib/logflare/alerting.ex
Ziinc:
```sql
select
t.request_url,
 countif(t.error = 'error') as errors
`my.logs` t
where timestamp_sub(current_timestamp(), INTERVAL 5 MINUTE)
group by t.request_url
having ( countif(t.error = 'error') / count(t.id) ) > 0.05
```
Something like this

lib/logflare/alerting.ex
Ziinc: Yup, though the current RecentLogsServer manages this it with some extra logic in there (like touching the sources with timestamps for last ingested timestamps).

Def like the idea of being able to direct these kind of alerts/querying to separate backends, the multi-backends flexibility should apply to this too.

lib/logflare/alerting.ex
Ziinc: We can handle do the error handling in the error handling as a task.
I do like the alert notifying, definitely would be helpful for debugging.
Sending the alert message would be good as well, though it might lead to noise on some types of errors such as connection issues, which might result in all alerts spamming their channels.

Ziinc: Since the max interval is 15 mins, i think it would be fine in such a case. It would be quite (or rather very) unlikely for every single alert to both execute at the same time and error out. 🤔

config/config.exs
Ziinc: Task for stalekenss check and streaming https://www.notion.so/supabase/Logflare-Alerts-optimize-init_task-for-loading-Citrine-Jobs-using-staleness-checks-a214a7aa7f1c45929bac25466e1721c9?pvs=4

lib/logflare/alerting.ex
Ziinc: TAsk https://www.notion.so/supabase/Logflare-Alerts-Error-handling-for-queries-with-errors-send-alert-to-user-061bb3f283bc405a8868e9505a34da9f?pvs=4

config/config.exs
chasers: Yeah `last_init` seams reasonable for now.

lib/logflare/alerting.ex
chasers: Maybe just make it overridable like we do with Endpoints and default it to 5 on the form. 

I guarantee the first alert someone gets with 1000 things is going to be WTF because it happened with the current alerts.

If they want more they can change it there.

chasers (approved):

chasers:
Overall tho, assuming tests pass I'm good with it!

PR #1410, opened 2023-03-22
feat: query endpoints by name

Ziinc:
adds a new api route `/api/endpoints/query/name/:name` for an endpoint by name. Requires an api key to be provided, and does not apply to endpoints where auth is disabled `enable_auth=false`

### DIFF
--- a/lib/logflare_web/controllers/plugs/fetch_resource.ex
+++ b/lib/logflare_web/controllers/plugs/fetch_resource.ex
@@ -14,22 +14,23 @@ defmodule LogflareWeb.Plugs.FetchResource do
   alias Logflare.Endpoints
   def init(_opts), do: nil
 
-  def call(%{assigns: %{resource_type: type}} = conn, _opts) when is_atom(type) do
-    conn = fetch_query_params(conn)
+  def call(%{assigns: %{resource_type: :source}, params: %{"source" => token}} = conn, _opts) do
+    source = Sources.get_source_by_token(token)
+    assign(conn, :source, source)
+  end
 
-    resource =
-      case type do
-        :source ->
-          token = Map.get(conn.params, "source")
-          Sources.get_source_by_token(token)
+  def call(%{assigns: %{resource_type: :endpoint}, params: %{"token" => token}} = conn, _opts) do
+    endpoint = Endpoints.get_query_by_token(token)
+    assign(conn, :endpoint, endpoint)
+  end
 
-        :endpoint ->
-          token = Map.get(conn.params, "token")
-          Endpoints.get_query_by_token(token)
-      end
-
-    conn
-    |> assign(type, resource)
+  def call(
+        %{assigns: %{resource_type: :endpoint, user: %{id: user_id}}, params: %{"name" => name}} =
+          conn,
+        _opts
+      ) do
+    endpoint = Endpoints.get_by(name: name, user_id: user_id)
+    assign(conn, :endpoint, endpoint)
   end
 
   def call(conn, _), do: conn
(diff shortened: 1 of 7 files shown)


lib/logflare_web/controllers/plugs/verify_resource_ownership.ex
```diff
@@ -11,23 +11,24 @@ defmodule LogflareWeb.Plugs.VerifyResourceOwnership do
   alias LogflareWeb.Api.FallbackController
   def init(_opts), do: nil
 
-  # no user set. (handles endpoints with no auth)
-  def call(%{assigns: assigns} = conn, _opts) when is_map_key(assigns, :user) == false, do: conn
+  def call(%{assigns: %{endpoint: %Query{enable_auth: false}}} = conn, _opts) do
+    conn
```
filipecabaco, line 15: we should rethink this at a later stage, having the fallback controller handle scenarios like this will be a tricky thing to debug at a later stage.

Fallback controllers should only handle errors since even their catch all behavior is already risky enough 😓 

But it can be a future us problem

lib/logflare_web/controllers/plugs/fetch_resource.ex
```diff
@@ -14,22 +14,23 @@ defmodule LogflareWeb.Plugs.FetchResource do
...
 
-  def call(%{assigns: %{resource_type: type}} = conn, _opts) when is_atom(type) do
-    conn = fetch_query_params(conn)
-
-    resource =
-      case type do
-        :source ->
-          token = Map.get(conn.params, "source")
-          Sources.get_source_by_token(token)
+  def call(%{assigns: %{resource_type: :source}, params: %{"source" => token}} = conn, _opts) do
```
filipecabaco, line 17: I feel that all of this is doing controller-level work. Plugs should be naive from the business domain and this is injecting a lot of business logic into them.

filipecabaco (approved):
I've approved but we need to rethink a couple of things IMO:
* Having plugs fetching data like we're doing feels like a smell, they should be as agnostic as possible from business logic but this approach it feels that we're doing "controller" work at the middleware layer.
* Fallback controller should handle errors/failures with their catch-all behavior and that doesn't seem to be the case here where we are just returning the conn without any action to handle said failures. Even if they bubbled up from other failure conditions we should have that be reflected in the fallback controller

All of this is discussions we can have async at a different time.

lib/logflare_web/controllers/plugs/verify_resource_ownership.ex
Ziinc: something looks off with the module and the file name, i'll check this. This logic isn't for fallback controller

Ziinc: This isn't the fallback controller, it is a separate plug that is checking for ownership. We do a passthrough if we don't need to check., which in thie case is when auth is disabled.

filipecabaco: 🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️ my bad that was stupid..

Ziinc:
Fair points, however we also need to minimize the data fetching, hence why they are placed in the pipeline and fetched, so that we don't have to do another fetch downstream. This was something that @chasers had pointed out in a previous PR, hence why we have this plug-fetching setup for a 1 time resource fetch.

Fundamentally there is no issue in placing the plugs in the controller instead of the router if it is easier to think about it on a controller basis, just that it is simpler to encapsulate the logic in a plug and apply them to many routes/scopes via a pipeline. Pushing the logic down into the controller action results in verbosity and duplicate logic across controllers.

filipecabaco:
The issue is that it gets really tricky to understand where stuff is being assigned.

Instead of searching for the controller for a specific route and you're done you need to check for the controller and all the plugs it goes through to potentially debug issues.

Verbosity is a tricky measure, sometimes it's better to have more code but be more explicit in the intention vs have implicit behavior scattered

PR #1413, opened 2023-03-23
chore: expose more ports for cli

Ziinc:
for cli testing

Replication:
1. build local docker image using `docker-compose build` without the ports exposed.
2. add the following line to the cli logflare service startup [here](https://github.com/supabase/cli/commit/573f69c7baa04bcdb3b11ce42c807ec53ec40850)
```
PortBindings:  nat.PortMap{"4002/tcp": []nat.PortBinding{{HostPort: strconv.FormatUint(4002, 10)}}},
```
3. Run the cli `INTERNAL_IMAGE_REGISTRY=docker.io go run . start`
4. Verify that there is no port mapping `docker-compose ps`. The supposedly exposed port will also not be available on localhost `https://localhost:4002`
5. Repeat steps 1-4 with the expose ports, and it will now bind.

On a separate note, it seems like all images that ports via the cli have an `EXPOSE` in their dockerfiles. Might be an issue with the docker api.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,5 +40,11 @@ RUN cp -r /opt/app/rel/logflare/bin/priv/static /opt/app/rel/logflare/lib/logfla
 RUN rm -r /opt/app/rel/logflare/bin/priv
 
 WORKDIR /opt/app/rel/logflare/bin
+
+# expose specific ports, because for some reason
+# when using docker api with supabase-cli, port bindings don't work without exposing them first
+EXPOSE 4000
+EXPOSE 4002
+
 COPY run.sh ./run.sh
 CMD ["sh", "run.sh"]


Dockerfile
```diff
@@ -40,5 +40,11 @@ RUN cp -r /opt/app/rel/logflare/bin/priv/static /opt/app/rel/logflare/lib/logfla
 RUN rm -r /opt/app/rel/logflare/bin/priv
 
 WORKDIR /opt/app/rel/logflare/bin
+
+# expose specific ports, because for some reason
+# when using docker api with supabase-cli, port bindings don't work without exposing them first
+EXPOSE 4000
+EXPOSE 4002
```
filipecabaco, line 47: we should be able to link them with port mapping from docker vs exposing another port 🤔

Ziinc: I've added in replication steps, let me know if you can replicate it.

filipecabaco:
I'm going to merge for the sake of creating a release so I can move with the vector changes.

Will try to replicate the steps later when we have more time since I still need to do the changes on the CLI ingestion side and will tackle it at that point in time

Ziinc:
Spoke with qiao and we can revert this case as it can be handled on cli side with [this](https://github.com/supabase/cli/pull/926/files#diff-1af1ea8d4dac7cf62987c89e0f1036ad192403d7135f445d719993f0f5303a96R486)

PR #1414, opened 2023-03-23
Fix/mailer fix for schema update single tenant

Ziinc:
if mailer api key is not present, it causes the schema update to crash, resulting in the source schema not updating.


filipecabaco (changes requested):
We can change `config :swoosh, local: <value>` to be configured from runtime.exs which removes all this challenges and allows people to check sent emails with swoosh mailbox

Ziinc:
ok switched to making it conditional config at runtime

PR #1415, opened 2023-03-23
feat: supabase mode schema update on startup

Ziinc:
On startup, the sources are created without schemas. However, this affects endpoint querying via studio as there are not schemas created on startup.

Hence, to trigger the schema update logic, we trigger some ingests based on samples from prod, but only after the bigquery tables have init (otherwise no schema change will be registered)


lib/logflare/single_tenant.ex
```diff
@@ -163,4 +169,37 @@ defmodule Logflare.SingleTenant do
...
     end
   end
+
+  def supabase_mode? do
+    if Application.get_env(:logflare, :supabase_mode) do
+      true
+    else
+      false
+    end
+  end
```
filipecabaco, line 179:
```suggestion
  def supabase_mode?, do: !! Application.get_env(:logflare, :supabase_mode)
```

Maybe a bit to similar to JS but does cut lines of code 😅

test/logflare/single_tenant_test.exs
```diff
@@ -60,6 +61,13 @@ defmodule Logflare.SingleTenantTest do
   describe "supabase_mode=true" do
     TestUtils.setup_single_tenant(seed_user: true, supabase_mode: true)
 
+    setup do
+      Logflare.Logs
+      |> stub(:ingest_logs, fn _batch, _source -> :ok end)
```
filipecabaco, line 66:
```suggestion
      stub( Logflare.Logs, :ingest_logs, fn _batch, _source -> :ok end)
```

test/logflare/single_tenant_test.exs
```diff
@@ -70,9 +78,16 @@ defmodule Logflare.SingleTenantTest do
     end
 
     test "startup tasks inserts log sources/endpoints" do
-      Logflare.Application.startup_tasks()
+      Logs
+      |> expect(:ingest_logs, 9, fn _batch, _source -> :ok end)
```
filipecabaco, line 82:
```suggestion
      expect(Logs, :ingest_logs, 9, fn _batch, _source -> :ok end)
```

filipecabaco (approved):
Readability suggestions

Maybe in the future when we are json schema driven we could recheck this to create the schemas based on the json schemas vs dummy data 👀

PR #1425, opened 2023-03-29
(fix) Moves permission request to parent workflow

filipecabaco:
Moves permissions request to parent workflow so mirror and docker ci can use this higher permissions token

https://github.com/Logflare/logflare/actions/runs/4554988706


Ziinc:
overlaps with https://github.com/Logflare/logflare/pull/1426

Ziinc:
closing first

PR #1436, opened 2023-04-03
fix: cut down schema update wait time by making init_table async

Ziinc:
Drops init wait time by around 10s by making BQ table initialization async. Kept a 3s buffer wait time in case of user latency issues or slow compute. In my local testing, 2s worked fine, but will leave it in to minimize race conditions.


filipecabaco (approved):

chasers (approved):
All good!

PR #1454, opened 2023-04-05
add: Set the type of authentication used by Management API

filipecabaco:
Sets the type of token expected to use the Management API


Ziinc (approved):
LGTM

PR #1455, opened 2023-04-05
fix: Cloudbuild script

filipecabaco:
Currently we're still using the inline trigger instead of the file to build the final docker image.

This will allow us to use the file in our repo vs inlined


Ziinc (approved):
LGTM

PR #1466, opened 2023-04-27
add - OTEL ingestion

filipecabaco:
Handle Open Telemetry data by using OTLP Collector protobufs


lib/logflare_grpc/endpoint.ex
```diff
@@ -0,0 +1,5 @@
+defmodule LogflareGrpc.Endpoint do
+  use GRPC.Endpoint
+
+  run(LogflareGrpc.Trace.Server, interceptors: [{GRPC.Logger.Server, level: :info}])
```
Ziinc, line 4: log level should match the log level env var. looks like this is a macro, so the value might not match the runtime log level config.

lib/logflare_grpc/trace/transform.ex
```diff
@@ -0,0 +1,100 @@
...
+  Converts a GRPC Trace to a Logflare Event
+  """
+
+  alias Opentelemetry.Proto.Trace.V1.ResourceSpans
+  alias Opentelemetry.Proto.Common.V1.KeyValue
+  alias Logflare.LogEvent
+  alias Logflare.Source
+
+  @spec to_log_events(list(ResourceSpans.t()), Source.t()) :: list(LogEvent.t())
+  def to_log_events(resource_spans, source) do
```
Ziinc, line 12: `@doc`

test/logflare_grpc/trace/transform_test.exs
```diff
@@ -0,0 +1,61 @@
...
+      le_event = Enum.find(log_events, fn le -> le.body["metadata"]["type"] == "event" end)
+
+      assert le_event.body["event_message"] == event.name
+      assert le_event.body["parent_span_id"] == Base.encode16(span.span_id)
+
+      assert DateTime.from_unix(le_event.body["timestamp"], :microsecond) ==
+               DateTime.from_unix(event.time_unix_nano, :nanosecond)
+    end
+
+    test "json parsable log event body", %{
```
Ziinc, line 53: if it is a map, it would be json parsable. this test can be removed.

Ziinc:
nice one, looks great overall 👍

lib/logflare_grpc/endpoint.ex
filipecabaco: Will check what I can do

lib/logflare_grpc/trace/transform.ex
```diff
@@ -0,0 +1,100 @@
...
+
+  defp handle_scope_span(%{scope: scope, spans: spans}, resource) do
+    %{name: name, version: version, attributes: attributes} = scope
+    resource = Map.merge(resource, %{"name" => name, "version" => version})
+    resource = Map.merge(resource, handle_attributes(attributes))
+
+    Enum.map(spans, &handle_span(&1, resource))
+  end
+
+  def handle_span(span, resource) do
```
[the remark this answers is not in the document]
filipecabaco: actually this should be a defp, will change

test/logflare_grpc/trace/server_test.exs
```diff
@@ -0,0 +1,83 @@
...
+      reject(Logs, :ingest, 1)
+
+      assert {:error, %GRPC.RPCError{message: "Invalid API Key or Source ID"}} =
+               Stub.export(channel, request)
+    end
+
+    test "returns an error if missing x-source-id header", %{
+      user: user
+    } do
+      headers = [{"x-api-key", user.api_key}]
```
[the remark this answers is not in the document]
filipecabaco: I was using access_tokens but from my understanding, we're still using a api_key in most integrations right?

test/logflare_grpc/trace/transform_test.exs
```diff
@@ -0,0 +1,61 @@
...
+    setup do
+      user = build(:user)
+      source = insert(:source, user: user)
+      %{resource_spans: TestUtils.random_resource_span(), source: source}
+    end
+
+    test "Creates LogEvent from a list with one Resource Span that contains an Event", %{
+      resource_spans: resource_spans,
+      source: source
+    } do
```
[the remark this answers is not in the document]
filipecabaco: here I'm testing that the received struct is properly converted into the correct amount and type of LogEvent (one span can have multiple events for example)

as such it's important to test if each type of event has the expected field values otherwise it will break our next steps of the implementation

test/support/test_utils.ex
```diff
@@ -374,4 +385,94 @@ defmodule Logflare.TestUtils do
...
+  def random_resource_span do
+    [
+      Opentelemetry.Proto.Trace.V1.ResourceSpans.new(
+        resource: Resource.new(),
+        scope_spans: random_scope_span()
+      )
+    ]
+  end
+
+  defp random_scope_span do
```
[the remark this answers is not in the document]
filipecabaco: Can do 👍

lib/logflare_grpc/endpoint.ex
filipecabaco: Yeah, during compilation there's no way but there will be a release later of grpc with telemetry and then we can plug into that for this type of information.

Honestly, this was only used for debugging, it doesn't offer much now.

PR #1468, opened 2023-05-01
chore: add erlang to asdf deps.

Ziinc:
Contributors would need to install the correct otp version with:
```bash
asdf install
```

This fixes the deps bump, which now requires otp 25.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,4 @@
 elixir 1.14.4-otp-25
 nodejs 18.13.0
 rust 1.64.0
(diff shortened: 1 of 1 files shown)


filipecabaco:
This is covered by `elixir 1.14.4-otp-25`, do we need this?

Ziinc:
Screenshot 2023-05-08 at 10 58 53 PM


asdf elixir doesn't include erlang

filipecabaco:
TIL!

PR #1497, opened 2023-05-07
fix: conditionally run grpc server if port is provided

Ziinc:
When `LOGFLARE_GRPC_PORT` is not set, it will result in server crash.  This makes the GRPC usage opt-in, and would not crash the standalone image when config is not provided.

runtime.exs parsing error:
Screenshot 2023-05-07 at 7 02 12 PM

application startup error:
Screenshot 2023-05-07 at 7 02 02 PM


Ziinc:
covered by @filipecabaco 's PR

PR #1500, opened 2023-05-10
fix - Send grpc trailer headers and change to use streams vs unary requests

filipecabaco:
Due to a timeout in staging when fetching trailers from a unary request, we'll implement streams to see if this mitigates the issue.

We'll need to improve error handling also since a failed span insert should return a better error to the consumer.

Note: Generated protobufs do not have streams and that change had to be mad by hand

### DIFF
--- a/lib/logflare_grpc/trace/server.ex
+++ b/lib/logflare_grpc/trace/server.ex
@@ -13,15 +13,21 @@ defmodule LogflareGrpc.Trace.Server do
 
   require Logger
 
-  def export(%{resource_spans: resource_spans}, stream) do
+  def export(events, stream) do
     with %{"x-api-key" => api_key, "x-source-id" => source_token} <-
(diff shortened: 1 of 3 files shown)


lib/logflare_grpc/trace/server.ex
```diff
@@ -13,15 +13,21 @@ defmodule LogflareGrpc.Trace.Server do
 
   require Logger
 
-  def export(%{resource_spans: resource_spans}, stream) do
+  def export(events, stream) do
     with %{"x-api-key" => api_key, "x-source-id" => source_token} <-
            GRPC.Stream.get_headers(stream),
          user when not is_nil(user) <- Users.get_by(api_key: api_key),
```
Ziinc, line 19: should not be using the `api_key` field, should abtstract and/or reuse out the logic in the VerifyApiAccess plug so that we can use both access tokens as well as api_key field.

PR #1509, opened 2023-05-11
add - Padding to tooltip; Minor env setup changes

filipecabaco:
* Adding padding to tooltip.
* Default a port to GRPC server
* Add proper port to docker .env file


assets/js/source_log_chart.jsx
```diff
@@ -76,7 +76,7 @@ const renderElixirLoggerTooltip = ({ data, color }) => {
     { c: brandGray, p: "other", t: "Other" },
   ]
   return (
-    <div style={{ backgroundColor: brandLightBlack }}>
+    <div style={{ backgroundColor: brandLightBlack, padding: "6px" }}>
```
Ziinc, line 79: This is fine, but can be replaced with a `tw-p-2` class

Ziinc (approved):

assets/js/source_log_chart.jsx
filipecabaco: would this work? since this is actually changing the style instead of the class I avoided tailwind

PR #1522, opened 2023-05-15
fix: retention date for team tier

kevcodez:
Let me know if there is anything else we need to adjust


kevcodez:
Might still make sense for consistency? Feel free to close otherwise - where do we need to make adjustments? Solely in Logflare through the UI? Any repo changes necessary?

Ziinc:
changes are only needed via logflare ui. we're going to script the endpoints exports so that it will be more consistent, but tiering will not affect self-hosted/cli setup

PR #1523, opened 2023-05-16
add - Production certificates required for GRPC server

filipecabaco:
Adds the required files so the GRPC server can properly connect to clients

Also adds instructions on how to setup everything to support it


DEVELOPMENT.md
```diff
@@ -76,3 +76,36 @@ Version bumping policy is as follows:
...
+
+- **Cloudflare:** Enable GRPC in the `Network` tab
+- **Cloudflare:** Generate a CA Certificate to be used on your Origin server:
+  - SSL/TLS -> Origin Server -> Create Certificate -> Create
+  - Create .`<env>`.cacert.pem with the content from the certificate field
+  - Create .`<env>`.cacert.key with the content from the key field
+- **Local:** Generate a self signed certificate for the origin server:
+  - `openssl req -newkey rsa:2048 -nodes -days 365000 -keyout .<env>.cert.key -out .<env>.req.pem` and set the email to your support email
+  - `openssl x509 -req -days 12783 -set_serial 1 -in .<env>.req.pem -out .<env>.cert.pem -CA .<env>.cacert.pem -CAkey .<env>.cacert.key`
+  - **Be extremely careful with the generated files**
```
Ziinc, line 94: can remove this disclaimer, they should know what they are doing.

DEVELOPMENT.md
```diff
@@ -76,3 +76,36 @@ Version bumping policy is as follows:
 Use the `:error_string` metadata key when logging, which is for additional information that we want to log but don't necessarily want searchable or parsed for schema updating.
 
 For example, do `Logger.error("Some error", error_string: inspect(params) )`
+
+## Deployment
+
+### GRPC Server
```
Ziinc, line 82: can we move this text block to the docs self-hosted.md file instead? this looks like self-hosting setup information. We can have a section called `## GRPC` and a subheader `### Example GRPC Setup on GCP`


For the DEVELOPMENT.md just need to note down the GRPC configs that we are using

Ziinc (approved):

DEVELOPMENT.md
Ziinc: we can slim down steps to just what our setup requires, though it isn't really development specific. would be more appropriate to go into our deployment notion doc actually.

filipecabaco: This is what our setup requires 😅 no slimming down of the steps

As for moving it out can I would rather start building a Deployment.md so we move that information out of Notion and closer to the code

Ziinc: Yup that sounds good

filipecabaco: Moved this to DEPLOYMENT.md . The docs in Notion are outdated so we can work on this document instead as we do the changes (e.g. moving to A B deployment)

PR #1539, opened 2023-05-19
fix: fully qualified names cte fixes

Ziinc:
This fixes the fully qualified naming issue where sources with the same naming pattern as bigquery entities were getting disallowed.

This PR changes the logic and takes a different approach. Instead of explicitly checking for disallowed specific patterns  (i.e. disallow-listing), we reverse the checking logic to allow-list certain patterns and treat everything else as an unknown entity. This makes the logic much easier to read and less prone to bugs, since the allow-list conditions are very clear.

I have added in tests for naming patterns that are similar to the bigquery entity naming pattern to ensure we transform everything correctly.

Included in this PR is also additional tests for single tenant mode and fully-quallified names, ensuring that users that set a global bigquery project will be able to reference entities within that project **ONLY** in single tenant mode.

### DIFF
--- a/lib/logflare/sql_v2.ex
+++ b/lib/logflare/sql_v2.ex
@@ -6,6 +6,7 @@ defmodule Logflare.SqlV2 do
   """
   alias Logflare.Sources
   alias Logflare.User
+  alias Logflare.SingleTenant
   alias __MODULE__.Parser
 
   @doc """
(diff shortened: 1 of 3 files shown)


chasers:
Based on the discussion we're having in Slack, instead of doing this we should take in GCP credentials and do queries with that.

chasers:
I think this is safe for now to get it deployed, but we need to follow this up with bringing your own service account with your own backend.

chasers (approved):

PR #1548, opened 2023-05-29
fix: fixes for billing accounts live page, with additional test coverage

Ziinc:
Fix for lingering leex usage in BillingAccountsLive, which is causing billing page to error out.


Ziinc:
@filipecabaco could you help with verification for this bug on staging/prod 🙏

filipecabaco (approved):

filipecabaco:
We're lacking a lot of testing to check the components and their interactions with the main code but we can tackle it later.

PR #1552, opened 2023-06-01
fix: oauth2 callback redirect fix

Ziinc:
This PR adds additional test coverage for the Oauth2 bug when cloudflare app redirects. 

Fix is implemented in our fork of PhoenixOauth2Provider, whose tests does not account for Application level configs.
The conn.private phoenix_view format had changed in Phoenix 1.7 and this had resulted in a breakage in behaviour.

the bugfix as implemented  in [this commit](https://github.com/Logflare/phoenix_oauth2_provider/commit/36e757920e54bffa0f603bf985aff91bb8b28f2d) essentially new pattern match for the new phoenix version, while retaining the old match format for older versions.

We can PR to upstream as well, but upstream seems unmaintained at the moment and it would only affect projects >= phoenix 1.7, which is not supported yet on upstream.

### DIFF
--- a/test/logflare_web/controllers/oauth_controller_tests.exs
+++ b/test/logflare_web/controllers/oauth_controller_tests.exs
@@ -12,4 +12,27 @@ defmodule LogflareWeb.OAuthControllerTests do
       assert get_flash(conn)["error"] =~ "Authentication error"
     end
   end
+
+  test "oauth2 callback redirect work correctly post-sign in", %{conn: conn} do
+    insert(:plan)
+    user = insert(:user)
+    conn = login_user(conn, user)
+
+    search =
+      URI.encode_query(%{
+        scope: "read write",
+        response_type: "code",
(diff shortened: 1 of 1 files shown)


filipecabaco (approved):

AkashRajpurohit:
Hey folks, I got this PR link from the support ticket I opened for facing 500 error when authenticating logflare on cloudflare apps, although I see that this has been released but I'm still getting the same 500 as before.

filipecabaco:
Hello @AkashRajpurohit , we tried to deploy the fix but the dependency didn't update properly, we're working on it actively and will deploy it ASAP.

Thank you for your patience.

AkashRajpurohit:
Kindly drop a text here once it's been fixed, will try to add logflare app post that again.

filipecabaco:
@AkashRajpurohit we just released the newer version 1.3.6 with this issue fixed. 

Thank you so much for your patience

PR #1553, opened 2023-06-01
add - Postgres Ingestion

filipecabaco:
* Adds a new postgres backend to pipeline v2
* Adds feature flag override using environment variables
* Connections with other repositories it's achieved by creating modules dynamically that use Ecto.Repo and starts them with the configs given by the source backend configuration under supervision of a new dynamic supervisor

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,14 +35,6 @@ services:
         read_only: true
     depends_on:
       - db
-  ingestion_db:
-    image: postgres:13.4-alpine
-    environment:
-      POSTGRES_PASSWORD: postgres
-      POSTGRES_USER: postgres
-      POSTGRES_DB: ingestion
-    ports:
-      - "5433:5432"
 
 volumes:
   pg-data:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,14 @@ services:
         read_only: true
     depends_on:
       - db
+  ingestion_db:
+    image: postgres:13.4-alpine
+    environment:
+      POSTGRES_PASSWORD: postgres
+      POSTGRES_USER: postgres
+      POSTGRES_DB: ingestion
+    ports:
+      - "5433:5432"
 
 volumes:
   pg-data:
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -60,7 +60,7 @@ defmodule Logflare.Application do
       {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.SourcesSup},
       {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.RecentLogsSup},
       {DynamicSupervisor,
-       strategy: :one_for_one, name: Logflare.Backends.Adaptor.Postgres.Supervisor},
+       strategy: :one_for_one, name: Logflare.Backends.Adaptor.PostgresAdaptor.Supervisor},
       {Registry, name: Logflare.Backends.SourceRegistry, keys: :unique},
       {Registry, name: Logflare.Backends.SourceDispatcher, keys: :duplicate}
     ] ++ common_children()
@@ -140,7 +140,7 @@ defmodule Logflare.Application do
       {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.SourcesSup},
       {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.RecentLogsSup},
       {DynamicSupervisor,
-       strategy: :one_for_one, name: Logflare.Backends.Adaptor.Postgres.Supervisor},
+       strategy: :one_for_one, name: Logflare.Backends.Adaptor.PostgresAdaptor.Supervisor},
       {Registry, name: Logflare.Backends.SourceRegistry, keys: :unique},
       {Registry, name: Logflare.Backends.SourceDispatcher, keys: :duplicate}
     ] ++ conditional_children() ++ common_children()
@@ -199,7 +199,9 @@ defmodule Logflare.Application do
     # loading the migrations properly, triggering errors when running Ecto.Migrator
 
     # Similar error found in Realtime: https://github.com/supabase/realtime/pull/520/files#diff-1de8846d5d70df4b816a1b2bca51468d6c8386bc81f9efe82df2bf837367497d
-    Logflare.Backends.Adaptor.Postgres.Repo.migrations()
+    # This is fixed in ecto_sql main branch, awaiting new version to be released
+    # https://github.com/elixir-ecto/ecto_sql/blob/d8eaabcf34d52fad1d286dab653b2f200ba99333/lib/ecto/migrator.ex#L360
+    Logflare.Backends.Adaptor.PostgresAdaptor.Repo.migrations()
     |> Enum.map(fn {_, migration} -> Code.ensure_loaded!(migration) end)
   end
 end
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -59,6 +59,9 @@ defmodule Logflare.Application do
       # v2 ingestion pipelines
       {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.SourcesSup},
       {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.RecentLogsSup},
+      {DynamicSupervisor,
+       strategy: :one_for_one,
+       name: Logflare.Backends.Logflare.Backends.Adaptor.Postgres.Supervisor},
       {Registry, name: Logflare.Backends.SourceRegistry, keys: :unique},
       {Registry, name: Logflare.Backends.SourceDispatcher, keys: :duplicate}
     ] ++ common_children()
@@ -134,8 +137,14 @@ defmodule Logflare.Application do
       # Startup tasks
       {Task, fn -> startup_tasks() end},
 
-      # Telemetry
-      Logflare.Telemetry
+      # v2 ingestion pipelines
+      {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.SourcesSup},
+      {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.RecentLogsSup},
+      {DynamicSupervisor,
+       strategy: :one_for_one,
+       name: Logflare.Backends.Logflare.Backends.Adaptor.Postgres.Supervisor},
+      {Registry, name: Logflare.Backends.SourceRegistry, keys: :unique},
+      {Registry, name: Logflare.Backends.SourceDispatcher, keys: :duplicate}
     ] ++ conditional_children() ++ common_children()
   end
 
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -14,7 +14,7 @@ defmodule Logflare.Application do
 
   def start(_type, _args) do
     env = Application.get_env(:logflare, :env)
-
+    ensure_loaded_postgres_backend_migrations()
     # TODO: Set node status in GCP when sigterm is received
     :ok =
       :gen_event.swap_sup_handler(
@@ -60,8 +60,7 @@ defmodule Logflare.Application do
       {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.SourcesSup},
       {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.RecentLogsSup},
       {DynamicSupervisor,
-       strategy: :one_for_one,
-       name: Logflare.Backends.Logflare.Backends.Adaptor.Postgres.Supervisor},
+       strategy: :one_for_one, name: Logflare.Backends.Adaptor.Postgres.Supervisor},
       {Registry, name: Logflare.Backends.SourceRegistry, keys: :unique},
       {Registry, name: Logflare.Backends.SourceDispatcher, keys: :duplicate}
     ] ++ common_children()
@@ -141,8 +140,7 @@ defmodule Logflare.Application do
       {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.SourcesSup},
       {DynamicSupervisor, strategy: :one_for_one, name: Logflare.Backends.RecentLogsSup},
       {DynamicSupervisor,
-       strategy: :one_for_one,
-       name: Logflare.Backends.Logflare.Backends.Adaptor.Postgres.Supervisor},
+       strategy: :one_for_one, name: Logflare.Backends.Adaptor.Postgres.Supervisor},
       {Registry, name: Logflare.Backends.SourceRegistry, keys: :unique},
       {Registry, name: Logflare.Backends.SourceDispatcher, keys: :duplicate}
     ] ++ conditional_children() ++ common_children()
@@ -193,5 +191,15 @@ defmodule Logflare.Application do
       :timer.sleep(3_000)
       SingleTenant.update_supabase_source_schemas()
     end
+  end
+
+  defp ensure_loaded_postgres_backend_migrations do
+    # Needed to ensure that the migrations are properly loaded
+    # This was found during tests where the use Ecto.Migration apparently wasn't
+    # loading the migrations properly, triggering errors when running Ecto.Migrator
+
+    # Similar error found in Realtime: https://github.com/supabase/realtime/pull/520/files#diff-1de8846d5d70df4b816a1b2bca51468d6c8386bc81f9efe82df2bf837367497d
+    Logflare.Backends.Adaptor.Postgres.Repo.migrations()
+    |> Enum.map(fn {_, migration} -> Code.ensure_loaded!(migration) end)
   end
 end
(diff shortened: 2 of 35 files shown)


lib/logflare/backends.ex
```diff
@@ -1,51 +1,51 @@
...
+      |> SourceBackend.changeset(%{config: config, type: type})
+      |> validate_config()
+      |> Repo.insert()
+
+    case source_backend do
       {:ok, updated} ->
-        {:ok,
-         updated
-         |> typecast_config_string_map_to_atom_map()}
+        restart_source_sup(source)
```
Ziinc, line 47: i don't think we should restart the entire source supervision tree here. it would unnecssarily restart unrelated source backend processes, as well as other processes under the SourcesSup.
It would instead be better to restart only the related source backend process (and its tree), via a kill or otherwise.

For example, say we ingest to postgres + bq + ch. updating the config for BQ would terminate processes for all 3 source backends.

lib/logflare/backends.ex
```diff
@@ -58,15 +58,18 @@ defmodule Logflare.Backends do
...
+      |> SourceBackend.changeset(%{config: config})
+      |> validate_config()
+      |> Repo.update()
+
+    case source_backend_config do
       {:ok, updated} ->
-        {:ok,
-         updated
-         |> typecast_config_string_map_to_atom_map()}
+        restart_source_sup(source_backend.source)
```
Ziinc, line 71: same as above.

lib/logflare/backends.ex
Ziinc: perhaps a `restart_source_backend_sup` function would help.

lib/logflare/backends/adaptor/postgres/repo/migrations/add_log_events.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Logflare.Backends.Adaptor.Postgres.Repo.Migrations.AddLogEvents do
+  use Ecto.Migration
+
+  def up do
+    create table(:log_events, primary_key: false) do
+      add(:id, :string, primary_key: true)
+      add(:metadata, :map)
+      add(:event_message, :string)
+      add(:timestamp, :utc_datetime_usec)
+    end
```
Ziinc, line 10: basic ingestion should only create 3 columns, id, event_message, timestamp.
metadata should be removed.

lib/logflare/backends/adaptor/postgres/repo/migrations/add_log_events.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Logflare.Backends.Adaptor.Postgres.Repo.Migrations.AddLogEvents do
+  use Ecto.Migration
+
+  def up do
+    create table(:log_events, primary_key: false) do
```
Ziinc, line 5: should dynamically set table based on config

test/logflare/backends/adaptor/postgres/pipeline_test.exs
```diff
@@ -0,0 +1,90 @@
...
+    %{username: username, password: password, database: database, hostname: hostname} =
+      Application.get_env(:logflare, Logflare.Repo) |> Map.new()
+
+    url = "postgresql://#{username}:#{password}@#{hostname}/#{database}"
+
+    source = insert(:source, user: insert(:user))
+    source_backend = insert(:source_backend, type: :postgres, config: %{url: url}, source: source)
+    repository_module = Repo.new_repository_for_source_backend(source_backend)
+    pipeline_name = Backends.via_source_backend(source_backend, Pipeline)
+    memory_buffer_pid = start_supervised!(MemoryBuffer)
```
Ziinc, line 21: on a side note, we actually need to remove the MemoryBuffer to insert directly into broadway.

lib/logflare/backends/adaptor/postgres.ex
```diff
@@ -0,0 +1,71 @@
+defmodule Logflare.Backends.Adaptor.Postgres do
```
Ziinc, line 1: lets change this to `Adaptor.PostgresAdaptor` for clarity and shift the file to same dir as webhook_adaptor to match the elixir module placement conventions.

test/logflare/backends_test.exs
```diff
@@ -50,6 +52,28 @@ defmodule Logflare.BackendsTest do
       assert %SourceBackend{config: %{url: "http" <> _}} =
                Backends.get_source_backend(source_backend.id)
     end
+
+    test "validates config correctly for postgres backends", %{source: source} do
+      assert {:ok, source_backend} =
+               Backends.create_source_backend(source, :postgres, %{url: "postgresql://host"})
+
+      assert %SourceBackend{config: %{url: "postgresql://host"}, type: :postgres} = source_backend
```
Ziinc, line 60: we could combine the match assertions above for brevity

test/logflare/backends_test.exs
```diff
@@ -50,6 +52,28 @@ defmodule Logflare.BackendsTest do
...
+
+      assert {:error, %Ecto.Changeset{}} =
+               Backends.create_source_backend(source, :postgres, %{url: nil})
+
+      assert {:ok, %SourceBackend{config: %{url: "postgresql://changed"}, type: :postgres}} =
```
Ziinc, line 65: shouldn't we assert for the user:password@host:port format?

test/support/test_utils.ex
```diff
@@ -387,4 +387,28 @@ defmodule Logflare.TestUtils do
   end
 
   def random_pos_integer(limit \\ 1000), do: :rand.uniform(limit)
+
+  @doc """
+  Fetches data until the asserts is able to run its asserts a certain amount of times (default 10 times).
+  """
+  @spec retry_fetch(
```
Ziinc, line 394: a nicer implementation that is also common in nodejs testing is to have a `waitFor` that accepts a callback, and the callback just keeps executing until it exceeds a timeout.

The interface for the function would be much simpler.

```
def assert_until(callback, opts) do
  opts = Enum.into(opts, %{start: DateTime.utc_now()} )

  //check if exceed timeout, break and throw error if exceeding 500ms
  ....

  try
    callback.()
  rescue
  _ -> 
    // recursively retry until break.
    assert_until(callback, opts)

end
```

Ziinc: The caller of this function then has a nicer interface:
```
assert_until fn -> 
  assert render(view) =~ "values"
end
```

lib/logflare/backends/adaptor/postgres/pipeline.ex
```diff
@@ -0,0 +1,65 @@
...
+    Message.update_data(message, &process_data(&1, adaptor_state))
+  end
+
+  defp process_data(log_event, %{repository_module: repository_module}) do
+    timestamp =
+      log_event.body["timestamp"]
+      |> DateTime.from_unix!(:microsecond)
+      |> DateTime.to_naive()
+
+    params = %{
```
Ziinc, line 43: should ingest the entire body as is. understand if you want to separate the schema/dynamic migration changes for now, but this would be a regression as it won't ingest other top level keys.

lib/logflare/backends/adaptor/postgres/pipeline.ex
```diff
@@ -0,0 +1,65 @@
...
+    new_identifier = Tuple.append(identifier, base_name)
+    {:via, module, {registry, new_identifier}}
+  end
+
+  def handle_message(_processor_name, message, adaptor_state) do
+    Message.update_data(message, &process_data(&1, adaptor_state))
+  end
+
+  defp process_data(log_event, %{repository_module: repository_module}) do
+    timestamp =
```
Ziinc, line 38: just wondering out loud - would it be better to store as utc_datetime in db vs naive ?

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -1,21 +1,21 @@
 defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
   @moduledoc false
   use GenServer
-  alias Logflare.Backends.{SourceBackend, Adaptor, Adaptor.WebhookAdaptor, SourceDispatcher}
-  alias Logflare.{Backends, Buffers.MemoryBuffer}
-  use Adaptor
```
Ziinc, line 6: why remove the macro?
the macro is supposed to stub certain functions (for example, throw on endpoint query, since this is ingest only.)

Ziinc (changes requested):

test/support/test_utils.ex
Ziinc: the current usage interface of the `retry_fetch` is quite complex and error prone i think.

lib/logflare/application.ex
```diff
@@ -187,4 +197,14 @@ defmodule Logflare.Application do
...
     end
   end
+
+  defp ensure_loaded_postgres_backend_migrations do
+    # Needed to ensure that the migrations are properly loaded
+    # This was found during tests where the use Ecto.Migration apparently wasn't
+    # loading the migrations properly, triggering errors when running Ecto.Migrator
+
+    # Similar error found in Realtime: https://github.com/supabase/realtime/pull/520/files#diff-1de8846d5d70df4b816a1b2bca51468d6c8386bc81f9efe82df2bf837367497d
+    Logflare.Backends.Adaptor.Postgres.Repo.migrations()
```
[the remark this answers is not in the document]
filipecabaco: still not released but can add it

docker-compose.yml
```diff
@@ -35,6 +35,14 @@ services:
         read_only: true
     depends_on:
       - db
+  ingestion_db:
```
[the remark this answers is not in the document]
filipecabaco: can do

lib/logflare/backends.ex
filipecabaco: This was a slight refactor for readability and this is something we can tackle in a later PR as an improvement of pipeline v2

lib/logflare/backends/adaptor/postgres/repo/migrations/add_log_events.ex
filipecabaco: do we want basic ingestion? the focus of this feature for now is to provide self service without BQ and for that effect we need metadata to replicate the queries we have in studio correct?

test/support/test_utils.ex
filipecabaco: just wanted something "quick and dirty" honestly and we can refactor this in a later PR as it's just for tests and reduce flakyness.

lib/logflare/backends/adaptor/webhook_adaptor.ex
filipecabaco: we're using the adaptor, just removed the alias but I will add it back so it can be properly called by typespecs

lib/logflare/backends/adaptor/postgres/repo.ex
```diff
@@ -0,0 +1,52 @@
...
+      %{url: url} = config
+      opts = [{:url, url} | opts]
+
+      {:ok, _} = DynamicSupervisor.start_child(Supervisor, repository_module.child_spec(opts))
+    end
+
+    :ok
+  end
+
+  @migrations [
```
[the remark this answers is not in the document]
filipecabaco: we're following the same ideas from Realtime here, we should have a properly defined schema for the postgres BE as to properly upgrade the schema if we see a need for it.

cc @chasers

lib/logflare/backends/adaptor/postgres/repo/migrations/add_log_events.ex
Ziinc: Minimally would need to be able to support top level keys, namely the `project` key for project ref.

lib/logflare/backends/adaptor/postgres/pipeline.ex
filipecabaco: wanted to make something similar to what we have on `lib/logflare/source/bigquery/pipeline.ex` but also wanted to avoid to have a single :map type for the full log since that would make the endpoint construction way more difficult for us (we would need a lot of jsonb queries 😅)

I can make the metadata to receive everything (aka the body)

lib/logflare/backends/adaptor/postgres/pipeline.ex
filipecabaco: maybe, just currently using the same as `lib/logflare/logs/log_event.ex` to not deviate too much

lib/logflare/backends/adaptor/postgres/pipeline.ex
filipecabaco: Did the change to ingest the full body

lib/logflare/backends/adaptor/postgres/repo/migrations/add_log_events.ex
filipecabaco: we now store the full body but also have event_message, timestamp and id as separate columns to be closer to log_events in BQ

lib/logflare/backends/adaptor/postgres/repo.ex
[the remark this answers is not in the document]
filipecabaco: Did the change to have a migration schema so we can update schemas on psql backends if need be plus we now set a different table name based on the source.

As for the data schema changes, there's no good way of creating dynamic migrations so that will be trickier and an something that we should tackle later

lib/logflare/backends.ex
```diff
@@ -93,12 +92,12 @@ defmodule Logflare.Backends do
   defp typecast_config_string_map_to_atom_map(nil), do: nil
 
   defp typecast_config_string_map_to_atom_map(%SourceBackend{type: type} = source_backend) do
```
Ziinc, line 94: an aside: we can consider using [Params](https://hexdocs.pm/params/readme.html) in the future for config typecasting, which would be safer and cleaner than what we are doing now.

lib/logflare/backends/adaptor/postgres_adaptor/log_event.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Logflare.Backends.Adaptor.PostgresAdaptor.LogEvent do
+  use Ecto.Schema
+  import Ecto.Changeset
+  @primary_key {:id, :string, []}
+  schema "log_events" do
```
Ziinc, line 5: the [dynamic table PR](https://github.com/Logflare/logflare/pull/1580) will remove this right?

lib/logflare/backends/adaptor/postgres_adaptor/repo.ex
```diff
@@ -0,0 +1,56 @@
...
+    Application.put_env(:logflare, name, migration_source: migration_table)
+
+    name
+  end
+
+  @spec connect_to_source_backend(Ecto.Repo.t(), SourceBackend.t(), Keyword.t()) :: :ok
+  def connect_to_source_backend(repository_module, %SourceBackend{config: config}, opts \\ []) do
+    unless Process.whereis(repository_module) do
+      %{url: url} = config
+      opts = [{:url, url} | opts]
```
Ziinc, line 32: small nit: `opts = [{:url, config.url} | opts]`

lib/logflare/backends/adaptor/postgres_adaptor/repo.ex
```diff
@@ -0,0 +1,56 @@
+defmodule Logflare.Backends.Adaptor.PostgresAdaptor.Repo do
```
Ziinc, line 1: can we add `@moduledoc` and doc tags for each function, so that its easier to understand how the adaptor connects and does migrations.

lib/logflare/backends/adaptor/postgres_adaptor.ex
```diff
@@ -0,0 +1,71 @@
+defmodule Logflare.Backends.Adaptor.PostgresAdaptor do
```
Ziinc, line 1: think we should also start documenting how each adaptor works as well, small issue. just a brief description of the ingestion flow works

Ziinc (approved):
Looks great. only minor docs stuff that would be helpful for maintenance.

Ziinc:
added the areas to refactor into the [catchall refactoring task](https://www.notion.so/supabase/Code-cleanup-test-coverage-refactoring-documentation-7088e54b06e94963adbc2a141c8f6992?pvs=4#7f3bc265367a4a34810b073aba1cc77d)

lib/logflare/backends/adaptor/postgres_adaptor/log_event.ex
filipecabaco: Yes 👍 and it will be the next one to be merged since it impacts everything

lib/logflare/backends/adaptor/postgres_adaptor.ex
filipecabaco: Agreed otherwise it's a lot of code to read

lib/logflare/backends.ex
filipecabaco: yes, it will also simplify a lot validation since we can just run a validate required

lib/logflare/backends/adaptor/postgres_adaptor/repo.ex
filipecabaco: good call will change

PR #1559, opened 2023-06-02
add - Move from SQL to Migrations

filipecabaco:
To properly support users, we need to migrate whatever changes are needed to support postgres backends

On connection, we will run migrations against the URL provided


filipecabaco:
Merging it to `postgres-ingestion` to guarantee the base feature works properly

PR #1580, opened 2023-06-14
add - Dynamic table names based on source

filipecabaco:
Adds a new way to setup table names based on the source id so we can better control where information goes which will simplify querying at the cost of extra dynamic module loading

### DIFF
--- a/lib/logflare/backends/adaptor/postgres_adaptor/repo.ex
+++ b/lib/logflare/backends/adaptor/postgres_adaptor/repo.ex
@@ -5,8 +5,10 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.Repo do
   Using the Source Backend source id we create a new Ecto.Repo which whom we will
   be able to connect to the configured PSQL URL, run migrations and insert data.
   """
-  alias Logflare.Backends.SourceBackend
+  alias Logflare.Repo
+  alias Logflare.Backends.Adaptor.PostgresAdaptor.Repo.Migrations.AddLogEvents
   alias Logflare.Backends.Adaptor.PostgresAdaptor.Supervisor
+  alias Logflare.Backends.SourceBackend
 
   require Logger
 
@@ -18,7 +20,8 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.Repo do
 
   @spec new_repository_for_source_backend(SourceBackend.t()) :: atom()
   def new_repository_for_source_backend(source_backend) do
-    name = Module.concat([Logflare.Repo.Postgres, "Adaptor#{source_backend.source_id}"])
+    source_backend = Repo.preload(source_backend, :source)
+    name = Module.concat([Logflare.Repo.Postgres, "Adaptor#{source_backend.source.token}"])
 
     case Code.ensure_compiled(name) do
       {:module, _} -> nil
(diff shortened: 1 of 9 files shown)


lib/logflare/backends/adaptor/postgres_adaptor/repo/migrations/add_log_events.ex
```diff
@@ -1,16 +1,36 @@
...
+    name = Module.concat([__MODULE__, "MigrationFor#{source_id}"])
+
+    ast =
+      quote do
+        use Ecto.Migration
+
+        def up do
+          create table("log_events_#{unquote(source_id)}", primary_key: false) do
+            add(:id, :string, primary_key: true)
+            add(:body, :map)
```
Ziinc, line 18: what is this field? it shouldn't be present by default for table creation, only id, event_message, and timestamp.
unless we are storing the entire log event as a jsonb? if so, then we would be duplicating fields.

test/logflare/backends/adaptor/postgres_adaptor/pipeline_test.exs
```diff
@@ -55,18 +57,29 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.PipelineTest do
...
 
       MemoryBuffer.add(memory_buffer_pid, log_event)
 
-      fetcher = fn -> repository_module.all(LogEvent) end
+      fetcher = fn ->
+        repository_module.all(
+          from(l in Repo.table_name(source_backend),
+            select: %LogEvent{
+              id: l.id,
+              body: l.body,
```
Ziinc, line 76: as above, not quite sure what this field refers to.

lib/logflare/backends/adaptor/postgres_adaptor/repo/migrations/add_log_events.ex
Ziinc: if schema changing is coming after, then it is fine to just focus on the 3 main fields (and drop body), since schema diffing would require generating migrations on each schema change.

lib/logflare/backends/adaptor/postgres_adaptor/repo.ex
```diff
@@ -57,5 +56,9 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.Repo do
       {:error, :failed_migration}
   end
 
-  def migrations, do: @migrations
+  @spec table_name(SourceBackend.t()) :: binary()
+  def table_name(%SourceBackend{source_id: source_id}), do: "log_events_#{source_id}"
```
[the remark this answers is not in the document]
filipecabaco: makes a lot of sense, will change

lib/logflare/backends/adaptor/postgres_adaptor/repo/migrations/add_log_events.ex
filipecabaco: this is to store the entire content which is the only way I can imagine doing endpoints properly since most of the queries will require that 🤔 event_message won't be enough to replicate the endpoints behaviours we have for CLI development

test/logflare/backends/adaptor/postgres_adaptor/pipeline_test.exs
filipecabaco: https://github.com/Logflare/logflare/pull/1580#discussion_r1242588817

lib/logflare/backends/adaptor/postgres_adaptor/repo/migrations/add_log_events.ex
Ziinc: having top level keys as columns would still be needed for partitioning for performance, as well as indexing, even if we can index jsonb keys. Something to consider, though it definitely is simpler to throw everything into a jsonb column.
since this is the mvp, we can stick with the `body` column approach, though if that is so then we would need to make the postgres table schema that we set as transparent to the user as possible, so that there is less confusion if they compare how data is stored to BigQuery.  we can probably do some sql parsing magic to achieve this.

Ziinc (approved):
looks good to merge once all issues resolved :)

lib/logflare/backends/adaptor/postgres_adaptor/repo/migrations/add_log_events.ex
filipecabaco: > though it definitely is simpler to throw everything into a jsonb column.

this is the body column at the moment, just a jsonb with everything for now and then we can refine

I do agree that for now we can stay with this since it's so walled off and iterate on it further

PR #1586, opened 2023-06-16
add - Live Dashboard for Dev and Staging

filipecabaco:
Adds Live Dashboard to profile application performance in Staging and Dev environment.

Currently the telemetry we're tracking is:
* Phoenix
* Ecto
* Cachex

Added also os_mon to get information of the machine for a given node

To access the dashboard it's required to be an admin user.


lib/logflare_web/router.ex
```diff
@@ -308,6 +311,8 @@ defmodule LogflareWeb.Router do
 
     delete("/accounts/:id", AdminController, :delete_account)
     get("/accounts/:id/become", AdminController, :become_account)
+
+    live_dashboard("/livedashboard", ecto_repos: [], metrics: Logflare.Telemetry)
```
Ziinc, line 315: should scope it under `/admin`

mix.exs
```diff
@@ -63,7 +64,7 @@ defmodule Logflare.Mixfile do
       # Phoenix stuff
       {:phoenix, "~> 1.7"},
       # {:phoenix_html, "~> 3.2"},
-      {:phoenix_live_view, "~> 0.18"},
+      {:phoenix_live_view, "~> 0.19"},
```
Ziinc, line 67: i think this version bump would lead to some errors, especially the `live_component` usage 
https://github.com/phoenixframework/phoenix_live_view/blob/main/CHANGELOG.md#deprecations

Ziinc: is the version bump really necessary? might be better to separate this from addition of live dashboard.

Ziinc (approved):

Ziinc:
mostly concerned with the `phoenix_live_view` version bump, will likely lead to some bugs.

lib/logflare_web/router.ex
filipecabaco: it's already under admin
Screenshot 2023-06-19 at 21 51 17

mix.exs
filipecabaco: can revert it and update later. from my testing no regressions have happened in the main flows but nothing against doing it at a separate time

lib/logflare_web/router.ex
Ziinc: Ah oops by bad 👍👍

PR #1591, opened 2023-06-20
Fix/default float type for numerical columnes

Ziinc:
Defaults to using floats in ambiguous event types


Ziinc:
Closing in favour of using new v2 ingest

PR #1600, opened 2023-06-22
feat: allow querying by name using route endpoints/query/:token_or_name

Ziinc:
allows querying by name via the `/api/endpoints/query/:token_or_name` route, to make routing simpler.
[ticket](https://www.notion.so/supabase/Update-endpoints-query-route-to-handle-names-378b08fac27641d8b38277c540a3e53d?pvs=4)


test/logflare/utils_test.exs
```diff
@@ -3,4 +3,6 @@ defmodule Logflare.UtilsTest do
   use ExUnit.Case, async: true
   import Logflare.EnumDeepUpdate
   doctest Logflare.EnumDeepUpdate
+  import Logflare.Utils
+  doctest Logflare.Utils
```
filipecabaco, line 7: I don't think we need this, we're testing the dependency

lib/logflare/utils.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Logflare.Utils do
```
filipecabaco, line 1: We're only using this in one piece of the code and it's calling a library directly, I don't see the need to have a utils library for this.

filipecabaco (approved):
Looks good but I don't see the need for the Logflare.Utils at this point so I would remove it.

Wi still approve as this is a minor thing that should not block you IMO

Ziinc:
Removed the Utils module and made the check a private function 👍

PR #1602, opened 2023-06-23
add - Suggestion field for sources

filipecabaco:
Suggestion fields can now be added to sources so users can define which fields are suggested to achieve the best performance.

This can be bypassed if the user desires.
Screenshot 2023-06-22 at 23 53 35
Screenshot 2023-06-22 at 23 52 53

### DIFF
--- a/lib/logflare/sources/source.ex
+++ b/lib/logflare/sources/source.ex
@@ -124,7 +124,7 @@ defmodule Logflare.Source do
     field(:drop_lql_filters, Ecto.Term, default: [])
     field(:drop_lql_string, :string)
     field(:v2_pipeline, :boolean, default: false)
-    field(:suggested_fields, :string, default: "")
+    field(:suggested_keys, :string, default: "")
     # Causes a shitstorm
     # field :bigquery_schema, Ecto.Term
 
@@ -168,7 +168,7 @@ defmodule Logflare.Source do
       :drop_lql_filters,
       :drop_lql_string,
       :v2_pipeline,
-      :suggested_fields
+      :suggested_keys
     ])
     |> cast_embed(:notifications, with: &Notifications.changeset/2)
     |> default_validations(source)
@@ -191,7 +191,7 @@ defmodule Logflare.Source do
       :drop_lql_filters,
       :drop_lql_string,
       :v2_pipeline,
-      :suggested_fields
+      :suggested_keys
     ])
     |> cast_embed(:notifications, with: &Notifications.changeset/2)
     |> default_validations(source)
--- a/lib/logflare/sources/source.ex
+++ b/lib/logflare/sources/source.ex
@@ -48,16 +48,16 @@ defmodule Logflare.Source do
              ]}
 
     typed_embedded_schema do
-      field :rate, :integer
-      field :latest, :integer
-      field :avg, :integer
-      field :max, :integer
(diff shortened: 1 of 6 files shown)


lib/logflare/sources/source.ex
```diff
@@ -101,43 +101,43 @@ defmodule Logflare.Source do
...
+    field(:bq_table_partition_type, Ecto.Enum, values: [:pseudo, :timestamp], default: :timestamp)
+    field(:custom_event_message_keys, :string)
+    field(:log_events_updated_at, :naive_datetime)
+    field(:notifications_every, :integer, default: :timer.hours(4))
+    field(:lock_schema, :boolean, default: false)
+    field(:validate_schema, :boolean, default: true)
+    field(:drop_lql_filters, Ecto.Term, default: [])
+    field(:drop_lql_string, :string)
+    field(:v2_pipeline, :boolean, default: false)
+    field(:suggested_fields, :string, default: "")
```
Ziinc, line 127: Maybe "suggested keys" for consistency

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -887,4 +891,31 @@ defmodule LogflareWeb.Source.SearchLV do
...
+  defp check_suggested_fields(
+         lql_rules,
+         %{suggested_fields: suggested_fields},
+         %{assigns: %{force_query: false}} = socket
+       ) do
+    suggested_present =
+      suggested_fields
+      |> String.split(",")
+      |> Enum.map(fn
+        "m." <> suggested_field -> "metadata." <> suggested_field
```
Ziinc, line 912: Don't think this is needed

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -887,4 +891,31 @@ defmodule LogflareWeb.Source.SearchLV do
...
+        suggested_field -> suggested_field
+      end)
+      |> Enum.all?(fn suggested_field ->
+        Enum.find(lql_rules, fn %{path: path} -> path == suggested_field end)
+      end)
+
+    if suggested_present, do: {:ok, socket}, else: {:error, :suggested_field_not_found}
```
Ziinc, line 919: So even if the user acknowledges it, they will still receive the pop-up? Might get noisy (and irritating) and overlap with other flashes. I think having a one time flag to show it once only will be good

Ziinc (approved):
I think showing the flash once would be good enough, not on every search (though we can see how the experience is like first then adjust)

Ziinc:
Will need to fix the check

lib/logflare_web/live/search_live/logs_search_lv.ex
filipecabaco: unfortunately we need it, if a user sets the keys as they use them (e.g. `m.project`) then it will not work since the LQL parser converts the `m` to `metadata`

lib/logflare_web/live/search_live/logs_search_lv.ex
filipecabaco: Will change

lib/logflare/sources/source.ex
filipecabaco: Will change

PR #1606, opened 2023-06-30
Fix/supabase sources ensure source started on create

Ziinc:
fix for https://github.com/supabase/cli/issues/1177 , handles startup of RLS processes for v1 ingest for when supabase sources are already inserted from a backup.

[ticket](https://www.notion.so/supabase/bug-logflare-supabase-mode-startup-does-not-start-source-processes-if-source-already-exists-609e66b4c4b340fc8c1ff62dfb7eb72d?pvs=4)

### DIFF
--- a/lib/logflare/single_tenant.ex
+++ b/lib/logflare/single_tenant.ex
@@ -154,7 +154,6 @@ defmodule Logflare.SingleTenant do
     end
   end
 
-
   @doc """
   Starts supabase sources if present.
   Note: not tested as `Logflare.Source.Supervisor` is a pain to mock.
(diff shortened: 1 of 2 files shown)


filipecabaco (approved):

lib/logflare/single_tenant.ex
```diff
@@ -155,6 +154,39 @@ defmodule Logflare.SingleTenant do
...
+  end
+
+  @doc """
+  Lists supabase sources pids if processes have been started
+
+  Note: not tested as `Logflare.Source.Supervisor` is a pain to mock.
+  TODO: add testing for v2
+  """
+  @spec list_supabase_sources_pids() :: [pid()]
+  def list_supabase_sources_pids do
```
filipecabaco, line 180: this one could be private but nitpick no need to do it

PR #1607, opened 2023-06-30
feat - Seed Supabase Mode to use Postgres Ingestion exclusively

filipecabaco:
To simplify Supabase Mode (self hosted) we will move to support ingestion with Postgres and not use BigQuery


Ziinc:
I am splitting out the changes of this PR, as there are too many things happening in this PR.

This PR should focus solely on usage of PG and BQ in supabase mode. However, due to the many blockers before this actually works as advertised, it is likely that this will be merged much further on in dev cycle.

PostgresAdaptor-specific changes are split out to this PR https://github.com/Logflare/logflare/pull/1618

This PR's scope should be maintained around the SingleTenant module and cli integration.

Ziinc:
Re-creating this PR to start on self-host version first.

PR #1608, opened 2023-07-04
fix: add in more schema keys

Ziinc:
adds in missing schema keys from the gotrue source.

source schema used as reference is [here](https://logflare.app/sources/16129/search?querystring=&tailing%3F=false)

### DIFF
--- a/priv/supabase/ingest_samples/gotrue.logs.prod.json
+++ b/priv/supabase/ingest_samples/gotrue.logs.prod.json
@@ -8,7 +8,44 @@
     "msg": "request started",
     "path": "/user",
     "remote_addr": "152.26.89.239",
-    "timestamp": "2023-03-23T14:36:26Z"
+    "timestamp": "2023-03-23T14:36:26Z",
+    "action": "some action",
+    "args": ["arg"],
+    "auth_event": {
+      "action": "action",
+      "actor_id": "123-123",
+      "actor_name": "some name",
+      "actor_username": "some username",
+      "actor_via_sso": false,
+      "log_type": "some type",
+      "traits": {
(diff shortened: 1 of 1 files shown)


priv/supabase/ingest_samples/gotrue.logs.prod.json
```diff
@@ -8,7 +8,44 @@
...
+    "project": "some123ref",
+    "metering": false,
+    "message": "some message",
+    "login_method": "some method",
+    "provider": "google",
+    "referer": "example.com",
+    "remote_addr": "123.123.123",
+    "request_id": "123-123",
+    "source_type": "some type",
+    "status": "some status",
```
sweatybridge, line 45:
```suggestion
    "status": 200,
```


Based on what I see on logflare prod, the status field is of integer type.

sweatybridge: Screenshot 2023-07-05 at 12 14 59 PM

To clarify, the json field logged by gotrue is an integer, but vector transforms it to a string. May be ok to ignore if the change is too complicated.

Ziinc: the seed payload should match the schema's types, since it skips vector transformations.  otherwise we'd have an integer column instead of string column being created.

Ziinc:
merging in first for testing.

priv/supabase/ingest_samples/gotrue.logs.prod.json
sweatybridge: Yea ideally gotrue log should also match the schema type and it currently doesn't. Small thing though

PR #1612, opened 2023-07-07
feat: add suggested keys to flash message

Ziinc:
This PR adds in rendering of the suggested keys to display the keys, for better ux.

This PR also adds in more documentation around the suggested search keys.


chasers (approved):
✅ assuming tests will pass

PR #1615, opened 2023-07-10
fix: move gce networking specific sleep to gce block

Ziinc:
This PR moves the GCE-related networking sleep to the gce if block so that it does not block self-hosting.


sweatybridge (approved):
🎉

PR #1618, opened 2023-07-12
feat: postgres adaptor querying and ingestion

Ziinc:
This PR focuses on enabling raw postgres querying and ingestion.

This means that the PostgresAdaptor module will accept either Ecto.Query or string SQL (in PG dialect) and execute it on the PG backend, and then return the results as is.

This PR does not focus on endpoints execution, though minor refactoring will be done.


lib/logflare/backends/adaptor/postgres_adaptor.ex
```diff
@@ -77,6 +69,42 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor do
...
+  # expose PgRepo functions
+  defdelegate create_repo(source_backend), to: PgRepo
+  defdelegate connect_to_repo(source_backend), to: PgRepo
+  defdelegate table_name(source_or_source_backend), to: PgRepo
+  defdelegate create_log_events_table(source_backend), to: PgRepo
+  defdelegate create_log_events_table(source_backend, override_migrations), to: PgRepo
+  defdelegate rollback_migrations(source_backend), to: PgRepo
+  defdelegate drop_migrations_table(source_backend), to: PgRepo
+  defdelegate migrations_table_name(source_backend), to: PgRepo
+  defdelegate insert_log_event(source_backend, log_event), to: PgRepo
```
filipecabaco, line 106: I don't fully agree with this since it's a layer of indirection but I can live with it 😄

filipecabaco (approved):

lib/logflare/backends/adaptor/postgres_adaptor.ex
```diff
@@ -23,8 +15,8 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor do
   alias Logflare.Backends.SourceBackend
   alias Logflare.Backends.SourceDispatcher
   alias Logflare.Buffers.MemoryBuffer
-  alias Logflare.Backends.Adaptor.PostgresAdaptor.Pipeline
-  alias Logflare.Backends.Adaptor.PostgresAdaptor.Repo
+  alias __MODULE__.Pipeline
+  alias __MODULE__.PgRepo
```
filipecabaco, line 19: This type of change can also make searching trickier (search all usages for a given module and would return all places where it's aliased) so non blocking but can reduce capacity to search for code usage

Ziinc: sure will adjust

lib/logflare/backends/adaptor/postgres_adaptor.ex
Ziinc: Reason for this is because ideally each Adaptor should expose and implement only the same callbacks as per the `Adaptor` behaviour, so that all logic for setup/teardown/migrations are encapsulated within the `*Adaptor` module namespace, so that internal implementation can always change without external api change. Whatever is exposed at Adaptor-level is essentially the public api for the backend.

In the long run, it would make our maintenance and testing much easier as the number of adaptors grow, since expected behaviour for each backend is always the same.

I'm leaning towards using `setup_ingest/1`, `setup_querying/1`. `teardown/1`, `insert_log_event/2` callbacks to `Adaptor`, but would like to see how the BigQueryAdaptor turns out first before extending Adaptor more with the commonalities.

filipecabaco: Ah got it!!! Makes sense 👍👍

PR #1624, opened 2023-07-17
feat: bq -> pg basic translation part 1

Ziinc:
This PR focuses BQ->PG syntax translation for logflare-specific querying.


filipecabaco (approved):
Approving so we're not blocking but we need a huge code cleanup here 😅

PR #1628, opened 2023-07-17
fix: shard source metrics pubsub

chasers:
Subscribing process could still be a bottleneck but this should be an improvement for sure.


lib/logflare/pubsub_rates/buffers.ex
```diff
@@ -16,7 +18,9 @@ defmodule Logflare.PubSubRates.Buffers do
   end
 
   def init(state) do
-    PubSub.subscribe(Logflare.PubSub, "buffers")
+    for shard <- 1..@pool_size do
+      PubSub.subscribe(Logflare.PubSub, "buffers:shard-#{shard}")
```
filipecabaco, line 22: 😍

filipecabaco (approved):

PR #1631, opened 2023-07-18
add - Single Tenant mode using Postgres Backend Adapter

filipecabaco:
When POSTGRES_BACKEND_URL is set, all sources will use the URL during source creation to set the Postgres Adaptor as the default source of data.

If the URL isn't set, Google Credentials will be required and be used


config/runtime.exs
```diff
@@ -186,10 +186,22 @@ config :stripity_stripe,
...
-  config :goth, json: File.read!("gcloud.json")
   config :grpc, port: System.get_env("LOGFLARE_GRPC_PORT", "50051") |> String.to_integer()
 end
 
+case {System.get_env("POSTGRES_BACKEND_URL"),
+      System.get_env("LOGFLARE_SINGLE_TENANT", "false") == "true"} do
+  {postgres_backend_url, true} when not is_nil(postgres_backend_url) ->
+    config :logflare, :postgres_backend_adapter,
+      url: postgres_backend_url,
+      pool_size: 3
```
Ziinc, line 197: is there any reason why 3, and not say 10 as is the usual default pool size?

config/runtime.exs
```diff
@@ -186,10 +186,22 @@ config :stripity_stripe,
        )
 
 if config_env() != :test do
-  config :goth, json: File.read!("gcloud.json")
   config :grpc, port: System.get_env("LOGFLARE_GRPC_PORT", "50051") |> String.to_integer()
 end
 
+case {System.get_env("POSTGRES_BACKEND_URL"),
```
Ziinc, line 192: might be clearer to use a cond/do

lib/logflare/backends/adaptor/postgres_adaptor/repo/migrations/add_log_events.ex
```diff
@@ -6,9 +6,10 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.Repo.Migrations.AddLogEvents
 
   alias Logflare.Backends.Adaptor.PostgresAdaptor.PgRepo
 
-  def generate_migration(source_backend) do
+  def generate_migration(%{source: %{token: token}} = source_backend) do
```
Ziinc, line 9: can we add docstring for this, i think we missed it in initial implementaiton. should mention that source needs to be preloaded.

lib/logflare/sources.ex
```diff
@@ -40,31 +41,50 @@ defmodule Logflare.Sources do
...
   @spec create_source(map(), User.t()) :: {:ok, Source.t()} | {:error, Ecto.Changeset.t()}
   def create_source(source_params, user) do
+    postgres_backend_adapter_url =
+      Application.get_env(:logflare, :postgres_backend_adapter, []) |> Keyword.get(:url)
+
+    source_params =
+      source_params
+      |> Enum.map(fn {k, v} -> {to_string(k), v} end)
+      |> Map.new()
+      |> then(&Map.put(&1, "v2_pipeline", !!postgres_backend_adapter_url))
```
Ziinc, line 51: would be cleaner to do these value setting at changeset level

lib/logflare/sources.ex
```diff
@@ -40,31 +41,50 @@ defmodule Logflare.Sources do
...
-      {:error, changeset} ->
-        {:error, changeset}
-    end
+    {:ok, source}
   end
 
+  defp create_big_query_schema_and_start_source({:error, changeset}), do: {:error, changeset}
+
+  defp create_backend_adaptor({:ok, source}, url) do
+    {:ok, _source_backend} = Backends.create_source_backend(source, :postgres, %{url: url})
```
Ziinc, line 82: should we compile the repo module here as well and create the conn?

lib/logflare/sources.ex
```diff
@@ -40,31 +41,50 @@ defmodule Logflare.Sources do
...
     source =
       user
       |> Ecto.build_assoc(:sources)
       |> Source.update_by_user_changeset(source_params)
       |> Repo.insert()
 
-    case source do
-      {:ok, source} ->
-        init_schema = SchemaBuilder.initial_table_schema()
+    case postgres_backend_adapter_url do
```
Ziinc, line 59: should also factor in the single tenant mode flag.

Ziinc (approved):
i think test cov is missing a few aspects for the single_tenant_mode flag:
- if `SINGLE_TENANT_MODE`=true, then either BQ or PG credentials must be provided
- if `SINGLE_TENANT_MODE=true and PG credentials provided, then all sources created should be v2 as well.
overall looks good 👍

config/runtime.exs
filipecabaco: just a smaller amount of conns since it will be a ton of conns going out (each source will have a pool of 10 😅 )

config/runtime.exs
filipecabaco: yes it is! will change it

lib/logflare/sources.ex
filipecabaco: good call, will change

lib/logflare/sources.ex
filipecabaco: unsure honestly, at the moment we're only doing all of that on first ingestion we can change this at a later date

lib/logflare/sources.ex
filipecabaco: 👍 makes sense

config/runtime.exs
Ziinc: Perhaps something to consider refactoring in future:
if the source backend repo is the same across multiple sources then the pool size can be increased correspondingly to reduce process overhead, instead of having dedicated repos for each source backend. Because 10 would be able to handle a large volume of concurrent queries, so having many redundant pools might not be optimal

filipecabaco: Agreed, maybe we should do the repo according to the URL instead of the source to simplify that process 🤔

PR #1634, opened 2023-07-21
fix - Fixes endpoints UI on save and on query run bugs

filipecabaco:
* On save, the lack of a default language on Query creation made a changeset fail
* On query test, the input would be deleted since the value was not being set and assigns were being removed

> Note: Endpoints are behaving weirdly due to other reasons in my dev environment so take that into account in the gifs below 😅 

## Before
before
## After
after

### DIFF
--- a/lib/logflare_web/live/endpoints/endpoints_live.ex
+++ b/lib/logflare_web/live/endpoints/endpoints_live.ex
@@ -1,16 +1,18 @@
 defmodule LogflareWeb.EndpointsLive do
   @moduledoc false
   use LogflareWeb, :live_view
+  use Phoenix.Component
+
   require Logger
+
   alias Logflare.Endpoints
   alias Logflare.Users
   alias LogflareWeb.Utils
-  use Phoenix.Component
-  embed_templates "actions/*", suffix: "_action"
-  embed_templates "components/*"
+
+  embed_templates("actions/*", suffix: "_action")
+  embed_templates("components/*")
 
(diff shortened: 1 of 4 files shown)


lib/logflare_web/live/endpoints/endpoints_live.ex
```diff
@@ -190,24 +178,18 @@ defmodule LogflareWeb.EndpointsLive do
...
-    {:noreply,
-     socket
-     |> put_flash(:info, "Successfully applied for the Endpoints beta. We'll be in touch!")}
+    message = "Successfully applied for the Endpoints beta. We'll be in touch!"
+    {:noreply, put_flash(socket, :info, message)}
   end
 
-  defp update_params_form(socket, parameters, query_string) do
+  defp assign_updated_params_form(socket, parameters, query_string) do
+    params = for(k <- parameters, do: {k, nil}, into: %{})
```
Ziinc, line 186: this would still clear the form values, i think.

Ziinc (commented):
could we have a screencap of the fix if it can't be tested easily

filipecabaco:
Posted a GIF with the before and after

lib/logflare_web/live/endpoints/endpoints_live.ex
filipecabaco: No, all good 👍 works as expected gifs in the PR

Ziinc (approved):
Awesome thanks @filipecabaco 👍😁

PR #1637, opened 2023-07-25
feat: bq-to-pg translation, cte handling and value casting

Ziinc:
this PR adds in two translation aspects:
- multiple from aliases field referencing, such that the table aliases are considered and omited/added to the json query syntax.
- cast handling, as json cannot be used to cast, hence we need to use LongArrow or HashLongArrow for the syntax


filipecabaco (approved):
CI failing but looks good to merge for now, later we can think about a code cleanup

PR #1659, opened 2023-08-01
fix: holy `:counters` batman

chasers:
Uses `:counters` to keep track of the buffer length instead of running everything through a GenServer.

- [x] tests pass
- [x] clean up


Ziinc:
Loving the PR name

filipecabaco (approved):

PR #1664, opened 2023-08-02
fix - Move PgRepo to a GenServer to avoid race conditions

filipecabaco:
Previously PgRepo was changing concurrently multiple times a single module configuration leading to issues when creating tables and migrations for a given source.

To fix this, in a weird, we added more concurrency.

PgRepo now is a GenServer per Repository which means that migrations are handled by a single mailbox that is able to properly coordinate actions.

### DIFF
--- a/lib/logflare/backends/adaptor/postgres_adaptor/pg_repo.ex
+++ b/lib/logflare/backends/adaptor/postgres_adaptor/pg_repo.ex
@@ -5,14 +5,16 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.PgRepo do
   Using the Source Backend source id we create a new Ecto.Repo which whom we will
   be able to connect to the configured PSQL URL, run migrations and insert data.
   """
-  alias Logflare.Backends.Adaptor.PostgresAdaptor.Repo.Migrations.AddLogEvents
-  alias Logflare.Backends.Adaptor.PostgresAdaptor.Supervisor
+  use GenServer
+
   alias Logflare.Backends.Adaptor.PostgresAdaptor
   alias Logflare.Backends.Adaptor.PostgresAdaptor.PgLogEvent
+  alias Logflare.Backends.Adaptor.PostgresAdaptor.Repo.Migrations.AddLogEvents
+  alias Logflare.Backends.Adaptor.PostgresAdaptor.PgRepoSupervisor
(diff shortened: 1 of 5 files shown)


lib/logflare/backends/adaptor/postgres_adaptor/pg_repo.ex
```diff
@@ -139,21 +118,17 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.PgRepo do
...
   Drops the migration table
   """
   @spec drop_migrations_table(SourceBackend.t()) :: :ok
   def drop_migrations_table(source_backend) do
-    repository_module = create_repo(source_backend)
-    migrations_table = migrations_table_name(source_backend)
-    Ecto.Adapters.SQL.query!(repository_module, "DROP TABLE IF EXISTS #{migrations_table}")
-    :ok
+    repo_module_name = get_repo_module(source_backend)
+    GenServer.call(:"PgRepo#{repo_module_name}", {:drop_migrations_table, source_backend})
```
Ziinc, line 131: We can use the backends process registry for this with a :via instead.

lib/logflare/backends/adaptor/postgres_adaptor/pg_repo.ex
```diff
@@ -198,4 +173,93 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.PgRepo do
 
     repo.insert(changeset)
   end
+
+  ## Genserver calls
+
+  def start_link(%{repo_module_name: repo_module_name} = state),
+    do: GenServer.start_link(__MODULE__, state, name: :"PgRepo#{repo_module_name}")
```
Ziinc, line 180: Could we  store the source backend config for the repo in state? It would result in less args being passed around

Ziinc: Also, we can register different name using :via without needing to use another atom.

Ziinc (approved):

lib/logflare/backends/adaptor/postgres_adaptor/pg_repo.ex
filipecabaco: can do 👍 decided to just send everything "organised" but can do that

and will do the change to use via 👍

PR #1673, opened 2023-08-07
fix - Update metadata from map to array

filipecabaco:
On query responses from postgres we're running a transformation where metadata keys are converted from maps to lists so they are closer to BigQuery responses
Screenshot 2023-08-07 at 02 47 58


lib/logflare/backends/adaptor/postgres_adaptor.ex
```diff
@@ -114,12 +119,32 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor do
...
+  defp nested_map_update(value) when is_struct(value), do: value
+
+  defp nested_map_update(value) when is_map(value),
+    do: Enum.reduce(value, %{}, &nested_map_update/2)
+
+  defp nested_map_update(value) when is_list(value), do: Enum.map(value, &nested_map_update/1)
+
+  defp nested_map_update(value), do: value
+
+  defp nested_map_update({"metadata", value}, acc) do
```
Ziinc, line 140: This isn't specific to the metadata key, it applies to alll key paths

test/logflare/backends/postgres_adaptor_test.exs
```diff
@@ -68,6 +68,72 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptorTest do
...
+
+      assert :ok = PostgresAdaptor.ingest(pid, [log_event])
+
+      expected_event = %{
+        "event_message" => "test-msg",
+        "id" => "93d0f6b4-91f9-4a08-938c-39f00cca7fdd",
+        "metadata" => [
+          %{
+            "host" => "db-default",
+            "parsed" => %{
```
Ziinc, line 102: This value should be wrapped in an array

Ziinc (changes requested):

lib/logflare/backends/adaptor/postgres_adaptor.ex
filipecabaco: by doing that I get the render error
Screenshot 2023-08-07 at 12 31 09

Ziinc (approved):

PR #1675, opened 2023-08-07
fix: query results should be nested in an array to mimic bigquery behaviour

Ziinc:
This PR adds in array nested to the postgres adaptor to mimic bigquery querying behaviour.

This might change in a future version as we may remove this array nested to standardize payload returning across backends (and follow the ingested json structure as close as possible)


filipecabaco:
Screenshot 2023-08-07 at 20 34 12
worked as expected 👍 🎉

filipecabaco (approved):

PR #1678, opened 2023-08-07
fix: handle buffer_full and respond to client

chasers:
Problem: 

Currently if the Broadway buffer is full it will error with a 5xx. This throws off a log message for every log event bogging things down. It also causes the load balancer to think the all nodes are not healthy and it won't route any traffic at all. 

Solution:

This surfaces ingest errors to the client in the form of a 406 status code with the normal message array clients expect.

The message will be different for a full buffer.

Screenshot 2023-08-08 at 3 11 21 PM


lib/logflare/logs/log_event.ex
```diff
@@ -19,7 +19,8 @@ defmodule Logflare.LogEvent do
     field :valid, :boolean
     field :drop, :boolean, default: false
     field :is_from_stale_query, :boolean
-    field :validation_error, {:array, :string}
+    field :validation_error, :string
+    field :ingest_error, :string
```
Ziinc, line 23: technically, an ingest error would cover validation as well. Maybe we can split it out to `pipeline` when referring to the broadway insert pipeline. We would need to start classifying other types of errors as well, like `backend` for backend errors
i would think that an array of embeds would be more versatile, such as 
```elixir
errors: [%IngestError{type: :validation, message: "..."}, %IngestError{type: :pipeline, message: "..."}]
```
other ideas are like `%IngestError{scope: :validation, ...}` or `%IngestError{stage :buffer, ...}`

test/logflare/log_event_test.exs
```diff
@@ -42,7 +43,8 @@ defmodule Logflare.LogEventTest do
              drop: false,
              # validity gets overwritten
              valid: true,
-             validation_error: "",
+             validation_error: nil,
+             ingest_error: nil,
```
Ziinc, line 47: would also need to update controller tests to cover rendering of these errors to api clients.

lib/logflare/logs/log_event.ex
Ziinc: Another perspective is that there can only be one error during the ingestion pipeline, so we could simplify it to:
```
ingestion_error: %{type: :validation, message: "..."}
```
```
ingestion_error: %{type: :buffer, message: "..."}
```

```
ingestion_error: %{type: :bigquery_stream, message: "..."}
```
which would also help with the rendering of the ingestion errors when I update the [rejected logs page](https://www.notion.so/supabase/bug-bigquery-execution-errors-are-not-surfaced-to-query-testing-form-faee2b3ed559429f8334a5538aaa441a?pvs=4)

chasers: Yeah good idea.

Added a LogEvent.PipelineError struct. The array of errors there I think was a mistake because the API call returns an array of errors.

A LogEvent should really only have one error, the first one where it errored at. 

And we should return an array of LogEvent.PipelineErrors in that API call (todo later to make sure we don't break clients).

A PipelineError should cover everything as we can include backend ingest errors there too at some point.

test/logflare/log_event_test.exs
chasers: Think I got all of them. If not link specific ones and I'll fix.

chasers:
FYI snuck in some more batchers in here.

filipecabaco (approved):

chasers:
@Ziinc feel free to comment after this is closed...

Ziinc:
changes look great 👍

PR #1689, opened 2023-08-22
Fix/postgres translation errors UI filtering for list comparisons

Ziinc:
This PR fixes postgres translations relating to the `InList` BinaryOp.
Comparisons should be typecasted to text to handle list value comparisons

### DIFF
--- a/lib/logflare/sql.ex
+++ b/lib/logflare/sql.ex
@@ -877,7 +877,7 @@ defmodule Logflare.Sql do
           |> String.downcase()
 
         field_arg =
-          if is_timestamp_identifier?(to_trunc) do
+          if timestamp_identifier?(to_trunc) do
             at_time_zone(to_trunc)
           else
             to_trunc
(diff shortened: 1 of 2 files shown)


lib/logflare/sql.ex
```diff
@@ -1390,6 +1414,12 @@ defmodule Logflare.Sql do
   defp is_identifier?(identifier),
     do: is_map_key(identifier, "CompoundIdentifier") or is_map_key(identifier, "Identifier")
 
+  defp is_numeric_value?(%{"Value" => %{"Number" => _}}), do: true
+  defp is_numeric_value?(_), do: false
+  defp is_json_access?(%{"Nested" => %{"JsonAccess" => _}}), do: true
+  defp is_json_access?(%{"JsonAccess" => _}), do: true
+  defp is_json_access?(_), do: false
```
hauleth, line 1421: By [Elixir naming convention](https://hexdocs.pm/elixir/naming-conventions.html#trailing-question-mark-foo) it should be `json_access?` and `numeric_value?`. `is_*` prefix (with no question mark at the end) is meant for guards.

lib/logflare/sql.ex
```diff
@@ -1433,24 +1463,48 @@ defmodule Logflare.Sql do
...
       }
     }
   end
+
+  defp json_access_arrow(data, hash) do
+    arrow =
+      cond do
+        data.in_binaryop or data.in_between or data.in_function_or_cast -> "LongArrow"
+        data.in_inlist -> "Arrow"
+        true -> "Arrow"
```
hauleth, line 1501:
```suggestion
        true -> "Arrow"
```

hauleth (approved):

PR #1692, opened 2023-09-02
fix: log dist buffer and increase dist buffer limit

chasers:
I verified on prod this won't create excess log messages in normal ops:

```elixir
iex> :erlang.system_monitor(self(), [:busy_dist_port, :busy_port, {:long_gc, 1_000}, {:long_schedule, 100}])
iex> flush()
```


filipecabaco (approved):

chasers:
FYI I moved the long gc back to 250ms and commented out the Logflare.SystemMetrics.Procs.Poller so it won't start.

PR #1696, opened 2023-09-06
fix: Slack hook message format

chasers:


chasers:
Everyone is asleep: merging

PR #1715, opened 2023-10-02
ft: support ingesting gzipped data for logs

### LINKED ISSUE #1699: Support gzip ingest

hauleth:
Close #1699


chasers:
@hauleth need to format some files. Check the Elixir CI action.

lib/logflare_web/controllers/plugs/compressed_body_reader.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Plug.Parsers.CompressedBodyReader do
+  def read_body(conn, opts \\ []) do
+    content_encoding = Plug.Conn.get_req_header(conn, "content-encoding")
+
+    with {:ok, body, conn} <- Plug.Conn.read_body(conn, opts),
+         do: {:ok, try_decompress(body, content_encoding), conn}
```
hauleth, line 6: Ok, that can cause memory exhaustion on the node, but I will prepare the fix for that.

Ziinc:
Screenshot 2023-10-05 at 11 45 56 PM

would need `mix format`

lib/logflare_web/controllers/plugs/compressed_body_reader.ex
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  defp try_decompress(data, []), do: {:ok, data}
+  defp try_decompress(data, ["gzip"]), do: safe_gunzip(data)
+
+  defp safe_gunzip(data) do
+    z = :zlib.open()
+    try do
+      :zlib.inflateInit(z, 31)
+      result = :zlib.safeInflate(z, data)
```
Ziinc, line 21: from my understanding, the safeInflate throw an exception if memory exceeds, which would not result in error tuple above at line 9

Ziinc: not really an issue since it would 500. Would be good to handle it with a 400 bad request though.

Ziinc (changes requested):

lib/logflare_web/controllers/plugs/compressed_body_reader.ex
hauleth: It will not, it will then return `{:continue, deflated}` and you need to call it again with new data. So it will never throw. And even if, we safely handle that there and it will either return 413 (most cases) or it will throw 500 (less common case).

lib/logflare_web/controllers/plugs/compressed_body_reader.ex
```diff
@@ -0,0 +1,33 @@
+defmodule Plug.Parsers.CompressedBodyReader do
```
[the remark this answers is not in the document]
hauleth: That was my plan, but other plugs was using that so I wanted to keep it consistent. I will change it.

Ziinc (approved):

Ziinc:
formatting and its good to merge

Ziinc (approved):

PR #1716, opened 2023-10-02
chore: cleanup Credo warnings

hauleth:
Not all warnings are cleaned up yet, but some low hanging fruits are fixed
there.


Ziinc (approved):
Looks good. Would prefer smaller PRs in the future where possible to make reviewing easier

hauleth:
If you want, I can split it into set of PRs

PR #1718, opened 2023-10-04
fix: sql parser error on save is not returning correct error tuple

Ziinc:
Errors when saving alerts is not thrown, resulting in a full page refresh instead of feedback to the user. 
occurs only when sql parser cannot parse the query.


Ziinc:
@chasers FYI

PR #1719, opened 2023-10-04
chore: use OTP 26

hauleth:
- chore: update Rustler to version 0.29.x
- chore: replace calls to Logger.warn with Logger.warning


hauleth:
I haven't updated `.tool-versions` (as I do not use that file personally) so I do not know if it should be handled by me, or someone else can do that.

Ziinc:
@hauleth .tool-versions would need to be bumped to otp26 as we use asdf for environment management.

Ziinc:
would need the tool-versions changes + `mix format`

hauleth:
It seems that there is flaky test.

Ziinc (approved):

hauleth:
Finally it pass everything. Merge it and I will update #1715 on top and it will pass as well.

chasers:
@hauleth merged!

PR #1720, opened 2023-10-04
fix: Alerts changeset bug referencing Endpoints instead of AlertQuery

Ziinc:
Alerts ui does not show correct changset data.

Fixes the changeset bug which was referencing Endpoints.


chasers:
Hmmm something else failing here @Ziinc

PR #1738, opened 2023-10-10
ft: use Bandit instead of Cowboy as a web server

hauleth:


Ziinc (approved):

chasers:
Well that was easy!

hauleth:
Yup, way easier than I thought

--

Łukasz Niemier
***@***.***

PR #1750, opened 2023-10-12
chore: reorganise backend adaptor behaviour

hauleth:
This moves dynamic calls to inside of behaviour module to have coherent view on how behaviour will be used. Additionally this remove the code duplication that happened with template functions and reduce the compilation times (no compile time dependency).

### DIFF
--- a/lib/logflare/backends/adaptor.ex
+++ b/lib/logflare/backends/adaptor.ex
@@ -31,30 +31,13 @@ defmodule Logflare.Backends.Adaptor do
   """
   @callback validate_config(changeset :: Ecto.Changeset.t()) :: Ecto.Changeset.t()
 
-  defmacro __using__(_opts) do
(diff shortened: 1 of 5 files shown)


Ziinc (approved):

lib/logflare/backends/adaptor.ex
```diff
@@ -31,30 +31,9 @@ defmodule Logflare.Backends.Adaptor do
...
-
-      def cast_and_validate_config(params) do
-        params
-        |> cast_config()
-        |> validate_config()
-      end
-
-      defoverridable unquote(__MODULE__)
-    end
+  def cast_and_validate_config(mod, params) do
```
Ziinc, line 34: lets add specs & docs for this

Ziinc (approved):

PR #1755, opened 2023-10-12
chore: reorganise buffers behaviour

hauleth:


Ziinc (commented):
the MemoryBuffer actually needs to get rewritten to how the v1 pipeline works, such that we insert directly into Broadway's internal queue. Having the separate buffer leads to situation where we have 2 sequential queues.

hauleth:
Ok, so we can close it and I can checkout how to do that. The whole point of all these changes is to have behaviour callbacks calls in single place, so this will allow me to add telemetry in one place instead of tracing through whole codebase to see where I need add them, or to be wary of adding them to each implementation independently. So that is why you can see some of these appearing, as it is all preparation to Telemetry addition.

Ziinc:
It's fine to merge this first, just wanted to give you more context. I'll have a final once over on Monday, but I like the api changes. We had the idea of a persistent disk-backed buffer previously to reduce risk of dropping events on node failure/shutdown, hence the pluggable design.

lib/logflare/buffers/memory_buffer.ex
```diff
@@ -1,66 +1,58 @@
...
+
+  def start_link(opts \\ []) do
+    proc_name = opts[:proc_name] || opts[:name]
+
+    GenServer.start_link(__MODULE__, %{proc_name: proc_name}, opts)
+  end
+
+  @impl GenServer
+  def init(%{proc_name: proc_name}) do
+    {:ok, %__MODULE__{queue: :queue.new(), proc_name: proc_name || inspect(self())}}
```
Ziinc, line 44: can remove the inspect as the proc_name should always be an identifier.
```suggestion
    {:ok, %__MODULE__{queue: :queue.new(), proc_name: proc_name || self()}}
```

lib/logflare/buffers/memory_buffer.ex
```diff
@@ -1,66 +1,58 @@
...
   end
 
-  @impl Buffer
+  @impl Logflare.Buffers.Buffer
   def length(identifier) do
     GenServer.call(identifier, :length)
   end
 
+  # GenServer state and init callbacks
+  defstruct queue: nil, size: 0, proc_name: nil
```
Ziinc, line 34: let used [TypedStruct](https://hexdocs.pm/typed_struct_uberbrodt/TypedStruct.html)

Ziinc (changes requested):

lib/logflare/buffers/memory_buffer.ex
hauleth: My reason for `inspect/1` there was to use it in logs and telemetry events, but maybe having it as an PID and changing when needed isn't that bad.

lib/logflare/buffers/memory_buffer.ex
hauleth: Yeah, I never really used it, so I can check it out. But honestly in this place that shouldn't matter much as it is used only internally.

lib/logflare/buffers/memory_buffer.ex
hauleth: Also, I planned to reduce the amount of cases where we need *named processes* and instead work with just PIDs when possible.

Ziinc (approved):

PR #1756, opened 2023-10-13
Fix/secrets management scripts, ssl for otp26

Ziinc:
This PR fixes the secrets encryption/decryption scripts. 

Behaviour is decoupled when running `make encrypt.*` and `make decrypt.*` to prevent overriding of local unencrypted/uncommited changes.

This PR also adds in ecto ssl_opts, and also adds in the required staging/prod db ssl certs to get ssl properly supported on otp26. 

I have also updated naming of the google cloud keys, so that it is easier to work with the makefile patterns.


Ziinc:
@hauleth would be good if you could have a look at the makefile stuff in case i missed something.

Makefile
```diff
@@ -71,28 +75,37 @@ encrypt.dev: .dev.env.enc
...
+	.$$*.cert.pem.enc \
+ 	.$$*.db-client-cert.pem.enc \
+ 	.$$*.db-client-key.pem.enc \
+ 	.$$*.db-server-ca.pem.enc
 
 .PHONY: $(addprefix encrypt.,${envs})
+.PHONY: $(addprefix decrypt.,${envs})
 
-# OpenTelemetry Protobufs
+# OpenTelemetry Protobufse
```
filipecabaco, line 108: Minor error here

filipecabaco (approved):

Makefile
```diff
@@ -48,19 +48,23 @@ __start__: decrypt.${ENV}
...
-	@gcloud kms encrypt --ciphertext-file=$@ --plaintext-file=$(@:.enc=) \
+%.enc: FORCE
+	@gcloud kms encrypt --ciphertext-file=cloudbuild/$@ --plaintext-file=$(@:.enc=) \
 		--location=${GCLOUD_LOCATION} \
 		--keyring=${GCLOUD_KEYRING} \
 		--key=${GCLOUD_KEY} \
 		--project=${GCLOUD_PROJECT}
+	@echo "$@ has been encrypted"
+
+FORCE:
```
hauleth, line 67: I do not get what that is supposed to do.

Makefile
```diff
@@ -48,19 +48,23 @@ __start__: decrypt.${ENV}
 #     make decrypt.{dev,staging,prod} # Decrypt secrets for given environment
 #     make encrypt.{dev,staging,prod} # Encrypt secrets for given environment
 
-%: %.enc
-	@gcloud kms decrypt --ciphertext-file=$< --plaintext-file=$@ \
+.%.env .%.json %.pem .%.key: FORCE
+	@gcloud kms decrypt --ciphertext-file=cloudbuild/$@.enc --plaintext-file=$@ \
```
hauleth, line 52: Why not:

```suggestion
%: cloudbuild/%.enc
	@gcloud kms decrypt --ciphertext-file=$< --plaintext-file=$@ \
```

That would make more sense to me.

config/runtime.exs
```diff
@@ -227,14 +210,33 @@ cond do
...
+      certfile: "cert.pem",
+      keyfile: "cert.key",
+      verify: :verify_peer
+    ]
+end
 
-  config :logflare, ssl: ssl_opts
+if(
+  System.get_env("DB_SSL") == "true" && File.exists?("db-server-ca.pem") &&
+    File.exists?("db-client-ca.pem") && File.exists?("db-client-key.pem")
```
hauleth, line 224: I am pretty sure that checking for paths like that can fail in release, as these files may not be where we want them to be. Maybe instead use `DB_SSL` as a path to where these files will be stored? That should make the configuration more flexible.

Makefile
Ziinc: yes just gave this a spin and it works great 👍 . I had put in the `FORCE` target as the previous iteration was not triggering the encryption recipe when I changed the decrypted file, but this one works better with the change tracking.

config/runtime.exs
Ziinc: agreed. yes we can update the certfile loading method for both db and grpc, can do this in a separate PR.

PR #1769, opened 2023-10-23
fix: gzip payload chunking for large batches

Ziinc:
This PR fixes errors occured when inflating gzipped payloads that are chunked by :zlib, but are not correctly handled.

If gzip payloads are too large, an error will be raised.

### DIFF
--- a/test/logflare_web/plugs/compressed_body_reader_test.exs
+++ b/test/logflare_web/plugs/compressed_body_reader_test.exs
@@ -15,19 +15,40 @@ defmodule LogflareWeb.Plugs.CompressedBodyReaderTest do
   end
 
   property "with no `content-encoding` header data is passed through as is" do
-    check all(data <- binary()) do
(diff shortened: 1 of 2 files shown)


test/logflare_web/plugs/compressed_body_reader_test.exs
```diff
@@ -15,19 +15,40 @@ defmodule LogflareWeb.Plugs.CompressedBodyReaderTest do
...
     end
+
+    check all(data <- gen_max_chunk_payloads()) do
+      compressed = :zlib.gzip(data)
+      conn = conn(compressed, [{"content-encoding", "gzip"}])
+
+      assert_raise RuntimeError, "max chunks reached", fn ->
+        @subject.read_body(conn)
+      end
+    end
```
hauleth, line 40: I would extract that to separate test though.

hauleth (approved):

PR #1775, opened 2023-10-23
feat: add access token support for ingest path

Ziinc:
This PR adds in ingestion access token support, which will allow easier rolling of used api keys.
This sets the stage for sunsetting the legacy user.api_key column in the future, as all api usage would use access tokens instead.


lib/logflare_web/controllers/plugs/set_verify_source.ex
```diff
@@ -14,76 +16,22 @@ defmodule LogflareWeb.Plugs.SetVerifySource do
...
-
-        name && is_browser_path ->
-          Sources.get_by_and_preload(name: name)
-
-        true ->
-          nil
-      end
-
+    source = Sources.get_by_and_preload(id: id)
     user_authorized? = &(&1.user_id === user.id || user.admin)
```
hauleth, line 20: Why have this as lambda instead of function or just plain variable? I see it is used only once.

hauleth (approved):

lib/logflare_web/controllers/plugs/set_verify_source.ex
Ziinc: will remove.

chasers (approved):

PR #1778, opened 2023-10-24
feat: query composition

Ziinc:
Nesting endpoints within endpoints, or endpoints in alerts....
Screenshot 2023-10-25 at 2 32 47 AM
Screenshot 2023-10-25 at 2 32 42 AM

### DIFF
--- a/lib/logflare/endpoints.ex
+++ b/lib/logflare/endpoints.ex
@@ -10,6 +10,7 @@ defmodule Logflare.Endpoints do
   alias Logflare.Backends
   alias Logflare.Backends.Adaptor.PostgresAdaptor
   alias Logflare.SingleTenant
+  alias Logflare.Alerting
 
   import Ecto.Query
   @typep run_query_return :: {:ok, %{rows: [map()]}} | {:error, String.t()}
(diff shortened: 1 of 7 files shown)


lib/logflare/endpoints.ex
```diff
@@ -159,10 +160,21 @@ defmodule Logflare.Endpoints do
...
 
     with {:ok, declared_params} <- Logflare.Sql.parameters(query_string),
+         {:ok, expanded_query} <-
+           Logflare.Sql.expand_subqueries(
+             endpoint_query.language,
+             query_string,
+             endpoints ++ alerts
+           ),
+         transform_input <-
+           if(sandboxable && sql_param, do: {expanded_query, sql_param}, else: expanded_query),
```
hauleth, line 177:
```suggestion
         transform_input =
           if(sandboxable && sql_param, do: {expanded_query, sql_param}, else: expanded_query),
```

Or even move it into `do` block, as it is not needed in the `with` patterns.

lib/logflare/sql.ex
```diff
@@ -9,6 +9,81 @@ defmodule Logflare.Sql do
...
+  defp replace_names_with_subqueries({k, v}, data) when is_list(v) or is_map(v) do
+    {k, replace_names_with_subqueries(v, data)}
+  end
+
+  defp replace_names_with_subqueries(kv, data) when is_list(kv) do
+    Enum.map(kv, fn kv -> replace_names_with_subqueries(kv, data) end)
+  end
+
+  defp replace_names_with_subqueries(kv, data) when is_map(kv) do
+    Enum.map(kv, fn kv -> replace_names_with_subqueries(kv, data) end) |> Map.new()
```
hauleth, line 81: Use `Map.new/2` here instead of that.

```suggestion
    Map.new(kv, &replace_names_with_subqueries(&1, data))
```

lib/logflare/sql/composition.ex
hauleth, line 1: I think it was accidental addition, as this file is empty.

test/logflare/sql_test.exs
```diff
@@ -1,6 +1,7 @@
 defmodule Logflare.SqlTest do
   @moduledoc false
   use Logflare.DataCase, async: false
+  use ExUnitProperties
```
hauleth, line 4: Was that a mistake? I do not see any property test in that file.

hauleth (dismissed):

lib/logflare/endpoints.ex
Ziinc: we want to catch any errors in the subsequent `Sql.transform/3` call, hence why it can't be moved into `do`

chasers (approved):

PR #1783, opened 2023-10-25
chore: use gcloud trigger to build image

Ziinc:
This build-trigger based flow is to be used if we cannot resolve the cloud build v2 repo issues.
Cloud build v2 repo does not seem to be able to connect to the github repo, resulting in an INTERNAL_ERROR when triggering through `gcloud builds submit`.
This workaround is hacky-er as we need to `sleep` in order to wait for the build to be complete, before triggering another build for updating the instance templates & groups

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,14 +117,11 @@ grpc.protoc:
 
 deploy.staging.main:
 	@gcloud config set project logflare-staging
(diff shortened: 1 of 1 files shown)


Ziinc:
seems to be working now #1784

PR #1791, opened 2023-10-27
fix: become user when cookies are set

Ziinc:
This fixes the admin functionality for taking control of a user account


Ziinc:
merging in first so that i can upgrade staging cluster.

chasers:
Approved.

PR #1793, opened 2023-10-27
chore: revert sandbox query id removal

Ziinc:
This PR reverts the non-backwards compatible migration where the `sandbox_query_id` column was removed without backwards compat for rolling updates


chasers (commented):
Hmm ... is this going to fail since we have that field on prod now? 

Should be fine as long as it doesn't block the deploy but might need to delete the field right before deploying to prod.

Ziinc:
switched to use `add_if_not_exists`

chasers (approved):

PR #1797, opened 2023-11-01
fix: preload sources for VerifyApiAccess plug

Ziinc:
This PR fixes a bug in the rate limiter which requires `:sources` key to be preloaded for the User struct
image


Ziinc:
@chasers FYI, will merge this in first as it is blocking test cluster setup

chasers (approved):

PR #1800, opened 2023-11-01
fix: adds goth partitioning, telemetry, adjustment in application children  position, options tweaks

Ziinc:
Adds in the following Goth tweaks:
- telemetry span for fetch execution
- partitioning, had to hack around Goth's internal use of registry.
- shifted Goth up in the sup children
- reduced retry cap to 10s, added logging as well
- made server startup to be sync so that it fetches token on startup before continuing.

### DIFF
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -185,6 +185,8 @@ defmodule Logflare.Application do
               source: source,
               refresh_before: 60 * 15,
               prefetch: :sync,
+              http_client: &goth_finch_http_client/1,
+
               retry_delay: fn
                 n when n < 3 ->
                   1000
@@ -227,6 +229,18 @@ defmodule Logflare.Application do
   def config_change(changed, _new, removed) do
     LogflareWeb.Endpoint.config_change(changed, removed)
     :ok
+  end
+
+  # tell goth to use our finch pool
+  # https://github.com/peburrows/goth/blob/master/lib/goth/token.ex#L144
+  defp goth_finch_http_client(options) do
+    {method, options} = Keyword.pop!(options, :method)
+    {url, options} = Keyword.pop!(options, :url)
+    {headers, options} = Keyword.pop!(options, :headers)
+    {body, options} = Keyword.pop!(options, :body)
+
+    Finch.build(method, url, headers, body)
+    |> Finch.request(Logflare.FinchDefault, options)
   end
 
   defp finch_pools do
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -32,38 +32,38 @@ defmodule Logflare.Application do
 
   defp get_children(:test) do
     finch_pools() ++
-    [
-      ContextCache,
-      Users.Cache,
-      Sources.Cache,
-      Billing.Cache,
-      SourceSchemas.Cache,
-      PubSubRates.Cache,
-      Logs.LogEvents.Cache,
-      Logs.RejectedLogEvents,
-      {Phoenix.PubSub, name: Logflare.PubSub},
-      Logflare.Repo,
-      {Registry,
-       name: Logflare.V1SourceRegistry, keys: :unique, partitions: System.schedulers_online()},
-      {Registry,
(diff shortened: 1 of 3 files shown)


lib/logflare/application.ex
```diff
@@ -90,89 +81,136 @@ defmodule Logflare.Application do
...
+         strategy: :one_for_one, name: Logflare.Backends.Adaptor.PostgresAdaptor.PgRepoSupervisor},
+        {Registry,
+         name: Logflare.Backends.SourceRegistry,
+         keys: :unique,
+         partitions: System.schedulers_online()},
+        {Registry, name: Logflare.Backends.SourceDispatcher, keys: :duplicate},
+
+        # citrine scheduler for alerts
+        Logflare.AlertsScheduler
+      ] ++ common_children()
```
chasers, line 161: Finch should start before Goth yeah?

lib/logflare/application.ex
```diff
@@ -90,89 +81,136 @@ defmodule Logflare.Application do
...
+        nil ->
+          []
+
+        json ->
+          # Setup Goth for GCP connections
+          credentials = Jason.decode!(json)
+          scopes = ["https://www.googleapis.com/auth/cloud-platform"]
+          source = {:service_account, credentials, scopes: scopes}
+
+          spec =
```
chasers, line 176: I think we need to tell it to specificaly use our `Logflare.FinchDefault` pool?

https://github.com/peburrows/goth/blob/master/lib/goth/token.ex#L209C13-L209C13de

lib/logflare/google/bigquery/gen_utils/gen_utils.ex
```diff
@@ -72,7 +72,15 @@ defmodule Logflare.Google.BigQuery.GenUtils do
   @typep conn_type :: :ingest | :query | :default
   @spec get_conn(conn_type()) :: Tesla.Env.client()
   def get_conn(conn_type \\ :default) do
-    Goth.fetch(Logflare.Goth)
+    # use pid as the partition hash
+    partition_count = System.schedulers_online()
+    partition = :erlang.phash2(self(), partition_count)
+
+    metadata = %{partition_count: partition_count}
```
hauleth, line 79: Why include that in metadata? Didn't you want to use `partition` there instead?

lib/logflare/application.ex
```diff
@@ -90,89 +81,136 @@ defmodule Logflare.Application do
...
-      # Context Caches
-      ContextCache,
-      Users.Cache,
-      Sources.Cache,
-      Billing.Cache,
-      SourceSchemas.Cache,
-      PubSubRates.Cache,
-      Logs.LogEvents.Cache,
+    # set goth early in the supervision tree
+    conditional_children() ++
```
hauleth, line 85: Maybe it is just me, but I find `Enum.concat/1` here more readable than `++`.

lib/logflare/application.ex
Ziinc: done

lib/logflare/application.ex
Ziinc: i think for smaller lists it would be alright, but for such a large list, the nesting of lists within lists would make it harder to read. i think its better for smaller lists though

hauleth (approved):

chasers:
@Ziinc I think this is a good stab at it. If we are seeing a Goth issue it should be logged but if one of the Goths is having issues it will continuously affect those same sources, which could be bad.

Future work here could be to make a process pull from multi-goths and cache the tokens for other procs to pull from. So we have a process to ensure there are multiple good tokens to use by any of our procs. And then we control the caching and access of that cache to ensure it's efficient.

chasers (approved):

chasers:
Approved assuming the tests pass that is.

wojtekmach:
@chasers Goth 1.3+ was designed to allow having multiple instances. We have an example here: https://github.com/peburrows/goth/blob/v1.4.2/README.md?plain=1#L65:L83. If that doesn't work, it's definitely a bug.

chasers:
@wojtekmach thanks! We're not entirely sure what the root cause is of what we're seeing.

Will report back!!

PR #1809, opened 2023-11-03
chore: allow versioned building on workflow dispatch

Ziinc:
fix for triggering rebuilds of versioned image on workflow dispatch.
Currently we don't have v1.5.4 image built yet on dockerhub, causing downstream cloudbuild workflow to fail.


Ziinc:
@chasers FYI

PR #1814, opened 2023-11-06
chore: use e2-standard-4 for load test cluster

Ziinc:
e2-standard-8 is quite expensive (almost US$200 per node per month, lets try with e2-standard-4 first before going with higher vcpus.

also adds in some ci improvements


hauleth (approved):
LGTM

PR #1816, opened 2023-11-08
fix: fetch cache keys only once

hauleth:
This change makes `ContextCache.bust_keys` function fetch keys from the cache only once instead repeatedly for each busted key individually.

This is still not perfect implementation, but at least should reduce the problem with `CacheBuster` stalling and making its inbox grow uncontrollably.

### DIFF
--- a/lib/logflare/cache_buster.ex
+++ b/lib/logflare/cache_buster.ex
@@ -20,9 +20,12 @@ defmodule Logflare.CacheBuster do
   end
 
   def handle_info(%Transaction{changes: changes}, state) do
-    for record <- changes do
-      handle_record(record)
+    for record <- changes,
+        record = handle_record(record),
+        record != :noop do
+      record
     end
+    |> ContextCache.bust_keys()
 
     {:noreply, state}
   end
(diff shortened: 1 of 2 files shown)


chasers (approved):

lib/logflare/cache_buster.ex
```diff
@@ -20,9 +20,12 @@ defmodule Logflare.CacheBuster do
   end
 
   def handle_info(%Transaction{changes: changes}, state) do
-    for record <- changes do
-      handle_record(record)
+    for record <- changes,
+        record = handle_record(record),
```
Ziinc, line 24: Can match on the tuple instead and remove the noop check.

lib/logflare/context_cache.ex
```diff
@@ -37,29 +37,23 @@ defmodule Logflare.ContextCache do
...
-          {:ok, true} ->
-            Cachex.del(@cache, key)
-
-          {:ok, nil} ->
-            :noop
-        end
-
-      _key ->
-        :noop
+    Enum.each(keys, fn {token, cache_key} = key ->
```
Ziinc, line 43: Let's add in some telemetry here to track the number of keys and exec duration, if it is large then Task.async_stream or Flow might help.

Ziinc:
@hauleth ContextCache also doesn't have any test coverage, adding some in would be rad

lib/logflare/cache_buster.ex
hauleth: No, I cannot, as it would change behaviour (there will be match failure exception on `:noop`).

lib/logflare/context_cache.ex
hauleth: That will be resolved in next PR. We discussed that on Slack with @chasers already.

PR #1820, opened 2023-11-13
fix: avg rates should be ints

chasers:
Noticed long floats on staging. This fixes it.


lib/logflare/source/rate_counter_server.ex
```diff
@@ -268,14 +268,28 @@ defmodule Logflare.Source.RateCounterServer do
...
+  """
+
+  @spec stats([integer()]) :: %{avg: integer(), sum: integer()}
   def stats(xs) when is_list(xs) do
     {total, count} =
       Enum.reduce(xs, {0, 0}, fn v, {total, count} ->
         {total + v, count + 1}
       end)
 
+    avg = Kernel.ceil(total / count)
```
hauleth, line 289: Ceil there is correct function? I would need to run benchmark whether this or `div(total + count - 1, count)` is faster.

Ziinc, line 289: Need some  checks for div by zero for either numerator denominator

Ziinc: I think it is for rounding

Ziinc (changes requested):
Needs an if check for division

lib/logflare/source/rate_counter_server.ex
```diff
@@ -268,14 +268,28 @@ defmodule Logflare.Source.RateCounterServer do
     Sources.Counters.get_inserts(source_id)
   end
 
+  @doc """
+  Takes a list of integers representing per second counts of events
+  of a source and returns the `avg` and `sum` of those counts.
+
+  ## Examples
+
+      iex> Logflare.Source.RateCounterServer.stats([12,5,30,5,1000])
```
Ziinc, line 277: Adding in a case for empty list should cover the div by zero issue

lib/logflare/source/rate_counter_server.ex
chasers: Ya it’s for rounding. You can take this over to get it in the latest version.

chasers: Long floats were on the dashboard.

hauleth: @Ziinc I know that this is for rounding. Just my point is whether `trunc/1` (faster) or `round/1` (more correct) would not work better.

PR #1825, opened 2023-11-14
fix: :buffer_counter_not_found bug

Ziinc:
This error gets thrown when the RLS does not start up fast enough for incoming requests. This change ensures that the startup of the critical genservers are performed in the init and is blocking.

### DIFF
--- a/lib/logflare/source/recent_logs_server.ex
+++ b/lib/logflare/source/recent_logs_server.ex
@@ -51,21 +51,6 @@ defmodule Logflare.Source.RecentLogsServer do
   @broadcast_every 500
   @pool_size Application.compile_env(:logflare, Logflare.PubSub)[:pool_size]
 
-  def start_link(%__MODULE__{source_id: source_id} = rls) when is_atom(source_id) do
-    GenServer.start_link(__MODULE__, rls, name: Source.Supervisor.via(__MODULE__, source_id))
-  end
-
-  ## Client
-  @spec init(RLS.t()) :: {:ok, RLS.t(), {:continue, :boot}}
-  def init(rls) do
-    Process.flag(:trap_exit, true)
-
-    touch()
-    broadcast()
-
-    {:ok, rls, {:continue, :boot}}
-  end
-
   @spec push(LE.t()) :: :ok
   def push(%LE{source: %Source{token: source_id}} = log_event) do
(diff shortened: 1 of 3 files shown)


lib/logflare/source/recent_logs_server.ex
```diff
@@ -51,21 +51,6 @@ defmodule Logflare.Source.RecentLogsServer do
...
   @pool_size Application.compile_env(:logflare, Logflare.PubSub)[:pool_size]
 
-  def start_link(%__MODULE__{source_id: source_id} = rls) when is_atom(source_id) do
-    GenServer.start_link(__MODULE__, rls, name: Source.Supervisor.via(__MODULE__, source_id))
-  end
-
-  ## Client
-  @spec init(RLS.t()) :: {:ok, RLS.t(), {:continue, :boot}}
-  def init(rls) do
-    Process.flag(:trap_exit, true)
```
chasers, line 61: Not trapping exits anymore?

chasers: the `terminate` callback won't fire now.

lib/logflare/source/recent_logs_server.ex
```diff
@@ -237,7 +240,7 @@ defmodule Logflare.Source.RecentLogsServer do
 
   def terminate(reason, state) do
     # Do Shutdown Stuff
-    Logger.info("Going Down - #{inspect(reason)} - #{state.source_id}", %{
+    Logger.error("Going Down - #{inspect(reason)} - #{state.source_id}", %{
```
chasers, line 243: I had `info` here because we may WANT to shut down an RLS process. When it gets deleted, restarted etc.

Not necessarily an error.

chasers (approved):

chasers:
@Ziinc approving for async's sake. Only thing I'm kind of concerned about is the exit trapping.

lib/logflare/source/recent_logs_server.ex
Ziinc: ah mistakenly removed. adding back

lib/logflare/source/recent_logs_server.ex
Ziinc: Gotcha. switching back

Ziinc:
adjusted based on review changes. will get this deployed today.

PR #1826, opened 2023-11-15
feat: use Postgres for clustering strategy

Ziinc:
This PR uses postgres for clustering, which will base the notify channel name on the magic cookie.


lib/logflare/cluster/postgres_strategy.ex
```diff
@@ -0,0 +1,117 @@
+defmodule Logflare.Cluster.PostgresStrategy do
```
hauleth, line 1: Maybe we could extract that and publish as a separate library? I think that quite a few people may find it useful.

hauleth (approved):

lib/logflare/cluster/postgres_strategy.ex
chasers: We definitely should.  Realtime and Supavisor are using it.

lib/logflare/cluster/postgres_strategy.ex
```diff
@@ -0,0 +1,117 @@
...
+
+  When a node comes online, it begins to broadcast its name in a "heartbeat" message to the channel. All other nodes that receive this message attempt to connect to it.
+
+  This strategy does not check connectivity between nodes and does not disconnect them
+
+  ## Options
+
+  * `heartbeat_interval` - The interval at which to send heartbeat messages in milliseconds (optional; default: 5_000)
+
+  The magic cookie is used as the channel name.
```
chasers, line 15: Did we check constraints around this from Postgres?

chasers: In Supavisor and Realtime this is passed in with the config.

If we want to use the magic cookie we should probably move this and do it there.

Ziinc: from my manual testing, pg doesn't accept a channel name with `:` prefix, hence why i converted it to use `Atom.to_string/1` to get the stringified atom. `Node.get_cookie/0` seems to return the cookie as an atom.
I can make this more foolproof though with some alphanumeric conversions

chasers: Roger.  Maybe default behavior (for the eventual library) is to use the cleaned magic cookie and override (or additionally) subscribe to the channel(s) passed in the config.

chasers: And yeah just double check because the cookie on dev could not be as long or complex as prod.

chasers (approved):

lib/logflare/cluster/postgres_strategy.ex
Ziinc: task [here](https://www.notion.so/supabase/Logflare-Clustering-Extract-out-Postgres-Clustering-Strategy-to-its-own-separate-library-42d88bf841ad4338a19daac2aed11e25?pvs=4), backlogged

PR #1830, opened 2023-11-15
fix: context cache fixes

chasers:
- [x] initial test

WIP


lib/logflare/cache_buster.ex
```diff
@@ -21,6 +21,7 @@ defmodule Logflare.CacheBuster do
 
   def handle_info(%Transaction{changes: changes}, state) do
     for record <- changes,
+        :ok = :telemetry.execute([:logflare, :context_cache, :handle_record], %{count: 1}),
```
hauleth, line 24: We could add metadata there about type of evicted record.

lib/telemetry.ex
```diff
@@ -84,7 +84,8 @@ defmodule Logflare.Telemetry do
         tags: [:processor],
         unit: {:native, :millisecond}
       ),
-      counter("logflare.context_cache.busted.count")
+      counter("logflare.context_cache.busted.count"),
+      counter("logflare.context_cache.handle_record.count")
```
hauleth, line 88: Here we could record count of evicted entries by type.

hauleth (approved):

test/logflare/context_cache_test.exs
```diff
@@ -0,0 +1,38 @@
...
+    %{source: source}
+  end
+
+  test "ContextCache works", %{source: source} do
+    context = Sources
+    fun_arity = {:get_by, 1}
+    args = [[token: source.token]]
+
+    # Cache our function call results
+    cached_source = ContextCache.apply_fun(context, fun_arity, args)
```
Ziinc, line 22: apply_fun could probably be simplfiied to receive `apply_fun({Sources, :get_by, [[token: source.token]]} )`, since the arity is implied through the arg count.

Ziinc: or even `apply_fun(Sources, :get_by, [...])`

test/logflare/context_cache_test.exs
```diff
@@ -0,0 +1,38 @@
...
+    assert %Logflare.Source{} = cached_source
+
+    # Make sure we have it in our context cache
+    assert {:cached, %Logflare.Source{}} = Cachex.get!(cache_name, cache_key)
+
+    # Should do this with the wal instead at some point
+    assert {:ok, :busted} = ContextCache.bust_keys([{context, cached_source.id}])
+
+    # Make sure we don't have it after it's busted
+    assert is_nil(Cachex.get!(cache_name, cache_key))
```
Ziinc, line 36:
```suggestion
    refute Cachex.get!(Sources.Cache, cache_key)
```

test/logflare/context_cache_test.exs
```diff
@@ -0,0 +1,38 @@
...
+  test "ContextCache works", %{source: source} do
+    context = Sources
+    fun_arity = {:get_by, 1}
+    args = [[token: source.token]]
+
+    # Cache our function call results
+    cached_source = ContextCache.apply_fun(context, fun_arity, args)
+
+    cache_key = {fun_arity, args}
+    cache_name = ContextCache.cache_name(context)
```
Ziinc, line 25:
```suggestion
    cache_name = ContextCache.cache_name(context)
    assert cache_name == Sources.Cache
```

Ziinc (approved):

lib/logflare/cache_buster.ex
hauleth: Expensiveness of both approaches is disputable. Many calls should not be super expensive there, but it depends on the handlers attached to Telemetry. `%{count: length(changes)}` needs to traverse whole list again, and additionally do not allows us to group by record type.

test/logflare/context_cache_test.exs
hauleth: I think I prefer `assert is_nil` than `refute` here. Especially that `Cachex.get!/2` call can return `false` as a completely correct response.

lib/logflare/cache_buster.ex
Ziinc: agreed, having it as is would make sense if we add in the record type.

hauleth (changes requested):
We should gather information about evicted entries by their types.

lib/logflare/cache_buster.ex
chasers: Added!

lib/telemetry.ex
chasers: Added

test/logflare/context_cache_test.exs
chasers: Pull the arity out of the cache key but you kind of want it here because it makes calling apply_fun easier.

lib/logflare/context_cache.ex
```diff
@@ -32,43 +41,51 @@ defmodule Logflare.ContextCache do
...
+
+    Enum.each(keys, fn {token, phash_key} = key ->
+      with true <- token in values,
+           {context, _} = token,
+           context_cache = cache_name(context),
+           {:ok, cache_key} = Cachex.get(@cache, key),
+           {:ok, true} <- Cachex.del(context_cache, cache_key) do
+        Cachex.del(@cache, key)
+      end
+    end)
```
hauleth: I wonder if counting how many entries were in fact evicted would be useful. Like:
```suggestion
    busted = Enum.count(keys, fn {token, phash_key} = key ->
      with true <- token in values,
           {context, _} = token,
           context_cache = cache_name(context),
           {:ok, cache_key} = Cachex.get(@cache, key),
           {:ok, true} <- Cachex.del(context_cache, cache_key) do
        Cachex.del(@cache, key)
        
        true
      else
        _ -> false
      end
    end)
    
    :telemetry.execute([:logflare, :context_cache], %{busted: busted}, %{})
```

lib/logflare/context_cache.ex
```diff
@@ -32,43 +41,51 @@ defmodule Logflare.ContextCache do
...
-  defp index_keys(context, cache_key, value) do
-    keys_key = {{context, select_key(value)}, cache_key}
+  @spec cache_name(atom()) :: atom()
+  def cache_name(context) do
+    Module.concat(context, Cache)
+  end
 
-    {:ok, key} = Cachex.get(@cache, keys_key)
+  defp index_keys(context, cache_key, value) do
+    keys_key = {{context, select_key(value)}, :erlang.phash2(cache_key)}
```
hauleth, line 86: What will happen if there will be conflict?

hauleth (approved):
Few ideas, other than that LGTM

lib/logflare/context_cache.ex
```diff
@@ -32,43 +41,51 @@ defmodule Logflare.ContextCache do
...
-      end)
-
-    :telemetry.execute(
-      [:logflare, :context_cache, :busted],
-      %{count: total},
-      %{}
-    )
+    for {context, primary_key} <- values do
+      match_spec =
+        {:entry, {{context, primary_key}, :_}, :_, :_, :"$1"}
```
hauleth, line 65: Maybe using some library like [Matcha](https://github.com/christhekeele/matcha) or [ex2ms](https://github.com/ericmj/ex2ms) would make it a little bit cleaner?

lib/logflare/context_cache.ex
```diff
@@ -32,43 +41,51 @@ defmodule Logflare.ContextCache do
...
+
+      keys = :ets.match(@cache, match_spec)
```
hauleth, line 67: Is such direct match against ETS table handled by `Cachex` allowed by the library?

lib/logflare/context_cache.ex
chasers: I think there is likely enough info here that it won't be. 

`{{Logflare.Users, 2334}, 12342335}` and phash2 is a combo of the function name and all arguments.

hauleth: And what if there will be something? What will be the impact? I think that will mean that the cache will be invalidated to early, isn't it? But we probably also should ad TTL to the cache entries just to be sure that we evict not needed entries that are super old.

lib/logflare/context_cache.ex
chasers: Yeah with `Cachex.Query.raw` but I need to figure out how to do that exactly. It's a thin wrapper over match specs it says.

lib/logflare/context_cache.ex
chasers: If there was a conflict it would override the cached results with the wrong e.g. User actually prior to hitting this bust code.

chasers: Well no it would reverse index the wrong key. So yeah it would end up busting the wrong record which is fine. And yeah I will add a TTL to all of it just in case.

lib/logflare/context_cache.ex
chasers: I was actually just thinking we should really be using fewer libs which is why I went direct 😭

hauleth: I agree with that sentiment. I was just wondering about others. I am pretty proficient with reading match specs, but others may not be.

lib/logflare/context_cache.ex
```diff
@@ -32,43 +41,52 @@ defmodule Logflare.ContextCache do
...
+        Cachex.del(context_cache, v)
+        Cachex.del(@cache, k)
       end)
-
-    :telemetry.execute(
-      [:logflare, :context_cache, :busted],
-      %{count: total},
-      %{}
-    )
+    end
```
hauleth, line 74: I wonder if we maybe can do everything in one step. So create match spec list for all possible matched values, and then evict them in single ETS query. I wonder if that would be faster rather than doing separate queries. I would need to benchmark that.

chasers: I think we could but right now the cache for the context (`Logflare.Users.Cache`) is a different ETS table than the reverse index (`Logflare.ContextCache`) so we'd need to refactor to use one ETS table for all of them (and probably partition them).

test/logflare/context_cache_test.exs
```diff
@@ -0,0 +1,34 @@
...
+    insert(:plan, name: "Free")
+    source = insert(:source, user: user)
+    args = [token: source.token]
+    source = Sources.Cache.get_by(args)
+    fun = :get_by
+    cache_key = {fun, [args]}
+    %{source: source, cache_key: cache_key}
+  end
+
+  test "cache_name/1", %{source: source} do
```
hauleth, line 20: We do not use `source` there.

test/logflare/context_cache_test.exs
```diff
@@ -0,0 +1,34 @@
...
+    assert Sources.Cache == ContextCache.cache_name(Sources)
+  end
+
+  test "apply_fun/3", %{source: source, cache_key: cache_key} do
```
hauleth, line 24: We do not use `source` there.

hauleth (approved):

test/logflare/context_cache_test.exs
chasers: Doh

PR #1837, opened 2023-11-21
chore: remove err trapping

Ziinc:
Seems like the error trapping breaks versioned retagging ci
https://github.com/Logflare/logflare/actions/runs/6948416156
might be due to the `-e` option set on GH actions by default, but since it is not really necessary, removing.


Ziinc:
Merging in to fix versioned staging deploys

PR #1839, opened 2023-11-21
feat: persist endpoint query params form values

Ziinc:
This persists endpoint query params form values for testing usability.


Ziinc:
currently failing due to compilation check. should pass with https://github.com/Logflare/logflare/pull/1842 merged.

chasers (approved):

PR #1841, opened 2023-11-22
feat: switch cluster warning to telemetry poller

Ziinc:
This PR moves the cluster warning to a period 10s check, together with a telemetry event.
This reduces log emission on high request load, as cluster warning log was being emitted on each rate limit check


Ziinc:
failing due to https://github.com/Logflare/logflare/pull/1842

PR #1843, opened 2023-11-22
fix: revert batch size

Ziinc:
This reverts the broadway max batch size, as discussed. We will add more metrics around this before deciding to increase the batch size as a means for increasing throughput.


Ziinc:
failing due to https://github.com/Logflare/logflare/pull/1842

chasers (approved):

PR #1845, opened 2023-11-22
feat: increase max search row return

Ziinc:
This PR increase the max resultset of the search query. It is a first step for adding in infinite scroll style pagination.
If amount of scrolling is a concern, we can drop it down to 500. 
I think having it at 1k reduces the amount of queries that a user would possibly make.


hauleth (approved):

Ziinc:
Closing this for now as it might degrade search performance due to payload size from BQ. Will revisit this again when search page gets improved.

PR #1853, opened 2023-11-23
chore: push all tags for retagging script

Ziinc:


Ziinc:
buildx step already pushes image, this is redundant

PR #1857, opened 2023-11-23
chore: intermediate docker image should use sha not version

Ziinc:


Ziinc:
@chasers Merging in first to test.

PR #1861, opened 2023-11-28
fix: main cluster deployment shitdown script

Ziinc:
Fixes test cluster deploment.


Ziinc:
@chasers FYI merging.

PR #1862, opened 2023-11-29
Support running on AWS Fargate

mats16:
It makes Logflare to detect host ip on AWS Fargate.

### DIFF
--- a/rel/env.sh.eex
+++ b/rel/env.sh.eex
@@ -14,6 +14,13 @@
 # RELEASE_DISTRIBUTION variable below.
 # export RELEASE_DISTRIBUTION=name
 # export RELEASE_NODE=<%= @release.name %>@127.0.0.1
-export RELEASE_DISTRIBUTION=name
+
 HOST="${LOGFLARE_NODE_HOST:-127.0.0.1}"
+
+# for AWS ECS Fargate
+if [ "$AWS_EXECUTION_ENV" = "AWS_ECS_FARGATE" ]; then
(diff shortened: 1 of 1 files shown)


Ziinc:
hi @mats16 , you should be able to set the hostname through `LOGFLARE_NODE_HOST` env var through a startup script or by modifying your container startup command. Is that not possible? we would prefer not to add in hosting platform specific code.

mats16:
HI @Ziinc

It is difficult to set the hostname through `LOGFLARE_NODE_HOST` on some environments such as Kubernetes or Amazon ECS, Fly.io.
This is because the hostname is defined at container startup.

In fact, [supabase/realtime](https://github.com/supabase/realtime/blob/main/rel/env.sh.eex) has settings for each environment.
https://github.com/supabase/realtime/blob/main/rel/env.sh.eex

I think it is better to use `$(hostname -i)` when `LOGFLARE_NODE_HOST` is not configured. 
(It means it is not just for AWS Fargate.)
However, changes in default behavior have the impact.

Ziinc:
Supporting platform specific code is not on Logflare's list of priorities at the moment, I'll leave this open so that we can circle back in the future.

Doing something like 
`LOGFLARE_NODE_HOST=$(hostname -i) ./run.sh` in the entry command is the recommended solution.

Ziinc:
Closing as wont fix. 
@mats16 would be great if you could PR into the docs file on self-hosting to add in some info on this. we can contribute our GCP scripts as an example as well.

PR #1863, opened 2023-11-29
chore: fix prod gce container registry path  to match staging

Ziinc:
This adjusts the cloudbuild substitutions to match staging, as seems like curly braces are needed for dynamic substitutions.


Ziinc:
@chasers fyi, merging in first for canary

chasers:
Approved

PR #1864, opened 2023-11-29
chore: fix shutdown scripts, drop min cluster size for prod, and increase staging min ready time

Ziinc:
Increase min ready time for staging to let nodes be more stable before shutting down old nodes.

Drop cluster size to allow for 2 node canarying.

fix shutdown script which is breaking the node shutdown.


Ziinc:
@chasers fyi

PR #1865, opened 2023-11-29
chore: gcloud should ignore deps folder

Ziinc:
deps folder is not ignored, so triggering the build locally causes issues.


Ziinc:
@chasers FYI

PR #1866, opened 2023-11-29
fix: don't hit SourceSchemas cache here

chasers:
Bug: if source schema gets cached incorrectly it can prevent you from querying a field even though the field is present when you view the schema.

This should fix that.


Ziinc:
@chasers with the latest cachebuster changes, is incorrect caching still an issue?

Ziinc (approved):

PR #1868, opened 2023-11-30
perf: reduce http2 pools to reduce chance of server throttling.

Ziinc:
Reduction of http2 pool count reduces chances of connection throttling
image


chasers:
We're turning too many knobs before we even have this version on production.

Let's get stable on prod for at least a week and then we can pull this in.

chasers (approved):

Ziinc:
yea we can stagger this out, no hurry on this one.

PR #1875, opened 2023-12-04
feat: cross-node cache warming

Ziinc:
This PR adds cross-node cache warming to Cachex caches.

- [x] persistent node selection, to prevent imports from different nodes
- [x] warm on cache startup in non-blocking way
- [x] stream batches to CacheWarmer genserver, avoiding costly full table export.

Possible future extensions:
- add to health check, to prevent request hits before cache is fully warmed.
- Source.Supervisor boot should wait for cache warming to be complete.


chasers:
Commenting for comms but lets wait on this one. Lots of edge cases to consider.

lib/logflare/cluster/cache_warmer.ex
```diff
@@ -0,0 +1,71 @@
...
+      else
+        Node.list()
+        |> Enum.map(fn node ->
+          case :rpc.call(node, Cachex, :count, [cache]) do
+            {:ok, count} when count > 0 -> node
+            _ -> nil
+          end
+        end)
+        |> Enum.filter(&(&1 != nil))
+        |> Enum.sort_by(&Atom.to_string/1)
```
hauleth, line 39: IIRC `Atom.to_string` call is not needed, as it is default behaviour of comparing atoms.

lib/logflare/cluster/cache_warmer.ex
```diff
@@ -0,0 +1,71 @@
...
+        |> List.first()
```
hauleth, line 40: Sorting and then fetching first is the same as finding minimum.

hauleth, line 40:
```suggestion
        |> Enum.filter(&(&1 != nil))
        |> Enum.min()
```

Ziinc:
Closing this, as:
1. db upgrades as allowed us to spin up multiple nodes at concurrently now. This makes this PR obsolete as it is built on the assumption of sequential node startups.
2. a more scalable and safer way would be to broadcast db data fetching across the cluster, instead of sharing cached values.
3. caches should pre-warm themselves

PR #1878, opened 2023-12-08
feat: Partners.UserController metadata filtering

Ziinc:
- renames AccountController to UserController to match resource name
- adds GET users metadata filtering, relies on #1879


lib/logflare/user.ex
```diff
@@ -61,7 +61,7 @@ defmodule Logflare.User do
 
   typed_schema "users" do
     field :email, :string
-    field :provider, :string
+    field :provider, :string, default: "email"
```
hauleth, line 64: Shouldn't that be an enum (user space, in DB it still can be string).

hauleth (approved):

PR #1880, opened 2023-12-08
feat: access tokens management api

Ziinc:
Adds in management api routes for access token management, needed for platform integration.

- Only private/public scopes are permitted.
- No updating of keys are permitted.


lib/logflare_web/controllers/api/access_token_controller.ex
```diff
@@ -0,0 +1,82 @@
...
+      {:owner, false} ->
+        conn
+        |> Plug.Conn.send_resp(404, [])
+        |> Plug.Conn.halt()
+    end
+  end
+
+  defp maybe_redact_token(%{scopes: scopes} = t) do
+    cond do
+      not (scopes =~ "public") -> %{t | token: nil}
```
hauleth, line 78: There should be `=~`? Because that feels a little bit wonky.

Ziinc: `scopes` would include resource scoping in the future as well, such as `public source:{id}` and `public endpoint:{id}`

hauleth: Then I would do parsing to not accidentally match on value like `public-access` or similar. And that would be much cleaner when parsed IMHO

hauleth, line 78:
```suggestion
      # XXX: Replace with scopes parser
      not (scopes =~ "public") -> %{t | token: nil}
```

Ziinc: added a simple scopes str split for doing the exact scope check.

lib/logflare_web/controllers/api/access_token_controller.ex
```diff
@@ -0,0 +1,82 @@
...
+         {:owner, true} <- {:owner, access_token.resource_owner_id == user.id},
+         :ok <- Auth.revoke_access_token(user, token) do
+      conn
+      |> Plug.Conn.send_resp(204, [])
+      |> Plug.Conn.halt()
+    else
+      {:owner, false} ->
+        conn
+        |> Plug.Conn.send_resp(404, [])
+        |> Plug.Conn.halt()
```
hauleth, line 72: I would do something like returning `{:error, :not_found}` there and then handle it in `fallback_controller`.

lib/logflare_web/controllers/api/access_token_controller.ex
```diff
@@ -0,0 +1,82 @@
...
+    }
+  )
+
+  def delete(%{assigns: %{user: user}} = conn, %{"token" => token}) do
+    with access_token <- Auth.get_access_token(user, token),
+         {:owner, true} <- {:owner, access_token.resource_owner_id == user.id},
+         :ok <- Auth.revoke_access_token(user, token) do
+      conn
+      |> Plug.Conn.send_resp(204, [])
+      |> Plug.Conn.halt()
```
hauleth, line 67:
```suggestion
      send_resp(conn, 204, [])
```
There should be no need to call `Plug.Conn.halt/1` there.

lib/logflare_web/controllers/api/access_token_controller.ex
```diff
@@ -0,0 +1,82 @@
...
+    scopes_input = Map.get(params, "scopes", "")
+
+    with {:scopes, true} <- {:scopes, not (scopes_input =~ "partner")},
+         {:ok, access_token} <- Auth.create_access_token(user, params) do
+      conn
+      |> put_status(201)
+      |> json(access_token)
+    else
+      {:scopes, false} ->
+        FallbackController.call(conn, {:error, :unauthorized})
```
hauleth, line 48:
```suggestion
        {:error, :unauthorized}
```

lib/logflare_web/controllers/api/access_token_controller.ex
```diff
@@ -0,0 +1,82 @@
...
+        |> Plug.Conn.send_resp(404, [])
+        |> Plug.Conn.halt()
+    end
+  end
+
+  defp maybe_redact_token(%{scopes: scopes} = t) do
+    cond do
+      "public" not in String.split(scopes) -> %{t | token: nil}
+      true -> t
+    end
```
hauleth, line 80:
```suggestion
    if "public" not in String.split(scopes) do
      %{t | token: nil}
    else
      t
    end
```

lib/logflare_web/controllers/api/access_token_controller.ex
```diff
@@ -0,0 +1,73 @@
...
+  )
+
+  def delete(%{assigns: %{user: user}} = conn, %{"token" => token}) do
+    with access_token <- Auth.get_access_token(user, token),
+         {:owner, true} <- {:owner, access_token.resource_owner_id == user.id},
+         :ok <- Auth.revoke_access_token(user, token) do
+      Plug.Conn.send_resp(conn, 204, [])
+    else
+      {:owner, false} -> {:error, :not_found}
+    end
```
hauleth, line 67:
```suggestion
    access_token = Auth.get_access_token(user, token)
    if access_token.resource_owner_id == user.id do
      :ok = Auth.revoke_access_token(user, token)
      Plug.Conn.send_resp(conn, 204, [])
    else
      {:error, :not_found}
    end
```

lib/logflare_web/controllers/api/access_token_controller.ex
```diff
@@ -0,0 +1,73 @@
...
+    scopes_input = Map.get(params, "scopes", "")
+
+    with {:scopes, true} <- {:scopes, not (scopes_input =~ "partner")},
+         {:ok, access_token} <- Auth.create_access_token(user, params) do
+      conn
+      |> put_status(201)
+      |> json(access_token)
+    else
+      {:scopes, false} -> {:error, :unauthorized}
+    end
```
hauleth, line 48:
```suggestion
    if "partner" in String.split(scopes_input) do
      {:error, :unauthorized}
    else
      {:ok, access_token} = Auth.create_access_token(user, params)
      
      conn
      |> put_status(201)
      |> json(access_token)
    end
```

lib/logflare_web/controllers/api/fallback_controller.ex
```diff
@@ -17,7 +17,9 @@ defmodule LogflareWeb.Api.FallbackController do
     |> halt()
   end
 
-  def call(conn, nil) do
+  def call(conn, nil), do: call(conn, {:error, :not_found})
```
hauleth, line 20: I would add task/todo comment to hunt occurrences of that and replace it with `{:error, :not_found}` tuple and remove this case.

lib/logflare_web/controllers/api/access_token_controller.ex
Ziinc: this would result in bad params not propagating up.

hauleth: It was not propagating previously either, as there was no `else` clause that would match incorrect return from `Auth.create_access_token/2`.

Ziinc: adjusted to forward changeset errors

lib/logflare_web/controllers/api/access_token_controller.ex
Ziinc: Auth.get_access_token might return nil, i'll refactor this.

Ziinc: can't use if as `Auth.get_access_token/1` might return nil.

lib/logflare_web/controllers/api/access_token_controller.ex
```diff
@@ -0,0 +1,74 @@
...
+
+    with {:scopes, true} <- {:scopes, "partner" not in String.split(scopes_input)},
+         {:ok, access_token} <- Auth.create_access_token(user, params) do
+      conn
+      |> put_status(201)
+      |> json(access_token)
+    else
+      {:scopes, false} -> {:error, :unauthorized}
+      {:error, _} = err -> err
+    end
```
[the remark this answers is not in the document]
Ziinc: don't really agree that it applies in this case. The "anti-pattern" is around handling ambiguous returns from multiple different functions, where there is possible overlap of error tuples when calling different functions.

In our case, the returns are known and clearly defined given the slimness of the function, and the possible error tuples are also known and handled by the fallback controller.

Also, if we were to follow the recommendation, it would add 2 lines to the stacktrace due to the additional wrapper functions, and make the whole function more verbose than it needs to be.

I have added more explicit tuple matching to make the error boundaries clearer. We can relook at this again in the future, but i think further refactoring has no meaningful difference.

Ziinc:
merging in first.

PR #1883, opened 2023-12-12
feat: partner api - allow impersonation of user

Ziinc:
This PR allows partners to impersonate users when querying the management api. Requires partners to have the user token uuid.

depends on #1878

### DIFF
--- a/lib/logflare/partners.ex
+++ b/lib/logflare/partners.ex
@@ -80,8 +80,7 @@ defmodule Logflare.Partners do
     query =
       from(p in Partner,
         join: u in assoc(p, :users),
-        where: p.token == ^token,
-        where: u.token == ^user_token,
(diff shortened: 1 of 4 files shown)


lib/logflare/partners.ex
```diff
@@ -77,15 +77,12 @@ defmodule Logflare.Partners do
...
-        select: u
-      )
-
-    Repo.one(query)
+    from(p in Partner,
+      join: u in assoc(p, :users),
+      where: p.token == ^token and u.token == ^user_token,
+      select: u
+    )
+    |> Repo.one()
```
hauleth, line 85: While I prefer the merged `:where` part I think I prefer having variable for `from` and then pass it to `Repo.one` instead of using pipe there.

hauleth (approved):

PR #1889, opened 2023-12-13
fix: add Auth context cache

chasers:
- [x] adds Auth.Cache context cache
- [x] Calls Auth.Cache in VerifyApiAccess plug
- [x] Expires Auth.Cache set to 30 seconds (this cache doesn't get busted from the write ahead log)
- [x] Current tests pass
- [x] Tests for Auth.Cache


hauleth (commented):
Do we have any possibility to revoke existing token? Like for example removing user from organisation or forceful session end due to security breach. If so, how does it work with that cache?

chasers:
> Do we have any possibility to revoke existing token? Like for example removing user from organisation or forceful session end due to security breach. If so, how does it work with that cache?

Just have to give it 30 seconds and the cache key for that user will expire, failing down to calling the fn again.

test/logflare/auth/cache_test.exs
```diff
@@ -0,0 +1,55 @@
+defmodule Logflare.Auth.CacheTest do
+  use Logflare.DataCase
+  alias Logflare.Auth
+  alias Logflare.Factory
```
Ziinc, line 4: should be auto imported already

test/logflare/auth/cache_test.exs
```diff
@@ -0,0 +1,55 @@
...
+  alias Logflare.Partners.Partner
+
+  setup do
+    user = Factory.insert(:user)
+    [user: user, team: Factory.insert(:team, user: user), partner: Factory.insert(:partner)]
```
Ziinc, line 9:
```suggestion
    [user: user, team: insert(:team, user: user), partner: insert(:partner)]
```

test/logflare/auth/cache_test.exs
```diff
@@ -0,0 +1,55 @@
+defmodule Logflare.Auth.CacheTest do
+  use Logflare.DataCase
+  alias Logflare.Auth
+  alias Logflare.Factory
+  alias Logflare.Partners.Partner
+
+  setup do
+    user = Factory.insert(:user)
```
Ziinc, line 8:
```suggestion
    user = insert(:user)
```

Ziinc (approved):

Ziinc:
merging in to get this to prod.

PR #1890, opened 2023-12-14
feat: query controller for executing one-off queries

Ziinc:
This PR adds in the `/api/query` route, which allows users execute one-off queries against their sources.

Currently, if accepts either `?bq_sql=`, `?sql=`, or `?pg_sql=`

`?sql=` is an alias for bq_sql and is the default.

Using `?pg_sql=` will result in the query executing on the first existing source backend created..


lib/logflare_web/controllers/api/query_controller.ex
```diff
@@ -0,0 +1,39 @@
...
+    responses: %{200 => List.response(QueryResult)}
+  )
+
+  def query(%{assigns: %{user: user}} = conn, %{"pg_sql" => sql}) do
+    with {:ok, %{rows: rows}} <- Endpoints.run_query_string(user, {:pg_sql, sql}) do
+      json(conn, %{result: rows})
+    else
+      {:error, err} ->
+        json(conn, %{errors: err})
+    end
```
hauleth, line 24: Why not handle that in `FallbackController`?

lib/logflare_web/controllers/api/query_controller.ex
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def query(conn, %{"sql" => sql}) do
+    query(conn, %{"bq_sql" => sql})
+  end
```
hauleth, line 29: I would put that before `pg_sql` clause.

lib/logflare_web/controllers/api/query_controller.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  def query(%{assigns: %{user: user}} = conn, %{"bq_sql" => sql}) do
+    with {:ok, %{rows: rows}} <- Endpoints.run_query_string(user, {:bq_sql, sql}) do
+      json(conn, %{result: rows})
+    else
+      {:error, err} ->
+        json(conn, %{errors: err})
+    end
+  end
```
hauleth, line 38: Shouldn't there also be a "catch all" function that sends error message about missing parameter?

test/logflare_web/controllers/api/query_controller_test.exs
```diff
@@ -0,0 +1,234 @@
...
+  #     endpoints: [endpoint | _]
+  #   } do
+  #     invalid_user = insert(:user)
+
+  #     assert conn
+  #            |> add_access_token(invalid_user, ~w(private))
+  #            |> delete("/api/endpoints/#{endpoint.token}", %{name: TestUtils.random_string()})
+  #            |> response(404)
+  #   end
+  # end
```
hauleth, line 233:
```suggestion
```

hauleth (changes requested):

lib/logflare_web/controllers/api/query_controller.ex
chasers: Should it all be just `sql` and then pick up the type of sql from the backend?

lib/logflare_web/controllers/api/query_controller.ex
Ziinc: good point. removed.

lib/logflare_web/controllers/api/query_controller.ex
Ziinc: @chasers one could have multiple backends attached to a source. Having different query params for each backend makes the caller explicitly specify which backend to be used, and would be the situation for example when they are ingesting into both pg and bq, but they want to query pg for certain queries and bq for certain queries.

Ziinc: moved the fn header up.

lib/logflare_web/controllers/api/query_controller.ex
Ziinc: Added.

hauleth (approved):

PR #1891, opened 2023-12-14
feat: partner upgrade/downgrade tiers

Ziinc:
This PR adds in upgrade/downgrade routes for partner provisioned users. When upgraded, users will be on the enterprise plan.


lib/logflare/partners.ex
```diff
@@ -117,4 +117,27 @@ defmodule Logflare.Partners do
...
+
+  def user_upgraded?(%User{id: id}) do
+    query =
+      from(pu in PartnerUser, where: pu.user_id == ^id, select: pu.upgraded)
+
+    Repo.one(query)
+    |> case do
+      nil -> false
+      value -> value
+    end
```
hauleth, line 129:
```suggestion
    case Repo.one(query) do
      nil -> false
      value -> value
    end
```

hauleth: Or:
```suggestion
    Repo.one(query) || false
```

lib/logflare/partners.ex
```diff
@@ -117,4 +117,27 @@ defmodule Logflare.Partners do
...
+  def upgrade_user(p, u), do: do_upgrade_downgrade(p, u, true)
+  def downgrade_user(p, u), do: do_upgrade_downgrade(p, u, false)
+
+  def do_upgrade_downgrade(%Partner{id: partner_id}, %User{id: user_id}, value) do
+    query =
+      from(pu in PartnerUser, where: pu.partner_id == ^partner_id and pu.user_id == ^user_id)
+
+    Repo.update_all(query, set: [upgraded: value])
+
+    {:ok, Repo.one(query)}
```
hauleth, line 141:
```suggestion
    query =
      from(pu in PartnerUser,
      	where: pu.partner_id == ^partner_id and pu.user_id == ^user_id,
      	select: pu)

    case Repo.update_all(query, set: [upgraded: value]) do
      {1, [partner]} -> {:ok, partner}
      _ -> :error
    end
```

lib/logflare/billing.ex
```diff
@@ -329,6 +330,9 @@ defmodule Logflare.Billing do
       SingleTenant.single_tenant?() ->
         get_plan_by(name: "Enterprise")
 
+      Partners.user_upgraded?(user) == true ->
```
hauleth, line 333:
```suggestion
      Partners.user_upgraded?(user) ->
```

hauleth (approved):

hauleth (approved):

PR #1894, opened 2023-12-15
fix: try_apply with Cachex.fetch

chasers:
Will now rescue if the applied fun throws an error vs crashing Cachex.


Ziinc:
could add in a mock function to `raise "some error"`

Ziinc:
something like 
```elixir
Logflare.Users
|> expect(:get, fn _id -> raise "err" end)


# in test
assert {:error, _} = Logflare.Users.Cache.get(123)
```
This might cause some negative side effects though on the callers, because it would not be according to specs.

PR #1897, opened 2023-12-19
fix: test cluster machine types, disk size

Ziinc:
Screenshot 2023-12-19 at 3 23 41 PM

Seems like there are some restrictions with e2 instances, this switches all staging instances to `c2d-standard-2`, and reduces disk size to match prod at 25GB. This brings the per-instance cost to around US$68, so we can spin up a few more for better simulation of prod.


Ziinc:
@chasers FYI.

PR #1899, opened 2023-12-22
fix: cloudbuild trigger SA roles

Ziinc:
This moves the `.dev.env.enc` file to correct location under `cloudbuild/` so that makefile script works.

This PR also adds in additional roles to the cloud build trigger SA, and updates staging secrets to match in use service account (as it seems like resource manager is removing all iam policies not declared).


Ziinc:
@chasers FYI, will merge this in a bit. This was the cause of our canary deploy failure, it is a temp fix. Cloud resource manager has quite a bit of tech debt, it fires off a lot of queries on each `set_iam_policy/1` call. Each time a user signs in, it gets called, resulting in db transaction and cpu spikes (+600/s txn, +5-8% cpu).
Ideally, we should call `get_iam_policy/0` first and only add/remove users to the retrieved list, instead of rebuilding each from scratch each time.

In pics, each spike is a sign in.

Screenshot 2023-12-23 at 12 35 58 AM
Screenshot 2023-12-23 at 12 35 52 AM

Ziinc:
Not urgent, can wait until team availability is higher.

Ziinc:
@chasers FYI, merged for testing.

PR #1907, opened 2024-01-05
feat: switch to debian, use otp26.2.1, elixir 1.16.0

Ziinc:
Refactoring of docker image so that we use glibc instead of musl.

update elixir to 1.16, fixes deprecation warnings.

Bumps version to v1.5.16

image build increases in size by around 120MB to 527MB. we can do more trimming in the future.


.docker.env
```diff
@@ -12,4 +12,4 @@ GOOGLE_PROJECT_ID=logflare-dev-238720
 GOOGLE_PROJECT_NUMBER=1023172132421
 LOGFLARE_GRPC_PORT=50051
 
-POSTGRES_BACKEND_URL=postgresql://postgres:postgres@db:5432/logflare_docker
+LOGFLARE_NODE_HOST=127.0.0.1
```
hauleth, line 15: We probably should add EditorConfig or something to force trailing newline on last line.

hauleth (approved):

PR #1908, opened 2024-01-08
chore: add node-problem-detector, bump to cos109, add in provisioning model.

Ziinc:
This adds the `node-problem-detector` service as per milelync suggestions.


Ziinc:
seems like we can use a metadata key instead https://cloud.google.com/container-optimized-os/docs/how-to/monitoring#metadata

Ziinc:
Added a bump to cos109 for latest LTS.

Ziinc:
Also sets the added disk to be the boot disk.

chasers (approved):

PR #1909, opened 2024-01-08
fix: tini installation on cloudbuild

Ziinc:
This fixes installation of tini on cloudbuild

https://github.com/Logflare/logflare/actions/runs/7446084745/job/20256142362
Screenshot 2024-01-08 at 6 47 53 PM


Ziinc:
@chasers FYI

PR #1910, opened 2024-01-08
chore: CD fixes for compute boot disk, version bump to 1.5.17

Ziinc:
GCP create similar cli generated should not have the `boot` option and should use `--boot-disk-*` options instead.


Ziinc:
@chasers merging in first for testing.

PR #1911, opened 2024-01-08
chore: remove provisioning-model option

Ziinc:
apparently it is not a valid option despite being generated in the gcp console.


Ziinc:
@chasers fyi

PR #1912, opened 2024-01-08
chore: use pd-balanced boot disks

Ziinc:
apparently local-ssd can't be used for boot disks. also bumps the size of the boot disk just in case.


Ziinc:
@chasers FYI

PR #1917, opened 2024-01-11
fix: restore http server options for bandit migration from cowboy

Ziinc:
This PR adds in equivalent cowboy http server options, for high load optimizations.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -31,10 +31,7 @@ config :logflare, LogflareWeb.Endpoint,
         # https://www.erlang.org/doc/man/inet
         # both reuseport and reuseport_lb should be provided for linux
         {:reuseport, true},
-        {:reuseport_lb, true},
-        # for so_reuseport flag for unix linux
-        # https://www.erlang.org/doc/man/gen_tcp.html#type-option
-        {:raw, 1, 15, <<1::32-native>>}
+        {:reuseport_lb, true}
         #
       ]
     ]
--- a/config/config.exs
+++ b/config/config.exs
@@ -20,8 +20,22 @@ config :logflare, LogflareWeb.Endpoint,
   adapter: Bandit.PhoenixAdapter,
   http: [
     thousand_island_options: [
+      # https://cloud.google.com/load-balancing/docs/https/#timeouts_and_retries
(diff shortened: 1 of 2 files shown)


config/config.exs
```diff
@@ -20,8 +20,22 @@ config :logflare, LogflareWeb.Endpoint,
...
+      # https://github.com/mtrudel/thousand_island/blob/ae733332892b1bb2482a9cf4e97de03411fba2ad/lib/thousand_island/transports/tcp.ex#L61
       transport_options: [
-        reuseport: true
+        # https://www.erlang.org/doc/man/inet
+        # both reuseport and reuseport_lb should be provided for linux
+        {:reuseport, true},
+        {:reuseport_lb, true},
+        # for so_reuseport flag for unix linux
+        # https://www.erlang.org/doc/man/gen_tcp.html#type-option
+        {:raw, 1, 15, <<1::32-native>>}
```
hauleth, line 37: Why use this while we use `{:reuseport, true}` already?

config/dev.exs
```diff
@@ -7,7 +7,14 @@ config :logflare,
 config :logflare, LogflareWeb.Endpoint,
   server: true,
   http: [
-    port: System.get_env("PORT") || 4000
+    port: System.get_env("PORT") || 4000,
+    thousand_island_options: [
+      transport_options: [
+        # https://stackoverflow.com/questions/41885427/use-so-reuseport-with-phoenix
+        # mac-specific flags
+        {:raw, 0xFFFF, 0x0200, <<1::32-native>>}
```
hauleth, line 15: Same here, why bother as we already have `{:reuseport, true}` set? Maybe you wanted `SO_REUSEADDR`, but that option is also exposed without using raw socket options.

config/config.exs
Ziinc: yea, seems redundant after closer reading. 
I think it was left in the cowboy config as there wasn't a `reuseport` option

hauleth: I think that was leftover from some older version of OTP, as Cowboy IIRC supports exactly the same set of options as `gen_tcp` as it pass them semi-transparently (it enforces some options).

hauleth (approved):

PR #1924, opened 2024-01-18
chore: revert FinchDefault changes, separate FinchGoth pool

Ziinc:
This PR adds in a FinchGoth pool that is separate from FinchDefault.
This also drops the FinchDefault count to 50, which is the same as v1.4.4.

The goal of this is to eliminate the possibility that the FinchDefault change to `count: 150` is causing the higher disconnections from the gcp apis, and also isolates the Finch changes since v1.4.4.

As Goth only uses the pool to refresh tokens, it is not needed to have high counts.

We had only customized it as by default it will use http

### DIFF
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -252,10 +252,10 @@ defmodule Logflare.Application do
   defp finch_pools do
     [
       # Finch connection pools, using http2
-      {Finch, name: Logflare.FinchIngest, pools: %{:default => [protocol: :http2, count: 200]}},
-      {Finch, name: Logflare.FinchQuery, pools: %{:default => [protocol: :http2, count: 100]}},
-      {Finch, name: Logflare.FinchGoth, pools: %{:default => [protocol: :http2, count: 1]}}
-      {Finch, name: Logflare.FinchDefault, pools: %{:default => [protocol: :http2, count: 50]}}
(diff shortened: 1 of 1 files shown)


lib/logflare/application.ex
```diff
@@ -246,15 +246,16 @@ defmodule Logflare.Application do
...
   end
 
   defp finch_pools do
     [
       # Finch connection pools, using http2
       {Finch, name: Logflare.FinchIngest, pools: %{:default => [protocol: :http2, count: 200]}},
       {Finch, name: Logflare.FinchQuery, pools: %{:default => [protocol: :http2, count: 100]}},
-      {Finch, name: Logflare.FinchDefault, pools: %{:default => [protocol: :http2, count: 150]}}
+      {Finch, name: Logflare.FinchGoth, pools: %{:default => [protocol: :http2, count: 1]}}
+      {Finch, name: Logflare.FinchDefault, pools: %{:default => [protocol: :http2, count: 50]}}
```
hauleth, line 258: I would use shorthand syntax in the map as well:
```suggestion
      {Finch, name: Logflare.FinchGoth, pools: %{default: [protocol: :http2, count: 1]}}
      {Finch, name: Logflare.FinchDefault, pools: %{default: [protocol: :http2, count: 50]}}
```

Just to clean it up a little.

hauleth (approved):

PR #1926, opened 2024-01-19
chore: revert max buffer length

Ziinc:
At 40k buffer length, the ram consumed can balloon to over >1GB.  

This can put the node at risk by consuming all available memory if there is sudden high load across many sources

SS was taken from load test cluster, where sources were at or near max buffer length. 

Screenshot 2024-01-19 at 8 04 03 PM


Ziinc:
@chasers FYI

PR #1929, opened 2024-01-22
ft: update BigQuery schema in pipeline v2

hauleth:


hauleth:
This *should* work, but I am trying to write tests for it to be sure that it does what it should do. However that requires a little bit more untangling of the V1 pipeline (I do not know where `source_id` atom comes from in `RLS`).

test/logflare/backends/bigquery_adaptor_test.exs
```diff
@@ -0,0 +1,55 @@
...
+
+    source_backend =
+      insert(:source_backend,
+        type: :bigquery,
+        source: source,
+        config: config
+      )
+
+    _rls = start_supervised %{
+      start: {RLS, :start_link, %RLS{source_id: source.id}}
```
Ziinc, line 42: should be `source_id: source.token`

Ziinc:
Relevant file for schema updating is here https://github.com/Logflare/logflare/blob/26c325d724d6479c646b17bf63f9de6b9edebcb2/lib/logflare/source/bigquery/schema.ex

Ziinc (approved):

PR #1930, opened 2024-01-23
chore: revert pipeline changes for processor concurrency

Ziinc:
This reverts processor concurrency and max_restarts options, which could be the cause of the `:long_schedule` issues


Ziinc:
@chasers FYI

PR #1934, opened 2024-01-25
fix: bandit max connnections

Ziinc:
This removes the bandit customizations.
This config results in max theoretical connections at `1_000 * 64_000` which is way more than we actually need since it is per acceptor.

Bandit already defaults to `16_384` per acceptor, so theoretical max connections by default is `100 * 16_384` which is plenty.

The settings was ported over from the cowboy config, but ranch works differently from thousand island hence does not make sense to use 64_000  max_connections nor 1_000 acceptors either since we only have 56 vCPUs on prod.

Dropping it down to bandit defaults so that we can see baseline performance. If acceptors are the bottleneck it would be quite evident in the network metrics.

https://hexdocs.pm/thousand_island/ThousandIsland.html

https://ninenines.eu/docs/en/ranch/2.0/manual/ranch/


Ziinc:
merging in first, @chasers FYI

PR #1936, opened 2024-01-26
fix: disable busy wait

Ziinc:
https://erlangforums.com/t/vm-tuning-guide/1945/4

as well as [rabbitmq advice](https://www.rabbitmq.com/runtime.html#busy-waiting)

COS is also linux based, which also has CFS since it was merged to linux kernel in 2007.
docker image is on debian, which uses CFS.

Some other discussions backing this up:
- https://stressgrid.com/blog/beam_cpu_usage/
- https://elixirforum.com/t/any-tips-on-tuning-the-beam-vm/51038/4?u=ziinc

### DIFF
--- a/rel/vm.args.eex
+++ b/rel/vm.args.eex
@@ -15,6 +15,10 @@
 +sbwtdcpu none
 +sbwtdio none
 
+## Reduce scheduler wake up threshold
++swt low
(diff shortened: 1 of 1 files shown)


Ziinc:
@chasers FYI

PR #1937, opened 2024-01-26
fix: revert bandit

Ziinc:
Reverts bandit usage, so that we can isolate whether it is bandit causing scheduler issues.


Ziinc:
merging in for more testing, @chasers FYI

PR #1938, opened 2024-01-26
fix: health check for ETS ingestion tables

Ziinc:
Health check for ETS tables is now needed, due to race conditions in startup time.


Ziinc:
@chasers FYI

PR #1942, opened 2024-01-30
fix: source routing bugs

Ziinc:
This fixes a bug where number fields were not getting correctly matched for regex filters.
It also fixes a page rendering bug. 

An old failing test has also been removed. Subsequent refactoring of the tests will come later.

### DIFF
--- a/lib/logflare/sources/source_routing.ex
+++ b/lib/logflare/sources/source_routing.ex
@@ -62,6 +62,8 @@ defmodule Logflare.Logs.SourceRouting do
 
         lql_filter_matches_any_of_nested_values? =
           Enum.reduce_while(le_values, false, fn le_value, _acc ->
+            le_str_value = if is_binary(le_value), do: le_value, else: inspect(le_value)
+
             lql_filter_matches? =
               cond do
                 is_nil(le_value) ->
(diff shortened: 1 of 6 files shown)


Ziinc:
change is large enough to warrant its own deploy.

Ziinc:
This pulls out all regex rules logic, which has been verified on prod that no users use regex anymore.

lib/logflare/sources/source_routing.ex
```diff
@@ -75,13 +63,13 @@ defmodule Logflare.Logs.SourceRouting do
...
                 operator == :string_contains ->
-                  String.contains?(le_value, value)
+                  String.contains?(le_str_value, value)
 
                 operator == := ->
                   apply(Kernel, :==, [le_value, value])
 
                 operator == :"~" ->
-                  apply(Kernel, :=~, [le_value, ~r/#{value}/u])
+                  apply(Kernel, :=~, [le_str_value, ~r/#{value}/u])
```
hauleth, line 72: I think that this would be more readable.
```suggestion
                  le_str_value =~ ~r/#{value}/u
```

lib/logflare_web/controllers/source_controller.ex
```diff
@@ -556,16 +556,9 @@ defmodule LogflareWeb.SourceController do
...
-        |> Map.take([:body, :via_rule, :origin_source_id])
-
-      if le.via_rule do
-        %{le | via_rule: %{regex: le.via_rule.regex}}
-      else
-        le
-      end
+      le
+      |> Map.from_struct()
+      |> Map.take([:body, :via_rule, :origin_source_id])
```
hauleth, line 561:
```suggestion
      Map.take(le, [:body, :via_rule, :origin_source_id])
```

Struct is map already, no need to do that double editing.

hauleth (approved):

PR #1944, opened 2024-01-30
fix: ratecounterserver exits

chasers:
RateCounterServer is failing to start with an already_started error which is putting the RecentLogServer in a crash loop.


lib/logflare/source/recent_logs_server.ex
```diff
@@ -240,11 +241,18 @@ defmodule Logflare.Source.RecentLogsServer do
     end
   end
 
+  def handle_info({:EXIT, _from, _reason}, state) do
```
Ziinc, line 244:
```suggestion
  def handle_info({:EXIT, _from, _reason}, _state) do
```

Ziinc (approved):

Ziinc:
would need to fix checks as well

lib/logflare/source/recent_logs_server.ex
chasers: Huh? Don't I need state in my :noreply tuple?

chasers: Maybe you were meaning the terminate...

hauleth (approved):

PR #1947, opened 2024-01-31
feat: ETS-backed BufferCounter

Ziinc:
This switches the BufferCounter table to use ets instead of `:counter` module.
This makes it non-blocking as well, so that if RLS tree is slow to come up, any calls to the BufferCounter module does not result in buffer_counter_not_found error.

The GenServer essentially only ensures that the counter is in sync with what Broadway has internally, and does the buffer broadcast. In event that Broadway slows down or provides an incorrect estimate, acks/push events will still inc/decr the current counter value, so some latency retrieving the estimate count is tolerable. Once retrieved, it will reset the counter value atomically.

Many operations make the `:ets.insert_new/2` call to ensure that the key is present. This can be optimized in the future with benchmarks.

ets table options are adjusted based on this:
- https://www.erlang.org/blog/scalable-ets-counters/

### DIFF
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -134,6 +134,7 @@ defmodule Logflare.Application do
         # init Counters before Supervisof as Supervisor calls Counters through table create
         Sources.Counters,
         Sources.RateCounters,
+        Sources.BufferCounters,
         PubSubRates.Rates,
         PubSubRates.Buffers,
         PubSubRates.Inserts,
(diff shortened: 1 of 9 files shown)


chasers:
Wait wait wait, how is this better than the counters module? I don't see how this fixes anything.

chasers:
e.g. :counters have the same write_concurrency option (which we're using) and are array based (like ETS) so I would think they use the same underlying internals.

Ziinc:
Motivation for this PR, in summary of [the slack thread yesterday](https://supabase.slack.com/archives/C032VRDHPDE/p1706664979079059):
- remove blocking of ingest on BufferCounter startup to allow for boot queue
- reduce error log emit

Usage of `:counters` works well here and may be [even more performant](https://elixirforum.com/t/is-ets-update-counter-efficient/20359/5?u=ziinc) than this implementation, would need some benchmarks to know.

Moving this to draft PR as a perf optimization, as we can still achieve the 2 goals with `:counters` mod.

Ziinc:
Closing as we don't need this anymore.

PR #1948, opened 2024-01-31
fix: buffer counter boot queue

Ziinc:
This PR adds in a boot queue for the broadway Pipeline. Boot queue length is set to 1k items, with fifo discards when queue is maxed (i.e. older items are discarded)

Boot queue is popped from the BufferProducer, and once the pipeline is up, the BufferCounter switches to inserting directly into the buffer.

depends on #1947


Ziinc:
moving back to draft.

Ziinc:
superceded with https://github.com/Logflare/logflare/pull/2036

PR #1949, opened 2024-01-31
fix: max buffer size for producer

Ziinc:
This reduces the max buffer size to double the hardcoded BufferCounter limit. This accommodates for bursting beyond the limit, but not to the point where the memory usage goes out of hand.


Ziinc:
@chasers FYI, merging in first.

PR #1955, opened 2024-02-05
chore: add prod instance template generation

Ziinc:
Adds instance template generation for prod clusters.

Also separates out instance group deployment to a separate file


Ziinc:
@chasers FYI

PR #1957, opened 2024-02-07
fix: makes primary key selection extendible for different structs

Ziinc:
cache value key selection is currently not extensible to other undeclared structs, which is likely interfering with caching.


chasers (approved):
Awesome!

PR #1959, opened 2024-02-07
fix: display rejected logs for BQ insert errors

Ziinc:
PR adds explicit display of BQ insert errors as well as buffer fulls, for user debugging.

This has not been prettified due to other priorities.

Screenshot 2024-02-07 at 5 57 35 PM

Screenshot 2024-02-07 at 5 58 55 PM


chasers:
I don't really understand what's going on in this UI. The red stuff doesn't have timestamps. Are those associated with the log events above them?

I was thinking this whole thing needed to be re-worked to pipe request and pipeline logs into a "source" they couldn't delete but do useful things with (like create alerts on).

Ziinc:
i'll make some tweaks to the error parsing and the ui so that it is more usable, as well as adding in some ui hints.

- [ ] ui hints & explainers
- [ ] documentation
- [ ] link from dashboard
- [ ] link to docs section
- [ ] ui fixes
  - [ ] wider gap between log events
  - [ ] display pipeline error type/stage
  - [ ] pretty format insert errors
  - [ ] add warning logs for non-string error messages that need to get stringified with `inspect`
  - [ ] add count summary of rejected logs

I agree on the reworking of the rejected logs, this is more of getting out what is already built so that we can refine it further towards the end goal of a `rejected_events` table. The task for it has already been created [here](https://www.notion.so/supabase/Logflare-Ingest-Rejected-logs-should-be-backed-by-a-separate-auto-generated-source-510b75212a494b3a8ae4f93100bdcde7?pvs=4), but i don't foresee capacity for working on it until later in the year

PR #1971, opened 2024-02-15
perf: v1 pipeline - hibernate _after pipeline opt, fullsweep_after spawn_opt

Ziinc:
This adds in more aggressive hibernation in the v1 BufferProducer, and adds in spawn_opt `fullsweep_after` for all processors and batchers, such that all binaries are discarded. Since these are long lived processes, we need to trigger GC faster, and as we don't need/store binaries across each pipeline work unit, we can set it to 0 and sacrifice some cpu for triggering GC more frequently.

On dev, these changes (mainly the spawn_opt adjustments) dropped  process memory usage by over half, from ~1.9GB to ~950GB, with local loadfest serving as load testing.

Since we're severly underutilizing our prod cpu capacity, this adjustment should balance the resource usage better.

### DIFF
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -262,12 +262,14 @@ defmodule Logflare.Application do
   end
 
   defp finch_pools do
+    # scales per core
+    base = System.schedulers_online()
     [
       # Finch connection pools, using http2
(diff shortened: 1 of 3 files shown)


chasers:
>  fullsweep_after: 0
Nice one, yeah seems reasonable here.

chasers (approved):

PR #1973, opened 2024-02-16
fix: alerts for slack

Ziinc:
Fixes for Alerts
- fixes for sending slack alerts
- fixes for oauth2 callback for slack alert webhook integration
- adds alert name to the outbound message


Ziinc:
@chasers merging in first.

PR #1979, opened 2024-02-26
fix: bq pipeline max length

Ziinc:
This PR adds in a custom batch_size splitter for handling large log events, and a timeout reduction for ingest to ensure that we catch rate limiting issues earlier.

### DIFF
--- a/lib/logflare/source/bigquery/pipeline.ex
+++ b/lib/logflare/source/bigquery/pipeline.ex
@@ -20,6 +20,9 @@ defmodule Logflare.Source.BigQuery.Pipeline do
   alias Logflare.Sources
   alias Logflare.Users
 
+  # each batch should at most be 5MB
+  @max_batch_length 5_000_000
+  @max_batch_size 250
   def start_link([%RLS{} = rls | opts]), do: start_link(rls, opts)
   def start_link(%RLS{} = rls), do: start_link(rls, [])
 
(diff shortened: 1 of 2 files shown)


lib/logflare/source/bigquery/pipeline.ex
```diff
@@ -42,8 +45,10 @@ defmodule Logflare.Source.BigQuery.Pipeline do
           batchers: [
             bq: [
               concurrency: System.schedulers_online() * 2,
-              batch_size: 250,
-              batch_timeout: 1_500
+              batch_size: bq_batch_size_splitter(),
```
chasers, line 48: Oh very cool, this can take a function for dynamic batch sizes!

lib/logflare/source/bigquery/pipeline.ex
```diff
@@ -206,4 +211,31 @@ defmodule Logflare.Source.BigQuery.Pipeline do
...
+    {
+      {@max_batch_size, @max_batch_length},
+      fn
+        # reach max count, emit
+        _message, {1, _len} ->
+          {:emit, {@max_batch_size, @max_batch_length}}
+
+        # check content length
+        message, {count, len} ->
+          payload = Jason.encode!(message.data.body)
```
chasers, line 228: Would be great if we could take the `payload` from here and use it in our BigQuery request so we don't have to encode twice.

chasers (approved):

chasers:
But let's give it a go!

lib/logflare/source/bigquery/pipeline.ex
Ziinc: the google api sdk doesn't take in raw json string, so don't think it is possible  
https://github.com/Logflare/logflare/commit/bdea277e759abea845cbdae32996febcde1423fd

PR #1984, opened 2024-02-26
fix: runtime option to log wal records

chasers:
Lets us optionally log wal records on production.


lib/logflare/cache_buster.ex
```diff
@@ -15,13 +15,47 @@ defmodule Logflare.CacheBuster do
...
+    Logger.put_process_level(self(), :error)
+
     Cainophile.Adapters.Postgres.subscribe(Logflare.PgPublisher, self())
     {:ok, state}
   end
 
+  @doc """
+  Sets the Logger level for this process. It's started with level :error.
+
+  To devug wal records set process to level :info and each transaction will be logged.
```
Ziinc, line 27:
```suggestion
  To debug wal records set process to level :info and each transaction will be logged.
```

lib/logflare/cache_buster.ex
```diff
@@ -15,13 +15,47 @@ defmodule Logflare.CacheBuster do
...
+          | :error
+          | :info
+          | :none
+          | :notice
+          | :warn
+          | :warning
+        ) :: :ok
+  def set_log_level(level) when is_atom(level) do
+    Process.whereis(__MODULE__)
+    |> GenServer.call({:put_level, level})
```
Ziinc, line 45:
```suggestion
    GenServer.call(__MODULE__, {:put_level, level})
```
Module name is already registered as process name, no need to find pid.

Ziinc (approved):

PR #1985, opened 2024-02-26
fix: fix LiveDashboard metrics

chasers:
- [x] Broadway metrics now work
- [x] More usable 

I was trying to use the metrics we added for WAL records but could not because the metrics page length kept changing as metrics for more sources came in, adding more charts to the page. 

We should be able to view these metrics by source_token, yes, but will likely need to to use PromEx to expose an endpoint for a Prometheus service to scrape separately.


VERSION
```diff
@@ -1 +1 @@
-1.5.32
\ No newline at end of file
+1.5.34
```
Ziinc, line 1: can put this out in 1.5.33 release

Ziinc: it is already bumped by the other PR

Ziinc (approved):

PR #1992, opened 2024-02-28
fix: backends user_id parsing for team users

Ziinc:
fix for situation where user_id is stored in session as string when it is a team user


Ziinc:
@chasers FYI

PR #1994, opened 2024-03-04
fix: alerts duplication bug, LOGFLARE_ALERTS_MIN_CLUSTER_SIZE

Ziinc:
This adds in two env vars, `LOGFLARE_ALERTS_MIN_CLUSTER_SIZE` and `LOGFLARE_ALERTS_ENABLED`.

If current cluster size is below the min, then alerts will not send. Alerts can also be disabled cluster-wide as well.


Ziinc:
@chasers fyi

PR #1996, opened 2024-03-04
ft: use dynamic repos for PG adaptor

hauleth:
**Stack:**

* https://github.com/Logflare/logflare/pull/1996

---

ft: extract all pipeline v2 processes to separate sub-tree


lib/logflare/backends/adaptor/postgres_adaptor/supervisor.ex
```diff
@@ -0,0 +1,32 @@
...
+    do: DynamicSupervisor.start_child(@repo_sup, spec)
+
+  def get(%Backend{id: id}) do
+    case Registry.lookup(@registry, {__MODULE__, id}) do
+      [{pid, meta}] -> {:ok, pid, meta}
+      [] -> :error
+    end
+  end
+
+  def via(%Backend{id: id}), do: {:via, Registry, {@registry, {__MODULE__, id}}}
```
Ziinc, line 31: doc + spec

lib/logflare/backends/adaptor/postgres_adaptor/supervisor.ex
```diff
@@ -0,0 +1,32 @@
...
+      {DynamicSupervisor, strategy: :one_for_one, name: @repo_sup}
+    ]
+
+    Supervisor.init(children, strategy: :one_for_one)
+  end
+
+  def start_child(spec),
+    do: DynamicSupervisor.start_child(@repo_sup, spec)
+
+  def get(%Backend{id: id}) do
```
Ziinc, line 24: doc + spec

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
```diff
@@ -0,0 +1,97 @@
...
+  alias Logflare.Backends.Adaptor.PostgresAdaptor.Repo.Migrations
+
+  def start(%Backend{} = backend) do
+    config = backend.config
+    pool_size = Application.get_env(:logflare, :postgres_backend_adapter)[:pool_size] || 10
+    schema = config["schema"] || config[:schema]
+
+    opts = [
+      url: config["url"] || config.url,
+      name: Supervisor.via(backend),
```
Ziinc, line 21: is the via function for the repo or for the supervisor? if for the repo then shouldn't it be in this module?

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
```diff
@@ -0,0 +1,97 @@
...
+
+  alias Logflare.Source
+  alias Logflare.Backends.Backend
+
+  alias Logflare.Backends.Adaptor.PostgresAdaptor.Supervisor
+  alias Logflare.Backends.Adaptor.PostgresAdaptor.Repo.Migrations
+
+  def start(%Backend{} = backend) do
+    config = backend.config
+    pool_size = Application.get_env(:logflare, :postgres_backend_adapter)[:pool_size] || 10
```
Ziinc, line 16: lets set default pool size in config.exs instead, and default needs to be small from filipe's prior testing around the pool sizes
see https://supabase.slack.com/archives/C032VRDHPDE/p1688281725600829?thread_ts=1688146906.543309&cid=C032VRDHPDE

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
```diff
@@ -0,0 +1,97 @@
...
+      sync_connect: true,
+      after_connect: {__MODULE__, :__after_connect__, [schema]}
+    ]
+
+    case Supervisor.start_child({__MODULE__, opts}) do
+      {:ok, pid} -> {:ok, pid}
+      # If there is already started process for this backend, then simply reuse
+      # it
+      {:error, {:already_started, pid}} -> {:ok, pid}
+      other -> other
```
Ziinc, line 33: there shouldn't be any other case

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
```diff
@@ -0,0 +1,97 @@
...
+    after
+      put_dynamic_repo(old_repo)
+    end
+  end
+
+  @doc """
+  Run migrations for currently selected repository
+  """
+  @spec migrate!(Source.t()) :: :ok | {:error, term()}
+  def migrate!(%Source{} = source) do
```
Ziinc, line 61: if it won't error due to rescuing then shouldn't use bang func

lib/logflare/backends/adaptor/postgres_adaptor/supervisor.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Logflare.Backends.Adaptor.PostgresAdaptor.Supervisor do
+  use Supervisor
```
Ziinc, line 2: @moduledoc

lib/logflare/backends/adaptor/postgres_adaptor/pg_repo.ex
```diff
@@ -98,49 +57,22 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.PgRepo do
...
-  @spec migrations_table_name(Source.t()) :: String.t()
-  def migrations_table_name(%Source{token: token}) do
-    token =
-      token
-      |> Atom.to_string()
-      |> String.replace("-", "_")
-
-    "schema_migrations_#{token}"
+  @spec destroy(Adaptor.source_backend(), timeout()) :: :ok
+  def destroy({source, backend}, timeout \\ 5000) do
```
Ziinc, line 64: scary word + vague, maybe can be more precise with `destroy_tables`

test/logflare/backends/postgres_adaptor_test.exs
```diff
@@ -131,22 +126,6 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptorTest do
   end
 
   describe "repo module" do
-    test "connected?/1 true when repo is connected", %{backend: backend} do
```
Ziinc, line 134: can we update the tests to cover the SharedRepo's conencted?/1 instead? most of the sharedrepo functions aren't covered with tests.

lib/logflare/backends/adaptor/postgres_adaptor/supervisor.ex
Ziinc: would also prefer to use `Backends.via_source/3` in general to ensure that via ids are consistent while we're doing v1->v2 transition

Ziinc (changes requested):
core implementation looks good, just some nits

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
hauleth: My idea for a bang there is that it is destructive operation that caller should avoid using and be aware of the consequences (not all banged functions throws on error, see `var!/1`). But I can change it.

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
hauleth: There should not, but just in case, I handle it there so it can be treated upstream if the requirements change in the future.

test/logflare/backends/postgres_adaptor_test.exs
hauleth: I wanted to remove `SharedRepo.connected?/1` as well, as it is not needed anywhere in the code.

lib/logflare/backends/adaptor/postgres_adaptor/supervisor.ex
hauleth: I will take a look.

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
Ziinc: OK cool, adding some docs on what is executed would be good

Ziinc: Yea sound reasoning for the bang, let's keep it

test/logflare/backends/postgres_adaptor_test.exs
Ziinc: Gotcha, let's remove it within this PR too then

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
hauleth: I did it in supervisor to hide the name of the `DynamicSupervisor` from the `Repo`, as it is irrelevant implementation detail for the `SharedRepo`.

lib/logflare/backends/adaptor/postgres_adaptor/supervisor.ex
hauleth: The thing is, where I use `Supervisor.via/1` I do not have access to `Source` as it is not relevant there (in `SharedRepo` start). This mean that I cannot use `via_source` as it expect me to pass `Source.t()`. This also would mean that multiple sources cannot reuse the same backend connection and instead each would spawn their own connection.

lib/logflare/backends/adaptor/postgres_adaptor/pg_repo.ex
hauleth: `destroy_instance`?

PR #1999, opened 2024-03-12
feat: add/remove of source backends, secrets redaction

Ziinc:
This PR adds in the adding/removal of source backends to management api.

It also adds in secrets redaction for postgres url when performing json encoding, to ensure that secrets within connection strings are not leaked.

### DIFF
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -87,7 +87,6 @@ defmodule Logflare.Backends do
 
   @doc """
   Updates the backends of a source wholly. Does not work on partial data, all backends must be provided.
-
   """
   @spec update_source_backends(Source.t(), [Backend.t()]) ::
           {:ok, Source.t()} | {:error, Ecto.Changeset.t()}
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -85,6 +85,10 @@ defmodule Logflare.Backends do
     end
   end
 
+  @doc """
+  Updates the backends of a source wholly. Does not work on partial data, all backends must be provided.
+
+  """
   @spec update_source_backends(Source.t(), [Backend.t()]) ::
           {:ok, Source.t()} | {:error, Ecto.Changeset.t()}
   def update_source_backends(%Source{} = source, backends) do
--- a/lib/logflare/backends/backend.ex
+++ b/lib/logflare/backends/backend.ex
@@ -49,7 +49,9 @@ defmodule Logflare.Backends.Backend do
         ])
         |> Map.update(:config, %{}, fn
           config when type == :postgres ->
-            Map.put(config, :url, "redacted")
+            url = Map.get(config, :url) || Map.get(config, "url")
+            updated = String.replace(url, ~r/(.+):.+\@/, "\\g{1}:REDACTED@")
+            Map.put(config, :url, updated)
 
           cfg ->
             cfg
--- a/lib/logflare/backends/backend.ex
+++ b/lib/logflare/backends/backend.ex
@@ -11,16 +11,6 @@ defmodule Logflare.Backends.Backend do
 
   @adaptor_types [:bigquery, :webhook, :postgres]
 
-  @derive {Jason.Encoder,
-           only: [
-             :name,
-             :token,
-             :description,
-             :type,
-             :id,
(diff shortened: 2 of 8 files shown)


lib/logflare/backends/backend.ex
```diff
@@ -42,4 +32,30 @@ defmodule Logflare.Backends.Backend do
...
+          :name,
+          :token,
+          :description,
+          :type,
+          :id,
+          :config
+        ])
+        |> Map.update(:config, %{}, fn
+          config when type == :postgres ->
+            Map.put(config, :url, "redacted")
```
chasers, line 52: Hmmm, should we store the individual fields instead of the URL?

Will be hard to debug if the whole URL is redacted.

With Realtime and Supavisor we store the fields individually. Could look at other service for prior art too.

hauleth: We could parse URL and the redact only fields that should be secret.

Ziinc: will need to keep url in for backwards compat and more flexibility (e.g. adding in query params), but can add the separate config fields as well.

will do the password redaction in string for now, and defer the extra fields to backlog

Ziinc: redaction now only removes in-url passwords.

lib/logflare/backends.ex
```diff
@@ -83,6 +83,10 @@ defmodule Logflare.Backends do
     end
   end
 
+  @doc """
+  Updates the backends of a source wholly. Does not work on partial data, all backends must be provided.
+
+  """
```
hauleth, line 89:
```suggestion
  """

```

lib/logflare_web/controllers/api/source_controller.ex
```diff
@@ -35,7 +36,7 @@ defmodule LogflareWeb.Api.SourceController do
 
   def show(%{assigns: %{user: user}} = conn, %{"token" => token}) do
     with source when not is_nil(source) <- Sources.get_by(token: token, user_id: user.id),
-         [source] <- Sources.preload_for_dashboard([source]) do
+         source <- Sources.preload_defaults(source) do
```
hauleth, line 39: It always match, do we need it to be part of `with`?

lib/logflare_web/controllers/api/source_controller.ex
```diff
@@ -93,4 +98,51 @@ defmodule LogflareWeb.Api.SourceController do
...
+    responses: %{
+      201 => Source.response(),
+      404 => NotFound.response()
+    }
+  )
+
+  def add_backend(conn, %{"source_token" => token, "backend_token" => backend_token}) do
+    with {:ok, backend} <- Backends.fetch_backend_by(token: backend_token),
+         {:ok, source} <- Sources.fetch_source_by(token: token),
+         source <- Sources.preload_backends(source),
```
hauleth, line 118:
```suggestion
         source = Sources.preload_backends(source),
```

lib/logflare_web/controllers/api/source_controller.ex
```diff
@@ -93,4 +98,51 @@ defmodule LogflareWeb.Api.SourceController do
...
+    responses: %{
+      200 => Source.response(),
+      404 => NotFound.response()
+    }
+  )
+
+  def remove_backend(conn, %{"source_token" => token, "backend_token" => backend_token}) do
+    with {:ok, source} <- Sources.fetch_source_by(token: token),
+         source <- Sources.preload_backends(source),
+         filtered <- Enum.filter(source.backends, &(&1.token != backend_token)),
```
hauleth, line 141:
```suggestion
         source = Sources.preload_backends(source),
         filtered = Enum.filter(source.backends, &(&1.token != backend_token)),
```

hauleth (approved):

chasers (approved):

lib/logflare_web/controllers/api/source_controller.ex
Ziinc: doesn't need to be, in this case it is more of a stylistic choice since we can't pipe in the `do`.

Ziinc: switched it to use `=`

PR #2000, opened 2024-03-13
test: use better way to retry tests

hauleth:
**Stack:**

* https://github.com/Logflare/logflare/pull/2000

---

test: use better way to retry tests

### DIFF
--- a/.github/workflows/elixir-ci.yml
+++ b/.github/workflows/elixir-ci.yml
@@ -13,15 +13,8 @@ on:
       - 'native/**'
       - 'priv/**'
       - 'test/**'
-      - '!.github/**'
-      - '!cloudbuild/**'
-      - '!assets/**'
-      - '!Dockerfile'
-      - '!*.yml'
-      - '!*.md'
-      - '!*.sh'
-      - '!*.env'
-      - '!Makefile'
+      # Run suite when this file is changed
+      - '.github/workflows/elixir-ci.yml'
 
 permissions:
   contents: read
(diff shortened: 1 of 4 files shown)


test/support/test_utils.ex
```diff
@@ -419,26 +419,20 @@ defmodule Logflare.TestUtils do
...
-  def retry_fetch(fetcher, asserts, times, sleep) do
-    with fetcher_result when fetcher_result != :retry <- fetcher.(),
-         assert_result when assert_result != :retry <- asserts.(fetcher_result) do
-      send(self(), :done)
-    else
-      :retry ->
-        :timer.sleep(sleep)
-        retry_fetch(fetcher, asserts, times - 1)
+  @spec retry_assert(times :: pos_integer(), sleep :: pos_integer(), func :: (-> any())) :: any()
+  def retry_assert(times \\ 10, sleep \\ 50, func) do
```
Ziinc, line 425: alternate implementation is to keep retrying with 50ms intervals for a set timeout, then caller can adjust the timeout with a similar api like `assert_receive`

hauleth: I can change the implementation, that shouldn't be too hard to implement.

.github/workflows/elixir-ci.yml
```diff
@@ -13,15 +13,8 @@ on:
       - 'native/**'
       - 'priv/**'
       - 'test/**'
-      - '!.github/**'
```
Ziinc, line 16: reason for this is because doesn't make sense to re-run elixir ci when non-elixir code changes, reduces iteration time when working on ci/cd related or docs stuff.
maybe a better glob pattern would work

Ziinc (changes requested):
looks good, lets keep the current ci path filters for now

.github/workflows/elixir-ci.yml
hauleth: That doesn't matter, as all these entries were doing nothing anyway. These filters works that this action will be fires only if any files matches paths defined there. So negative matches that aren't within positive matches aren't needed at all.

Ziinc (approved):

PR #2006, opened 2024-03-18
ft: support accessing other top-level fields

hauleth:
**Stack:**

* https://github.com/Logflare/logflare/pull/2006

---

ft: support accessing other top-level fields


Ziinc (approved):
For the cloud_events event message we should also hardcode the message rewrite in the processor

PR #2014, opened 2024-03-26
fix: PubSubRates application tree for non-test env

Ziinc:
This PR fixes the application tree for PubSubRates, due to refactoring around the cluster-wide pubsub logic in #2012


Ziinc:
@chasers FYI.

PR #2020, opened 2024-03-29
ft: DataDog adaptor

hauleth:
This introduces wrapper adaptor over Webhook adaptor to send logs to
DataDog HTTP log ingest endpoint.


test/logflare/backends/adaptor/datadog_adaptor_test.exs
```diff
@@ -0,0 +1,117 @@
...
+          },
+          source: %{name: "sample"}
+        }
+      ]
+
+      @subject.ingest(pid, events, [])
+
+      assert_receive {:ok, [event]}, 200
+      assert %{"ddtags" => tags} = event
+      assert tags =~ ~r/foo:foo/
```
Ziinc, line 113: tags need to be a certain format
https://docs.datadoghq.com/getting_started/tagging/#define-tags
would need to handle restrictions for them.

lib/logflare/backends/adaptor/datadog_adaptor.ex
```diff
@@ -0,0 +1,190 @@
...
+    defp send_logs(%__MODULE__{} = state) do
+      middlewares = [
+        Tesla.Middleware.Telemetry,
+        Tesla.Middleware.JSON
+      ]
+
+      client = Tesla.client(middlewares)
+
+      Task.start(fn ->
+        case Tesla.request(client,
```
Ziinc, line 97: retry with backoff would be great

Ziinc: we can add it after we add in batching for webhook adaptor

lib/logflare/backends/adaptor/datadog_adaptor.ex
```diff
@@ -0,0 +1,190 @@
...
+              timeout: 1000,
+              timer: nil,
+              url: nil,
+              method: :post,
+              headers: []
+
+    def start_link(opts), do: GenServer.start_link(__MODULE__, Map.new(opts))
+
+    def ingest(_pid, []), do: :ok
+    def ingest(pid, logs), do: GenServer.cast(pid, {:events, logs})
```
Ziinc, line 30: This would get called for even small requests or singular event requests, as `Backends.ingest_logs/2` doesn't handle batching.
would need to handle batching at adaptor level, which is why i've mostly been using Broadway to auto-handle them. This won't survive at prod scale, which is the issues facing the current WebhookAdaptor and PostgresAdaptor.

Ziinc (changes requested):

test/logflare/backends/adaptor/datadog_adaptor_test.exs
hauleth: Ok, I will take a look. Maybe there should a way to define mappings for tags to avoid excessive tag amount as docs mention:

> Tags should not originate from unbounded sources, such as epoch timestamps, user IDs, or request IDs. Doing so may infinitely [increase the number of metrics](https://docs.datadoghq.com/metrics/) for your organization and impact your billing.

lib/logflare/backends/adaptor/datadog_adaptor.ex
hauleth: Yeah. That module probably should be extracted to be independent of the DataDog adaptor, or even to separate library, but for now I wanted to have everything simple and in single place.

lib/logflare/backends/adaptor/datadog_adaptor.ex
hauleth: Yesm that would get called for singular events, that is why that module *is* implementing batching in a way that it will wait until there is given amount of events in queue **or** timeout was reached. The goal of the timeout is to not batch messages indefinitely if there will be not enough of them, but rather send messages in more or less uniform periods even if there is less than N messages.

Ziinc: this is the same as the `:batch_timeout` option for broadway

test/logflare/backends/adaptor/datadog_adaptor_test.exs
Ziinc: yea i'm thinking of implementing a allowlist for schema paths that are permitted for schema updates, it can take a similar form:
```
metadata\..+
my
field
```
was thinking of making it regex based, can do a similar allowlist for allowed tag mappings

lib/logflare/backends/adaptor/datadog_adaptor.ex
hauleth: Oh, TIL. I can change it to Broadway to make everything more uniform

lib/logflare/backends/adaptor/datadog_adaptor.ex
```diff
@@ -0,0 +1,74 @@
...
+      start: {__MODULE__, :start_link, [arg]}
+    }
+  end
+
+  @impl Logflare.Backends.Adaptor
+  def start_link({source, backend}) do
+    backend = %{
+      backend
+      | config: %{
+          url: "https://http-intake.logs.datadoghq.com/api/v2/logs",
```
Ziinc, line 29: nit: move to module attribute with inline comments to datadog documentation

lib/logflare/backends/adaptor/datadog_adaptor.ex
```diff
@@ -0,0 +1,74 @@
...
+    {%{}, %{api_key: :string}}
+    |> Ecto.Changeset.cast(params, [:api_key])
+  end
+
+  @impl Logflare.Backends.Adaptor
+  def validate_config(changeset) do
+    import Ecto.Changeset
+
+    changeset
+    |> validate_required([:api_key])
```
Ziinc, line 61:
```suggestion
    changeset
    |> Ecto.Changeset.validate_required([:api_key])
```

Ziinc (changes requested):
Implementation looks good, small nits, good to merge after they're cleared.

For further improvements down the road, we'd need to consider their batch sizes:
Screenshot 2024-04-25 at 1 32 51 AM
we'd likely need configurable max batch size and max batch length for webhook adaptor, similar to what [BQ adaptor is doing](https://github.com/Logflare/logflare/commit/8e9d51a7e40501f42c248d5181e82e648e023adf)
will backlog this until their api starts rejecting us

Ziinc (dismissed):

hauleth:
@Ziinc what happened with all these commits?

Ziinc:
Ah I rebased it in the log drain PR

hauleth:
Should I clean this up with my local state?

--

Łukasz Niemier
***@***.***

Ziinc:
Have pushed up the region changes

Will need to do some more testing before merge

- [ ] manual test for DD drain rule

Ziinc (approved):

hauleth:
💚 💙 💜 💛 ❤️

PR #2023, opened 2024-04-03
chore: tweak deployment configs

Ziinc:
- reduce deployment min ready time between node startup
- tweak alerting min cluster nodes, 7 for prod, 3 for staging


hauleth (approved):
LGTM, but I am not sure if I am competent in this field enough to merge.

PR #2024, opened 2024-04-03
perf: reduce bootup recent update window

Ziinc:
- tweaks the busy sources bootup window to reduce time-to-healthy

### DIFF
--- a/lib/logflare/source/supervisor.ex
+++ b/lib/logflare/source/supervisor.ex
@@ -42,7 +42,7 @@ defmodule Logflare.Source.Supervisor do
     # Starting sources only when we've seen an event in the last 6 hours
     # Plugs.EnsureSourceStarted makes sure if a source isn't started, it gets started for ingest and the UI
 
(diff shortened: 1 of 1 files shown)


lib/logflare/source/supervisor.ex
```diff
@@ -42,15 +42,15 @@ defmodule Logflare.Source.Supervisor do
...
     # Plugs.EnsureSourceStarted makes sure if a source isn't started, it gets started for ingest and the UI
 
-    milli = :timer.hours(6)
+    milli = :timer.hours(1)
     from_datetime = DateTime.utc_now() |> DateTime.add(-milli, :millisecond)
 
     query =
       from(s in Source,
         order_by: s.log_events_updated_at,
         where: s.log_events_updated_at > ^from_datetime,
```
hauleth, line 51: Wouldn't [`ago/2`](https://hexdocs.pm/ecto/3.11.2/Ecto.Query.API.html#ago/2) be more readable there?

PR #2025, opened 2024-04-04
fix: slack invalid blocks

Ziinc:
This fixes the invalid_blocks errors that slack webhooks return due to invalid markdown. This affects both Legacy Recent Events Alerts and Query Alerts.
New block formatting will look something like this:
Screenshot 2024-04-04 at 9 50 34 AM

For each row, a new block will be used, allowing for intra-slack linking to specific events.

SlackHookServer is also adjusted to use this new block generation logic, as many events are not getting sent to slack channels due to the invalid formatting.

### DIFF
--- a/lib/logflare/backends/adaptor/slack_adaptor.ex
+++ b/lib/logflare/backends/adaptor/slack_adaptor.ex
@@ -23,7 +23,7 @@ defmodule Logflare.Backends.Adaptor.SlackAdaptor do
       end
 
     view_url = url(~p"/alerts/#{id}")
-    context = "🔊 *#{name}*#{rows_text} | [View alert](#{view_url})"
+    context = "🔊 *#{name}*#{rows_text} | #{view_url}"
 
     body =
       payload
--- a/lib/logflare/backends/adaptor/slack_adaptor.ex
+++ b/lib/logflare/backends/adaptor/slack_adaptor.ex
@@ -7,7 +7,7 @@ defmodule Logflare.Backends.Adaptor.SlackAdaptor do
   alias __MODULE__.Client
 
   alias Logflare.Alerting.AlertQuery
-
+  @endpoint LogflareWeb.Endpoint
   @doc """
   Sends a given payload to slack.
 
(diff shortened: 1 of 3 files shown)


lib/logflare/backends/adaptor/slack_adaptor.ex
```diff
@@ -39,17 +43,50 @@ defmodule Logflare.Backends.Adaptor.SlackAdaptor do
...
+                 accessory: %{
+                   type: "button",
+                   text: %{type: "plain_text", text: button_link.text},
+                   url: button_link.url,
+                   style: "primary"
+                 }
+               }
+             )
+           ])
+        |> Enum.filter(&(&1 != nil))
```
hauleth, line 89: Ok, that looks awful. Maybe some extraction to some helper functions would make it easier to read.

lib/logflare/backends/adaptor/slack_adaptor.ex
```diff
@@ -59,33 +96,64 @@ defmodule Logflare.Backends.Adaptor.SlackAdaptor do
...
+    %{type: "text", text: v}
   end
+
+  defp stringify(v) when is_integer(v) do
+    Integer.to_string(v)
+  end
+
+  defp stringify(v) when is_float(v) do
+    Float.to_string(v)
+  end
```
hauleth, line 155: These two aren't really needed as `Jason.encode!/1` should work on these as well. It doesn't work properly in such case?

lib/logflare/backends/adaptor/slack_adaptor.ex
```diff
@@ -59,33 +96,64 @@ defmodule Logflare.Backends.Adaptor.SlackAdaptor do
...
+          [text("#{k}:"), line_break(), text(v_str)]
 
-  def to_markdown(%{} = row) do
-    bullet = "• "
-    indent = "    "
+        true ->
+          [text("#{k}:"), space(), text(v_str)]
+      end
+    end)
+    |> Enum.intersperse([line_break()])
```
hauleth, line 129: Use [`Enum.map_intersperse/3`](https://hexdocs.pm/elixir/Enum.html#map_intersperse/3).

lib/logflare/backends/adaptor/slack_adaptor.ex
Ziinc: this is needed for the checks here, as the functions are are strict typed
https://github.com/Logflare/logflare/commit/0f884ec71dae96827f055e494ae30d5d71afb6d1

hauleth (approved):

PR #2028, opened 2024-04-04
perf: cache TeamUsers table

Ziinc:
Cache TeamUsers table which is used for Schema update notifications


hauleth (approved):
I feel like we have too many separate caches in the application. Maybe we should think about some unification of them?

Ziinc:
would be a little tricky as some caches have differing ttl time. Not impossible though, extending the ContextCache and having a shared DataCache is possible too.
Can defer this to EOY

PR #2036, opened 2024-04-09
feat: dynamic horizontal scaling of Broadway pipeline

Ziinc:
This is a reworked dynamic scaling of the broadway pipeline, which allows for horizontal scaling at runtime.

The DynamicPipeline module is written generically to apply to any broadway pipeline.

I have also adjusted down the proc counts for the base bigquery pipeline.


chasers (dismissed):

Ziinc:
on hold for now.

Ziinc:
- [ ] reduce process counts for BQ pipeline

Ziinc:
To merge after v1.7.2 deploy.

chasers:
> producer might silently shed load

Oh, I'd been meaning to look for this in the logs and it seems like we are dropping some stuff.

We'll need to check the GenStage buffer and respond. We could respond with 429 vs a 406. In the case of a 429 Vector will retry. We may need to tune it's retry frequency. Search for `"has discarded" "GenStage"`

Ziinc:
Refactored and remove the pushing of messages, it solely increases the number of broadway producers. It does not handle anything else besides that. From my local testing this has helped to increase ingest capacity, and works best with steady loads, but the spinning up of nodes also factors in any sudden rise in buffer counts.

The logic is also greatly simplified as it does not need to factor estimating of the internal buffer of the broadway pipelines anymore..

However, it is possible that the number of pipelines is not steadily increased, but instead spun up suddenly (or stopped suddenly). Broadway will still gracefully drain, but having spikey increases and decreases is not desirable. Will have a further think on how to reduce this.

Ziinc:
- increases max buffer len to 50k
- dynamically scales pipelines based on ingest count, increases rapidly if buffer is filling up, decreases slowly when buffer is emptying out. 
- currently only applies to bigquery adaptor

-

PR #2038, opened 2024-04-12
fix: slack hooks adaptor alerts and recent alerts uses message with CTA.

Ziinc:
Fixes slack button link block generation, which affects recent events notifications to slack

Screenshot 2024-04-12 at 8 52 21 PM


Ziinc:
@chasers FYI

PR #2040, opened 2024-04-15
feat: postgres adaptor improvements

Ziinc:
Adds following improvements to PostgresAdaptor:
- connection url is now optional, connection config can be provided instead, fixing bug where connection string cannot be parsed when password has special characters
- PostgresAdaptor process does not crash now if it cannot connect to the underlying PG db. The shared repo will now fail to connect with `{:error, :cannot_connect}` instead of bringing down all linked processes.
  - database connection errors are still logged, we might need to suppress this down the road if it gets too noisy.
- ingestion now does not go through the genserver, removing buffer insert bottleneck
- broadway now batches the inserts

With local testing with a staging supabase instance, we can insert at full loadfest throughput quite easily.


Ziinc:
- [ ] configurable pool size

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
```diff
@@ -76,9 +94,15 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.SharedRepo do
...
-    :ok
+    try do
+      Migrations.migrate(source)
+    rescue
+      _e in DBConnection.ConnectionError ->
+        {:error, :cannot_connect}
+    catch
+      :exit, {:killed, _} -> {:error, :cannot_connect}
+      :exit, {:noproc, _} -> {:error, :cannot_connect}
+    end
```
hauleth, line 105: Why nest `try`?

hauleth (approved):

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
Ziinc: will remove

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
```diff
@@ -62,6 +74,12 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.SharedRepo do
 
     try do
       func.()
+    rescue
+      _e in DBConnection.ConnectionError ->
+        {:error, :cannot_connect}
```
hauleth, line 79:
```suggestion
      DBConnection.ConnectionError ->
        {:error, :cannot_connect}
```

Technically that `_e in` thingy is not needed. I do not have strong opinion which one is more readable though.

hauleth (approved):

PR #2042, opened 2024-04-16
test: silence Finch errors

hauleth:
This adds Logger filter, that will silence Finch disconnection errors in tests to reduce amount of pollution in the test log.


Ziinc:
wouldn't this hide issues with tests, such as requests being made in background when they should be stubbed?

hauleth:
If that hides that test is incorrect, then IMHO the test is moot/invalid. Also that was just a lot of noise in tests output which made test results hard to read.

Ziinc (approved):

PR #2055, opened 2024-04-25
fix: fork gax to add retries and compression

Ziinc:
This PR adds override for Gax and uses our fork instead https://github.com/Logflare/elixir-google-gax

The fork adds in retries and gzip compression, which covers a lot of errors that we are seeing.


Ziinc:
@chasers FYI

PR #2057, opened 2024-04-26
chore: refactor logs search liveview

Ziinc:
Performs refactoring that cleans up the following:
- remove connected/disconnected differing logic for simplicity sake.
- Remove duplicate searching on mount, as search already gets triggered in the `handle_params/3` callback. This was causing lql query to get parsed 3 times.
- reorganizes assigns and removed duplicates.


Ziinc:
@chasers FYI

PR #2059, opened 2024-05-01
feat: log drains rules, extends rules implementation

Ziinc:
This implements drain rules, for draining events from a source to a given backend.
UI for managing rules is based on a per-backend perspective.

In the demo, we can see that ingested data from `dev.logs.up2` are forwarded (evident from the rule icon) to downstream backends, and that `loadfest.test.1` and `loadfest.test.2` are both receiving ingestion requests (there is no initial events being sent to these two sources).

This demo is done with webhooks adaptor.

https://github.com/Logflare/logflare/assets/22714384/be082d83-057b-4e31-88de-5395f83fe7b8


lib/logflare/backends/source_sup.ex
```diff
@@ -39,14 +46,19 @@ defmodule Logflare.Backends.SourceSup do
...
+      ([
+         %Backend{
+           type: :bigquery,
+           config: %{
+             project_id: project_id,
+             dataset_id: dataset_id
+           }
+         }
+         | ingest_backends
+       ] ++ rules_backends)
```
hauleth, line 59: I find

```suggestion
      [
         %Backend{
           type: :bigquery,
           config: %{
             project_id: project_id,
             dataset_id: dataset_id
           }
         }
         | ingest_backends
       ]
       |> Enum.concat(rules_backends)
```

To be a little bit cleaner?

lib/logflare/logs/lql/lql_parser.ex
```diff
@@ -121,6 +165,29 @@ defmodule Logflare.Lql.Parser do
...
+      cond do
+        match?({_val, ""}, int) ->
+          elem(int, 0)
+
+        match?({_val, ""}, flt) ->
+          elem(flt, 0)
+
+        true ->
+          v
+      end
```
hauleth, line 186:
```suggestion
    parsed =
      case Integer.parse(v) do
        {num, ""} ->
          num
        
        _ ->
      	  case Float.parse(v) do
      	    {num, ""} -> num
      	    _ -> v
      	  end
       end
```

lib/logflare/sources/source_routing.ex
```diff
@@ -64,7 +75,7 @@ defmodule Logflare.Logs.SourceRouting do
                   apply(Kernel, :==, [le_value, value])
```
hauleth, line 75:
```suggestion
                  le_value == value
```

lib/logflare/utils.ex
```diff
@@ -28,4 +28,27 @@ defmodule Logflare.Utils do
...
+    %{"test" => "data}
+  """
+  @spec stringify_keys(map()) :: map()
+  def stringify_keys(map = %{}) do
+    map
+    |> Enum.map(fn
+      {k, v} when is_atom(k) -> {Atom.to_string(k), stringify_keys(v)}
+      {k, v} when is_binary(k) -> {k, stringify_keys(v)}
+    end)
+    |> Enum.into(%{})
```
hauleth, line 46:
```suggestion
    Map.new(map, fn
      {k, v} when is_atom(k) -> {Atom.to_string(k), stringify_keys(v)}
      {k, v} when is_binary(k) -> {k, stringify_keys(v)}
    end)
```

hauleth (approved):

hauleth (approved):

Ziinc:
extra edge case to consider for follow up PRs

- [ ] when there are two same rules and one gets deleted, the backend child will get stopped instead of being left as is.

PR #2065, opened 2024-05-09
feat: elastic adaptor

Ziinc:
This PR adds in filebeat http handling
It also adds in http1 handling, which the filebeat server requires.

- [x] docs
- [x] filebeat testing
- [x] ui adjustments

deferred:
- gzip

https://github.com/Logflare/logflare/assets/22714384/0e8ce768-837a-465d-baa6-fd2126d3da80

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -33,7 +33,6 @@ config :logflare, Logflare.Repo,
 
 config :logflare, :postgres_backend_adapter, pool_size: 1
 config :grpc, start_server: false
-config :tesla, adapter: Tesla.Mock
 
 defmodule LogflareTest.LogFilters do
   def ignore_finch_disconnections(%{meta: %{mfa: {Finch.HTTP2.Pool, :disconnected, _}}}, _opts) do
--- a/config/test.exs
+++ b/config/test.exs
@@ -33,6 +33,7 @@ config :logflare, Logflare.Repo,
 
 config :logflare, :postgres_backend_adapter, pool_size: 1
 config :grpc, start_server: false
+config :tesla, adapter: Tesla.Mock
 
 defmodule LogflareTest.LogFilters do
   def ignore_finch_disconnections(%{meta: %{mfa: {Finch.HTTP2.Pool, :disconnected, _}}}, _opts) do
--- a/lib/logflare/backends/adaptor/elastic_adaptor.ex
+++ b/lib/logflare/backends/adaptor/elastic_adaptor.ex
@@ -38,6 +38,7 @@ defmodule Logflare.Backends.Adaptor.ElasticAdaptor do
       backend
       | config: %{
           url: backend.config.url,
+          http: "http1",
           headers:
             if basic_auth do
               %{"Authorization" => "Basic #{basic_auth}"}
--- a/lib/logflare/backends/adaptor/elastic_adaptor.ex
+++ b/lib/logflare/backends/adaptor/elastic_adaptor.ex
@@ -1,0 +1,98 @@
+defmodule Logflare.Backends.Adaptor.ElasticAdaptor do
+  @moduledoc """
+
+  Ingestion uses Filebeat HTTP input.
+
+  https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-http_endpoint.html
+
+  Basic auth implementation reference:
+  https://datatracker.ietf.org/doc/html/rfc7617
+
+  """
+
+  use TypedStruct
+
+  alias Logflare.Backends.Adaptor.WebhookAdaptor
+
+  typedstruct enforce: true do
+    field(:url, String.t())
+    # basic auth username and password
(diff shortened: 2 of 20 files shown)


config/test.exs
```diff
@@ -33,6 +33,7 @@ config :logflare, Logflare.Repo,
 
 config :logflare, :postgres_backend_adapter, pool_size: 1
 config :grpc, start_server: false
+config :tesla, adapter: Tesla.Mock
```
hauleth, line 36: I have tried that in the past but there were a lot of issues. I am glad that we can land it right now. It will reduce amount of mocking that we need for HTTP clients.

lib/logflare/backends/adaptor/elastic_adaptor.ex
```diff
@@ -0,0 +1,99 @@
...
+    |> Ecto.Changeset.cast(params, [:username, :password, :url])
+    |> validate_user_pass()
+  end
+
+  defp validate_user_pass(changeset) do
+    user = Ecto.Changeset.get_field(changeset, :username)
+    pass = Ecto.Changeset.get_field(changeset, :password)
+    user_pass = [user, pass]
+
+    if user_pass != [nil, nil] and Enum.any?(user_pass, &is_nil/1) do
```
hauleth, line 81:
```suggestion
    if is_nil(user) != is_nil(pass) do
```

This should have the same effect.

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -80,14 +81,27 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
...
-      opts
-      |> Keyword.put_new(:method, :post)
-      |> Keyword.update(:headers, [], &Map.to_list/1)
-      |> request()
+      adaptor =
+        if Keyword.get(opts, :http) == "http1" do
+          {Tesla.Adapter.Finch, name: Logflare.FinchDefaultHttp1, receive_timeout: 5_000}
+        else
+          {Tesla.Adapter.Finch, name: Logflare.FinchDefault, receive_timeout: 5_000}
+        end
```
hauleth, line 90: `Finch` do not take care of it automatically? Because that is weird that downstream is needs to know which protocol it should use.

lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -31,7 +31,7 @@
   </div>
 
   <div :if={@live_action == :new}>
-    <%= select(f, :type, ["Select a backend type...", Webhook: :webhook, Postgres: :postgres, BigQuery: :bigquery, Datadog: :datadog],
+    <%= select(f, :type, ["Select a backend type...", Webhook: :webhook, Postgres: :postgres, BigQuery: :bigquery, Datadog: :datadog, Elastic: :elastic],
```
hauleth, line 34: We probably should generate that list there.

test/logflare/backends/elastic_adaptor_test.exs
```diff
@@ -0,0 +1,120 @@
...
+      :timer.sleep(500)
+      [pid: pid, backend: backend, source: source]
+    end
+
+    test "adds authorization header", %{pid: pid, source: source, backend: backend} do
+      this = self()
+      ref = make_ref()
+
+      Tesla.Mock.mock_global(fn req ->
+        assert {_k, "Basic" <> _} = Enum.find(req.headers, fn {k, _v} -> k == "Authorization" end)
```
hauleth, line 108: [`List.keyfind/3`](https://hexdocs.pm/elixir/List.html#keyfind/4)

lib/logflare/backends/adaptor/elastic_adaptor.ex
Ziinc: it would only satisfy first condition, as the latter condition (e.g. if user is given but password is `nil`, left side would evaluate to `false` and right side would evaluate to `true`) would go throguh.

lib/logflare/backends/adaptor/webhook_adaptor.ex
Ziinc: Finch can handle both but without multiplexing
https://hexdocs.pm/finch/Finch.html#start_link/1-pool-configuration-options

Currently all our pools are http2 to leverage the multiplexing, so we'd need to have a separate http1 pool for that, separate from the http2 pools.
We could adjust the default pool to be both http1 and http2, but then we won't be able to get any multiplexing if the backend server supports it

Ziinc:
Coudln't get tesla mocking to work because we would need to mock on a per-client basis, but we dynamically build the client at runtime. Tesla docs doesn't have any guidance on mocking for dynamic clients, and imo would be too much work to get running for the benefit as compared to using mimic.

Ziinc:
@chasers FYI

PR #2066, opened 2024-05-09
chore: fix flaky topic sharded broadcasting tests

Ziinc:
This fixes flaky broadcast tests, where topic sharding was causing flaky subscription behaviour. 
Topic sharding is no longer required as the broadcast topics are now appropriately partitioned by the PubSub pool size config.


hauleth (approved):

hauleth (approved):

Ziinc:
@chasers btw, in case you have comments on the topic sharding changes.

PR #2067, opened 2024-05-11
ft: TCP adaptor

hauleth:


lib/logflare/backends/adaptor/tcp_adaptor/syslog.ex
```diff
@@ -0,0 +1,81 @@
...
+      header(le, options),
+      " ",
+      structured_data(le, options),
+      " ",
+      Jason.encode!(le.body),
+      "\n"
+    ]
+
+    # TODO: Add support for non-transparent framing
+    [to_string(IO.iodata_length(msg)), ?\s, msg]
```
hauleth, line 27: Currently only octet-counting framing is implemented as I have no idea how to do nested configuration there. Non-transparent framing (with newline endings) is simple to implement, as all that is needed is simply removing byte count and space.

lib/logflare/backends/adaptor/tcp_adaptor/syslog.ex
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  defp header(%LogEvent{} = le, options) do
+    level = to_level(le.body["level"] || le.body["metadata"]["level"])
```
hauleth, line 31: I did not have an idea how to extract level from log event in better way. So I am open to having better solution there.

lib/logflare/backends/adaptor/tcp_adaptor/syslog.ex
```diff
@@ -0,0 +1,81 @@
...
+    facility = options[:facility] || 16
```
hauleth, line 32: Facility is currently not configurable, but honestly I do not know if there is much need for that.

lib/logflare/backends/adaptor/tcp_adaptor/syslog.ex
```diff
@@ -0,0 +1,81 @@
...
+
+    ingested_at = DateTime.from_naive!(le.ingested_at, "Etc/UTC")
+
+    id = Ecto.UUID.dump!(le.id) |> Base.encode32(case: :lower, padding: false)
```
hauleth, line 36: Because of 32 characters limit I decided to encode UUIDs with Base32. Theoretically we could just remove `-` (dashes) from canonical representation, but I think that this approach is effective enough as well as it reduces amount of data that needs to be transferred.

lib/logflare/backends/adaptor/tcp_adaptor/syslog.ex
```diff
@@ -0,0 +1,81 @@
...
+
+    [
+      # Level and facility
+      "<#{facility * 8 + level}>1 ",
+      DateTime.to_iso8601(ingested_at),
+      # XXX: Unknown hostname?
+      " -",
```
hauleth, line 43: I left hostname field undefined, though if someone has proposal what could be in this field, then I am open to filling it.

lib/logflare/backends/adaptor/tcp_adaptor/syslog.ex
```diff
@@ -0,0 +1,81 @@
...
+  It uses [octet-counting framing][].
+
+  [RFC5424]: https://www.rfc-editor.org/rfc/rfc5424
+  [octet-counting framing]: https://www.rfc-editor.org/rfc/rfc6587#section-3.4.1
+  """
+
+  alias Logflare.LogEvent
+
+  # TODO: Change it to real value
+  @pen 62137
```
hauleth, line 14: This need to be updated as soon as Supabase or Logflare will have assigned [Private Enterprise Number](https://www.iana.org/assignments/enterprise-numbers/) from IANA.

lib/logflare/backends/adaptor/tcp_adaptor/syslog.ex
```diff
@@ -0,0 +1,81 @@
...
+      " -",
+      " ",
+      id
+    ]
+  end
+
+  defp structured_data(%LogEvent{} = le, _options) do
+    [
+      "[source@#{@pen} name=#{inspect(le.source.name)} id=\"#{le.source.id}\"]"
+    ]
```
hauleth, line 56: If there are more metadata informations that should be available there, then I am open to suggestions.

test/test_helper.exs
```diff
@@ -42,6 +42,15 @@ Mimic.copy(ExTwilio.Message)
 Mimic.stub(Goth)
 Mimic.stub(Finch)
 
-ExUnit.configure(exclude: [integration: true, failing: true, benchmark: true])
+has_telegraf? = not is_nil(System.find_executable("telegraf"))
```
hauleth, line 45: Telegraf is currently not installed on CI, so we will automatically skip these tests. There are some GitHub Actions to install it, but I didn't know what are the policies around adding 3rd party actions, so I have left it for the future.

PR #2068, opened 2024-05-15
feat: replace :global with :syn

Ziinc:
Replace `:global` with `:syn` to prevent locking bottlenecks


lib/logflare/endpoints/cache.ex
```diff
@@ -102,6 +100,11 @@ defmodule Logflare.Endpoints.Cache do
 
     unless state.disable_cache, do: refresh(proactive_querying_ms(state))
 
+    # register on syn
+    pid = self()
+    :syn.register(:endpoints, {query.id, params}, pid)
```
abc3, line 105: out of curiosity, why don't you register the process with {:via, :syn, SOME_ID}?

Ziinc: wanted to put all the syn code together for readability, since it would have to join the group too.
placing it as a via tuple is fine as well, no issues with that.

lib/logflare/endpoints/resolver.ex
```diff
@@ -4,35 +4,35 @@ defmodule Logflare.Endpoints.Resolver do
...
-          case DynamicSupervisor.start_child(Cache, spec) do
-            {:ok, pid} -> pid
-            {:error, {:already_started, pid}} -> pid
-          end
-
-        pid ->
-          Cache.touch(pid)
-          pid
-      end
+    :syn.lookup(:endpoints, {id, params})
```
chasers, line 22: jfyi you [can attach a callback](https://hexdocs.pm/syn/syn_event_handler.html) to do something if two procs end up getting spun up.

We use this in Supavisor and Realtime for a custom resolver. Syn defaults to keeping the last process started (killing the first) and in both Supavisor and Realtime we want to keep the first. 

Just sharing for comms, in case you have to do something similar in the future. We might want to keep the first process but not necessary atm.

chasers (approved):

PR #2071, opened 2024-05-21
fix: add guard for empty email form submissions on email login

Ziinc:
Fix for empty email login submissions when email is empty


Ziinc:
@chasers FYI

PR #2072, opened 2024-05-21
fix: copy-to-clipboard for log event modal

Ziinc:
Fixes copy to clipboard event modal

https://github.com/Logflare/logflare/assets/22714384/a334137b-8c9e-48fc-ac57-49dce040a006


Ziinc:
@chasers FYI

PR #2076, opened 2024-06-05
chore: remove prod/staging cacert and self-signed cert

Ziinc:
This PR removes prod/staging cacert and self-signed cert, leaving the grpc endpoint as is without ssl.


Ziinc:
@chasers FYI

PR #2087, opened 2024-06-07
fix: rule child is started on source routing at ingest time

Ziinc:
Fixes log drains that do not have the rule child started at ingest time, for situations such as multi-cluster setups (as we have in prod), where a rule is created on one cluster and the startup does not propagate across the cluster.

This also adds in an `:rpc.multicall/3` to ensure that usage rules are started/stopped across the cluster

### DIFF
--- a/lib/logflare/sources/rules.ex
+++ b/lib/logflare/sources/rules.ex
@@ -11,6 +11,7 @@ defmodule Logflare.Rules do
   alias Logflare.SourceSchemas
   alias Logflare.Backends.Backend
   alias Logflare.Backends.SourceSup
+  alias Logflare.Cluster
 
   @doc """
   Lists rules for a given Source or Backend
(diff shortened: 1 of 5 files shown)


lib/logflare/sources/rules.ex
```diff
@@ -38,7 +37,9 @@ defmodule Logflare.Rules do
     |> Repo.insert()
     |> case do
       {:ok, %Rule{backend_id: backend_id} = rule} = result when backend_id != nil ->
-        if Backends.source_sup_started?(rule.source_id), do: SourceSup.start_rule_child(rule)
+        # expected to fail on nodes where SourceSup is not started
+        :rpc.multicall(SourceSup, :start_rule_child, [rule])
```
abc3, line 41: just a reminder that the default timeout is set to infinity

Ziinc: refactored to add a default timeout, thanks for the catch 🙏

PR #2092, opened 2024-06-10
feat: add display timezone switcher, remove local timezone toggle

Ziinc:
This PR adds in local timezone display switching, which provides better ux when users want to switch their display timezone while keeping the search tied to a different timezone. 

It forces a differentiation between "search timezone" and "display timezone", where display timezone only changes the displayed timezone, whereas search timezone will affect the search queries that the user will query in.

Adjustments to the search timezone button and moving it closer to the form will come in a follow up PR.

https://github.com/Logflare/logflare/assets/22714384/bd342007-9277-471c-8dfc-d08ba718eff3

Also added in a UTC convenience button

https://github.com/Logflare/logflare/assets/22714384/5d6ae26f-4a36-4bdd-a43c-ea36592ec65a

### DIFF
--- a/assets/js/LogEventsChart.jsx
+++ b/assets/js/LogEventsChart.jsx
@@ -217,11 +217,10 @@ const LogEventsChart = ({
   loading,
   chart_data_shape_id: chartDataShapeId,
   chart_period: chartPeriod,
-  use_local_time: useLocalTime,
-  user_local_timezone: userTz,
+  display_timezone: userTz,
   pushEvent,
 }) => {
-  const tz = useLocalTime ? userTz : "Etc/UTC";
+  const tz = userTz
   const onClick = (event) => {
     pushEvent("soft_pause", {});
     const utcDatetime = event.data.datetime;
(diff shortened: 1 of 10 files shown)


Ziinc:
- [ ] remove timezone preference modal
- [ ] add a remember me button
- [ ] move switcher to subhead row

zoedsoupe:
I'm looking into this one currently

Ziinc:
@msmithstubbs

msmithstubbs:
Working on this one. Added a test, and will work through [check list above ](https://github.com/Logflare/logflare/pull/2092#issuecomment-2162506561)

@Ziinc you happy for me to merge in `main` as well?

Ziinc:
Some pointers for this:
- I think we have enough horizontal space in the subheader to move up the drop down to the same line as the other buttons. We can also reduce the width of the drop down as well as text size. 
- for the "remember" button, it should only appear when there is a change in the tz value (i.e. Only save it when prompted, as opposed to initial behaviour when it saves it all the time)
- working off this branch targeting main is fine. You will likely need to rebase.

PR #2096, opened 2024-06-13
feat: add sampling to open telemetry, exclude ingestion routes.

Ziinc:
Adds in the following safeguards for OpenTelemetry:
- Adds in sampling to opentelemetry tracing. Defaults to `0.001`. 
- All ingestion and endpoint querying routes are dropped and excluded from recording
- Conditionally starts the cowbow tracing only if an otel endpoint is provided.


lib/logflare_web/open_telemetry_sampler.ex
```diff
@@ -0,0 +1,67 @@
+defmodule LogflareWeb.OpenTelemetrySampler do
+  require OpenTelemetry.Tracer, as: Tracer
+  @behaviour :otel_sampler
```
chasers, line 3: Should check how this is implemented in case we can overload it.

Otherwise probably good for now.

chasers (approved):

lib/logflare_web/open_telemetry_sampler.ex
Ziinc: Read through the src for it, it uses the generated trace id to filter based on the probability distribution (i think it a uniform dist), if its in the upper bound tail end of the distribution then it will decide. It is is a pure function so cost is low, plus this is decided on span emission so it is quite safe.

PR #2118, opened 2024-06-25
perf: use :ets to cache recent logs

Ziinc:
This PR refactors the RecentLogsServer to move caching outside of the process. This will help with reducing memory footprint and gc-ing.

I have also optimized the code to have nearly the same reductions as before, so it should be just as performant while being fully stateless without a process involved.


lib/logflare/backends.ex
```diff
@@ -478,21 +479,62 @@ defmodule Logflare.Backends do
...
+    :ets.delete(:recent_events, source_id)
+
+    # sort newest to oldest
+    objects =
+      for event <- Enum.sort_by(log_events, fn %{body: %{"timestamp" => ts}} -> ts end, &<=/2) do
+        # don't cache source information
+        {source_id, %{event | source: nil}}
+      end
+
+    :ets.insert(:recent_events, Enum.take(objects ++ existing, 100))
```
chasers, line 496: Can maybe sort and take 100 in one go with reduce?

lib/logflare/backends.ex
```diff
@@ -478,21 +479,62 @@ defmodule Logflare.Backends do
...
-    RecentLogsServer.list(source.token)
+  @spec list_recent_events_local(Source.t()) :: [LogEvent.t()]
+  def list_recent_events_local(%Source{id: source_id}) do
+    ms =
+      Ex2ms.fun do
+        {id, event} when id == ^source_id -> event
+      end
+
+    :ets.select(:recent_events, ms)
+    |> Enum.sort_by(fn %{body: %{"timestamp" => ts}} -> ts end, &<=/2)
```
chasers, line 524: So we need to sort here if we're sorting in insert?

lib/logflare/backends/recent_events.ex
```diff
@@ -0,0 +1,16 @@
...
+
+  @ets_table_name :recent_events
+
+  ## Server
+  def start_link(_args) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__, hibernate_after: 1_000)
+  end
+
+  def init(_args) do
+    :ets.new(@ets_table_name, [:public, :named_table, :duplicate_bag])
```
chasers, line 13: Lets be obvious here and set the `decentralized_counters`, `write_concurrency` and `read_concurrency` opts even if we think the defaults are good.

lib/logflare/source/recent_logs_server.ex
```diff
@@ -6,147 +6,41 @@ defmodule Logflare.Source.RecentLogsServer do
...
-  end
-
   ## Server
   def start_link(args) do
     GenServer.start_link(__MODULE__, args,
       name: Backends.via_source(args[:source], __MODULE__),
-      spawn_opt: [
-        fullsweep_after: 20
-      ],
+      spawn_opt: [fullsweep_after: 50],
```
chasers, line 22: Not sure when you added the `fullseep_after` exactly but will definitely contribute to CPU usage.

chasers: However, will be much less now that log events are not in state.

lib/logflare/backends.ex
chasers: I guess unless it's a `sorted_set` we should make sure to re-sort.

lib/logflare/source/recent_logs_server.ex
```diff
@@ -229,26 +113,6 @@ defmodule Logflare.Source.RecentLogsServer do
     {:ok, current_cluster_inserts, current_inserts}
   end
 
-  def load_init_log_message(source_token) do
```
chasers, line 232: This is useful to see in some cases, but not married to it. Can add it back later if we want.

chasers (changes requested):
Only major thing:
- [ ] Let's specify the ets table opts even if the defaults are what we want

lib/logflare/backends.ex
```diff
@@ -478,21 +479,62 @@ defmodule Logflare.Backends do
     end
   end
 
+  def push_recent_events(%Source{}, []), do: :ok
+
+  def push_recent_events(%Source{id: source_id}, log_events) do
+    existing = :ets.lookup(:recent_events, source_id)
+
+    :ets.delete(:recent_events, source_id)
```
chasers, line 487: This seems hairy. We're going to have many processes doing this at once.

chasers: My feeling is that we're going to want to do only inserts here and then a a janitor process doing the deletes in the background, similar to how Cachex does maintains ets tables.

But, since this powers only the initial logs list view we can punt if the perf is there.

chasers, line 487: At least use `:ets.take` here yeah?

chasers (changes requested):

lib/logflare/backends.ex
chasers: Yeah so for very busy sources the snapshot of the "last 100" you get when you load the page is not really going to be the last 100 with  many processes doing this or even doing `:ets.take`.

But again, might be close enough.

lib/logflare/backends.ex
Ziinc: The existing items don't need to be sorted as it is ordered_set, sorting only the incoming reduces need to double sort

Ziinc: Doing it separately cuts down reductions when there is a full buffer

lib/logflare/backends.ex
Ziinc: Yea we can remove this since it's an ordered_set

lib/logflare/source/recent_logs_server.ex
Ziinc: Yea we can cut it out since there isn't anything stored in process state anymore

lib/logflare/backends.ex
Ziinc: Ah yes we can do that since we're already limiting count on insert

chasers: Actually @Ziinc if the goal here is to reduce GC, then we need to do the janitor process.

Because every time we select data from ETS we're copying it and generating double the garbage we need to. 

If we have a janitor process and delete without returning any data than I think we'd significantly reduce copying.

PR #2142, opened 2024-07-16
feat: one ets queue per buffer producer

Ziinc:
This PR adjusts the DynamicPipelines to use one ets queue per buffer producer, ensuring no race conditions while also allowing "sharding" of the queues to make traversals faster.
This required implementing of a "startup queue" concept, which is a default queue where events are temporarily stored if there are no active pipelines.
Items are at most stored in the startup queue for 5s, until the dynamic pipeline resolver picks up the startup items queue count and starts new pipelines. Sudden bursts are also accounted for in the resolving logic.

I've also implemented a "chunked round robin" utility, which is used to round robin a batch across many queues until the batch is empty.

I have also bumped up the max number of pipelines, so that really large sources can scale higher.


lib/logflare/backends/ingest_event_queue/queue_janitor.ex
```diff
@@ -53,20 +52,12 @@ defmodule Logflare.Backends.IngestEventQueue.QueueJanitor do
...
-      Backends.get_and_cache_local_pending_buffer_len(state.source_id, state.backend_id)
-
-    if pending_size > state.remainder do
-      # drop all ingested
-      IngestEventQueue.truncate(sid_bid, :ingested, 0)
-    else
-      IngestEventQueue.truncate(sid_bid, :ingested, state.remainder)
-    end
-
+    IngestEventQueue.truncate_queues(sid_bid, :ingested, state.remainder)
```
Ziinc, line 55: would leave a large total remainder across all queues if there are many pipelines.
can drop all ingested if there is sufficient pending count.

PR #2152, opened 2024-07-19
perf: set max dirty allocator threads

chasers:
seeing erts_literal_area_collector in observer a lot … ran into [this thread](https://github.com/erlang/otp/issues/4728) … which leads to [this comment](https://github.com/erlang/otp/issues/4728#issuecomment-829199755) … which seems like something we could be seeing


chasers:
Merging to see if I can get this on staging

PR #2162, opened 2024-08-02
fix: default value for webhook gzipping

Ziinc:
Fixes default value setting for gzip field for webhook config


Ziinc:
@chasers fyi

PR #2163, opened 2024-08-02
feat: initial cloak implementation

Ziinc:
This PR adds in initial cloak implementation for encryption at rest.
Adds in a new env var that is required `LOGFLARE_DB_ENCRYPTION_KEY`, and for key rotation the env var is `LOGFLARE_DB_ENCRYPTION_KEY_RETIRED`.

Only backend config is encrypted for now.

Migration is non-destructive, this version should only add the new column `:config_encrypted`

Pre-merge todos:
- [x] manual verification of secrets rolling
- [x] manual verification of config column migration
- [ ] create PR for config column removal, v 1.8.1
- [x] add encryption key for all envs


docs/docs.logflare.com/docs/self-hosting/index.md
```diff
@@ -56,6 +58,25 @@ Additional environment variable configurations for the OpenTelemetry libraries u
...
+
+In order to roll encryption keys and migrate existing encrypted data, use the `LOGFLARE_DB_ENCRYPTION_KEY_RETIRED` environment variable.
+
+Steps to perform the migration are:
+
+1. Move the retired encryption key from `LOGFLARE_DB_ENCRYPTION_KEY` to `LOGFLARE_DB_ENCRYPTION_KEY_RETIRED`.
+2. Generate a new encryption key and set it to `LOGFLARE_DB_ENCRYPTION_KEY`.
+3. Restart or deploy the server with the new environment variables.
+4. Upon successful server startup, an `info` log will be emitted that says that an retired encryption key is detected, and the migration will be initiated to transition all data encrypted with the retired key to be encrypted with the new key.
+5. Once the migration is complete, the retired encryption key can be safely removed.
```
staaldraad, line 78: "obvious" question, is there something that indicates migration is complete?

lib/logflare/vault.ex
```diff
@@ -0,0 +1,90 @@
...
+      [
+        default:
+          {Cloak.Ciphers.AES.GCM, tag: "AES.GCM.V1." <> hash(default_key), key: default_key},
+        retired:
+          if(retired_key != nil,
+            do: {Cloak.Ciphers.AES.GCM, tag: "AES.GCM.V1" <> hash(retired_key), key: retired_key},
+            else: nil
+          )
+      ]
+      |> Enum.filter(fn {_k, v} -> v != nil end)
```
staaldraad, line 42: this is a really nice feature in cloak 💪🏼

lib/logflare/vault.ex
```diff
@@ -0,0 +1,90 @@
...
+
+    config = Keyword.put(config, :ciphers, ciphers)
+
+    Task.start_link(fn ->
+      # wait for genserver config to be saved, see https://github.com/danielberkompas/cloak/blob/v1.1.4/lib/cloak/vault.ex#L186
+      :timer.sleep(1_000)
+
+      if Keyword.has_key?(ciphers, :retired) do
```
staaldraad, line 50: is it worth having a check that `default != retired` to skip the migration if for some reason identical keys are passed? Maybe Cloak does this internally already but if not, saving cycles on not "re"-encrypting data would be nice

staaldraad (approved):
just some questions and a possible change, but nothing blocking

chasers (approved):

docs/docs.logflare.com/docs/self-hosting/index.md
Ziinc: yea we'll need to add a log line for it, good catch 👍

lib/logflare/vault.ex
Ziinc: the migration won't run for data that is on the default key, as cloak tags the encrypted string and filters based on it. hence why the sha256 hash is needed.

PR #2166, opened 2024-08-06
feat: source.service_name for setting the service name on backends

Ziinc:
This PR adds in a `service_name` field to the `sources` table, allowing grouping of sources by that field. It is used in the logflare UI to group sources together.

This PR uses an existing commit for improving the dashboard template to heex.

For the datadog backend, it will be used as the `ddservice`, and if not set, the source name will be used. The entire event body will be set on the `data` key on the datadog payload, to prevent key clashing based on their reserved events.

It also changes the `ddsource` to `supabase`

Screenshot 2024-08-06 at 7 14 35 PM
Screenshot 2024-08-06 at 7 19 03 PM

Screenshot 2024-08-06 at 7 40 01 PM


Ziinc:
@chasers FYI

PR #2167, opened 2024-08-08
feat: private access token provisioning for single tenant mode

Ziinc:
This PR adds in access token provisioning in single tenant mode, where the initial private access token is now seeded on startup in order to allow management api access.

This allows users to use the management API to programmatically interact with the server without having to go through the user interface.


Ziinc:
@chasers FYI

PR #2177, opened 2024-08-20
fix: don't flood logs on bad user urls

Ziinc:
This fixes bug where bad user input urls will flood logs from finch
Screenshot 2024-08-21 at 2 49 31 AM

We should technically stop attempting requests and disable the backend if this occurs too much...

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -40,7 +40,11 @@ config :logflare, Logflare.PubSub, pool_size: 10
 config :logger,
   handle_otp_reports: true,
   handle_sasl_reports: false,
-  level: :info
+  level: :info,
+  compile_time_purge_matching: [
+    # to prevent pool connection attempts for bad user urls from flooding logs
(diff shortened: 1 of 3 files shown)


chasers:
Well, yeah how do we know this is happening then? 

I'm afraid we'll merge this and then not fully understand the impact, or let bad drains exist and build up for a long time.

Ziinc:
agreed, this would affect all finch pools so not ideal.

Issue is that the log gets emitted at the pool level
Looks like they changed this to a [warning level](https://github.com/sneako/finch/blob/80c2d423e3310549bd35749fb7f21ea96dadaef4/lib/finch/http2/pool.ex#L285)... perhaps i can switch the log level at runtime just before the call.

> let bad drains exist and build up for a long time.
we can handle the Tesla.Env to disable the drains automatically after x number of retries, so that would still be fine, so long as we suppress the finch log

PR #2179, opened 2024-08-21
fix: event message formatting for datadog adaptor

Ziinc:
This PR fixes a formatting error for datadog adaptor
image


Ziinc:
@chasers FYI

PR #2181, opened 2024-08-22
fix: remove saved search counters usage

Ziinc:
This PR removes usage of the SavedSearchCounters, which is causing db timeouts when deleting a source (due to the very large table that builds up over time).

This table and the related analytics dashboard is obsolete and increases maintenance burden, it is easier to do something with telemetry if we want to analyze LQL usage, which is currently not a business concern.

The underlying db table is left intact, but all application code using it will be removed.


Ziinc:
@chasers FYI

PR #2187, opened 2024-08-28
perf: remove RecentLogsServer cache

Ziinc:
This removes the RecentLogsServer cache and only uses it for last ingest ingest timestamp updating.

https://github.com/user-attachments/assets/1df79036-2118-42a5-ad66-18d1e1fd98a6


lib/logflare/source/recent_logs_server.ex
```diff
@@ -191,10 +92,6 @@ defmodule Logflare.Source.RecentLogsServer do
   end
 
   ## Private Functions
-  defp do_push({:push, _source_token, [le | _] = log_events}, state) do
-    new_queue = (Enum.to_list(state.recent) ++ log_events) |> LQueue.from_list(100)
```
chasers, line 195: Yeah I think this was a huge GC offender. Take a look at the LQueue source code.

chasers (approved):

lib/logflare/source/recent_logs_server.ex
Ziinc: omg its in an Agent 🤯

chasers: Yeah fml

PR #2214, opened 2024-09-24
perf: reduce retries gax

Ziinc:
Reduce number of retries for gax
https://github.com/Logflare/elixir-google-gax/commit/6772193c438dd72ac253ea34ab13e9c02842cba8


Ziinc:
@chasers FYI

PR #2226, opened 2024-10-11
Fix IPv6 support and support listening to a specific IP

tvogel:
For running logflare on fly.io with their IPv6 based private networking, I had to fix/improve a few things. Please take a look and pick what you like!

(My first exposure to Elixir/Erlang, so thanks for that!)


config/runtime.exs
```diff
@@ -211,7 +219,8 @@ cond do
            :postgres_backend_adapter,
            filter_nil_kv_pairs.(
              url: System.get_env("POSTGRES_BACKEND_URL"),
-             schema: System.get_env("POSTGRES_BACKEND_SCHEMA")
+             schema: System.get_env("POSTGRES_BACKEND_SCHEMA"),
+             socket_options: [:inet6]
```
Ziinc, line 223: should be a flag or from ip parsing

mix.exs
```diff
@@ -226,7 +226,10 @@ defmodule Logflare.Mixfile do
       {:opentelemetry_exporter, "~> 1.6"},
       {:opentelemetry_phoenix, "~> 1.1"},
       {:opentelemetry_cowboy, "~> 0.2"},
-      {:live_monaco_editor, "~> 0.1"}
+      {:live_monaco_editor, "~> 0.1"},
+
+      # IP address parsing
+      {:net_address, "~> 0.3.1"}
```
Ziinc, line 232: dependency not necessary, can use :inet module
https://www.erlang.org/doc/apps/kernel/inet.html#parse_address/1

config/runtime.exs
Ziinc: can use :inet to check the type of address
https://www.erlang.org/doc/apps/kernel/inet.html#is_ipv4_address/1
maybe can have a helper function for this, would make the code clearer

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
```diff
@@ -44,6 +44,7 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.SharedRepo do
 
         opts ++ fields
       end
+    opts = opts ++ (Map.take(config, [:socket_options]) |> Map.to_list())
```
Ziinc, line 47:
```suggestion
    opts = opts ++ [socket_options: config[:socket_options]]
```

Ziinc (changes requested):

Ziinc:
Thanks for the PR, looks like a good addition for self hosting setup

config/runtime.exs
tvogel: One problem is that `is_ipv4_address/1` only checks for the type of the IP passed in (whether the list of numbers has length 4, I think), so it must first be resolved and if there are both IPv4 and IPv6 available for a given host name, we must pick. It would actually be a lot better if the connection logic would determine the socket type and not the runtime configuration. I am convinced, you could implement that a lot quicker than me.

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
tvogel: The advantage of `take` would be that it also works if the `:socket_options` are omitted, right?

config/runtime.exs
```diff
@@ -77,6 +84,7 @@ config :logflare,
          hostname: System.get_env("DB_HOSTNAME"),
          password: System.get_env("DB_PASSWORD"),
          username: System.get_env("DB_USERNAME"),
+         socket_options: [ :inet6 ],
```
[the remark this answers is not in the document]
tvogel: Is now resolving and checking `DB_HOSTNAME` to determine IP type.

config/runtime.exs
tvogel: Is now parsing the URL, resolving the host and determining IP type.

mix.exs
tvogel: Done!

Ziinc (changes requested):
thanks for the adjustments, was OOO. I can take over the postgres adaptor backend adjustments needed.

config/runtime.exs
```diff
@@ -204,14 +233,22 @@ if config_env() != :test do
   config :grpc, port: System.get_env("LOGFLARE_GRPC_PORT", "50051") |> String.to_integer()
 end
 
+socket_options_for_url = fn url when is_binary(url) ->
+  case URI.parse(url) do
+    %URI{host: host} -> socket_options_for_host.(host)
+    _ -> raise "Failed to parse URL: #{inspect(url)}"
+  end
+end
+
```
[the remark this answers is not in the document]
tvogel: As I am an Elixir newbie, I don't fully understand the fallback-clause. Shouldn't the `nil`-case be excluded by the `when is_binary()` clause? Why can a fallback cover for it in this "overload"?

tvogel:
(Sorry for the late replies - I was traveling.)

6ixfalls:
@Ziinc, any update on this pull request? Hitting this issue currently as well and would love for this to be merged.

config/runtime.exs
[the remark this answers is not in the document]
Ziinc: the nil is excluded in the first match clause, but it still will need a fallback clause to mat h on, otherwise the MatchError will raise.
https://hexdocs.pm/elixir/case-cond-and-if.html

lib/logflare/backends/adaptor/postgres_adaptor/shared_repo.ex
```diff
@@ -45,6 +45,8 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor.SharedRepo do
         opts ++ fields
       end
 
+    opts = opts ++ [socket_options: config[:socket_options]]
```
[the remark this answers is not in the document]
Ziinc: In this case, with inserts into one postgres instance it is ok. However, it will not apply in the situation where there is 2 postgres instances in a multi-backend setup, 1 on ipv6 and the other on ipv4. The same socket options would not apply

Ziinc:
@tvogel could you adjust the two remaining issues for the merge?
1. socket opts only on single tenant mode
2. case clause fallback

ruslandoga:
👋 

Just FYI: https://github.com/plausible/analytics/pull/4245

Doing it this way removes the need for users to configure IP version in outgoing connections. And note that the server can always listen with `:inet6` since that option makes it bind both to IPv4 and IPv6 :)

PR #2232, opened 2024-10-21
fix: partition task supervisor

chasers:
- all Task.Supervisor calls are now ran through Utils.Tasks module
- Tasks Supervisor is now partitioned
- big one: doesn't spawn a task for every log event


lib/logflare/backends.ex
```diff
@@ -517,16 +515,16 @@ defmodule Logflare.Backends do
...
     task =
-      Task.async(fn ->
+      Tasks.async(fn ->
         nodes
         |> Enum.map(
-          &Task.Supervisor.async({TaskSupervisor, &1}, __MODULE__, :list_recent_logs_local, [
-            source
-          ])
+          &Tasks.async(fn ->
+            :erpc.call(&1, __MODULE__, :list_recent_logs_local, [source], 10_000)
```
filipecabaco, line 522: Isn't this timeout too high?

filipecabaco (approved):
Just one question regarding the erpc timeout. Wondering if it needs to be so high and what could be the impact of accumulating erpc calls

PR #2247, opened 2024-11-04
perf: add back fullsweep_after to Task.Supervisor

Ziinc:
Added back spawn_opt for Task.Supervisor, prevents garbage buildup.
Was removed in https://github.com/Logflare/logflare/commit/40db716e82cb3bc4041709f2aec8de6fe811cd53#diff-14a0c2e2cb2fe12cfea0794ab7c068ea2f25fa332f3fb7e0351e9d63b0266adfL81

### DIFF
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -51,7 +51,9 @@ defmodule Logflare.Application do
          name: Logflare.V1SourceRegistry,
          keys: :unique,
          partitions: max(round(System.schedulers_online() / 8), 1)},
-        {PartitionSupervisor, child_spec: Task.Supervisor, name: Logflare.TaskSupervisors},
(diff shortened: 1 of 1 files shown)


Ziinc:
the spawn_opt gets forwarded to the underlying process
https://github.com/elixir-lang/elixir/blob/v1.17/lib/elixir/test/elixir/task/supervisor_test.exs#L31

chasers (changes requested):
Removing this didn't change the memory usage profile at all.

And I think we need to do less GC'ing.

Ziinc:
As discussed, leaving this open for observation.

chasers (approved):

Ziinc:
Closing as memory isn't an issue now.

PR #2250, opened 2024-11-04
perf: tweak LogEventsCache size and ttl

Ziinc:
Reduces size of LogEventsCache further


chasers:
This is really not taking up a lot of ram:

Screenshot 2024-11-07 at 6 50 32 AM

chasers (approved):

chasers:
Yeah probably better to do more records and lower expiration.,

PR #2258, opened 2024-11-19
fix: env var renaming, OTEL fix attempt

Ziinc:
- renames OTEL sample ratio env var to `LOGFLARE_OTEL_SAMPLE_RATIO`
- updates self-signed ssl cert flow to not use CF as cacert (🤞 ) as the GCP load balancer seems quite picky in requiring a TLS connection between load balancer and origin server for http/2 protocol.

Screenshot 2024-11-19 at 4 27 16 PM


Ziinc:
@chasers fyi

PR #2259, opened 2024-11-19
feat: log levels in search page, and removal of UDFs

Ziinc:


msmithstubbs:
Taking a look at this one. Is there any other discussion around requirements here?

So far, my understanding is:

1. We are going to pull out the `metadata.level` value and use that to style the log line in the search page
2. Is there a list of log levels to support? Or do we use the standard syslog ones?
3. I'm not sure about 'remove UDFs in generated code' and how that fits in. Any more details on this one?

Ziinc:
1. yes, we'll use `metadata.level` for now. 
2. yes, syslog levels works.
3. Right now, generated queries rely a lot on UDFs that has the `LF_` prefix. We want to move away from these type of generated queries. It is fine to ditch the current ones and hand-craft the ecto queries that then get executed on the BQ backend.

msmithstubbs:
First pass attempt at this:

CleanShot 2025-07-17 at 16 02 41@2x

Note the first three events don't have a log level, they default to the `info` colour but don't show a log level. Last one *does* have level `info` so it is shown.

I'm using the colours from LogEventsChart for consistency. 

@Ziinc Let me know if any feedback.

Ziinc:
- for debug we can switch to blue, purple looks like it clashes with the other colors. Not sure if there is already a nice blue in the color palette.
- could we remove the tz display from the timestamp? Since the preference tz is now more in focus, we don't need to show the tz part for the timestamp and can use that space to display the level.
- not too sure how nice this will look, but can we try using a colored text instead of coloring the entire level text background? The disjointed box feels a little weird. If you think it looks like crap when trying it out then we'll stick to the floating boxes 😁

msmithstubbs:
Tried a few variations.


### 1. Timezone removed, debug changed to light blue

CleanShot 2025-07-22 at 13 46 14@2x


### 2. Logs levels with coloured text on black background

CleanShot 2025-07-22 at 13 52 22@2x


### 3. Also tried a pill shape for the log levels

CleanShot 2025-07-22 at 14 00 39@2x



I like 2 or 3, depending on how much attention we want the log levels to have. No. 2 is definitely more subtle.

Ziinc:
Let's go with 2 so that it doesn't distract too much from the data 👍

PR #2266, opened 2024-12-04
feat: cache warmer for Users context

Ziinc:
This PR adds in a cache warmer for the Users context, which will populate the cache on startup using active ingesting sources.


lib/logflare/application.ex
```diff
@@ -37,15 +37,15 @@ defmodule Logflare.Application do
   defp get_children(:test) do
     finch_pools() ++
       [
+        Logflare.Repo,
+        Logflare.Vault,
         Counters,
         RateCounters,
         Logs.LogEvents.Cache,
         ContextCache.Supervisor,
```
chasers, line 45: ContextCache needs to be started before the Repo otherwise cache requests will fail on startup if it starts getting requests right away.

chasers: But I guess you need the Repo to warm the Cache when it starts?

idk, I've definitely had the issue where the Cache was not started first and things exploded.

Ziinc: hmm lets adjust the health checks to only accept requests once all caches + repo is ready, no reason to accept requests early before startup is complete anyway

Ziinc: - added db health checks
- added cachex health checks

Didn't make the warmer blocking, though it is something to consider.

PR #2277, opened 2024-12-10
chore: staging tweaks

Ziinc:
- turns of cache stats collection on staging
- bumps instance types to c2d-highcpu-16


Ziinc:
@chasers FYI

PR #2280, opened 2024-12-11
perf: increase buffer len to 500k

Ziinc:
increase it further for load testing


Ziinc:
not the right solution for the bug.

PR #2283, opened 2024-12-16
perf: increase max ets tables allocation

Ziinc:
Screenshot 2024-12-16 at 11 36 23 AM

    
    > If large amounts of named tables are used and `ERL_MAX_ETS_TABLES`
    hasn't been increased, the performance of named table lookup will degrade.
    > 
    
    https://www.erlang.org/doc/apps/stdlib/ets.html#module-table-traversal


Ziinc:
@chasers FYI

PR #2289, opened 2024-12-20
perf: acks delete ingested events from queue

Ziinc:
immediately delete events from ets queue if queue if rate is fast enough to recent logs from incoming events


Ziinc:
@chasers FYI

PR #2291, opened 2025-01-02
fix: removes task spawning for listing recent logs

Ziinc:
Removes task spawning and uses :erpc.mutlicall/5 instead for better exception handling.

Issue: https://linear.app/supabase/issue/ANL-622/fix-list-recent-logs-task-match-error


Ziinc:
@chasers FYI

PR #2295, opened 2025-01-08
feat: emit egress telemetry via otel span

Ziinc:
This PR emits egress telemetry via an OTEL span.
This will get emitted for each outbound http request that is sent, covering Alerts and Backends. The backend metadata will get merged into the emitted attribute, with the `backend.*` attribute. If sent to a logflare otel endpoint, it will get converted into `_backend_my_metadata`.

This also switches the otel http server instrumentation for phoenix to bandit.

This PR also defaults the sampling to 1.0, as ingest and endpoint routes are already excluded by the sampler.

Screenshot 2025-01-08 at 2 20 54 PM


lib/logflare/application.ex
```diff
@@ -24,8 +24,8 @@ defmodule Logflare.Application do
     children = get_children(env)
 
     if Application.get_env(:logflare, :opentelemetry_enabled?) do
-      :opentelemetry_cowboy.setup()
-      OpentelemetryPhoenix.setup(adapter: :cowboy2)
+      OpentelemetryBandit.setup()
```
chasers, line 27: You were worried about too much data from ingest traces.

This gets us otel for egress but does not ship traces for all ingest requests?

Ziinc: I realised that I had excluded the ingest and endpoint route traces in the sampler 😂
I think I'll add back the routes but with a different sample ratio env var

Ziinc:
@chasers FYI

PR #2298, opened 2025-01-17
fix+feat: bandit attribute for path checking, separate ingest/endpoint sample ratio

Ziinc:
This PR fixes the path checking for the sampler, which was previously using cowboy instrumentation attributes.

It also separates out the ingest/endpoint-related paths to a separate sample ratio, `LOGFLARE_OTEL_INGEST_SAMPLE_RATIO` and `LOGFLARE_OTEL_ENDPOINT_SAMPLE_RATIO`. Behaviour is such that it defaults to the general sample ratio. This allows different sampling for high volume routes.

Staging and prod ingest sampling is dropped to `0.1` and `0.01` respectively. 
Staging and prod endpoint sampling is dropped to `0.2` and `0.1` respectively.


Ziinc:
@chasers FYI merging first for testing

chasers:
lgtm

PR #2302, opened 2025-01-22
perf: drop db connection pool to 15

Ziinc:
Drops db connection pool to 15 to be more conservative with connections


Ziinc:
@chasers FYI

PR #2305, opened 2025-01-24
feat: http protobuf otel ingestion

Ziinc:
This PR adds in http_protobuf ingestion at the `/v1/traces` route. This PR introduces a new protobuf parser for decoding collector trace export requests.

It also simplifies the number of accepted source uuid headers to just `x-source` and `x-collection`


Ziinc:
@chasers FYI

PR #2308, opened 2025-02-03
perf: add index for partner_users :upgraded field

Ziinc:
Performance optimization for when checking if tenants are upgraded


Ziinc:
@chasers FYI

PR #2311, opened 2025-02-05
 grpc-encoding `identity` should be handled, x-source-id removal 

Ziinc:
`identity` is set by some opentelemetry SDKs, it should be handled as a passthrough

This PR also fixes a traces exporting bug where `x-source-id` has since been removed.


Ziinc:
@chasers FYI

PR #2316, opened 2025-02-17
fix: prevent old logs from getting sent in slack notiifications

Ziinc:
Fixes SlackHookServer to filter out events that are stale.

Also fixes ANL-701 race condition by ensuring that source is always fetchable, in situations where source is deleted when slack notification is being triggered concurrently.

Fixes ANL-700 where nil kv pairs were not getting correctly displayed.

Also fixes ANL-702 by resulting in a no-op if ContextCache is not yet available on instance.


lib/logflare/source/slack_hook_server/slack_hook_server.ex
```diff
@@ -76,4 +74,84 @@ defmodule Logflare.Source.SlackHookServer do
...
+      0 == rate ->
+        []
+
+      rate in 1..3 ->
+        recent_events
+        |> Enum.take(-rate)
+
+      true ->
+        recent_events
+        |> Enum.take(-3)
```
drobnyd, line 148: Does it mean that we can't get more than 3 events forwarded within a given 60 second interval?

chasers (approved):

lib/logflare/source/slack_hook_server/slack_hook_server.ex
Ziinc: No, it would only ensure that freshness of events sent is within past 60s, and truncates full number of events to <=3.
it will still follow the set notification interval.
Due to cluster size, it is likely that more than 1 message is sent within the notification interval.
Notifications are not for funnelling all received events, as it is not optimized for that. We have a slack backend planned in the future that would be able to handle such a use case.

PR #2320, opened 2025-02-24
feat: remove usage of :partner_users table

Ziinc:
This optimizes the partner information storage to a 1-many relationship and removes the intermediary table that was previously used for storing tier upgrading information.


chasers (approved):

priv/repo/migrations/20250221154207_create_user_partner_details_column.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Logflare.Repo.Migrations.CreateUserPartnerDetailsColumn do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :partner_details, :map
```
chasers, line 6: Actually, if we have the partner_id now on the user do we really need this partner_details json?

Looking up the partner with the partner_id should be fast.

chasers (changes requested):

priv/repo/migrations/20250221154207_create_user_partner_details_column.exs
Ziinc: Needed to store the :upgraded value, otherwise can't differentiate between free and enterprise plan.

chasers (approved):

PR #2322, opened 2025-02-25
feat: adjust default chart grouping to hourly

Ziinc:
This should avoid users having the issue where the default chart grouping causes issues when doing filtering, since default query range is 48 hours.


kevcodez (approved):

kamilogorek (approved):

Ziinc:
Closing for now due to cost reasons. will move to user preference-based

PR #2339, opened 2025-03-06
fix: set jobs to run locally for quantum run strategy

Ziinc:
This PR sets the run strategy for newly created quantum jobs


Ziinc:
@chasers fyi

PR #2340, opened 2025-03-06
fix: correctly check for whether rule backend is started.

Ziinc:
Causing ingestion with backend rules to fail due to bottleneck


Ziinc:
@chasers FYI

PR #2343, opened 2025-03-06
chore: add timeout error handling

Ziinc:
adds timeout handling for syn


Ziinc:
Occurs when scheduler is overworking, response should 500 in any case.

PR #2351, opened 2025-03-25
chore: Split RecentLogsServer by functions

Ziinc:
This PR splits the RLS module into two separate genservers, for better separation of concerns and easier debugging. They have also been moved to the Backends module namespace to indicate that is has been refactored for v2 ingestion.

This PR also cleans up some old redundant test files, and removes the outdated architecture diagram.
Test coverage for the RecentEventsTouch server has been added, which was previously not available.
I have also simplified the logic for determining when to touch the Source.


chasers (approved):

lib/logflare/backends/recent_events_touch.ex
```diff
@@ -0,0 +1,99 @@
...
+    max = :timer.minutes(55)
+    Enum.random(min..max)
+  end
+
+  def handle_info(:touch, %{source_id: source_id} = state) do
+    source =
+      source_id
+      |> Sources.Cache.get_by_id()
+
+    Backends.list_recent_logs_local(source)
```
chasers, line 57: btw I think this is one thing that kills us.

Can we just do this logic with an ETS query instead?

Because when we return all the log events from ETS it is copying these large JSONs every time.

chasers: Also, I am on prod and this fn is giving me 200 log events when it seems like it should only give me 100.

```
source_token |> Logflare.Sources.get_source_by_token() |> Logflare.Backends.list_recent_logs_local() |> Enum.count()
200
```

Ziinc: > this fn is giving me 200 log events when it seems like it should only give me 100.
This function takes 100 events from each queue, so if there are two broadway pipelines there would be 200 events.
We could do the sort+take within the func but decided to let the caller handle it instead...


Yea placing it in a Task is fine in this case

Ziinc: `Backends.list_recent_logs_local(source)` is actually multiple sequential ets queries, have benchmarked it and it isn't slow. The variable binding would have some minor memory impact ofc, definitely can fire off a task for this.  Would be able to tone down the GCing if so.

Ziinc: fixed both issues.

chasers (approved):

PR #2352, opened 2025-03-25
feat: GenSingleton

Ziinc:
Generic cluster-wide singleton GenServer, to allow for logic reuse. This has been added to the RecentEventsTouch server to allow for a single genserver cluster-wide to perform the record updated, instead of every node in the cluster performing the update.

This module should only be used for non-critical functions or functions that are not time-sensitive, as it uses a naive interval checking logic.

It is generic enough to be applicable to both syn and global.

importantly, this reduces the total transactions occuring across the cluster, thereby also reducing cainophile messages broadcasted.

### DIFF
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -397,7 +397,6 @@ defmodule Logflare.Backends do
         [{pid, _}] -> {:ok, pid}
         _ -> {:error, :not_started}
       end
-
     end
   end
 
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -6,6 +6,7 @@ defmodule Logflare.Backends do
   alias Logflare.Backends.SourceRegistry
   alias Logflare.Backends.SourcesSup
   alias Logflare.Backends.SourceSup
+  alias Logflare.Backends.RecentEventsTouch
   alias Logflare.Backends.IngestEventQueue
   alias Logflare.LogEvent
   alias Logflare.Repo
@@ -369,6 +370,10 @@ defmodule Logflare.Backends do
   def via_source(source_id, mod, id), do: via_source(source_id, {mod, id})
 
   def via_source(%Source{id: id}, process_id), do: via_source(id, process_id)
+
+  def via_source(id, RecentEventsTouch) when is_number(id) do
+    {:via, :syn, {:core, RecentEventsTouch}}
+  end
 
   def via_source(id, process_id) when is_number(id) do
     {:via, Registry, {SourceRegistry, {id, process_id}}}
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -388,11 +388,16 @@ defmodule Logflare.Backends do
   end
 
   def lookup(module, %Source{} = source) do
-    {:via, _registry, {registry, via_id}} = via_source(source, module)
+    {:via, registry_mod, {registry, via_id}} = via_result = via_source(source, module)
 
-    case Registry.lookup(registry, via_id) do
-      [{pid, _}] -> {:ok, pid}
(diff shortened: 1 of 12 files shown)


Ziinc:
Changes to implement: 
- instead of a child spec, use an anon function or accept a module + args
- switch to genserver and link the proc
- if already started, each GenSingleton process will monitor the started child proc
- on exit, do a random wait  before attempting to start the proc. Do a syn.lookup check first before doing so
- rely on syn's conflict resolution instead of using a lock
- add nanosecond timestamp to syn proc metadata for 

why:
- don't need another layer of supervisor if it is already going to be supervised
- short duplication of processes is ok
- avoids complexity & message passing of obtaining/releasing locks

@Logflare/dashbit please advise if proposed changes & thought process is sound

lib/logflare/backends.ex
```diff
@@ -383,11 +388,15 @@ defmodule Logflare.Backends do
   end
 
   def lookup(module, %Source{} = source) do
-    {:via, _registry, {registry, via_id}} = via_source(source, module)
+    {:via, registry_mod, {registry, via_id}} = via_result = via_source(source, module)
 
-    case Registry.lookup(registry, via_id) do
-      [{pid, _}] -> {:ok, pid}
-      _ -> {:error, :not_started}
+    if registry_mod == :syn do
```
josevalim, line 393: I believe you should be able to do `GenServer.whereis(via_result)` and it will dispatch to the mod accordingly :)

lib/logflare/gen_singleton.ex
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def count_children(pid) do
+    GenServer.call(pid, :count_children)
+  end
+
+  @impl true
+  def init(args) do
+    interval = args[:interval] || @default_check
+    {:ok, pid} = Supervisor.start_link([], strategy: :one_for_one)
```
josevalim, line 25: You typically never want to start a supervisor inside another process (except another supervisor). This is because otherwise you have to reimplement the shutdown semantics. So idea here is to have GenSingleton be a supervisor that starts the GenServer. And then the GenServer can start the `child_spec` as a sibling in the supervision tree.

The tricky here is for the GenServer to capture the supervisor by passing self on `start_link`:

```elixir
  def start_link(args) do
    # Self here is the supervisor PID
    GenServer.start_link(__MODULE__, {self(), args}, [])
  end
```

josevalim:
> I believe you don't need an explicit syn.lookup, you can just try starting the process as usual. Since you use name: {:via, :sync, ...}, it will fail with :already_started, which is effectively the lookup. 

Agreed. The explicit lookup can be helpful if you are starting children dynamically all the time, because it avoid spawning a process only for it to realize the name is taken, but in this case it is all static, so you are likely not gaining anything. :)

lib/logflare/gen_singleton.ex
Ziinc: ah very nice didn't think of that 👍

lib/logflare/backends/source_sup.ex
```diff
@@ -50,7 +51,7 @@ defmodule Logflare.Backends.SourceSup do
     children =
       [
         {RateCounterServer, [source: source]},
-        {RecentEventsTouch, [source: source]},
+        {GenSingleton, child_spec: {RecentEventsTouch, source: source}},
```
josevalim, line 54: I would consider having a separate Supervisor for  `GenSingleton` with `strategy: :rest_for_one`. This way, if the GenSingleton "watcher" process (the current GenServer) terminates, it would also terminate its child, and you keep their failures contained/related.

josevalim (commented):
One final optional suggestion and it looks good to go!

PR #2367, opened 2025-04-04
perf: reduce net tick time

Ziinc:
Reduces the time taken to mark nodes as disconnected (if a nodedown is not received)


josevalim (commented):
LGTM :)

PR #2369, opened 2025-04-07
perf: ingestion tweaking for bigquery

Ziinc:
- doubles processor and batcher counts for BigQuery pipeline, in a real world scenario it would likely be limited by batcher request speed, so there are diminishing returns in increasing processor counts. This will roughly 2.6x output per pipeline, so that we can burn down queues faster. see benchmark for details 
- increases http1 pools to increase overall max http throughput. Likely redundant once we move to Storage Write API
- tweak producer schedule scaling based on avg ingest rate, reducing it significantly for high rate sources. 
- Increase Ingest event queue to 15k, and reduce round robin chunking to 50. This should spread out chunks to more queues now.
- Reduce QueueJanitor purge ratio to 0.05, to account for larger ingest queue permitted.

BencheeAsync benchmark of Pipeline. only sample size is relevant
```
Name                           minimum        maximum    sample size                     mode
:"test benchmarks 8-16"       0.0410 μs     7228.54 μs       864.00 K                  1.75 μs
:"test benchmarks 12-16"       0.0830 μs     5178.04 μs       935.25 K                  1.38 μs
:"test benchmarks 8-8"       0.0840 μs    11146.08 μs       401.00 K                  1.75 μs
:"test benchmarks 4-8"        0.125 μs     9210.75 μs       326.50 K                  1.50 μs
```

### DIFF
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -213,7 +213,7 @@ defmodule Logflare.Application do
        pools: %{
          "https://bigquery.googleapis.com" => [
            protocols: [:http1],
-           size: max(base * 100, 150),
+           size: max(base * 125, 150),
            count: http1_count,
            start_pool_metrics?: true
          ]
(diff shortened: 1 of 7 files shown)


chasers (approved):

josevalim (commented):
LGTM :) Increasing the batcher can be problematic if the API you are pushing to has rate limits but I suspect BigQuery will be more than fine :)

PR #2371, opened 2025-04-09
feat: threshold-based DynamicPipeline resolving

Ziinc:
This adjusts the dynamic Broadway pipeline creation logic to be threshold based -- if any queue hits a certain threshold, it would add an additional scaled number of pipelines.

It also adjusts the gradual decrease to make it much harder for it to scale down. Scale to zero condition is unchanged.

Threshold chosen is 500 items, which seems to be the eyeballed number in prod that queues are hovering at for busy sources. Most spike to 1-2k even for very busy sources. 

This should help avoid load shedding by the QueueJanitor, as load shedding is still getting triggered due to request spikes on only one pipeline, as previous resolve logic would only create more pipelines when pipeline is at least half of max (7.5k) at point of resolving. What is likely happening is a race condition where:
- the Queue Janitor triggers load sheding
- the BufferProducer takes majority of remainder of the queue
- the queue drops to below 50%
- the resolve triggers but doesn't increase pipeline count

On staging, seems like pipelines only get spun up at saturation at over >80% cpu util, and only 4 get spun up. So definitely a red flag.

Some other changes include:
- adding small randomizer to the pipeline resolve interval
- adding some debugging utils for runtime debugging

### DIFF
--- a/lib/logflare/utils/debugging.ex
+++ b/lib/logflare/utils/debugging.ex
@@ -3,22 +3,20 @@ defmodule Logflare.Utils.Debugging do
   alias Logflare.Backends.IngestEventQueue
 
   def list_counts(source_id) do
-    :erpc.multicall(
-      [Node.self() | Node.list()],
-      fn ->
-        {Node.self(), IngestEventQueue.list_counts({source_id, nil})}
-      end,
-      5000
-    )
+    :erpc.multicall(all_nodes(), __MODULE__, :list_counts_callback, [source_id], 5000)
+  end
+
+  def list_counts_callback(source_id) do
+    {Node.self(), IngestEventQueue.list_counts({source_id, nil})}
   end
 
   def list_pending_counts(source_id) do
-    :erpc.multicall(
-      [Node.self() | Node.list()],
-      fn ->
-        {Node.self(), IngestEventQueue.list_pending_counts({source_id, nil})}
-      end,
-      5000
-    )
+    :erpc.multicall(all_nodes(), __MODULE__, :list_pending_counts_callback, [source_id], 5000)
+  end
+
+  defp all_nodes, do: [Node.self() | Node.list()]
+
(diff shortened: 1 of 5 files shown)


chasers (approved):

lib/logflare/utils/debugging.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Logflare.Utils.Debugging do
+  @moduledoc false
+  alias Logflare.Backends.IngestEventQueue
+
+  def list_counts(source_id) do
+    :erpc.multicall(
+      [Node.self() | Node.list()],
+      fn ->
+        {Node.self(), IngestEventQueue.list_counts({source_id, nil})}
+      end,
```
josevalim, line 10: How do you perform deployments? Do you keep the old cluster and new cluster together for a while to migrate that or fully separate?

The reason I ask is because anonymous functions do not work if could have mixed versions, then in this case, you should do something like this instead:

```elixir
  def list_counts(source_id) do
    :erpc.multicall(
      [Node.self() | Node.list()], __MODULE__, list_counts_callback, [source_id]
    )
  end

  def list_counts_callback(source_id) do
    {Node.self(), IngestEventQueue.list_counts({source_id, nil})}
  end
```

I also understand this is debug only but perhaps you want to debug counts during deployments and if they are mixed clusters, then you wouldn't be able to :)

josevalim (commented):
Hi @Ziinc! I dropped one comment that may or may not apply to you, otherwise it looks great to me!

lib/logflare/utils/debugging.ex
Ziinc: yea right now there is a chance for that to happen with rolling deployments.
we do a mix of rolling  and green-blue depending on risk tolerance for that specific cluster.
Adjusted, thanks! 💟

PR #2372, opened 2025-04-09
feat: add labels for BQ jobs for billing association

Ziinc:
For billing. Uses id instead of uuid for consistency with user id labelling.

### DIFF
--- a/lib/logflare/ecto/bigquery/bq_repo.ex
+++ b/lib/logflare/ecto/bigquery/bq_repo.ex
@@ -31,7 +31,7 @@ defmodule Logflare.BqRepo do
   def query_with_sql_and_params(%User{} = user, project_id, sql, params, opts \\ [])
       when not is_nil(project_id) and is_binary(sql) and is_list(params) and is_list(opts) do
     override = Map.new(opts)
-
+    {labels, override} = Map.pop(override, :labels, %{})
     %Plan{name: plan} = Billing.Cache.get_plan_by_user(user)
 
     query_request = %QueryRequest{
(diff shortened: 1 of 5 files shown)


lib/logflare/ecto/bigquery/bq_repo.ex
```diff
@@ -42,11 +42,15 @@ defmodule Logflare.BqRepo do
...
+          %{
+            "managed_by" => "logflare",
+            "logflare_plan" => GenUtils.format_key(plan),
+            "logflare_account" => GenUtils.format_key(user.id)
+          },
+          labels
+        )
     }
 
     query_request = Map.merge(query_request, override)
```
jonatanklosko, line 56: `:labels` is still treated as overrides, the only difference is that it is deep-merged. One way is to handle it explicitly with `Map.pop/3` as you already do. An alternative you can consider is using [`Map.merge/3`](https://hexdocs.pm/elixir/Map.html#merge/3) with a custom merging function, in this case:

```elixir
query_request =
  Map.merge(query_request, override, fn
    _key, left, right -> right
    :labels, left, right -> Map.merge(left, right)
  end)
```

If you always want deep-merging behaviour, you can go even further and define a helper function:

```elixir
query_request = map_deep_merge(query_request, override)

# ...

defp map_deep_merge(%{} = left, %{} = right) do
  Map.merge(left, right, fn _key, left, right -> map_deep_merge(left, right) end)
end

defp map_deep_merge(_left, right), do: right
```

jonatanklosko (commented):
Hey @Ziinc! Looks good to me, just one suggestion regarding the overrides.

lib/logflare/ecto/bigquery/bq_repo.ex
Ziinc: opted to use DeepMerge as we're already using the lib in the codebase.

jonatanklosko:
Looks good to me!

PR #2373, opened 2025-04-14
perf: set busy wait to long from none

Ziinc:
This PR increases busy-wait for the scheduler. When we initially reduced it to `none`, it was with the understanding that we wanted to reduce system cpu % to cater for scaling down our instance sizes. 
Since we want to now saturate CPU usage fully on our now cheaper instances, this should be optimal since the workloads are constant and we want to maximize throughput. It is unlikely that we will need to reduce instance size more, unless we really can't saturate the instance even at 32 cores.

### DIFF
--- a/rel/vm.args.eex
+++ b/rel/vm.args.eex
@@ -15,9 +15,10 @@
 -env ERL_MAX_ETS_TABLES 250000
 
 ## Set busy-wait. Options are: none|very_short|short|medium|long|very_long
-+sbwt none
-+sbwtdcpu none
-+sbwtdio none
+## Longer is better to maximize CPU saturation
++sbwt long
++sbwtdcpu long
++sbwtdio long
 
 ## load balance on scheduler utilization
 +sub true


Ziinc:
Closes https://linear.app/supabase/issue/ANL-859/logflare-reduce-busy-wait

jonatanklosko:
Hey @Ziinc! What is the deployment situation, does the app have the whole machine to itself? Is it inside Docker?

An important concern with busy-wait is that the app CPU usage is inflated, which is problematic if the resources are shared with any other process. It has been discussed [here](https://elixirforum.com/t/performance-of-erlang-elixir-in-docker-kubernetes/21493/13). Also, [this benchmark](https://stressgrid.com/blog/beam_cpu_usage/#conclusion) indicated that busy-wait may have no tangible improvement in terms of application performance, but it may very well depend on application specifics.

Ziinc:
We have the whole machine to the docker image, with no cpu limts.
I was previously following [this comment](https://erlangforums.com/t/vm-tuning-guide/1945/4) in my previous [PR](https://github.com/Logflare/logflare/pull/1936), but it turns out that there are no cpu limits set by docker/VM.  The hypothesis that CFS was interacting with the scheduler busy-wait and limiting max utilization, which seems false.
I had also looked at the stressgrid benchmark before, but i now suspect that network i/o plays a large factor in their http benchmark. Furthermore, as they didn't open-source their benchmark code, it is quite hard to tell how much actual blocking work  the scheduler would be getting, especially because they `sleep` the process and are also simulating the db latency, which i assume is also a `sleep` call.  
Also, intuitively, busy-wait would have more of an impact when there is real-world non-uniform workloads, as i would rather that the schedulers are constantly awake and responsive vs having the small wakeup tax, however minute.

In a uniform saturated workload scenario, i would assume that busy-waiting would have a negligible effect since they would always have work in their run queue.

And after more thought i'll also set `+sub false` to switch back to load compaction and fully load all schedulers at all times, as i would rather that the scheduler doesn't run out of work and steal less work from another run queue.

So the thought process is:
1. add back busy-wait to cater for uneven request load and improve responsiveness, and to fully use all cpu available, and avoid schedulers from sleeping if their run queue runs out
2. remove scheduler utilization load balancing and switch back to load compaction, so that we fully load all run queues and in the event that they do run out of work, busy wait kicks in, and they can optionally steal work from another run queue. From my understanding, this should not be expensive as it is now [lock-free](https://www.erlang.org/doc/apps/erts/processmanagementoptimizations.html)

jonatanklosko:
> In a uniform saturated workload scenario, i would assume that busy-waiting would have a negligible effect since they would always have work in their run queue.

I think so! If the schedulers are busy for real, the busy-wait setting should not matter at all. But if that's the scenario you are expecting, it may also be an argument to keep busy-wait low (in the sense that, you don't expect low saturation, but if it happens, it's more evident through CPU utilisation).

That said, you are aware of the nuances, so go with your intuition. In cases like this, ideally you would have a specific metrics that you can observe over time to know if the change made positive/negative/any impact for your target.

PR #2376, opened 2025-04-16
fix: routing bug when checking if source sup is started

Ziinc:
bugfix for source routing for v2 pipeline


josevalim (commented):
Looks good to me too :)

PR #2377, opened 2025-04-16
fix: foreign key constraint on user delete

Ziinc:
Bugfix for cascading user deletion


lib/logflare/user.ex
```diff
@@ -90,8 +90,8 @@ defmodule Logflare.User do
 
     has_many :billing_counts, Logflare.Billing.BillingCount, on_delete: :delete_all
     has_many :sources, Source
-    has_many :endpoint_queries, Logflare.Endpoints.Query
-    has_many(:alert_queries, AlertQuery)
+    has_many :endpoint_queries, Logflare.Endpoints.Query, on_delete: :delete_all
+    has_many(:alert_queries, AlertQuery, on_delete: :delete_all)
```
jonatanklosko, line 94: To quote [`has_many/3`](https://hexdocs.pm/ecto/Ecto.Schema.html#has_many/3) docs:

> * `:on_delete` - The action taken on associations when parent record is deleted. May be `:nothing` (default), `:nilify_all` and `:delete_all`. **Using this option is DISCOURAGED for most relational databases. Instead, in your migration, set `references(:parent_id, on_delete: :delete_all)`.** Opposite to the migration option, this option cannot guarantee integrity and it is only triggered for `c:Ecto.Repo.delete/2` (and not on `c:Ecto.Repo.delete_all/2`) and it never cascades. If posts has many comments, which has many tags, and you delete a post, only comments will be deleted. If your database does not support references, cascading can be manually implemented by using `Ecto.Multi` or `Ecto.Changeset.prepare_changes/2`

What you most likely want is to change the `:user_id` column to `references(:users, on_delete: :delete_all)`, for both "endpoint_queries" and "alert_queries". This way deleting a user row will cascade-delete entries in those tables on the database level.

Is there a reason you wouldn't do that?

Ziinc: refactored 😄

lib/logflare/user.ex
jonatanklosko, line 1: The extra parentheses are probably not intended? There is no actual change, so can just remove the file form the diff altogether :)

jonatanklosko:
Looks good to me!

PR #2378, opened 2025-04-16
feat: move to LibclusterPostgres.Strategy, remove legacy strategies

Ziinc:
Removes old gce strategy, and initial pg strategy prototype.


josevalim:
Looks good to me! I assume you can't do DNS based discovery in GCE?

Ziinc:
> I assume you can't do DNS based discovery in GCE?
we didn't really explore further beyond using the gce api, as we didn't want the clustering strategy to be too tied to gcp...just in case 😄

PR #2379, opened 2025-04-16
fix: handle multiple union all

Ziinc:
Fixes https://linear.app/supabase/issue/ANL-863/logflare-endpoints-fix-union-all-for-2-expressions

Handles multiple union all in sql statements.
Moves tree traversal code from anon function to separate function definition for recursion. I should probably extract this out to a macro soon...


Ziinc:
merging this in first to prep for release.

PR #2380, opened 2025-04-16
perf: revert busy wait

Ziinc:
Reverting busy-wait, seems like this will cause an issue with load balancer as gcp load balancer calculates max loading based cpu utilization, causing load to be taken on and off nodes in a non-consistent way, since scheduler utilization (and thusly cpu utilization) jumps a lot when large amounts of load is taken on and off (despite gcp load balancer already having high utilization threshold).
I suspect that this also results in load balancer level pushback on requests resulting in higher latency on our load test cluster.
Screenshot 2025-04-16 at 7 45 25 PM

### DIFF
--- a/rel/vm.args.eex
+++ b/rel/vm.args.eex
@@ -15,10 +15,9 @@
 -env ERL_MAX_ETS_TABLES 250000
 
 ## Set busy-wait. Options are: none|very_short|short|medium|long|very_long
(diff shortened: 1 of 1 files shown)


jonatanklosko:
@Ziinc makes sense, anything that relies on the CPU usage is going to be tricked by the artificial inflation, that was one of the concerns. Good catch!

PR #2387, opened 2025-04-23
feat: disable tailing source

Ziinc:
Two main improvements:
- source-level tailing disable option, to prevent tailing for large sources
- source-level required keys, which prevents users from forcing a query when it is required.

minor perf fix:
- moves the query kickoff from handle_params, which results in duplicate queries being triggered on initial page load.

Also adds in source names to the page title as a minor usability improvement.

https://github.com/user-attachments/assets/81dba8a6-242c-4c40-8d7e-b01ea17cd22a

Screenshot 2025-04-23 at 6 51 49 PM


Ziinc:
tasks:
- https://linear.app/supabase/issue/ANL-205/include-logflare-source-name-in-page-title
- https://linear.app/supabase/issue/ANL-869/logflare-search-addressing-expensive-queries
- https://linear.app/supabase/issue/ANL-870/add-source-option-to-disable-logflare-search-auto-tailing
- https://linear.app/supabase/issue/ANL-877/prevent-users-from-forcing-queries-when-suggested-keys-are-needed

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -837,17 +897,31 @@ defmodule LogflareWeb.Source.SearchLV do
...
+    suggested_present =
+      Enum.all?(suggested, fn suggested_field ->
         Enum.find(lql_rules, fn %{path: path} -> path == suggested_field end)
       end)
 
-    if suggested_present, do: {:ok, socket}, else: {:error, :suggested_field_not_found}
+    required_present =
+      Enum.all?(required, fn required_field ->
+        Enum.find(lql_rules, fn %{path: path} ->
+          path == String.trim_trailing(required_field, "!")
```
josevalim, line 917: Trim once instead of per rule:

```suggestion
        required_field = String.trim_trailing(required_field, "!")

        Enum.find(lql_rules, fn %{path: path} ->
          path == required_field
```

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -139,10 +142,13 @@ defmodule LogflareWeb.Source.SearchLV do
           |> assign(:search_op_log_events, search_op_log_events)
           |> assign(chart_aggregate_enabled?: search_agg_controls_enabled?(lql_rules))
 
-        kickoff_queries(source.token, socket.assigns)
+        Process.send_after(self(), :kickoff_queries, 500)
```
josevalim, line 145: Instead of `send_after`, you can use `if connected?(socket) do` and only start the queries if connected. :)

lib/logflare_web/templates/source/edit.html.eex
```diff
@@ -284,11 +284,25 @@
...
     <%= submit "Save", class: "btn btn-primary form-button" %>
   <% end %>
+
+  <%= section_header("Search Tailing") %>
+  <%= form_for @changeset, Routes.source_path(@conn, :update, @source),fn f -> %>
+    <div class="form-group">
+      <%= label f, :disable_tailing do %>
+        Toggle Source search tailing for this source. Currently <%= if(@source.disable_tailing, do: "disabled", else: "enabled") %>.
+      <% end %>
+      <%= hidden_input f, :disable_tailing, value: if(@source.disable_tailing, do: "false", else: "true") %>
```
josevalim, line 299: This may work, as booleans would be encoded to string (but not that important anyway):

```suggestion
      <%= hidden_input f, :disable_tailing, value: !@source.disable_tailing %>
```

test/logflare_web/live/search_live/logs_search_lv_test.exs
```diff
@@ -621,6 +626,106 @@ defmodule LogflareWeb.Source.SearchLVTest do
...
+      expect(GoogleApi.BigQuery.V2.Api.Jobs, :bigquery_jobs_query, 4, fn _conn, _proj_id, opts ->
+        {:ok, TestUtils.gen_bq_response()}
+      end)
+
+      {:ok, view, _html} =
+        live(conn, ~p"/sources/#{source.id}/search?querystring=something123&tailing%3F=true")
+
+      refute get_view_assigns(view).tailing?
+
+      :timer.sleep(400)
```
josevalim, line 711: It would be nice to not rely on timers. I believe the BigQuery API may be async but you could perhaps wrap the API in a version that immediately returns the results by sending a message to `self()`. This means it is still ultimately async but during tests the mocked responses are always sent immediately and you don't need timers. Anyway, some food for thought! :)

josevalim (commented):
Hi @Ziinc! Looking good, I have dropped some comments :)

chasers (approved):

test/logflare_web/live/search_live/logs_search_lv_test.exs
Ziinc: Ah this is due to the message passing between the centralized querying in the legacy architecture. The BQ api is already mocked entirely using mimic, so this is a stopgap measure for reducing flakiness until we refactor the searching logic (soon-ish).

PR #2388, opened 2025-04-23
fix: SQL query gets truncated by LiveMonacoEditor and query params

Ziinc:
This fixes a bug where queries were getting  truncated before getting run, due to a bug (i think) with LiveMonacoEditor and also the omnibox on handle_params.

### DIFF
--- a/lib/logflare_web/live/query_live.ex
+++ b/lib/logflare_web/live/query_live.ex
@@ -47,9 +47,9 @@ defmodule LogflareWeb.QueryLive do
                 "language" => "sql",
                 "fontSize" => 12,
                 "padding" => %{
-                  "top" => 14
+                  "top" => 14,
+                  "bottom" => 14
                 },
-                "fixedOverflowWidgets" => false,
                 "contextmenu" => false,
                 "hideCursorInOverviewRuler" => true,
                 "smoothScrolling" => true,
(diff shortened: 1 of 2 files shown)


lib/logflare_web/live/query_live.ex
```diff
@@ -163,37 +161,41 @@ defmodule LogflareWeb.QueryLive do
   end
 
   def handle_params(params, _uri, socket) do
-    query_string = params["q"] || socket.assigns.query_string
+    socket =
+      if !socket.assigns.query_string,
+        do: assign(socket, :query_string, params["q"]),
```
josevalim, line 166: There may be a potential bug lurking here. If `params["q"]` is nil, then the `socket.assigns.query_string` will be `nil`. This means `!= ""` below will be true, and you will send a message to parse it. Parsing calls `Endpoints.parse_query_string(query_string)` which, at least according to its typespec, expects a string.

It may make sense to normalize the query string here to be either always an empty string or always nil. Perhaps the second is more idiomatic?

Ziinc: 👍  adjusted to normalize in the handle_params/2, which is much cleaner as the query string determination is now localised instead of spread out in the both mount/3 and handle_params/2

PR #2391, opened 2025-05-06
feat: logflare endpoints support large sql queries

Ziinc:
This PR adds in GET body read support as fallback for very large sql queries. It also increases the max request line length for http1 requests by 10x from the default.

### DIFF
--- a/lib/logflare_web/controllers/endpoints_controller.ex
+++ b/lib/logflare_web/controllers/endpoints_controller.ex
@@ -9,6 +9,12 @@ defmodule LogflareWeb.EndpointsController do
   alias LogflareWeb.OpenApi.ServerError
   alias LogflareWeb.OpenApiSchemas.EndpointQuery
 
+  @plug_parsers_init Plug.Parsers.init(
+                       parsers: [JsonParser],
+                       json_decoder: Jason,
+                       body_reader: {PlugCaisson, :read_body, []}
+                     )
+
   action_fallback(LogflareWeb.Api.FallbackController)
   tags(["Public"])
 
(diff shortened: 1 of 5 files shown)


lib/logflare_web/controllers/endpoints_controller.ex
```diff
@@ -55,4 +58,28 @@ defmodule LogflareWeb.EndpointsController do
...
+
+  # only parse body for get when ?sql= is empty and it is sandboxable
+  # passthrough for all other cases
+  defp parse_get_body(
+         %{assigns: %{endpoint: %_{sandboxable: true}}, query_params: qp} = conn,
+         _opts
+       )
+       when is_map_key(qp, "sql") == false do
+    args =
+      Plug.Parsers.init(
```
wojtekmach, line 70: just for completeness, if we did:

```
plug Plug.Parsers, ...
```

this `Plug.Parsers.init/1` would have been called at compile-time. Looking at the implementation, this all looks cheap:

-  https://github.com/elixir-plug/plug/blob/v1.17.0/lib/plug/parsers.ex#L253
- https://github.com/Logflare/logflare/blob/734ffe6b9a6a79f0ca138af9c3b5b4f190cf40d1/lib/logflare_web/controllers/plugs/json_parser.ex#L11
- https://github.com/elixir-plug/plug/blob/v1.17.0/lib/plug/parsers/json.ex#L32

however if this ever shows up in any performance work, a trivial change is to move it to compile-time:

```elixir
@plug_parsers_opts Plug.Parsers.init(
                     parsers: [JsonParser],
                     json_decoder: Jason,
                     body_reader: {PlugCaisson, :read_body, []}
                   )

defp parse_get_body(...) do
  # ...
  |> Plug.Parsers.call(@plug_parsers_opts)
  # ...
end
```

wojtekmach (commented):
Looks good to me!

lib/logflare_web/controllers/endpoints_controller.ex
Ziinc: shifted it to compile time :D i'll take any perf gain i can get haha

PR #2393, opened 2025-05-08
feat: add testing ui for checking verifying query expanding for nested fields

Ziinc:
This PR fixes Logflare UI to correctly handle nested parameters as well as allowing easy testing of nested parameters within the UI.
Screenshot 2025-05-08 at 3 47 09 PM

### DIFF
--- a/lib/logflare/endpoints.ex
+++ b/lib/logflare/endpoints.ex
@@ -11,6 +11,7 @@ defmodule Logflare.Endpoints do
   alias Logflare.Backends.Adaptor.PostgresAdaptor
   alias Logflare.SingleTenant
   alias Logflare.Alerting
+  alias Logflare.Alerting.Alert
   alias Logflare.Backends
 
   import Ecto.Query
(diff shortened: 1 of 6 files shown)


lib/logflare_web/live/query_live.ex
```diff
@@ -142,13 +143,18 @@ defmodule LogflareWeb.QueryLive do
   def mount(%{}, %{"user_id" => user_id}, socket) do
     user = Users.get(user_id)
 
+    endpoints = Endpoints.list_endpoints_by(user_id: user.id)
+    alerts = Alerting.list_alert_queries_by_user_id(user.id)
```
jonatanklosko, line 147: Currently these variables are not used, I expect they should be put in assigns below.

lib/logflare/endpoints.ex
```diff
@@ -126,10 +127,18 @@ defmodule Logflare.Endpoints do
     iex> parse_query_string("select @testing from date")
     {:ok, %{parameters: ["testing"]}}
   """
-  @spec parse_query_string(String.t()) :: {:ok, %{parameters: [String.t()]}} | {:error, any()}
-  def parse_query_string(query_string) do
-    with {:ok, declared_params} <- Logflare.Sql.parameters(query_string) do
-      {:ok, %{parameters: declared_params}}
+  @spec parse_query_string(:bq_sql | :pg_sql, String.t(), [Query.t()], [Alert.t()]) ::
+          {:ok, %{parameters: [String.t()], expanded_query: String.t()}} | {:error, any()}
+  def parse_query_string(language, query_string, endpoints \\ [], alerts \\ [])
```
jonatanklosko, line 132: Instead of multiple default arguments, you can also consider a single list of options:

```suggestion
  def parse_query_string(language, query_string, opts \\ [])
    opts = Keyword.validate!(opts, [endpoints: [], alerts: []])
```

Or, if you expect both to be generally passed, you can skip the default `\\ []` values, so the call is always explicit.

It mostly depends on how you expect to be called most of the time, and if it's likely to have more options in the future.

jonatanklosko (commented):
Two small comments :)

lib/logflare/endpoints.ex
Ziinc: Removed the default values as it should always be explicit in the foreseeable future

Ziinc: thanks for the recommendation for the Keyword.validate!/2, it'll be definitely something to adopt once this function becomes more and more complex

jonatanklosko:
Looks good to me!

PR #2395, opened 2025-05-12
feat: add support for arrays for otel traces

Ziinc:
This PR adds support for otel array attributes, as needed by storage team.

Screenshot 2025-05-12 at 6 58 36 PM

### DIFF
--- a/lib/logflare/logs/otel_trace.ex
+++ b/lib/logflare/logs/otel_trace.ex
@@ -98,18 +98,16 @@ defmodule Logflare.Logs.OtelTrace do
 
   defp extract_key_value(%KeyValue{
          key: key,
-         value: %AnyValue{value: {:array_value, %ArrayValue{values: values}}}
+         value: value
        }) do
-    {key, Enum.map(values, &extract_key_value/1)}
+    {key, extract_value(value)}
   end
 
-  defp extract_key_value(%KeyValue{key: key, value: %{value: {_, value}}}) do
-    {key, value}
+  defp extract_value(%AnyValue{value: {:array_value, %ArrayValue{values: values}}}) do
+    Enum.map(values, &extract_value/1)
   end
 
-  defp extract_key_value(%AnyValue{value: {_type, value}}) do
-    value
-  end
-
-  defp extract_key_value(v), do: v
+  defp extract_value(%_{value: {_type, value}}), do: value
+  defp extract_value(nil), do: nil
+  defp extract_value(value), do: value
 end
--- a/lib/logflare/logs/otel_trace.ex
+++ b/lib/logflare/logs/otel_trace.ex
@@ -14,7 +14,8 @@ defmodule Logflare.Logs.OtelTrace do
 
   alias Opentelemetry.Proto.Trace.V1.ResourceSpans
   alias Opentelemetry.Proto.Common.V1.KeyValue
-
+  alias Opentelemetry.Proto.Common.V1.AnyValue
+  alias Opentelemetry.Proto.Common.V1.ArrayValue
   @behaviour Logflare.Logs.Processor
 
   def handle_batch(resource_spans, _source) when is_list(resource_spans) do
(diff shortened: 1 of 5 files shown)


test/support/test_utils_grpc.ex
```diff
@@ -75,8 +77,90 @@ defmodule Logflare.TestUtilsGrpc do
...
+      array_of_strings,
+      array_of_numbers,
+      array_of_booleans
+    ]
+  end
+
+  # this is not supported by traces/metrics, only logs
+  # https://github.com/open-telemetry/opentelemetry-specification/pull/2888
+  # https://github.com/open-telemetry/opentelemetry-specification/issues/376
+  defp random_key_value(:array_of_kv) do
```
jonatanklosko, line 98: This is not used anywhere. Perhaps expected, I am just noting to double-check :)

test/logflare_grpc/trace/server_test.exs
```diff
@@ -5,6 +5,11 @@ defmodule LogflareGrpc.Trace.ServerTest do
   alias Opentelemetry.Proto.Collector.Trace.V1.ExportTraceServiceResponse
   alias Opentelemetry.Proto.Collector.Trace.V1.TraceService.Stub
   alias Logflare.SystemMetrics.AllLogsLogged
+  require OpenTelemetry.Tracer
+
+  require Record
+  @fields Record.extract(:span_ctx, from_lib: "opentelemetry_api/include/opentelemetry.hrl")
+  Record.defrecordp(:span_ctx, @fields)
```
jonatanklosko, line 12: This doesn't seem to be used anywhere?

test/logflare/logs/otel_trace_test.exs
```diff
@@ -9,6 +9,25 @@ defmodule Logflare.Logs.OtelTraceTest do
...
+      assert Enum.any?(Map.values(a), &is_binary/1)
+      assert Enum.any?(Map.values(a), &is_boolean/1)
+    end
+
+    test "array handling", %{
+      resource_spans: resource_spans,
+      source: source
+    } do
+      batch = OtelTrace.handle_batch(resource_spans, source)
+    end
```
jonatanklosko, line 29: This doesn't assert anything and `handle_batch` is already called above with the same arguments. I would expect this test to be either expanded or removed :)

lib/logflare/logs/otel_trace.ex
```diff
@@ -95,7 +96,20 @@ defmodule Logflare.Logs.OtelTrace do
 
   defp extract_key_value(%KeyValue{key: key, value: nil}), do: {key, nil}
 
+  defp extract_key_value(%KeyValue{
+         key: key,
+         value: %AnyValue{value: {:array_value, %ArrayValue{values: values}}}
+       }) do
+    {key, Enum.map(values, &extract_key_value/1)}
```
jonatanklosko, line 103: The `Enum.map` here actually extracts just values. Perhaps it's worth separating a `extract_value` function, like this:

```elixir
defp extract_key_value(%KeyValue{key: key, value: value}) do
  {key, extract_value(value)}
end

defp extract_value(%AnyValue{value: {:array_value, %ArrayValue{values: values}}}) do
  Enum.map(values, &extract_value/1)
end

defp extract_value(%AnyValue{value: {_type, value}}), do: value
defp extract_value(nil), do: nil
defp extract_value(value), do: value
```

I added the last clause, because currently you have `defp extract_key_value(v), do: v`, though I am not sure if it's actually necessary, is it ever something other than `%AnyValue{}`?

test/support/test_utils_grpc.ex
Ziinc: yea i intend to use it again when we support logs 😄

test/logflare_grpc/trace/server_test.exs
Ziinc: ah remnants from testing. my bad, removed.

test/logflare/logs/otel_trace_test.exs
Ziinc: removed as it array is already covered in above test.

lib/logflare/logs/otel_trace.ex
Ziinc: checked the otel spec and seems like AnyValue only gets used for arbitrary data values (this situation), but it is a bit vague if this is always used in the protobuf payload. 
I'll keep the last clause as fallback in case there is an unknown edge case, should be clearer if it is needed once we have better tooling around the protobuf struct generation for tests, in current state it is very much observational as opposed to using the otel sdk internals directly to generate the pb struct.

PR #2396, opened 2025-05-12
Initial ClickHouse Log Drain Functionality

amokan:
Super simple/naive approach to doing some basic smoke testing on sinking logs into Clickhouse by fronting the existing webhook adaptor.

Includes just enough UI to configure a Clickhouse backend and some simple tests based off the Loki backend test.

Long-term, we'll want to rework this to be its own first-class adaptor that leverages something like [ch](https://hex.pm/packages/ch) under the hood.

I did have to introduce a _workaround_ for dynamically providing a URL to the webhook adaptor given that configs are merged within that adaptor at runtime - paving over the URL manipulation performed to inject the table name into the querystring. This will go away at a later time when we roll out a more robust implementation.

Worth noting that to use this backend, it does expect the table to exist in CH. See the module doc in the clickhouse adaptor for an example `CREATE TABLE` statement.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
@@ -10,15 +10,14 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
 
   ```sql
   CREATE TABLE "default"."supabase_log_ingress" (
-      "id" UUID,
-      "event_message" String,
-      "metadata" String,
-      "service" String,
-      "timestamp" DateTime64(6)
-  )
-  ENGINE MergeTree()
-  ORDER BY ("timestamp")
-  SETTINGS index_granularity = 8192 SETTINGS flatten_nested=0
+        "id" UUID,
+        "event_message" String,
+        "body" String,
+        "timestamp" DateTime64(6)
+    )
+    ENGINE MergeTree()
+    ORDER BY ("timestamp")
+    SETTINGS index_granularity = 8192 SETTINGS flatten_nested=0
   ```
   """
 
@@ -29,10 +28,11 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
 
   typedstruct enforce: true do
     field(:url, String.t())
+    field(:url_override, String.t(), default: nil)
     field(:username, String.t())
     field(:password, String.t())
     field(:database, String.t(), default: "default")
-    field(:table, String.t())
+    field(:table, String.t(), default: "supabase_log_ingress")
     field(:port, non_neg_integer(), default: 8443)
   end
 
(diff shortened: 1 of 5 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -0,0 +1,163 @@
...
+
+  Currently this adaptor expects a table to be created with the following schema.
+
+  Adjust the database / table names as needed.
+
+  ```sql
+  CREATE TABLE "default"."supabase_log_ingress" (
+        "id" UUID,
+        "event_message" String,
+        "body" String,
```
chasers, line 15: Lets change this to `metadata` and just go ahead and make it a json field.

This would let us play with how efficient json querying is while we load test the ingest side of things.

chasers (changes requested):
See above

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
chasers: Hmm, perhaps nevermind:

```
However, if you are using ClickHouse Cloud, you must first get in touch with support to enable the usage of the JSON type.
```

chasers: Yeah lets just keep it as is and start writing some datas

chasers (approved):

test/logflare/backends/clickhouse_adaptor_test.exs
```diff
@@ -0,0 +1,102 @@
...
+      [backend: backend, source: source]
+    end
+
+    test "payload is a properly formatted log event", %{source: source} do
+      this = self()
+      ref = make_ref()
+
+      @client
+      |> expect(:send, fn req ->
+        send(this, {ref, req[:body]})
```
Ziinc, line 85: could you add a url assertion to check if the request is correctly sent with the dynamically built url

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -165,7 +165,8 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
         end
 
       Client.send(
-        url: config.url,
+        # if a `url_override` key is available in the merged config, use that before falling back to `url`
```
Ziinc, line 168: could you add docs for this behaviour, even if temporary

Ziinc: The stored_config should probably get transformed by the respective backend type instead of getting inserted wholesale, that might allow us to avoid this url_overriding as the runtime config used can differ from the stored config.

Ziinc (approved):

lib/logflare/backends/adaptor/webhook_adaptor.ex
amokan: > The stored_config should probably get transformed by the respective backend type instead of getting inserted wholesale, that might allow us to avoid this url_overriding as the runtime config used can differ from the stored config.

Let's see where the ClickHouse stuff goes in the short-term. I do agree that the bandaid I put in place could be improved though.

PR #2399, opened 2025-05-15
Connection pool workaround for the sake of Clickhouse cloud

amokan:
The current implementation of the Clickhouse log drain relies on the existing `WebhookAdaptor` and the underlying Finch-based connection pooling.

Due to the way Clickhouse cloud URLs are with dynamic subdomain parts - this makes leveraging traditional Finch pooling a bit tricky without doing some sort of runtime management of pools since the pools are keyed by a `{<scheme>, <host>, <port>}` tuple.

Rather than deal with that extra complexity, this change simply defines a new pool dedicated to Clickhouse for now that is specified for use by leveraging a new config opts in the `WebhookAdaptor`.

Again, this should be temporary to push a bit harder on clickhouse logs draining until a deeper implementation with [`:ch`](https://hex.pm/packages/ch) is finished.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
@@ -85,7 +85,7 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
     %{
       url: updated_url,
       url_override: updated_url,
-      pool_override: Logflare.FinchIngest,
+      pool_name: Logflare.FinchIngest,
       database: config.database,
       table: config.table,
       port: config.port,
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
@@ -85,11 +85,12 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
     %{
       url: updated_url,
       url_override: updated_url,
+      pool_override: Logflare.FinchClickhouse,
       database: config.database,
       table: config.table,
       port: config.port,
       headers: headers,
-      http: "http2",
+      http: "http1",
       gzip: true
     }
   end
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
@@ -85,7 +85,7 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
     %{
       url: updated_url,
       url_override: updated_url,
-      pool_override: Logflare.FinchClickhouse,
+      pool_override: Logflare.FinchIngest,
       database: config.database,
       table: config.table,
       port: config.port,
--- a/lib/logflare/backends/adaptor/webhook_adaptor.ex
+++ b/lib/logflare/backends/adaptor/webhook_adaptor.ex
@@ -59,12 +59,25 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
     @moduledoc false
     use Tesla, docs: false
 
+    defguardp is_possible_pool(value)
+              when not is_nil(value) and not is_boolean(value) and is_atom(value)
+
     def send(opts) do
+      http_opt = Keyword.get(opts, :http)
+      pool_override_opt = Keyword.get(opts, :pool_override)
+
       adaptor =
-        if Keyword.get(opts, :http) == "http1" do
+        if http_opt == "http1" do
           {Tesla.Adapter.Finch, name: Logflare.FinchDefaultHttp1, receive_timeout: 5_000}
         else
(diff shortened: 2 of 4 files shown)


test/logflare/backends/clickhouse_adaptor_test.exs
```diff
@@ -82,8 +82,13 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptorTest do
...
       |> expect(:send, fn req ->
-        send(this, {ref, req[:body]})
-        %Tesla.Env{status: 200, body: ""}
+        expected_url =
+          "http://localhost:8443?query=INSERT%20INTO%20supabase_log_ingress%20FORMAT%20JSONEachRow"
+
+        if req[:url] == expected_url do
+          send(this, {ref, req[:body]})
+          %Tesla.Env{status: 200, body: ""}
+        end
```
Ziinc, line 91:
```suggestion
        assert req[:url] =~ ~r"localhost:8443.+query=INSERT.+EachRow"
        send(this, {ref, req[:body]})
        %Tesla.Env{status: 200, body: ""}
```

Ziinc: on a side note, since this only sets additional query params to the end url, we can add a query_params config map as part of the WebhookAdaptor config, which makes it more generalized and removes the need for overriding the entire url

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -59,14 +78,27 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
...
-        if Keyword.get(opts, :http) == "http1" do
+        if http_opt == "http1" do
           {Tesla.Adapter.Finch, name: Logflare.FinchDefaultHttp1, receive_timeout: 5_000}
         else
           {Tesla.Adapter.Finch, name: Logflare.FinchDefault, receive_timeout: 5_000}
         end
 
+      adaptor =
+        if is_possible_pool(pool_override_opt) do
+          {Tesla.Adapter.Finch, name: pool_override_opt, receive_timeout: 5_000}
```
Ziinc, line 97:
```suggestion
          {Tesla.Adapter.Finch, name: pool_name, receive_timeout: 5_000}
```

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -59,14 +78,27 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
     @moduledoc false
     use Tesla, docs: false
 
+    defguardp is_possible_pool(value)
+              when not is_nil(value) and not is_boolean(value) and is_atom(value)
+
     def send(opts) do
+      http_opt = Keyword.get(opts, :http)
+      pool_override_opt = Keyword.get(opts, :pool_override)
```
Ziinc, line 86:
```suggestion
      pool_name = Keyword.get(opts, :pool_name)
```

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -1,5 +1,24 @@
...
+  @moduledoc """
+  Backend adaptor for webhooks / HTTP posts.
+
+  A number of other adaptors (_ClickHouse, DataDog, Elastic, Loki, etc_) leverage this to handle the final HTTP transaction.
+
+  ### Finch Pool Selection
+
+  By default the pool will be selected automatically based on the `:http` configuration option.
+
+  If you want to manually select a specific Finch pool, you can use the `:pool_override` option and provide the module name.
```
Ziinc, line 11:
```suggestion
  If you want to manually select a specific Finch pool, you can use the `:pool_name` option and provide the module name.
```

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -85,11 +85,12 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
     %{
       url: updated_url,
       url_override: updated_url,
+      pool_override: Logflare.FinchIngest,
```
Ziinc, line 88:
```suggestion
      pool_name: Logflare.FinchIngest,
```

lib/logflare/backends/adaptor/webhook_adaptor.ex
Ziinc: using a cond/do would make the `adaptor` binding clearer.
is_possible_pool/1 is also unnecessary as we can use `is_module/1`

Ziinc (changes requested):

lib/logflare/backends/adaptor/webhook_adaptor.ex
amokan: Tell me more about `is_module/1`. I'm only familiar with the compiled/loaded variants in the Code namespace, which are not ideal candidates here.

chasers (approved):

PR #2402, opened 2025-05-21
[WIP] Clickhouse Backend Improvements

amokan:
Clickhouse implementation that uses [`ch`](https://hex.pm/packages/ch) rather than hand-crafted HTTP requests.

Still needs some polish and I plan to remove the `ClickhouseWebhookAdaptor` tomorrow. If you spot any glaring issues, let me know @Ziinc or @chasers 

## TODO

- [ ] `after_connect` hook for DbConnection so that credentials are validated ASAP
- [ ] Roll out ingest table to CH after validating connection if it does not exist already
- [ ] Dynamic ingest table naming rather than table config attribute?
- [ ] Unit tests


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -134,6 +84,47 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
...
+    state = %__MODULE__{
+      config: backend.config,
+      backend: backend,
+      backend_token: if(backend, do: backend.token, else: nil),
+      source_token: source.token,
+      source: source,
+      pipeline_name: Backends.via_source(source, Pipeline, backend.id)
+    }
+
+    {:ok, _pipeline_pid} = Pipeline.start_link(state)
```
Ziinc, line 124: instead of having a separate DynamicSupervisor to manage all CH connections, would it be feasible to start the connection here together with Broadway here? 
Name registry can then use the usual `Backends.via_source/3` functions.
Although the PostgresAdaptor uses a separate DynamicSupervisor, but it would be much better to encapsulate all related procs under the same Adaptor process tree so that when we reset proc state, it would not leave stale procs hanging in the DynSup.

Ziinc: this has revealed a architecture issue with the PostgresAdaptor :P

amokan: @Ziinc I have updated this implementation by removing those additional layers. Still using a GenServer for the adaptor though.

amokan: I've added some logic to leverage the `DBConnection` [`:after_connect`](https://hexdocs.pm/db_connection/DBConnection.html#start_link/2) option for provisioning the ClickHouse table.

This works fine, but I notice this logic fires as many times as pools that are configured. Will investigate how to prevent that or create a workaround when I have a chance.

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
```diff
@@ -0,0 +1,27 @@
...
+      when is_non_empty_binary(database) and is_non_empty_binary(table) do
+    """
+    CREATE TABLE IF NOT EXISTS "#{database}"."#{table}" (
+      "id" UUID,
+      "event_message" String,
+      "body" String,
+      "timestamp" DateTime64(6)
+    )
+    ENGINE MergeTree()
+    ORDER BY ("timestamp")
```
chasers, line 23: Lets add a partition by day and ttl here for default 90 days for now

lib/logflare/guards.ex
```diff
@@ -0,0 +1,48 @@
+defmodule Logflare.Guards do
+  @moduledoc """
+  Custom guard functions.
+
+  Utilize these by adding `import Logflare.Guards` in your module.
```
chasers, line 5: Cool!!!

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -0,0 +1,66 @@
...
+          {BufferProducer,
+           [source_id: adaptor_state.source.id, backend_id: adaptor_state.backend.id]},
+        transformer: {__MODULE__, :transform, []},
+        concurrency: 1
+      ],
+      processors: [
+        default: [concurrency: 5, min_demand: 1]
+      ],
+      batchers: [
+        ch: [concurrency: 5, batch_size: 350]
```
chasers, line 30: Probably going to need to be able configure the batch size but we can punt this for a bit

chasers (commented):
looking go to me

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
amokan: I have dropped the `after_connect` approach because it does fire per connection - so a pool size of 3 kicks off that logic 3 times.

Went with a temp GenServer approach instead 🤷

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
amokan: Sounds good. I did switch them to module attributes so they are easy to spot/change in the meantime.

Ziinc:
> Handle v2 source flag somehow
this backend will only be available for v2_pipeline sources, so no need to handle the v2_pipeline flag anywhere within the adaptor.

> Permission/Grant check upon config?
Could you add a new `test_connection/?` optional callback to the Adaptor behaviour?

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
```diff
@@ -0,0 +1,152 @@
...
+    db_table_string =
+      if is_non_empty_binary(database) do
+        "#{database}.#{table}"
+      else
+        "#{table}"
+      end
+
+    Enum.join([
+      """
+      CREATE TABLE IF NOT EXISTS #{db_table_string} (
```
Ziinc, line 43: needs some sanitization checks if we're interpolating it directly into the query. maybe just a simple check to ensure no DDL keywords.

lib/logflare/guards.ex
```diff
@@ -0,0 +1,48 @@
+defmodule Logflare.Guards do
```
Ziinc, line 1: maybe move to Logflare.Utils.Guards

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
amokan: I have now removed that old `table` attribute from the adaptor to minimize confusion.

amokan:
> > Permission/Grant check upon config?

> Could you add a new `test_connection/?` optional callback to the Adaptor behaviour?

I have added the callback @Ziinc. Let me know if you were thinking of something else.

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
amokan: But at the end of the day, if you really want DDL keywords to be filtered - just tell me.

lib/logflare/backends/adaptor/clickhouse_adaptor/provisioner.ex
```diff
@@ -0,0 +1,69 @@
...
+  @spec start_link(ClickhouseAdaptor.source_backend_tuple()) :: GenServer.on_start()
+  def start_link({%Source{}, %Backend{}} = args) do
+    GenServer.start_link(__MODULE__, args)
+  end
+
+  @doc false
+  @impl GenServer
+  def init({%Source{} = source, %Backend{} = backend} = args) do
+    state = %__MODULE__{
+      source: source,
```
Ziinc, line 41: seems like we don't need to store the entire structs in state, can just store the source_id source_token backend_id.
this would consume some memory footprint, though minimal.

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
```diff
@@ -0,0 +1,208 @@
...
+
+  - `:database` - (Optional) Will produce a fully qualified `<database>.<table>` string when provided with a value. Defaults to `nil`.
+  - `:ttl_days` - (Optional) Will add a TTL statement to the table creation query. Defaults to `90`. `nil` will disable the TTL.
+
+  """
+  @spec create_log_ingest_table_statement(table :: String.t(), opts :: Keyword.t()) :: String.t()
+  def create_log_ingest_table_statement(table, opts \\ [])
+      when is_non_empty_binary(table) and is_list(opts) do
+    database = Keyword.get(opts, :database)
+    ttl_days_temp = Keyword.get(opts, :ttl_days, @default_ttl_days)
```
Ziinc, line 52: the ttl should be based off the bigquery retention value on the retention_days value on Source.
should default to 3 days as per free accounts.

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
Ziinc: ah made the comment as the database value comes from the user but doesn't seem to have any validation. 
i tihnk its fine to ignore then

lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -214,6 +212,10 @@
               The port number for the ClickHouse server.
             </small>
           </div>
+          <div class="form-group">
+            <%= label(f_config, :port, "Connection pool size") %>
+            <%= text_input(f_config, :pool_size, class: "form-control", default: 1) %>
```
Ziinc, line 217: does this actually propagate to the ch pool? seems like we're hardcoding it to 3

lib/logflare/backends/adaptor/clickhouse_adaptor/provisioner.ex
amokan: Minor change made to maintain only those 3 attributes of the structs

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
amokan: @Ziinc I think I have this covered now, but please check when you have a moment.

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
ruslandoga: 👋 

ClickHouse supports "identifier" params for tables names and such.

```elixir
iex> Ch.query(pool, "create table {table:Identifier} (name String) engine MergeTree order by ()", %{"table" => "some_random_table (name String) engine MergeTree order by (); drop table Students; --"})

iex> Ch.query!(pool, "select name from system.tables order by metadata_modification_time desc limit 1").rows
#==> [["some_random_table (name String) engine MergeTree order by (); drop table Students; --"]]
```

Thank you for giving :ch a try!

PR #2403, opened 2025-05-27
feat: add headers to alert webhooks

itslenny:
Adds ability to set headers in addition to URL for alert webhooks


Ziinc:
sorry @itslenny will be closing this... 💯/💯 on the attempt though 😆

PR #2408, opened 2025-06-02
feat: otel metrics ingestion through http/proto

v0idpwn:
Adds support for metrics ingestion through http/proto in the API.

Does some refactoring in the otel traces ingestion, since a lot of code can be shared with metrics. Also fixes some issues in traces: for example, it didn't return the expected response on success according to the spec. This could cause issues in some clients.

### DIFF
--- a/test/logflare_web/controllers/log_controller_test.exs
+++ b/test/logflare_web/controllers/log_controller_test.exs
@@ -147,7 +147,7 @@ defmodule LogflareWeb.LogControllerTest do
         |> put_req_header("content-type", "application/x-protobuf")
         |> post(Routes.log_path(conn, :otel_traces), body)
 
-      assert TestUtils.protobuf_response(conn, 200, ExportTraceServiceResponse) ==
+      assert protobuf_response(conn, 200, ExportTraceServiceResponse) ==
                %ExportTraceServiceResponse{partial_success: nil}
 
       assert_receive {^ref, [event1, event2]}, 4000
@@ -178,7 +178,7 @@ defmodule LogflareWeb.LogControllerTest do
         |> put_req_header("content-type", "application/x-protobuf")
         |> post(Routes.log_path(conn, :otel_metrics), body)
 
-      assert TestUtils.protobuf_response(conn, 200, ExportMetricsServiceResponse) ==
+      assert protobuf_response(conn, 200, ExportMetricsServiceResponse) ==
                %ExportMetricsServiceResponse{partial_success: nil}
 
       assert_receive {^ref, data_points}, 4000
@@ -511,5 +511,11 @@ defmodule LogflareWeb.LogControllerTest do
     conn
     |> put_req_header("content-type", "application/json")
     |> post(Routes.log_path(conn, :create, source_name: source.name), input)
+  end
+
+  defp protobuf_response(conn, expected_status, protobuf_schema) do
+    body = Phoenix.ConnTest.response(conn, expected_status)
+
+    protobuf_schema.decode(body)
   end
 end
--- a/test/support/test_utils.ex
+++ b/test/support/test_utils.ex
@@ -204,17 +204,6 @@ defmodule Logflare.TestUtils do
   end
 
   @doc """
-  Asserts a response has a certain status and decodes the protobuf body.
-
-  Similar to Phoenix's `json_response/2` function.
-  """
-  def protobuf_response(conn, expected_status, protobuf_schema) do
-    body = Phoenix.ConnTest.response(conn, expected_status)
-
-    protobuf_schema.decode(body)
-  end
-
-  @doc """
   Used to retrieve a nested BigQuery field schema from a table schema. Returns nil if not found.
 
   ### Example


lib/logflare/logs/otel.ex
```diff
@@ -0,0 +1,54 @@
+defmodule Logflare.Logs.Otel do
```
v0idpwn, line 1: This module was mostly extracted from the traces module. I still have to improve the APIs since some functions that were private are now public.

lib/logflare/logs/otel_metric.ex
```diff
@@ -0,0 +1,69 @@
...
+        "event_message" => metric.name,
+        "metadata" => metadata,
+        "start_time" => Logflare.Logs.Otel.nano_to_iso8601(data_point.start_time_unix_nano),
+        "unit" => metric.unit,
+        "value" => value,
+        "metric_type" => metric_type,
+        "attributes" => Logflare.Logs.Otel.handle_attributes(data_point.attributes),
+        "timestamp" => Logflare.Logs.Otel.nano_to_iso8601(data_point.time_unix_nano),
+        "project" => metadata["name"]
+      }
```
v0idpwn: What I'm not very happy with is the duplication of data for every data point. The metadata is huge:
```json
  "metadata": {
    "host": {
      "arch": "arm64",
      "id": "254B8545-12DB-5BCC-A702-E0F2872D105A",
      "name": "Felipes-MacBook-Pro.local"
    },
    "name": "@opentelemetry/instrumentation-runtime-node",
    "process": {
      "command": "/Users/v0idpwn/sb/otel_test_app/node_modules/.bin/ts-node",
      "command_args": [
        "/Users/v0idpwn/.local/share/mise/installs/node/24.1.0/bin/node",
        "/Users/v0idpwn/sb/otel_test_app/node_modules/.bin/ts-node",
        "--require",
        "./instrumentation.ts",
        "app.ts"
      ],
      "executable": {
        "name": "node",
        "path": "/Users/v0idpwn/.local/share/mise/installs/node/24.1.0/bin/node"
      },
      "owner": "v0idpwn",
      "pid": 97856,
      "runtime": {
        "description": "Node.js",
        "name": "nodejs",
        "version": "24.1.0"
      }
    },
    "service": {
      "name": "unknown_service:node"
    },
    "telemetry": {
      "sdk": {
        "language": "nodejs",
        "name": "opentelemetry",
        "version": "2.0.1"
      }
    },
    "type": "metric",
    "version": "0.15.0"
  },
```
Not sure if I'm bothered due still being in a "OLTP" mindset :)

lib/logflare_web/router.ex
```diff
@@ -461,7 +461,24 @@ defmodule LogflareWeb.Router do
...
+        protobuf_schema: Opentelemetry.Proto.Collector.Trace.V1.ExportTraceServiceRequest
+      }
+    )
+
+    post(
+      "/metrics",
+      LogController,
+      :otel_metrics,
+      assigns: %{
+        protobuf_schema: Opentelemetry.Proto.Collector.Metrics.V1.ExportMetricsServiceRequest
```
Ziinc: can we alias this to make it more concise

test/support/test_utils_grpc.ex
```diff
@@ -36,12 +53,16 @@ defmodule Logflare.TestUtilsGrpc do
     ]
   end
 
-  defp random_scope_span do
-    scope = %InstrumentationScope{
+  defp random_scope() do
```
v0idpwn, line 56:
```suggestion
  defp random_scope do
```

lib/logflare/logs/otel.ex
```diff
@@ -0,0 +1,54 @@
...
+
+  def extract_value(%AnyValue{value: {:array_value, %ArrayValue{values: values}}}) do
+    Enum.map(values, &extract_value/1)
+  end
+
+  def extract_value(%_{value: {_type, value}}), do: value
+  def extract_value(nil), do: nil
+  def extract_value(value), do: value
+
+  # TODO: rename
```
v0idpwn:
```suggestion
```
I'll leave as it is

test/logflare_web/controllers/log_controller_test.exs
```diff
@@ -143,14 +147,48 @@ defmodule LogflareWeb.LogControllerTest do
...
+        |> put_req_header("x-source", Atom.to_string(source.token))
+        |> put_req_header("content-type", "application/x-protobuf")
+        |> post(Routes.log_path(conn, :otel_metrics), body)
+
+      assert TestUtils.protobuf_response(conn, 200, ExportMetricsServiceResponse) ==
+               %ExportMetricsServiceResponse{partial_success: nil}
+
+      assert_receive {^ref, [dp1, dp2, dp3, dp4]}, 4000
+
+      assert %{"metadata" => _, "event_message" => _, "metric_type" => _} = dp1
```
Ziinc: can we do an Enum.all

test/support/test_utils.ex
```diff
@@ -203,6 +203,17 @@ defmodule Logflare.TestUtils do
     }
   end
 
+  @doc """
```
Ziinc: if this is only used in one location we don't need a separate function in utils for it, can simply be a private function in log_controller.ex module.

test/support/test_utils_grpc.ex
```diff
@@ -213,4 +234,131 @@ defmodule Logflare.TestUtilsGrpc do
       value: %AnyValue{value: {:double_value, :rand.uniform(100) + 1.00}}
     }
   end
+
```
Ziinc, line 237: would be good if we can eventually migrate this to use StreamData for generation for better property testing. not a priority

lib/logflare_web/controllers/log_controller.ex
```diff
@@ -229,7 +232,29 @@ defmodule LogflareWeb.LogController do
...
+        %ExportMetricsServiceRequest{resource_metrics: resource_metrics}
+      ) do
+    case Processor.ingest(resource_metrics, Logs.OtelMetric, source) do
+      {:ok, _} ->
+        resp = Protobuf.encode_to_iodata(%ExportMetricsServiceResponse{})
+        send_resp(conn, 200, resp)
+
+      # Not sure what should be done here
+      {:error, errors} ->
+        raise RuntimeError, errors
```
Ziinc: should return as an 400 resp error message. you can just let the fallback controller handle it, we should have a catch-all error for bad requests.
can do `{:error, "Bad Protocol Buffer Request"}` or something like that.

Ziinc: responding in plain text might be ideal, might need to have a separate function head for that in fallback controller since we usualy use it for json response.

lib/logflare/logs/otel_metric.ex
```diff
@@ -0,0 +1,162 @@
+defmodule Logflare.Logs.OtelMetric do
+  @moduledoc """
+  Converts a Otel metric to event parameters
+
+  Converts a list of ResourceMetrics to a list of Logflare Events
+
+  Important details about the conversation:
```
Ziinc, line 7: llm leak

lib/logflare_web/controllers/log_controller.ex
Ziinc: we can leave it as a json response for now, don't need to bother with text err response

lib/logflare/logs/otel_metric.ex
v0idpwn: Actually copy pasted from the trace module :), I'm not gonna use LLMs until I'm comfortable with the codebase.

test/support/test_utils.ex
v0idpwn: I'm anticipating that it will be used for gRPC as well, so I followed the same approach that was in place for Traces. Happy to change if you think it should be private anyway.

lib/logflare_web/controllers/log_controller.ex
v0idpwn: I need to see what would be OTLP compliant...

v0idpwn: I guess we need to give a `Status` response: https://opentelemetry.io/docs/specs/otlp/#failures-1

v0idpwn: I pushed something... Not very happy with it but it's conformant...

test/support/test_utils.ex
v0idpwn: Ah, my bad, I just realised you were talking about this one and not the ones in TestUtilsGrpc. I'll move it around.

v0idpwn: Good call, fixed!

PR #2410, opened 2025-06-04
feat: gRPC support for Otel metrics

v0idpwn:
This should also probably support HTTP/json (through transcoding from the library), although I haven't tested this alternative and it may not be fully conformant.

### DIFF
--- a/lib/logflare_web/controllers/plugs/verify_resource_access.ex
+++ b/lib/logflare_web/controllers/plugs/verify_resource_access.ex
@@ -74,6 +74,11 @@ defmodule LogflareWeb.Plugs.VerifyResourceAccess do
   def check_resource(%Source{}, nil), do: true
 
   def check_resource(%Source{} = resource, token) do
-    :ok == Auth.check_scopes(token, ["ingest", "ingest:source:#{resource.id}", "ingest:collection:#{resource.id}"])
+    :ok ==
+      Auth.check_scopes(token, [
+        "ingest",
+        "ingest:source:#{resource.id}",
+        "ingest:collection:#{resource.id}"
+      ])
   end
 end


test/logflare_grpc/verify_api_resource_access_test.exs
```diff
@@ -0,0 +1,135 @@
...
+      %EmptyResponse{}
+    end
+  end
+
+  defmodule TestEndpoint do
+    use GRPC.Endpoint
+
+    intercept(LogflareGrpc.Interceptors.VerifyApiResourceAccess)
+    run(MockServer)
+  end
```
v0idpwn, line 21: We can maybe split this, or make it smarter, if we add more interceptors.

lib/logflare_grpc/interceptors/verify_api_resource_access.ex
```diff
@@ -0,0 +1,87 @@
...
+    end
+  end
+
+  defp identify_requestor(api_key, scopes) do
+    case VerifyApiAccess.identify_requestor(api_key, scopes) do
+      {:ok, access_token, user} ->
+        {:ok, access_token, user}
+
+      {:error, _reason} ->
+        {:error, GRPC.RPCError.exception(status: :unauthenticated, message: "Invalid API key")}
```
Ziinc: shouldn't reveal the reason. Just say `Unauthorized`

lib/logflare_grpc/interceptors/verify_api_resource_access.ex
```diff
@@ -0,0 +1,87 @@
...
+    end
+  end
+
+  defp fetch_api_key(stream) do
+    case GRPC.Stream.get_headers(stream) do
+      %{"authorization" => "Bearer " <> token} ->
+        {:ok, token}
+
+      %{"x-api-key" => key} ->
+        {:ok, key}
```
Ziinc:
```suggestion
      %{"x-api-key" => token} ->
        {:ok, token}
```

lib/logflare_grpc/interceptors/verify_api_resource_access.ex
```diff
@@ -0,0 +1,87 @@
...
+    end
+  end
+
+  defp fetch_source_id(stream) do
+    case GRPC.Stream.get_headers(stream) do
+      %{"x-collection" => token} ->
+        {:ok, token}
+
+      %{"x-source" => token} ->
+        {:ok, token}
```
Ziinc:
```suggestion
      %{"x-collection" => source_uuid ->
        {:ok, source_uuid}

      %{"x-source" => source_uuid ->
        {:ok, source_uuid}
```
just a note, the uuid is a string at this point, while internally on Source.token it is an atom. renaming the bound variable will help avoid confusion

lib/logflare_grpc/interceptors/verify_api_resource_access.ex
```diff
@@ -0,0 +1,87 @@
...
+         GRPC.RPCError.exception(
+           status: :unauthenticated,
+           message: "Missing source id"
+         )}
+    end
+  end
+
+  defp fetch_source(user, source_token) do
+    if Sources.valid_source_token_param?(source_token) do
+      case Sources.get_by_and_preload(user_id: user.id, token: source_token) do
```
Ziinc: should use Sources.Cache

test/logflare_grpc/metrics/server_test.exs
```diff
@@ -0,0 +1,48 @@
...
+    setup do
+      user = insert(:user)
+      source = insert(:source, user: user)
+
+      {:ok, _pid, port} = GRPC.Server.start_endpoint(LogflareGrpc.Endpoint, 0)
+      on_exit(fn -> GRPC.Server.stop_endpoint(LogflareGrpc.Endpoint) end)
+      {:ok, %{source: source, user: user, port: port}}
+    end
+
+    defp emulate_request(channel, request) do
```
Ziinc: we can move this out of the describe block to the bottom of the file for readability.

test/logflare_grpc/metrics/server_test.exs
```diff
@@ -0,0 +1,48 @@
...
+      Stub.export(channel, request)
+    end
+
+    test "returns a success response and starts log event ingestion using access token", %{
+      source: source,
+      user: user,
+      port: port
+    } do
+      access_token = insert(:access_token, resource_owner: user)
```
Ziinc:
```suggestion
      access_token = insert(:access_token, resource_owner: user, scopes: "ingest")
```

Ziinc: it currently defaults to `public` per config.exs,  which is considered legacy and deprecated

test/logflare_grpc/verify_api_resource_access_test.exs
```diff
@@ -0,0 +1,135 @@
...
+      assert {:ok, %EmptyResponse{}} = request_with_headers(headers, port)
+    end
+
+    test "success using access token for specific source", %{
+      source: source,
+      user: user,
+      port: port
+    } do
+      access_token =
+        insert(:access_token, resource_owner: user, scopes: "ingest ingest:source:#{source.id}")
```
Ziinc, line 57:
```suggestion
        insert(:access_token, resource_owner: user, scopes: "ingest:source:#{source.id}")
```
having `ingest` in the scope will allow it to ingest to any source, making the specific source scope redundant.

test/logflare_grpc/verify_api_resource_access_test.exs
Ziinc: this approach is fine, but i wonder if we are introducing additional complexity with the TestProtobuf modules, when testing the interceptor would also be covered under the server_test.exs setup.

Another approach to test the interceptors in isolation is to mock the callbacks that are passed to the interceptor, or to abstract out and expose only the core wrapping logic. 

in the end, a lot of the auth logic actually lives in Plugs.VerifyApiAccess, so having a very extensive interceptor test would be redundant when we are actually testing for whether the interceptor wraps VerifyApiAccess correctly.

just some food for thought

Ziinc:
👍 do test the transcoding and update the otel docs with example config for it.
Would also be beneficial to add in an example to the docs on using the electricsql exporter as reference for the elixir community that uses us.

PR #2411, opened 2025-06-05
fix: single tenant seeding bug, update dev docs

Ziinc:
This PR fixes a bug with local seeding of billing plans, where the server errors out during data seeding when there are multiple Enterprise plans present.

This also adds in some additional scripts to manage starting in single tenant mode.


lib/logflare/single_tenant.ex
```diff
@@ -186,18 +186,51 @@ defmodule Logflare.SingleTenant do
...
+          for plan <- plans, plan.id != to_keep_and_update.id do
+            Billing.delete_plan(plan)
+          end
+
+          {:ok, updated}
+        end
+
+      [] ->
+        Billing.create_plan(@plan_attrs)
     end
```
josevalim, line 234: Hi @Ziinc! I took at stab at unifying the different clauses. If you don't need to return `{:error, :already_created}`, then it could be even a bit simpler. :) Feel free to use it or discard it.

```suggestion
    case plans do
      [_ | _] = plans ->
        keys = Map.keys(@plan_attrs)

        # Find a matching plan or get the first one
        to_keep =
          Enum.find(plans, List.first(plans), fn plan ->
            attrs = Map.take(plan, keys)
            attrs == @plan_attrs
          end)

        # Delete everything else
        for plan <- plans, plan.id != to_keep.id do
          Billing.delete_plan(plan)
        end
        
        cond do
          Map.take(to_keep, keys) != @plan_attrs ->
            Billing.update_plan(plan, @plan_attrs)
          
          length(plans) == 1 ->
            {:error, :already_created}
            
          true ->
            {:ok, to_keep}
        end

      [] ->
        Billing.create_plan(@plan_attrs)
    end
```

PR #2412, opened 2025-06-05
feat: export some metrics through otel

v0idpwn:
Since #2408 we support ingestion of otel metrics through oltp/http. This PR starts "dogfooding" this feature, exporting a few metrics via the OtelMetricExporter library.


lib/logflare_web/router.ex
```diff
@@ -72,6 +72,19 @@ defmodule LogflareWeb.Router do
...
+  pipeline :otlp_api do
+    plug(Plug.RequestId)
+
+    plug(Plug.Parsers,
+      parsers: [ProtobufParser],
+      body_reader: {PlugCaisson, :read_body, []},
+      length: 12_000_000
+    )
+
+    plug(:accepts, ["protobuf"])
```
v0idpwn, line 84: While the NodeJS exporter doesn't care, the Elixir one conflicts with our previous accepts config. I made it a separate pipeline since the main API endpoints don't support protobuf (and the otlp ones don't support json/bert)

Ziinc: Would be better to split the accepts to different pipeline functions while keeping the common plugs within the same pipeline, makes it harder for configs to deviate over time.

Ziinc (approved):
You'll also need to do a minor version bump the VERSION file at root

PR #2413, opened 2025-06-06
feat: export all metrics through otel

v0idpwn:
This is a follow up to #2412, and is dependent on it working well. Instead of exporting a selected subset of metrics, we export all of them.

A bunch of metrics had to be converted from summaries to distributions, since the exporter doesn't support summaries (and distributions are generally better anyway). Some of them had wrong types (e.g.: we probably aren't interested in a distribution of memory but in the last value available).

### DIFF
--- a/lib/telemetry.ex
+++ b/lib/telemetry.ex
@@ -70,7 +70,9 @@ defmodule Logflare.Telemetry do
 
     database_metrics = [
       distribution("logflare.repo.query.total_time", unit: {:native, :millisecond}),
-      distribution("logflare.repo.query.decode_time", unit: {:native, :millisecond}),
+      # TODO: decode_time is `nil` in some of the ecto queries
+      # In most telemetry adapters this is fine, but it causes issues in OtelMetricExporter
+      # distribution("logflare.repo.query.decode_time", unit: {:native, :millisecond}),
       distribution("logflare.repo.query.query_time", unit: {:native, :millisecond}),
       distribution("logflare.repo.query.queue_time", unit: {:native, :millisecond}),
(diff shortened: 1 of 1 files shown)


lib/telemetry.ex
```diff
@@ -64,59 +64,67 @@ defmodule Logflare.Telemetry do
...
 
     database_metrics = [
-      summary("logflare.repo.query.total_time", unit: {:native, :millisecond}),
-      summary("logflare.repo.query.decode_time", unit: {:native, :millisecond}),
-      summary("logflare.repo.query.query_time", unit: {:native, :millisecond}),
-      summary("logflare.repo.query.queue_time", unit: {:native, :millisecond}),
-      summary("logflare.repo.query.idle_time", unit: {:native, :millisecond})
+      distribution("logflare.repo.query.total_time", unit: {:native, :millisecond}),
+      # TODO: decode_time is `nil` in some of the ecto queries
+      # In most telemetry adapters this is fine, but it causes issues in OtelMetricExporter
```
Ziinc, line 74: what issues specifically? might be worth opening up an issue on the repo.

lib/telemetry.ex
```diff
@@ -194,4 +188,8 @@ defmodule Logflare.Telemetry do
       :telemetry.execute([:cachex, metric], metrics)
     end)
   end
+
+  defp batch_size_reporter_opts do
+    [buckets: [0, 5, 10, 25, 50, 100, 150, 250, 350, 1000]]
```
```suggestion
    [buckets: [0, 1, 5, 10, 50, 100, 150, 250, 500, 1000, 2000]]
```

Ziinc (approved):

PR #2415, opened 2025-06-08
fix(otel): split scope and resource; project name

v0idpwn:
This PR does two things:

## Split resource and scope into separate fields

OTel traces and metrics always have two entities: resource and scope. We are currently merging both of them and putting them in metadata. This is a bit limiting. This PR changes it to keep them separate maps, and moves the data from `metadata` to root (not sure about this part, though).

Before:
```json
{
  "metadata": {
    "type": "span",
    "service": {
      "name": "Logflare",
      "version": "1.14.2"
    },
    "name": "opentelemetry_bandit",
    "version": "1.2.3",
    "asd": "bsd"
  },
  "event_message": "POST /a/b",
  "span_id": 391283921,
  "trace_id": 30921803120
}
```

After:
```json
{
  "metadata": {
    "type": "span"
  },
  "resource": {
    "service": {
      "name": "Logflare",
      "version": "1.14.2"
    }
  },
  "scope": {
    "name": "opentelemetry_bandit",
    "version": "1.2.3",
    "attributes": {
       "asd": "bsd"
    }
  },
  "event_message": "POST /a/b",
  "span_id": 391283921,
  "trace_id": 30921803120
}
```

I'm happy to keep it **inside** metadata, but I'd really prefer to avoid merging them.

## Project name

Currently, we read the `resource.name` field to determine the project name. Instead, we should probably use `resource.service.name` , according to OTel docs:

>There are attributes provided by the OpenTelemetry SDK. One of them is the `service.name`, which represents the logical name of the service. By default, SDKs will assign the value `unknown_service` for this value, so it is recommended to set it explicitly, either in code or via setting the environment variable `OTEL_SERVICE_NAME`.

In my understanding, there’s currently no “standardizing” for the `resource.name` field, so we shouldn’t assume that it's there. But we can probably rely on `resource.service.name`

### DIFF
--- a/lib/logflare/logs/otel.ex
+++ b/lib/logflare/logs/otel.ex
@@ -16,6 +16,10 @@ defmodule Logflare.Logs.Otel do
     end)
   end
 
+  def resource_project(handled_resource) do
+    handled_resource["service"]["name"]
+  end
+
   def handle_scope(scope) do
     %{name: name, version: version, attributes: scope_attributes} = scope
 
--- a/lib/logflare/logs/otel.ex
+++ b/lib/logflare/logs/otel.ex
@@ -6,7 +6,20 @@ defmodule Logflare.Logs.Otel do
   alias Opentelemetry.Proto.Common.V1.KeyValue
   alias Opentelemetry.Proto.Common.V1.AnyValue
   alias Opentelemetry.Proto.Common.V1.ArrayValue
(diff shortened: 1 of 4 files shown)


Ziinc (approved):
keeping them at root is fine. Would be good to add in documentation around the rationale for the key transformations in the otel docs, as well as examples of the before/after transformations.

PR #2416, opened 2025-06-08
fix(otel-ingestion): compatibility with v1 pipelines

v0idpwn:
While v2 pipelines return `{:ok, count}` on success, v1 pipelines return just `:ok`. Hence we need to accept both.


VERSION
```diff
@@ -1 +1 @@
-1.15.0
\ No newline at end of file
+1.15.1
```
Ziinc, line 1: don't need to bump this as we haven't done a canary or prod deploy yet

lib/logflare_web/controllers/log_controller.ex
```diff
@@ -244,14 +245,23 @@ defmodule LogflareWeb.LogController do
...
+    resource_metrics
+    |> Processor.ingest(Logs.OtelMetric, source)
+    |> protobuf_response(conn, %ExportMetricsServiceResponse{})
   rescue
     exception ->
       send_proto_error(conn, 500, "Internal server error")
       reraise exception, __STACKTRACE__
   end
 
+  defp protobuf_response(ingest_result, conn, success_response) do
```
Ziinc, line 257: would be cleaner to have multiple function heads and match on that instead.

Ziinc: and you can match on error first as well to reduce the conditions by 1

PR #2420, opened 2025-06-09
feat: First pass at S3 parquet log sink

amokan:
Implements a super simple S3 log sink using parquet files.

Batch timeout is adjustable in the backend edit form. Currently defaults to 5000 (5 seconds) or a max batch size of 10k.

Explorer and its Rust NIF handle the parquet files and streaming to S3.

Schema is currently fixed at a four-column setup similar to ClickHouse.


lib/logflare/backends/adaptor/s3_parquet_adaptor.ex
```diff
@@ -0,0 +1,233 @@
...
+       aws_access_key_id: :string,
+       aws_secret_access_key: :string,
+       batch_timeout: :integer
+     }}
+    |> Changeset.cast(params, [
+      :s3_bucket,
+      :aws_region,
+      :aws_access_key_id,
+      :aws_secret_access_key,
+      :batch_timeout
```
Ziinc: lets remove the aws-specific prefixes to avoid confusion in case it points to a different provider

lib/logflare/backends/adaptor/s3_parquet_adaptor.ex
```diff
@@ -0,0 +1,233 @@
...
+  end
+
+  @doc false
+  @impl Logflare.Backends.Adaptor
+  def validate_config(%Changeset{} = changeset) do
+    import Ecto.Changeset
+
+    changeset
+    |> validate_required([:s3_bucket, :aws_region, :aws_access_key_id, :aws_secret_access_key])
+    |> Changeset.validate_number(:batch_timeout, greater_than_or_equal_to: @min_batch_timeout)
```
Ziinc: lets do a @max_batch_timeout of 5s

lib/logflare/backends/adaptor/s3_parquet_adaptor.ex
```diff
@@ -0,0 +1,233 @@
...
+        region: backend.config.aws_region
+      ]
+    )
+  end
+
+  @doc false
+  @impl Supervisor
+  def init({%Source{} = source, %Backend{config: %{} = config} = backend}) do
+    pipeline_state = %__MODULE__{
+      config: config,
```
Ziinc: shouldn't store config in state in event of config changes at runtime, pulling from Backends.Cache and transforming is the best way to avoid stale config.

Also, storing the entire structs results in higher proc memory, storing the IDs is enough.

Pipeline names can be built from ids, IIRC, so no real need for the entire struct.

lib/logflare/backends/adaptor/s3_parquet_adaptor.ex
```diff
@@ -0,0 +1,233 @@
...
+
+  If the process is not located in the registry or does not exist, this will return `nil`.
+  """
+  @spec pipeline_pid(source_backend_tuple()) :: pid() | nil
+  def pipeline_pid({%Source{}, %Backend{}} = args) do
+    case find_pipeline_pid_in_source_registry(args) do
+      {:ok, pid} -> pid
+      _ -> nil
+    end
+  end
```
Ziinc:
```suggestion
  def pipeline_pid({%Source{}, %Backend{}} = args) do
    args
    |> pipeline_via()
    |> GenServer.whereis()
  end
```

lib/logflare/backends/adaptor/s3_parquet_adaptor.ex
```diff
@@ -0,0 +1,233 @@
...
+  @doc """
+  Determines if a particular Broadway pipeline process is alive based on a `Source` and `Backend` pair.
+  """
+  @spec pipeline_alive?(source_backend_tuple()) :: boolean()
+  def pipeline_alive?({%Source{}, %Backend{}} = args) do
+    case pipeline_pid(args) do
+      nil -> false
+      pid -> Process.alive?(pid)
+    end
+  end
```
Ziinc:
```suggestion
  def pipeline_alive?(args), do: !!pipeline_pid(args)
```
pipeline_pid/1 already does match check so we can skip over that

lib/logflare/backends/adaptor/s3_parquet_adaptor.ex
```diff
@@ -0,0 +1,233 @@
...
+
+  @doc """
+  Generates the S3 path for a new parquet file based on a `Source` struct.
+  """
+  @spec new_s3_filename(source_backend_tuple()) :: String.t()
+  def new_s3_filename({%Source{} = source, %Backend{} = backend}) do
+    bucket_name = backend.config.s3_bucket
+    source_token = s3_parquet_source_token(source) |> String.replace("-", "_")
```
Ziinc:
```suggestion
    source_token = s3_parquet_source_token(source)
```
`s3_parquet_source_token/1` already does the replacement

lib/logflare/backends/adaptor/s3_parquet_adaptor.ex
```diff
@@ -0,0 +1,233 @@
...
+
+  @doc """
+  Converts a list of `LogEvent` structs to a parquet file and uploads it to S3.
+  """
+  @spec push_log_events_to_s3(source_backend_tuple(), LogEvent.t() | [LogEvent.t()]) ::
+          :ok | {:error, term()}
+  def push_log_events_to_s3({%Source{} = source, %Backend{} = backend}, %LogEvent{} = le) do
+    push_log_events_to_s3({source, backend}, [le])
+  end
+
```
Ziinc: i think we can remove the singular event fn head, doubt it will be ever used.

lib/logflare/backends/adaptor/s3_parquet_adaptor/pipeline.ex
```diff
@@ -0,0 +1,78 @@
...
+        transformer: {__MODULE__, :transform, []},
+        concurrency: @producer_concurrency
+      ],
+      processors: [
+        default: [concurrency: @processor_concurrency, min_demand: 1]
+      ],
+      batchers: [
+        s3_parquet: [
+          concurrency: 1,
+          batch_size: @batch_size,
```
Ziinc: should use a batch size function that caps the total batch size to say 8MB (by estimating the byte size of each log event body), similar to how the BQ pipeline does it.

Ziinc: https://hexdocs.pm/broadway/Broadway.html#start_link/2-batchers-options

lib/logflare/backends/backend.ex
```diff
@@ -17,7 +17,8 @@ defmodule Logflare.Backends.Backend do
     postgres: Adaptor.PostgresAdaptor,
     bigquery: Adaptor.BigQueryAdaptor,
     loki: Adaptor.LokiAdaptor,
-    clickhouse: Adaptor.ClickhouseAdaptor
+    clickhouse: Adaptor.ClickhouseAdaptor,
+    s3parquet: Adaptor.S3ParquetAdaptor
```
Ziinc: Lets call this S3Adaptor for more flexibility so that we can just extend the inner modules it if we build out s3 tables etc instead of re-creating another adaptor.
```suggestion
    s3: Adaptor.S3Adaptor
```

Ziinc (changes requested):

lib/logflare/backends/adaptor/s3_parquet_adaptor/pipeline.ex
amokan: I hadn't ever looked closely at the function option for batch_size before. Very cool.

lib/logflare/backends/adaptor/s3_parquet_adaptor.ex
amokan: @Ziinc I think this is in a better spot now. I am not doing anything in the transform config callback as it seems to be unnecessary. Please let me know if I'm off on that though!

PR #2421, opened 2025-06-10
fix: add artifact repository role to compute engine SA, add  grafana SA

Ziinc:
fixes artifact repository issues for the IAM policy management.
Also adds in `GOOGLE_GRAFANA_SA` env var for allowing a Service Account specifically for the Grafana BQ plugin

It also does a minor refactor of the resource manager policy binding building for service accounts, to factor in service accounts that are not set.


lib/logflare/google/resource_manager.ex
```diff
@@ -85,80 +85,44 @@ defmodule Logflare.Google.CloudResourceManager do
...
+             "roles/cloudbuild.serviceAgent"
+           ]},
+          {env_cloud_build_trigger_sa(),
+           [
+             "roles/cloudbuild.builds.editor",
+             "roles/iam.serviceAccountUser",
+             "roles/cloudbuild.serviceAgent"
+           ]},
+          {env_api_sa(), ["roles/editor", "roles/cloudbuild.builds.editor"]},
+          {env_grafana_sa(), ["roles/bigquery.dataViewer", "roles/bigquery.jobUser"]}
```
v0idpwn, line 117:
```suggestion
          {env_grafana_sa(), ["roles/bigquery.dataViewer", "roles/bigquery.jobUser"]}
```
Should it maybe be optional? It's gonna crash when it's not set (binary <> nil on line 122)

Ziinc: it gets filtered out with the comprehension at line 119 if it returns nil 😄

v0idpwn: missed in the diff!

PR #2422, opened 2025-06-11
feat(otlp): logs ingestion

v0idpwn:
Adds support for ingesting [OpenTelemetry Logs](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-resource) through http/protobuf and gRPC. In general, OTLP logs are still not widely adopted, but this is a good step for OTLP compliance and future adoption.

Alongside the automated tests, I tested this with [this node application](http://github.com/v0idpwn/otel_test_app) and it works fine.


lib/logflare/logs/otel_log.ex
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  defp handle_log_record(%LogRecord{} = log_record, base_event) do
+    Map.merge(base_event, %{
+      "event_message" => event_message(log_record),
+      "body" => Otel.extract_value(log_record.body),
+      "attributes" => Otel.handle_attributes(log_record.attributes),
+      "severity_number" => SeverityNumber.value(log_record.severity_number),
+      "severity_text" => log_record.severity_text,
+      "trace_id" => Ecto.UUID.cast!(log_record.trace_id),
```
Ziinc: this would not be a uuid, leaving it as is is fine.
Also, doing a bang function on a hot code path is a recipe for disaster 😄 
```suggestion
      "trace_id" => log_record.trace_id,
```

Ziinc (approved):

lib/logflare/logs/otel_log.ex
Ziinc: Resolved in latest commit

v0idpwn: Got it from traces, thought it was by design :)

PR #2425, opened 2025-06-13
chore: upgrade sql parser rust crate to prepare for clickhouse queries

amokan:
Started digging into the sql parsing logic and realized there are quite a few updates needed to ideally support ClickHouse queries. Many of which came around the v0.48 release last summer.

This effort moved us from v0.33 of the rust crate up to v0.48, which gets us a large chunk of CH-specific data types and handling. This seemed like a sane place to stop for the time being.

Admittedly the changes to the AST parsing and manipulation were a bit heavy and I leaned on AI to work out some of those deltas during the big v0.47 jump. I don't love the code around the translate/transform logic, but the tests are passing and we can refactor that stuff later.

You can see the relevant changelog for the rust sql parser [here](https://github.com/apache/datafusion-sqlparser-rs/blob/main/changelog/0.51.0-pre.md) as we go from v0.33.0.


mix.exs
```diff
@@ -177,7 +177,7 @@ defmodule Logflare.Mixfile do
       {:floki, "~> 0.29.0", only: [:test]},
 
       # Rust NIFs
-      {:rustler, "~> 0.34.0"},
+      {:rustler, "~> 0.36.2", override: true},
```
Ziinc, line 182: why is the override necessary?

lib/logflare/sql/parser.ex
```diff
@@ -1,20 +1,46 @@
...
+
+  import Logflare.Utils.Guards
+
+  @valid_dialects ~w(bigquery clickhouse postgres)
+
   defmodule Native do
+    @moduledoc """
+    Module used for interacting with the `sqlparser_ex` NIF.
+    """
+
```
Ziinc:
```suggestion
    @moduledoc false

```

lib/logflare/sql/parser.ex
```diff
@@ -1,20 +1,46 @@
 defmodule Logflare.Sql.Parser do
+  @moduledoc """
+  Provides functionality for converting SQL queries into an AST and back to SQL strings.
+
+  This leverages the [sqlparser](https://crates.io/crates/sqlparser) Rust crate to handle SQL parsing and AST generation.
+
+  Valid dialects are: `"bigquery"`, `"clickhouse"`, and `"postgres"`.
```
Ziinc:
```suggestion
```

native/sqlparser_ex/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "sqlparser_ex"
-version = "0.1.0"
+version = "0.48.0"
```
Ziinc: is this intention to keep it in sync with sqlparser version number? or llm decided it? 
we can leave it as 0.1.0 to avoid the additional work keeping it in sync
```suggestion
version = "0.1.0"
```

native/sqlparser_ex/src/lib.rs
```diff
@@ -19,40 +19,85 @@ struct Response {
...
 #[rustler::nif]
 fn parse(dialect_str: &str, query: &str) -> NifResult<Response> {
     let result = match dialect_str {
         "bigquery" => Parser::parse_sql(&BigQueryDialect {}, query),
+        "clickhouse" => Parser::parse_sql(&ClickHouseDialect {}, query),
         "postgres" => Parser::parse_sql(&PostgreSqlDialect {}, query),
-        _ => Err(ParserError(
-            "Parser for this dialect is not supported.".to_string(),
-        )),
+        _ => return Ok(Response::error("Parser for this dialect is not supported.")),
```
Ziinc: a bit apprehensive on this approach, because now any nif errors results in it becoming log messages and won't get sent to the user.
The error message sometimes contains the raw query, which may leak and expose logic to end users as the error message bubbles up via API.
Not against this though as we can properly handle the errors in the elixir layer, but that would mean that we need to sanitize the error message that gets returned to the user instead of showing the SQL string.
Would need to add that into parser.ex or sql.ex to clean the message before we merge this.

native/sqlparser_ex/src/lib.rs
```diff
@@ -19,40 +19,85 @@ struct Response {
...
+    let mut value: serde_json::Value = match serde_json::from_str(json) {
+        Ok(val) => val,
+        Err(e) => return Ok(Response::error(&format!("JSON parsing error: {}", e))),
+    };
 
-    let mut parts = vec![];
-    for node in nodes {
-        parts.push(format!("{}", node))
-    }
+    ensure_required_fields(&mut value);
```
Ziinc: i think we can revert the this validation logic, seems unnecessary and makes the function more restrictive.

Ziinc (changes requested):
biggest issue is the error handling changes for parsing the ast. would need to add in a test to verify that we don't expose the AST to the end user.

test/logflare/sql_test.exs
```diff
@@ -263,9 +262,6 @@ defmodule Logflare.SqlTest do
             },
             {{"with src as (select a from my_table) select c from src",
               "select a from b; select c from d;"}, "Only singular query allowed"},
-            # no source name in query
-            {"select datetime() from light-two-os-directions-test",
-             "can't find source light-two-os-directions-test"},
```
Ziinc, line 268: why is this removed?

mix.exs
amokan: explorer is hardcoded upstream in main to rely on 34

native/sqlparser_ex/Cargo.toml
amokan: I kept it in sync to make it obvious at comp/runtime what version of the underlying rust crate I was dealing with since really the internal rust code is just a wrapper.

test/logflare/sql_test.exs
amokan: Its no longer a parser issue.

native/sqlparser_ex/src/lib.rs
amokan: ok

test/logflare/sql_test.exs
amokan: https://github.com/apache/datafusion-sqlparser-rs/blob/main/changelog/0.51.0-pre.md#0450-2024-04-12

> Support unquoted hyphen in table/view declaration for BigQuery (#1178)

@Ziinc

native/sqlparser_ex/Cargo.toml
amokan: ~~I guess I do not understand why this would be an issue? The code wasn't updated in 2-3 years, so I struggle to imagine it's going to be maintained on a regular basis going forward.~~

~~Tell me if I'm wrong though.~~

I'll just change it to 0.1.0 again and I can always tweak it locally when dealing with this.

mix.exs
amokan: And I should note that the changes are safe between 0.34 and 0.36.2 in context of explorer.

https://hexdocs.pm/rustler/changelog.html

native/sqlparser_ex/src/lib.rs
amokan: I'll just go back to the old approach to keep it simple for now. Was mostly using this as a crutch to understand problems during the transition. We can revisit a better approach later.

lib/logflare/sql/parser.ex
```diff
@@ -1,20 +1,43 @@
...
     end
   end
 
+  @doc """
+  Converts an AST or list of ASTs back into a SQL string.
+  """
+  @spec to_string(map()) :: String.t()
   def to_string(ast) when is_map(ast), do: __MODULE__.to_string([ast])
 
+  @spec to_string([map()]) :: String.t()
```
Ziinc:
```suggestion
  @spec to_string(map() | [map()]) :: String.t()
  def to_string(ast) when is_map(ast), do: __MODULE__.to_string([ast])
```

lib/logflare/sql/parser.ex
```diff
@@ -1,20 +1,41 @@
...
     end
   end
 
+  @doc """
+  Converts an AST or list of ASTs back into a SQL string.
+  """
+  @spec to_string(map() | [map()]) :: String.t()
   def to_string(ast) when is_map(ast), do: __MODULE__.to_string([ast])
-
   def to_string(asts) when is_list(asts) do
```
Ziinc, line 40:
```suggestion

  def to_string(asts) when is_list(asts) do
```

PR #2426, opened 2025-06-15
perf: use PartitionSupervisor for Endpoints

chasers:
- uses PartitionSupervisor for Endpoints

### DIFF
--- a/lib/logflare/endpoints/resolver.ex
+++ b/lib/logflare/endpoints/resolver.ex
@@ -30,7 +30,9 @@ defmodule Logflare.Endpoints.Resolver do
         spec = {Cache, {query, params}}
         Logger.debug("Starting up Endpoint.Cache for Endpoint.Query id=#{id}", endpoint_id: id)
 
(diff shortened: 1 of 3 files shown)


lib/logflare/endpoints/resolver.ex
```diff
@@ -19,7 +19,8 @@ defmodule Logflare.Endpoints.Resolver do
   Returns the resolved pid.
   """
   def resolve(%Logflare.Endpoints.Query{id: id} = query, params) do
-    :syn.lookup(:endpoints, {id, params})
+    {:via, :syn, {:endpoints, {query.id, params}}}
+    |> GenServer.whereis()
```
nelsonmestevao, line 23: @chasers with this change I believe we now only have two possible responses, it should be either a pid or nil value :warning: 

cc/ @Ziinc

nelsonmestevao: I open a PR https://github.com/Logflare/logflare/pull/2440 to improve this :pray:

PR #2431, opened 2025-06-16
fix: endpoints caches should always shutdown

chasers:
- fixes tests
- catches cases where Endpoints processes don't always shutdown
- added `hibernate_after: 5_000`

### DIFF
--- a/lib/logflare/endpoints/cache.ex
+++ b/lib/logflare/endpoints/cache.ex
@@ -159,11 +159,10 @@ defmodule Logflare.Endpoints.Cache do
   end
 
   def handle_cast(:touch, state) do
-    state = %{state | last_query_at: DateTime.utc_now()}
     :timer.cancel(state.shutdown_timer)
     timer = state.query |> cache_duration_ms() |> shutdown()
 
-    {:noreply, %{state | shutdown_timer: timer}}
+    {:noreply, %{state | shutdown_timer: timer, last_query_at: DateTime.utc_now()}}
   end
 
   def handle_info(:refresh, state) do
--- a/lib/logflare/endpoints/cache.ex
+++ b/lib/logflare/endpoints/cache.ex
@@ -35,7 +35,7 @@ defmodule Logflare.Endpoints.Cache do
 
     name = {:via, :syn, {endpoints, {query.id, params}}}
 
-    GenServer.start_link(__MODULE__, {query, params}, name: name)
+    GenServer.start_link(__MODULE__, {query, params}, name: name, hibernate_after: 5_000)
   end
 
   @doc """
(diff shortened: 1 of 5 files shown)


lib/logflare/endpoints/cache.ex
```diff
@@ -170,29 +177,32 @@ defmodule Logflare.Endpoints.Cache do
     {:noreply, %{state | query_tasks: tasks}}
   end
 
+  def handle_info(:shutdown, state) do
+    {:stop, :normal, state}
```
Ziinc, line 157:
```suggestion
    Logger.debug("#{__MODULE__} : shutting down cache")
    {:stop, :normal, state}
```

lib/logflare/endpoints/cache.ex
```diff
@@ -154,8 +159,10 @@ defmodule Logflare.Endpoints.Cache do
   end
 
   def handle_cast(:touch, state) do
-    state = %{state | last_query_at: DateTime.utc_now()}
-    {:noreply, state}
+    :timer.cancel(state.shutdown_timer)
+    timer = state.query |> cache_duration_ms() |> shutdown()
+
+    {:noreply, %{state | shutdown_timer: timer, last_query_at: DateTime.utc_now()}}
```
Ziinc: this changes the behaviour to extend the max lifetime of the cache by the cache duration on each endpoint query. Can result in stale data.
Also no test coverage for the touch behaviour change.
imo should only extend cache duration if proactive requerying is set

lib/logflare/endpoints/cache.ex
```diff
@@ -170,29 +177,32 @@ defmodule Logflare.Endpoints.Cache do
...
+  end
+
   def handle_info({from_task, {:ok, results}}, state) do
     tasks = Enum.reject(state.query_tasks, &(&1.pid == from_task))
-    {:noreply, %{state | cached_result: results, query_tasks: tasks}}
+
+    :timer.cancel(state.shutdown_timer)
+    timer = refresh(proactive_querying_ms(state))
+
+    {:noreply, %{state | cached_result: results, query_tasks: tasks, shutdown_timer: timer}}
```
Ziinc: This means that if results are received, the Cache will never shutdown through the `:shutdown` callback and will continuously keep refreshing...

Ziinc: also rather confusing to set the `timer` to `:shutdown_timer`, doesn't quite make sense.
 setting `shutdown_timer: nil` and having separate key like `:refresh_timer` would be clearer

Ziinc (changes requested):
Most problematic is the `:touch` issue.

Also, the `:shutdown_timer` key usage has code smell, feels off and harder to reason

test/logflare/endpoints_cache.test.ex
```diff
@@ -143,8 +154,30 @@ defmodule Logflare.EndpointsCacheTest do
       end)
 
       # After proactive_requerying_seconds, should return updated response
-      Process.sleep(600)
+      Process.sleep(endpoint.proactive_requerying_seconds * 1000 + 100)
       assert {:ok, %{rows: [%{"testing" => "456"}]}} = Endpoints.run_cached_query(endpoint)
```
Ziinc, line 158: This test doesn't cover the shutdown of the cache **after** the proactive requerying, hence why the above issue wasn't caught by the test

lib/logflare/endpoints/cache.ex
Ziinc: added a separate `:refresh_timer` to handle this.

test/logflare/endpoints_cache.test.ex
Ziinc: added more tests to cover this

lib/logflare/endpoints/cache.ex
Ziinc: removed the touch/1 api, we aren't using it anyways.

PR #2432, opened 2025-06-16
fix: remove unreachable code path

nelsonmestevao:


nelsonmestevao:
Hi @Ziinc, I removed the case statement because backends in AlertQuery will always be a list after the preload (even if empty), so the other case is not possible to happen as far as I understand :pray: Let me know if you disagree, please.

Ziinc (approved):
👍

PR #2433, opened 2025-06-17
feat: managed service account querying strategy

Ziinc:
This PR adds in managed service account querying, to get around Google's 100 rps rate limit on core REST api calls.

### DIFF
--- a/lib/logflare/backends/adaptor/bigquery_adaptor.ex
+++ b/lib/logflare/backends/adaptor/bigquery_adaptor.ex
@@ -175,7 +175,7 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
     project_id = project_id || Application.get_env(:logflare, Logflare.Google)[:project_id]
 
     get_next_page(project_id, nil)
-    |> Enum.filter(&(&1.name =~ @service_account_prefix))
+    |> Enum.filter(&(&1.name =~ @service_account_prefix <> "@"))
   end
 
   defp handle_response({:ok, response}, project_id) do
@@ -309,16 +309,7 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
   @spec goth_child_spec(String.t(), String.t()) :: Supervisor.child_spec()
   def goth_child_spec(json, sub \\ nil) do
     credentials = Jason.decode!(json)
-    scopes = ["https://www.googleapis.com/auth/cloud-platform"]
-
-    source =
-      if sub,
-        do:
-          {:service_account, credentials,
-           [
-             sub: sub
-           ]},
-        else: {:service_account, credentials, scopes: scopes}
+    source = {:service_account, credentials, if(sub, do: [sub: sub], else: [])}
 
     {
       Goth,


Ziinc:
Hi @wojtekmach , would need some advice on the service account impersonation via Goth [here](https://github.com/Logflare/logflare/pull/2433/files#diff-cc0ef6d91b0fc5b3a064a2c4972cd2f807643b9485580d5ed6bf012e5e67bc43R241-R254). I've tried achieving the SA impersonation using the custom claims map as [documented](https://hexdocs.pm/goth/1.4.5/Goth.Token.html#fetch/1-source), but it does not seem to work

This [issue](https://github.com/peburrows/goth/issues/155) mentions the docs discrepancy, but i suspect why they're able to successfully impersonate is due to domain-wide delegation vs the gcp-specific iam role granted as documented [here](https://cloud.google.com/iam/docs/create-short-lived-credentials-direct) in the GCP docs.

Any ideas?

Ziinc:
@wojtekmach thanks for the hint, I was able to get the impersonation up and running with our fork 🎉 
i've put out a draft PR for the fork for now 
https://github.com/peburrows/goth/pull/188

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -135,4 +138,265 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
...
+
+
+    https://hexdocs.pm/google_api_iam/0.45.0/GoogleApi.IAM.V1.Model.ServiceAccount.html
+  """
+  @spec list_managed_service_accounts(String.t()) :: [GoogleApi.IAM.V1.Model.ServiceAccount.t()]
+  def list_managed_service_accounts(project_id \\ nil) do
+    project_id = project_id || Application.get_env(:logflare, Logflare.Google)[:project_id]
+
+    get_next_page(project_id, nil)
+    |> Enum.filter(&(&1.name =~ @service_account_prefix))
```
wojtekmach: nitpick: would this work as well? I think it's a bit stricter, ensures prefix is in fact a prefix.

```suggestion
    |> Enum.filter(&String.starts_with?(&1.name, @service_account_prefix))
```

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -135,4 +138,265 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
...
+    scopes = ["https://www.googleapis.com/auth/cloud-platform"]
+
+    source =
+      if sub,
+        do:
+          {:service_account, credentials,
+           [
+             sub: sub
+           ]},
+        else: {:service_account, credentials, scopes: scopes}
```
wojtekmach:
```suggestion
    options =
      if sub do
        [sub: sub]
      else
        [scopes: scopes]
      end

    source = {:service_account, credentials, options}
```

btw, I think what you're using is the default scope so setting it explicitly shouldn't be needed. If it is needed, please open up an issue because that looks like a bug!

wojtekmach (commented):
Looks good to me, I left a couple minor comments below.

PR #2439, opened 2025-06-17
chore: add validation tests for source_schema

nelsonmestevao:
This change also makes `Ecto.Term` not convert null values to binary, this change brings the necessary tests to make sure it works properly.


test/logflare/source_schemas_test.exs
```diff
@@ -4,6 +4,36 @@ defmodule Logflare.SourceSchemasTest do
...
+
+  test "create_source_schema/2 missing required attributes returns error changeset" do
+    user = insert(:user)
+    source = insert(:source, user: user)
+
+    attrs = %{}
+
+    assert {:error, %Ecto.Changeset{} = changeset} =
+             SourceSchemas.create_source_schema(source, attrs)
+
```
Ziinc:
```suggestion
    assert {:error, %Ecto.Changeset{} = changeset} =
             SourceSchemas.create_source_schema(source, %{})
```

test/logflare/source_schemas_test.exs
```diff
@@ -4,6 +4,36 @@ defmodule Logflare.SourceSchemasTest do
...
+  test "create_source_schema/2 with valid attributes" do
+    user = insert(:user)
+    source = insert(:source, user: user)
+
+    attrs = %{
+      bigquery_schema: TestUtils.build_bq_schema(%{"test" => 123}),
+      schema_flat_map: %{"test" => "integer"}
+    }
+
+    assert {:ok, %Logflare.SourceSchemas.SourceSchema{} = schema} =
```
Ziinc: alias

test/logflare/source_schemas_test.exs
```diff
@@ -4,6 +4,36 @@ defmodule Logflare.SourceSchemasTest do
 
   alias Logflare.SourceSchemas
 
+  test "create_source_schema/2 with valid attributes" do
+    user = insert(:user)
+    source = insert(:source, user: user)
+
+    attrs = %{
+      bigquery_schema: TestUtils.build_bq_schema(%{"test" => 123}),
+      schema_flat_map: %{"test" => "integer"}
```
Ziinc: can we also add a test to cover validation for nested maps, it should reject multiple levels

Ziinc (changes requested):

test/logflare/source_schemas_test.exs
nelsonmestevao: Done in last push :pray:

test/logflare/source_schemas_test.exs
nelsonmestevao: Just to make sure I understood correctly, this is currently not rejecting multiple nested levels, right? I will need to drop any key that the value is a map or should I error in that case?

nelsonmestevao: I will do it today :pray:

nelsonmestevao:
Hi @Ziinc, I used `StreamData` in tests where I though was appropriate, and I added a validation to `:schema_flat_map` making sure it is a flat map. I don't know if I should also validate all values are only atoms or strings too, WDYT? :thinking:

Ziinc (approved):

PR #2440, opened 2025-06-17
fix: correctly match on the response for pid via syn in Endpoint.Resolver

nelsonmestevao:


Ziinc (approved):
@chasers might be reason why so many caches were getting created 💀 

@nelsonmestevao nice catch

Ziinc:
@nelsonmestevao could you add a follow up PR to cover this specific Resolver code in the case where:
if an Endpoint.Cache already exists for a given Endpoint.Query and param, it should return the same pid
🙏

Ziinc:
Haha yea adding in dialyzer would be nice, IIRC I put it off when I saw the mountain of errors that dialyzer spit out 😂

PR #2443, opened 2025-06-18
chore: improve dev env

Ziinc:
Adds `make start.{st|sb}.{pg|bq}` variants and `make reset`. 
Adds clearer documentation
also does some minor fixes to run.sh script so that it allows overriding of file env vars (i.e. only sets env vars from file if not already present), allowing docker-compose `environment` to override env vars (see `docker-compose.pg.yml`


Makefile
```diff
@@ -30,6 +33,7 @@ compile.check:
 .PHONY: test test.only compile.check
 
 setup: setup.node
+	-epmd -daemon
```
josevalim, line 36: I may be something some Makefile sourcery but should this be:

```suggestion
	epmd -daemon
```

nelsonmestevao: I think the `-` means continue even if `epdm` command fails :pray:

josevalim: TIL, thank you!

PR #2447, opened 2025-06-19
fix: set clickhouse log event table to be ordered descending

amokan:
Order newly created log ingest tables to be sorted descending by default


amokan:
closing this since the descending sort feature is experimental.

https://clickhouse.com/docs/operations/settings/merge-tree-settings#allow_experimental_reverse_key

PR #2458, opened 2025-06-23
test: improve coverage on BillingAccountLive.Edit

nelsonmestevao:
The `BillingAccountLive.Edit` template is now colocated as the new versions of Phoenix LiveView support and promote (so I also deleted the old `BillingAccountView` module).

Fixed the bug on `send_update/2` calls where the id was an atom and it should have been a string.

This PR also takes the opportunity to replace the old Routes usage for path with the new sigil for billing routes.


lib/logflare_web/live/billingaccount_live/edit.ex
nelsonmestevao, line 1: This is the main LiveView being tested :pray:

lib/logflare_web/live/billingaccount_live/edit.ex
```diff
@@ -79,15 +81,15 @@ defmodule LogflareWeb.BillingAccountLive do
...
+    send_update(LogflareWeb.BillingAccountLive.ChartComponent, id: "chart", counter: counter)
 
     {:noreply, socket}
   end
 
   def handle_info({:update_payment_methods, callback, method}, socket) do
     send_update(
       LogflareWeb.BillingAccountLive.PaymentMethodComponent,
-      id: :payment_method,
+      id: "payment_method",
```
nelsonmestevao, line 92: The atom version was actually not calling the component with the updated assigns :warning:

test/logflare_web/live/billingaccount_live/edit_test.exs
nelsonmestevao, line 1: These are the tests that I am adding for the [main LiveView](https://github.com/Logflare/logflare/pull/2458/files#r2162572138) :muscle:

lib/logflare_web.ex
```diff
@@ -127,6 +127,12 @@ defmodule LogflareWeb do
     end
   end
 
+  def routes do
```
Ziinc, line 130: if this is simply using the path_helpers macro, then we can make the path_helpers non-private and use that direcltly

Ziinc: maybe renaming `path_helpers/0` to `routes/0` would be nicer.

nelsonmestevao: I did that in last push, WDYT? :muscle:

Ziinc: looks great 💟

PR #2459, opened 2025-06-23
test: add tests for Logflare.Sql.contains_cte?/2 and fix typo

nelsonmestevao:


lib/logflare/sql.ex
```diff
@@ -609,7 +609,7 @@ defmodule Logflare.Sql do
 
   defp find_all_source_names(_kv, acc, _data), do: acc
 
-  defp extract_cte_alises(ast) do
+  defp extract_cte_aliases(ast) do
```
nelsonmestevao, line 612: Just took the opportunity to also fix this typo :pray:

test/logflare/sql_test.exs
```diff
@@ -487,6 +487,112 @@ defmodule Logflare.SqlTest do
...
+      query_with_cte = """
+      WITH users_summary AS (
+        SELECT user_id, COUNT(*) as total_events
+        FROM events
+        GROUP BY user_id
+      )
+      SELECT * FROM users_summary WHERE total_events > 10
+      """
+
+      assert Sql.contains_cte?(query_with_cte)
```
nelsonmestevao, line 501: I noticed this function is not used anywhere but I already had the tests. I am not sure if I should just delete it though :thinking:

Ziinc: lets keep it for now, might be useful in future.

PR #2461, opened 2025-06-25
fix: invalid google user iam bug

Ziinc:
This PR fixes a bug in the iam policy provisioning, where invalid (deleted) google users are not removed from the provisioning, causing iam policy creation to fail.

Depends on #2433


lib/logflare/google/resource_manager.ex
```diff
@@ -68,20 +64,48 @@ defmodule Logflare.Google.CloudResourceManager do
...
-                "#{caller}": %{
-                  accounts: Enum.count(members),
-                  response: :error,
-                  response_message: "#{GenUtils.get_tesla_error_message(response)}"
-                }
-              }
-            }
-          }
-        )
+      {:error, %Tesla.Env{} = response} ->
```
SteffenDE, line 67: It may be worth to extract that branch into a helper function.

lib/logflare/google/resource_manager.ex
```diff
@@ -68,20 +64,48 @@ defmodule Logflare.Google.CloudResourceManager do
...
+                  |> TeamUsers.update_team_user(%{valid_google_account: false})
+
+                true ->
+                  :noop
+              end
+
+            if result == :noop do
+              Logger.error(
+                "Could find user #{captured} in the database. Set IAM policy error: #{message}"
+              )
```
SteffenDE, line 92: Previously, you also passed the logflare metadata when logging - is it not needed any more? Otherwise, you could do something like

```elixir
log_meta = fn map ->
  [
    logflare: %{
      google: %{
        cloudresourcemanager: %{
          "#{caller}": Map.merge(%{accounts: Enum.count(members)}, map)
        }
      }
    }
  ]
end
```
and use that when logging (could also be passed to the helper function I suggested above)

```elixir
Logger.error(
  "Could find user #{captured} in the database. Set IAM policy error: #{message}",
  log_meta.(%{response: :error, response_message: message})
)
```

lib/logflare/google/resource_manager.ex
```diff
@@ -138,46 +162,26 @@ defmodule Logflare.Google.CloudResourceManager do
...
-          %Billing.Plan{name: "Free"} -> false
-          _plan -> true
-        end
+    emails =
+      Users.list_users(paying: true, provider: :google)
+      |> Users.preload_valid_google_team_users()
+      |> Enum.flat_map(fn user ->
+        for tu <- user.team.team_users do
+          tu.email
+        end ++ [user.email]
```
SteffenDE, line 171: Is the order of users important? Otherwise, I'd build the list like this to prevent the extra traversal for appending:

```suggestion
        [user.email | for tu <- user.team.team_users do
          tu.email
        end]
```

lib/logflare/users/users.ex
```diff
@@ -117,6 +130,17 @@ defmodule Logflare.Users do
     Repo.preload(user, :team)
   end
 
+  def preload_team_users(user) do
+    Repo.preload(user, team: [:team_users])
+  end
```
SteffenDE, line 135: I don't see this function used anywhere, do you need to keep it?

lib/logflare/google/resource_manager.ex
Ziinc: yea we don't need the metadata for now...maybe a telemetry event would cleaner.

lib/logflare/users/users.ex
Ziinc: removed 😄

PR #2463, opened 2025-06-25
test: improve coverage on SourceController

nelsonmestevao:
- add `:phoenix_test` (that required upgrading `floki` to `0.38.0`) to make it easier to test static views (non-liveviews) to more easily allow upgraging to LiveViews.

- use sigil_p whenever we used `Routes` helper module before in these tests.


test/logflare_web/controllers/source_controller_test.exs
```diff
@@ -417,34 +408,16 @@ defmodule LogflareWeb.SourceControllerTest do
...
-        |> get(source_path(conn, :show, s1.id))
-
-      assert html_response(conn, 403) =~ "403"
-      assert html_response(conn, 403) =~ "Forbidden"
-    end
-  end
-
   describe "create" do
-    setup [:old_setup, :expect_user_plan]
+    setup [:create_plan, :old_setup, :expect_user_plan]
```
Ziinc, line 412: would be good if we can refactor away the :old_setup which makes each test case less straightforward and harder to read

Ziinc (approved):

PR #2464, opened 2025-06-26
perf: add Cachex for Logflare Endpoints 

Ziinc:
This PR adds in Cachex for Logflare Endpoints, allowing us to remove the Query storage in the GenServer, and ensuring that no cross-cluster stale queries are present.
This renames Endpoints.Cache to Endpoints.ResultsCache, and uses the convention Endpoints.Cache module for context caching as per other contexts in the codebase.
This also hashes the params when creating the Endpoint.ResultsCache name, allowing us to reduce the impact of large params on the proc registration tables.

### DIFF
--- a/lib/logflare/endpoints/cache.ex
+++ b/lib/logflare/endpoints/cache.ex
@@ -21,7 +21,7 @@ defmodule Logflare.Endpoints.Cache do
                  if(stats, do: Utils.cache_stats()),
                  Utils.cache_limit(100_000)
                ]
-               |> Enum.filter(& &1),
+               |> Enum.reject(&is_nil/1),
              expiration: Utils.cache_expiration_min()
            ]
          ]}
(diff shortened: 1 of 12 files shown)


lib/logflare/endpoints/results_cache.ex
```diff
@@ -0,0 +1,241 @@
...
+  end
+
+  def endpoints_part(query_id) do
+    part = :erlang.phash2(query_id, System.schedulers_online())
+    "endpoints_#{part}" |> String.to_existing_atom()
+  end
+
+  def endpoints_partitions() do
+    for n <- 0..System.schedulers_online(), do: "endpoints_#{n}" |> String.to_existing_atom()
+  end
```
SteffenDE: This function seems unused?

lib/logflare/endpoints/cache.ex
```diff
@@ -1,231 +1,37 @@
...
+        {Cachex, :start_link,
+         [
+           __MODULE__,
+           [
+             hooks:
+               [
+                 if(stats, do: Utils.cache_stats()),
+                 Utils.cache_limit(100_000)
+               ]
+               |> Enum.filter(& &1),
```
SteffenDE: nit: I usually find `Enum.reject(&is_nil/1)` clearer - this is completely fine of course

SteffenDE (commented):
Looks good to me!

PR #2465, opened 2025-06-26
perf: revert endpoints partition supervisor change

Ziinc:
This reverts the partition supervisor usage for Endpoints.ResultsCache, as it results in duplicate queries getting executed due to n procs getting created by the PartitionSupervisor, where n is number of schedulers

depends on #2464 

todo: add test

### DIFF
--- a/lib/logflare/endpoints/resolver.ex
+++ b/lib/logflare/endpoints/resolver.ex
@@ -31,11 +31,7 @@ defmodule Logflare.Endpoints.Resolver do
         spec = {ResultsCache, {query, params}}
         Logger.debug("Starting up Endpoint.Cache for Endpoint.Query id=#{id}", endpoint_id: id)
 
-        via =
-          {:via, PartitionSupervisor,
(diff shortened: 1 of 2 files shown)


lib/logflare/endpoints/resolver.ex
```diff
@@ -31,11 +31,7 @@ defmodule Logflare.Endpoints.Resolver do
         spec = {ResultsCache, {query, params}}
         Logger.debug("Starting up Endpoint.Cache for Endpoint.Query id=#{id}", endpoint_id: id)
 
-        via =
-          {:via, PartitionSupervisor,
-           {Logflare.Endpoints.ResultsCache.PartitionSupervisor, {id, params}}}
```
chasers, line 36: It's key'd on the `{id, params}` so the same query ID with the same params should be going to the same DynamicSupervisor.

Where are you seeing this not working as expected? 

We need to partition this.

Ziinc: ah i only just realised that the partitioning is for the dynamic supervisor and not the actual spec 🤦

PR #2467, opened 2025-06-27
dynamic pipelines tracing

WolfDan:
Add OTEL tracing for scaling of DynamicPipeline

### DIFF
--- a/lib/logflare/backends/dynamic_pipeline.ex
+++ b/lib/logflare/backends/dynamic_pipeline.ex
@@ -302,6 +302,7 @@ defmodule Logflare.Backends.DynamicPipeline do
             for _ <- 1..diff do
               DynamicPipeline.add_pipeline(state.name)
             end
+            |> do_telemetry(:increment, state, pipeline_count)
 
             new_state
 
@@ -311,6 +312,7 @@ defmodule Logflare.Backends.DynamicPipeline do
             for _pipeline <- 1..diff do
               DynamicPipeline.remove_pipeline(state.name)
             end
+            |> do_telemetry(:decrement, state, pipeline_count)
 
             new_state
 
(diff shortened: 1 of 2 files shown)


test/logflare/backends/dynamic_pipeline_test.exs
```diff
@@ -177,4 +177,64 @@ defmodule Logflare.DynamicPipelineTest do
...
+       %{name: name, pipeline_args: pipeline_args} do
+    ref_increment =
+      :telemetry_test.attach_event_handlers(self(), [
+        [:logflare, :backends, :dynamic_pipeline, :increment]
+      ])
+
+    ref_decrement =
+      :telemetry_test.attach_event_handlers(self(), [
+        [:logflare, :backends, :dynamic_pipeline, :decrement]
+      ])
```
WolfDan, line 191: I added two different events for the dynamic pipeline, 

Based on [Backends](https://github.com/Logflare/logflare/blob/main/lib/logflare/backends.ex#L571) we add a telemetry for each event, which what I was trying to replicate here

But I'm unsure if this is the right way to do it since we count those values on `Logflare.Telemetry`

WolfDan (commented):
Should I add this to `Logflare.Telemetry`? If so what should be the metric aggregation for it?

lib/logflare/backends/dynamic_pipeline.ex
```diff
@@ -327,5 +329,26 @@ defmodule Logflare.Backends.DynamicPipeline do
...
+      :telemetry.execute(
+        [:logflare, :backends, :dynamic_pipeline, action],
+        %{
+          error_count: error_count,
+          success_count: length(actions) - error_count,
+          from_pipeline_count: from_pipeline_count
+        },
+        %{
+          source_id: state.pipeline_args[:source].id,
+          source_token: state.pipeline_args[:source].token
```
Ziinc: lets also include backend information to this, namely:
backend_id
backend_token
backend_type

test/logflare/backends/dynamic_pipeline_test.exs
Ziinc: yes this is correct, please also add metrics for Logflare.Telemetry and verify that they get shipped via otel (you can point logflare to itself in dev mode, you'd need to dev with a BQ backend to verify it locally.

PR #2470, opened 2025-06-27
test: improve test coverage for Api.EndpointController

nelsonmestevao:
This commit adds missing tests for the put operation (in update endpoint). It also validates the response against the OpenAPI schema format by introducing a new helper function `assert_schema/2` to avoid boilerplate (essentially repeating the api_spec function call).

In the `LogController`, added operation missing definitions to avoid all the warnings from the OpenApiSpex validator.

Start using the `sigil_p` in the testing routes to have a better DX when making a typo.

Also fixed a repeated typo on "endpoint" word.

### DIFF
--- a/lib/logflare_web/controllers/log_controller.ex
+++ b/lib/logflare_web/controllers/log_controller.ex
@@ -85,6 +85,8 @@ defmodule LogflareWeb.LogController do
     |> handle(conn)
   end
 
+  operation :cloudflare, false
+
   def cloudflare(%{assigns: %{source: source}} = conn, %{"batch" => batch}) when is_list(batch) do
     batch
     |> Processor.ingest(Logs.Raw, source)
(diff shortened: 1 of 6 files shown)


lib/logflare_web/controllers/log_controller.ex
nelsonmestevao, line 1: This was necessary, otherwise OpenApiSpex would pollute the console with warnings about the missing spec operations.

lib/logflare_web/open_api_schemas.ex
```diff
@@ -60,17 +60,17 @@ defmodule LogflareWeb.OpenApiSchemas do
...
-      source_mapping: %Schema{type: :object},
-      sandboxable: %Schema{type: :boolean},
+      source_mapping: %Schema{type: :object, nullable: true},
+      sandboxable: %Schema{type: :boolean, nullable: true},
       cache_duration_seconds: %Schema{type: :integer},
       proactive_requerying_seconds: %Schema{type: :integer},
       max_limit: %Schema{type: :integer},
-      enable_auth: %Schema{type: :boolean},
-      inserted_at: %Schema{type: :string, format: :"date-time"},
-      updated_at: %Schema{type: :string, format: :"date-time"}
```
nelsonmestevao, line 73: These keys are not in the JSON derive protocol implementation. So I removed them from the spec as well.

Ziinc (approved):

PR #2472, opened 2025-06-27
test: add missing test for put request in TeamController and validate all schema responses

nelsonmestevao:
Following up on the work of PR https://github.com/Logflare/logflare/pull/2470, we can now validate the response schemas according to the OpenAPI spec making sure a consistent behavior is enforced for all common operations (not found, unprocessable entity, etc).

### DIFF
--- a/lib/logflare_web/open_api_schemas.ex
+++ b/lib/logflare_web/open_api_schemas.ex
@@ -164,19 +164,19 @@ defmodule LogflareWeb.OpenApiSchemas do
       api_key: %Schema{type: :string},
       email_preferred: %Schema{type: :string},
       name: %Schema{type: :string},
-      image: %Schema{type: :string},
+      image: %Schema{type: :string, nullable: true},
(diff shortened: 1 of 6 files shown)


lib/logflare_web/open_api_schemas.ex
```diff
@@ -164,19 +164,19 @@ defmodule LogflareWeb.OpenApiSchemas do
...
+      bigquery_dataset_id: %Schema{type: :string, nullable: true},
       api_quota: %Schema{type: :integer},
-      company: %Schema{type: :string},
+      company: %Schema{type: :string, nullable: true},
       token: %Schema{type: :string}
     }
     use LogflareWeb.OpenApi,
       properties: @properties,
-      required: [:email, :provider, :token, :provider_uid, :api_key]
+      required: [:email, :provider, :token, :api_key]
```
nelsonmestevao, line 179: We are not even deriving this key on the JSON encoding, that it is way I am removing it.

PR #2476, opened 2025-06-30
fix: remove dbgs and replace them with asserts

nelsonmestevao:


Ziinc:
@nelsonmestevao can we add this credo check please https://hexdocs.pm/credo/Credo.Check.Warning.Dbg.html 🙇

Ziinc (approved):

PR #2478, opened 2025-06-30
test: add testing for sanitize_backend_config

nelsonmestevao:


Ziinc (approved):
jfyi we should be redacting via Adaptor behaviour, so this will be refactored off soon.

PR #2479, opened 2025-07-01
feat: add new telemetry metrics

WolfDan:


lib/logflare/backends/dynamic_pipeline.ex
```diff
@@ -294,6 +294,20 @@ defmodule Logflare.Backends.DynamicPipeline do
       pipelines = DynamicPipeline.list_pipelines(state.name)
       pipeline_count = Enum.count(pipelines)
 
+      :telemetry.execute(
+        [:logflare, :backends, :dynamic_pipeline],
+        %{
+          pipeline_count: pipeline_count
+        },
+        %{
+          source_id: state.pipeline_args[:source].id,
```
Ziinc, line 303: lets use this opportunity to refactor the pipeline_args to only hold the bare minimum values, and as storing the source/backend structs in the args results in higher memory and increased risk of stale values.

Ziinc: We can use Backends.Cache and Sources.Cache to fetch the structs when needed, so we can just store the id/token/type in the args, values that we know won't change.

Ziinc: lets do it in a follow up PR

PR #2485, opened 2025-07-01
test: improve coverage in Backend adaptors

nelsonmestevao:
Worked on improving coverage for missing cases:

- Datadog Adaptor
- Postgres Adaptor
- Slack Adaptor


lib/logflare/backends/adaptor/datadog_adaptor.ex
```diff
@@ -17,11 +17,13 @@ defmodule Logflare.Backends.Adaptor.DatadogAdaptor do
...
     "US1-FED" => "https://http-intake.logs.ddog-gov.com/api/v2/logs"
   }
+  @regions Map.keys(@api_url_mapping)
 
-  def api_url_mapping, do: @api_url_mapping
+  def regions, do: @regions
 
   typedstruct enforce: true do
     field(:api_key, String.t())
+    field(:region, String.t())
```
nelsonmestevao: I don't see any of these structs being used, but in this one the region key was missing.

lib/logflare_web.ex
```diff
@@ -90,10 +90,6 @@ defmodule LogflareWeb do
 
       use Phoenix.LiveView, opts
 
-      # declare endpoint and router for Phoenix.VerifiedRoutes
-      @endpoint LogflareWeb.Endpoint
-      @router LogflareWeb.Router
```
nelsonmestevao, line 95: This should have been removed when the `path_helpers` was introduced 🙏

test/logflare/backends/adaptor/postgres_adaptor_test.exs
```diff
@@ -49,13 +50,12 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptorTest do
...
       assert {:ok, _} = Backends.ingest_logs([log_event], source)
 
-      # TODO: replace with a timeout retry func
-      :timer.sleep(2_500)
-
       # query by Ecto.Query
       query = from(l in PostgresAdaptor.table_name(source), select: l.body)
 
-      assert {:ok, [%{"test" => "data"}]} = PostgresAdaptor.execute_query(backend, query)
+      TestUtils.retry_assert(fn ->
```
nelsonmestevao, line 56: To allow remove the sleep and TODO 💪

test/logflare/backends/adaptor/datadog_adaptor_test.exs
```diff
@@ -17,8 +18,19 @@ defmodule Logflare.Backends.Adaptor.DatadogAdaptorTest do
     :ok
   end
 
+  test "regions/1 return valid datadog regions list" do
```
Ziinc: looks like testing constants, necessary?
if its just for coverage, testing for strings is sufficient.

test/logflare/backends/adaptor/postgres_adaptor_test.exs
```diff
@@ -204,20 +203,37 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptorTest do
...
-  defp cast_config(attrs) do
+  defp cast_and_validate_config(attrs) when is_list(attrs) do
     attrs
     |> Map.new()
-    |> PostgresAdaptor.cast_config()
-    |> PostgresAdaptor.validate_config()
+    |> cast_and_validate_config()
+  end
+
+  defp cast_and_validate_config(attrs) when is_map(attrs) do
```
Ziinc: seems quite unnecessary to have two function heads.

test/logflare/backends/adaptor/postgres_adaptor_test.exs
```diff
@@ -204,20 +203,37 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptorTest do
...
 
-    refute cast_config(url: "://localhost:5432").valid?
-    refute cast_config(url: "//localhost:5432").valid?
-    refute cast_config(url: "/localhost:5432").valid?
-    refute cast_config(url: "postgres//localhost:5432").valid?
+    test "requires either url or hostname" do
+      assert errors_on(cast_and_validate_config(url: "postgresql://localhost:5432")) == %{}
+      assert errors_on(cast_and_validate_config(hostname: "localhost"))
+
+      assert errors_on(cast_and_validate_config(%{})) == %{
```
Ziinc: this could just be an empty list and we can then remove the 2nd function head

test/logflare/backends/adaptor/slack_adaptor_test.exs
```diff
@@ -0,0 +1,265 @@
...
+               ]
+
+        {:ok, %Tesla.Env{}}
+      end)
+
+      assert {:ok, _} = SlackAdaptor.send_message(alert, [])
+
+      Logflare.Backends.Adaptor.SlackAdaptor.Client
+      |> expect(:send, fn _url, body ->
+        assert body.blocks == [
```
Ziinc: lets send the body to the test proc to perform the assertions in the test case, otherwise errors in the separate proc won't cause the test to fail.

test/logflare/backends/adaptor/slack_adaptor_test.exs
```diff
@@ -0,0 +1,265 @@
...
+                 %{"another title" => "another body"}
+               ])
+    end
+
+    test "can work with any payload" do
+      test_url = "http://example.com/test"
+
+      Logflare.Backends.Adaptor.SlackAdaptor.Client
+      |> expect(:send, fn url, body ->
+        assert url == test_url
```
Ziinc: see above.

test/logflare/backends/adaptor/slack_adaptor_test.exs
```diff
@@ -0,0 +1,265 @@
...
+
+  doctest Logflare.Backends.Adaptor.SlackAdaptor, import: true
+
+  describe "send_message/1" do
+    test "can work with alert query" do
+      alert = insert(:alert)
+
+      Logflare.Backends.Adaptor.SlackAdaptor.Client
+      |> expect(:send, fn _url, body ->
+        assert body.blocks == [
```
Ziinc: see below comment

lib/logflare/backends/adaptor/datadog_adaptor.ex
Ziinc: we can remove structs that aren't used.

Ziinc: originally was planning to use this for type validations in the ecto validations, but seems unnecessary.

Ziinc (changes requested):

lib/logflare/backends/adaptor/datadog_adaptor.ex
nelsonmestevao: Removed all structs I noticed not being used in last commit (fe2a141330816fe55fbbbc61eb560ac2de48d7f8) :pray:

test/logflare/backends/adaptor/datadog_adaptor_test.exs
nelsonmestevao: These are the list of options a user can choose, I found it easier to make sure it sees the correct list by testing the function. Not sure if you agree it needs to be tested. 

I am going to remove the test and if you find important to make sure the user sees the correct list, I will do it in the LiveView test :pray:

test/logflare/backends/adaptor/postgres_adaptor_test.exs
nelsonmestevao: I removed it in last pus :pray:  If you agree I can even make it simpler by using maps in the tests instead a keyword list, that way I avoid the Map.new call, WDYT?

test/logflare/backends/adaptor/slack_adaptor_test.exs
nelsonmestevao: I am not sure what you mean, if there's something wrong in the blocks this test fails :thinking: 

Example of me making the test fail on purpose :point_down: Sorry if I miss understood you :pray: 
 
image

test/logflare/backends/adaptor/datadog_adaptor_test.exs
nelsonmestevao: I did it in this test https://github.com/Logflare/logflare/pull/2491/files#diff-601e4c12d18c1a76bf168f94f21328bcfd44b7b24ee5d10529ba7db4875f11c7R193 :pray:

Ziinc: imo would be meaningless to test these and adds additional unnecessary code, esp since the values are external and not under our control

test/logflare/backends/adaptor/slack_adaptor_test.exs
Ziinc: we shouldn't assume that the client would be called in the same process of the test case, esp if things get refactored down the road, so it would be more conservative to send the body to the test case pid.

```elixir
pid = self()
Client
|> expect(:send , fn ... -> 
  send(pid, data) 
end)

assert_receive %{} = data
# do assertions
``` 

this would make it more future proof

Ziinc: we use this pattern quite extensively in other tests

nelsonmestevao: Got it, done in last pushed commit :pray: Thank you :rocket:

Ziinc (approved):

PR #2487, opened 2025-07-02
feat: add display timezone switcher, remove local timezone toggle

msmithstubbs:
Finishing up https://github.com/Logflare/logflare/pull/2092

Allows temporary timezone switching with an option to permanently save the preference when it differs from the user's stored timezone.

https://github.com/user-attachments/assets/630ed056-5883-4961-b548-1f63bb1729a3

* Adds a display timezone dropdown for selecting timezone logs are displayed in
* Dropdown will reflect a timezone set by `tz` query param
* If selected timezone is different to users stored timezone preference a checkbox is provided to store the selection.
* Checkbox is hidden if selected timezone matches the user preference timzone
* Button to jump to UTC
* Removes the timezone dialog from search heading.

## Notes:
1. If updating the user preference timezone fails a flash message is displayed, however no current validations would actually trigger this (timezone field is always provided)
2. I've left **display timezone** as the label from the original PR, but maybe simply **timezone** is sufficient?
3. Changing the timezone will update query params and trigger a new search. Saving the timezone won't.


lib/logflare/users/user_preferences.ex
```diff
@@ -5,11 +5,12 @@ defmodule Logflare.Users.UserPreferences do
 
   typed_embedded_schema do
     field :timezone, :string
+    field :display_timezone, :string
```
Ziinc: so when does the search tz and the display tz differ?
if the user chooses to "remember" current timezone, wouldn't that apply for their searches? when would the search timezone and the display timezone differ?
we should only need to store one timezone value (the value that the user wants us to remember), and then any changes to the dropdown only adjusts local state and does not immediately store it unless the user explicitly asks for it (with the remember)

msmithstubbs: Totally agree - storing only one timezone would make more sense!

The instructions were to add test coverage to PR 2092, and it already had separate search and display timezones so I proceeded with that. Rationale for tracking two timezones is described [here](https://github.com/Logflare/logflare/pull/2092#issue-2343900097): 

> It forces a differentiation between "search timezone" and "display timezone", where display timezone only changes the displayed timezone, whereas search timezone will affect the search queries that the user will query in.

I don't know the reasoning but figured there was a good but non-obvious use case?

If I were to do this from first principles of "a user needs to be able to quickly switch between timezones" I'd suggest:
1. Use the existing `user.preferences.timezone` field as you suggest. I don't think it's used elsewhere.
2. Keep dropdown with the timezones. Considering listing them City name first and sorted alphabetically to make it easier to read.
3. Remove the 'remember' checkbox and just persist the timezone whenever the user changes it. Timezone is simply whatever the user last used. 
4. Keep the UTC button, and consider adding one for 'local' timezone (eg. whatever timezone the browser uses).

I'm not sure what the use case here is though so that may not cover off the requirements.

Ziinc: yup, the display timezone should not be stored, because it could come from the query param as well when sharing links.
😆 funnily enough, pts 1-4 were actually the original version before we added in query params timezones.

But due to the query param timezone handling case, we need to factor for the following situation:
When user1 shares link to user2, it will contain user1's display tz, and user2 tz would then load the page using user1's display tz (not user2's preferred tz). Note that user1's display tz could be different from their saved default tz.


Some other considerations is that we want to minimize the db and server impact as much as possible, so causing a db txn on each pref change would not be great:
When user1 changes their viewing tz to utc for example, it could be for a temporary quick check, causing a db update would be unnecessary as the user might not want it to persist (e.g. if it auto-persists, they would need to change back their tz when viewing another page), causing another db txn.

msmithstubbs: Just to clarify:

1. If 'Remember' checkbox is changed to `true` the timezone in the dropdown is stored in `user.preferences.timezone`. 
2. If 'Remember' checkbox is changed to `false` nothing is updated.
3. When the user changes timezone dropdown:
  - the checkbox is set to `false`. (So the user can do a quick of a different time zone and switch back), and
  - The timezone is the URL is updated (I think?)
5. If the user selects the same timezone as stored in their preferences the checkbox is set to `true` to indicate it is stored.
6. The timezone dropdown defaults to the tz in the URL param or user.preferences.timezone.

Is that right? Have I missed anything?

Ziinc: yes this sounds good 👍

msmithstubbs: Looks like we'll have to rethink the checkbox. Per the discussion above:
1. it currently shows as checked (`true`) when the selected timezone matches the user preferences stored timezone.
2. If a user unchecks the checkbox (`false`), we view is updated, but since they still have the same timezone selected as their stored pref the checkbox is rendered as checked. 


https://github.com/user-attachments/assets/9cd45da2-d82f-4753-a030-ea47d7c64a5a



We previously discussed making it disappear checked, making it behave more like a button (_click to store the timezone..._). Should I go back to this?

Ziinc: > We previously discussed making it disappear checked, making it behave more like a button 

yes this will work as well.
i think it will also be a bit more intuitive

PR #2109, opened 2024-06-18
fix: refactors and reworks the LogEventLive component and viewer

Ziinc:

PR #2492, opened 2025-07-03
fix: refactor log event live and component

msmithstubbs:
Picking up from https://github.com/Logflare/logflare/pull/2109

Fixed compilation errors and got the modal working. Timestamp is passed through. Still need to finish up tests but wanted to check this functionality is in the right direction.

https://github.com/user-attachments/assets/3fc6b4c9-dc7a-48c1-adeb-08d87affeefe

The original PR added a formatted timestamp to the permalink link - should it say? 

https://github.com/Logflare/logflare/pull/2109/files#diff-e60a006782f6833fe1fd1b41951363b779be33cd8b80090bb4b2699b25a11c8cR27

Otherwise does this cover the functional requirements? If so I'll clean it up and add tests.


lib/logflare/utils.ex
```diff
@@ -80,6 +80,29 @@ defmodule Logflare.Utils do
   def stringify(v) when is_binary(v), do: v
   def stringify(v), do: inspect(v)
 
+  @doc """
```
Ziinc, line 73: lets merge these with the function heads from :72

lib/logflare_web/live/log_event_live.ex
```diff
@@ -1,4 +1,4 @@
-defmodule LogflareWeb.LogEventLive.Show do
+defmodule LogflareWeb.LogEventLive do
   @moduledoc """
   Handles all user interactions with the source logs search
```
Ziinc:
```suggestion
  Handles all user interactions with a single event
```

lib/logflare_web/live/log_event_live.ex
```diff
@@ -13,7 +13,9 @@ defmodule LogflareWeb.LogEventLive.Show do
   require Logger
 
   def mount(%{"source_id" => source_id} = params, session, socket) do
-    source = Sources.get_source_for_lv_param(source_id)
+    source_id = String.to_integer(source_id)
+    ts = Map.get(params, "timestamp")
+    source = Sources.get(source_id)
```
Ziinc:
```suggestion
    ts = Map.get(params, "timestamp")
    source = Sources.get(source_id)
```

lib/logflare_web/live/log_event_live.ex
```diff
@@ -30,25 +32,23 @@ defmodule LogflareWeb.LogEventLive.Show do
...
+      |> assign(:log_event_id, params["uuid"])
+      |> case do
+        socket when ts != nil ->
+          {:ok, dt, _} = DateTime.from_iso8601(ts)
+          assign(socket, :timestamp, dt)
+
+        socket ->
+          socket
       end
+      |> assign(:log_event, LogEvents.Cache.get!(token, {"uuid", params["uuid"]}))
```
Ziinc: bang version would cause this to error if empty

lib/logflare_web/live/log_event_live/search_log_event_viewer_component.ex
```diff
@@ -7,17 +7,21 @@ defmodule LogflareWeb.Search.LogEventViewerComponent do
...
 
   @impl true
-  def update(%{error: _} = assigns, socket) do
+  def update(%{log_event: nil} = assigns, socket) do
     socket =
-      assign(socket, :error, "Log event not found!")
+      socket
       |> assign_defaults(assigns)
+      |> start_async(:load, fn ->
+        load_event(assigns)
```
Ziinc: this results in some unnecessary data copying

lib/logflare_web/live/log_event_live.ex
Ziinc: Lets start using Sources.Cache for search related ui

lib/logflare_web/live/search_live/logs_search_lv_utils.ex
```diff
@@ -34,4 +34,11 @@ defmodule LogflareWeb.SearchLV.Utils do
   def pid_source_to_string(pid, source) do
     "#{pid_to_string(pid)} for #{source.token}"
   end
+
+  def iso_timestamp(timestamp) do
```
Ziinc: might want to move this to Logflare.Utils since it is quite generic and useful

lib/logflare_web/live/log_event_live.ex
Ziinc: we can get rid of the bang version in the LogEvents.Cache i think

lib/logflare/utils.ex
```diff
@@ -80,6 +80,29 @@ defmodule Logflare.Utils do
...
+    iex> stringify(122)
+    "122"
+    iex> stringify("something")
+    "something"
+    iex> stringify(%{})
+    "%{}"
+    iex> stringify([])
+    "[]"
+  """
+  def stringify(v) when is_atom(v), do: Atom.to_string(v)
```
amokan: Probably overkill, but worth noting that there is a guard I added recently in `Logflare.Utils.Guards` called `is_atom_value` which will exclude atoms such as `nil` or boolean values which can sometimes leak through to general utility functions like this 🤷‍♂️ 

Up to @Ziinc if that level of concern is needed here though.

Ziinc: yea would be a good addition as well 🔥

test/logflare_web/live_views/log_event_live_test.exs
```diff
@@ -0,0 +1,101 @@
...
+    expect(GoogleApi.BigQuery.V2.Api.Jobs, :bigquery_jobs_query, 1, fn _conn, _proj_id, opts ->
+      send(pid, {:query, ref, opts[:body].query})
+
+      {:ok,
+       TestUtils.gen_bq_response([%{"id" => le.id, "event_message" => le.body["event_message"]}])}
+    end)
+
+    {:ok, view, _html} = live(conn, ~p"/sources/#{source.id}/event?#{%{uuid: le.id}}")
+
+    :timer.sleep(1500)
```
Ziinc, line 36: lets use retry_assert so that the test doesn't take too long as this sets the min time

test/logflare_web/live_views/log_event_live_test.exs
```diff
@@ -0,0 +1,101 @@
...
+       TestUtils.gen_bq_response([%{"id" => le.id, "event_message" => le.body["event_message"]}])}
+    end)
+
+    {:ok, _view, _html} =
+      live(
+        conn,
+        ~p"/sources/#{source.id}/event?#{%{timestamp: "2024-01-10T20:13:03Z", uuid: le.id}}"
+      )
+
+    :timer.sleep(1500)
```
Ziinc, line 63: same as above

test/logflare_web/live_views/log_event_live_test.exs
```diff
@@ -0,0 +1,101 @@
...
+
+    expect(GoogleApi.BigQuery.V2.Api.Jobs, :bigquery_jobs_query, 2, fn _conn, _proj_id, _opts ->
+      {:error, "some error"}
+    end)
+
+    logs =
+      capture_log(fn ->
+        {:ok, view, _html} = live(conn, ~p"/sources/#{source.id}/event?#{%{uuid: le.id}}")
+
+        :timer.sleep(400)
```
Ziinc, line 90: we can remove this since retry_assert already has some sleeping in built

PR #2496, opened 2025-07-03
chore: sql module refactoring

amokan:
This breaks the `Logflare.Sql` module apart a bit in a first pass of refactoring to reduce the cognitive load (reduces the LOC by over 1400) and paves the way for some dialect specific handling within smaller dedicated modules.

I've kept the `Logflare.Sql` tests as-is to avoid any regressions while also adding new tests for the supporting modules that focus on their specific use cases.


lib/logflare/sql/dialect_transformer/bigquery.ex
```diff
@@ -0,0 +1,81 @@
...
+      |> Atom.to_string()
+      |> String.replace("-", "_")
+
+    # byob bq
+    project_id =
+      if is_nil(data.user_project_id), do: data.logflare_project_id, else: data.user_project_id
+
+    # byob bq
+    dataset_id =
+      if is_nil(data.user_dataset_id), do: data.logflare_dataset_id, else: data.user_dataset_id
```
Ziinc, line 33: what if we did the byob check when building the data instead? and then store only `project_id` and `dataset_id` in the data map?
might be cleaner and we won't "leak" the byob concept into other functions.

lib/logflare/sql/dialect_translation.ex
```diff
@@ -0,0 +1,1135 @@
...
+    end
+  end
+
+  defp traverse_convert_identifiers(
+         {"CompoundIdentifier" = k, [%{"value" => head_val}, tail] = v},
+         data
+       ) do
+    cond do
+      # Use match?/2 there to check if list has at least 2 values in it. It is
+      # faster than `langth(list) > 2` as it do not need to traverse whole list
```
Ziinc:
```suggestion
      # faster than `length(list) > 2` as it do not need to traverse whole list
```

test/logflare/sql/dialect_transformer/clickhouse_test.exs
```diff
@@ -0,0 +1,66 @@
...
+
+      result = Clickhouse.transform_source_name("source_two", data)
+      expected = ClickhouseAdaptor.clickhouse_ingest_table_name(source2)
+
+      assert result == expected
+    end
+
+    test "handles source with special characters in token" do
+      user = insert(:user)
+      source = insert(:source, name: "special_source", user: user)
```
Ziinc, line 53: we can consider building these structs instead of inserting so that we can make the tests async (since there is nothing repo related for the sql stuff)
low priority

test/logflare/sql_test.exs
```diff
@@ -346,6 +346,77 @@ defmodule Logflare.SqlTest do
...
+    test "parser can handle tuple definitions" do
+      user = insert(:user)
+
+      for input <- [
+            "select tuple(1,2,3) as foo",
+            "select tuple(\'abc\')",
+            "select tuple(1, t.str_col) from values(\'str_col String\', (\'hello\'), (\'world\'), (\'test\')) t",
+            "SELECT tuple(t.str_col) AS abc from values(\'str_col String\', (\'hello\'), (\'world\'), (\'test\')) t"
+          ] do
+        assert {:ok, _v2} = Sql.transform(:ch_sql, input, user)
```
Ziinc, line 359: something that i've noticed is that sometimes the parser can silently ignore some syntax without panicking, might want to make these tests a bit more resilient by making it spit the sql back out again from the ast

Ziinc: if these panic on parsing then its fine

PR #2497, opened 2025-07-03
fix: alerting improvements

Ziinc:
This PR includes the following improvements for alerts:
- moves to using GenSingleton supervision tree
- adds in proper :erpc usage for managing singleton calls
- adds in tracing for alerts firing for debugging of alert duplication
- adds in logging for AlertsScheduler init
- fixes redirection for deleted alerts


lib/logflare_web/live/alerts/alerts_live.ex
```diff
@@ -61,19 +61,31 @@ defmodule LogflareWeb.AlertsLive do
...
     socket =
-      if socket.assigns.live_action == :edit do
+      if socket.assigns.live_action == :edit and alert != nil do
         assign(socket, :changeset, Alerting.change_alert_query(alert))
       else
         assign(socket, :changeset, nil)
       end
 
-    {:noreply, socket}
+    if !!alert_id and !alert do
```
jonatanklosko: You can make it a bit more explicit, which may be more readable than the negations:

```suggestion
    if alert_id != nil and alert == nil do
```

lib/logflare/alerting.ex
```diff
@@ -150,15 +155,7 @@ defmodule Logflare.Alerting do
...
-    AlertsScheduler.jobs()
-    |> Enum.find_value(fn {_ref,
-                           %Quantum.Job{
-                             task: {_module, _function, [%AlertQuery{id: query_id}, _]}
-                           } = job} ->
-      if query_id == id do
-        job
-      end
-    end)
+    remote_call_scheduler(:find_job, ["#{id}"])
```
jonatanklosko:
```suggestion
    remote_call_scheduler(:find_job, [Integer.to_string(id)])
```

(doesn't involve protocol dispatch and is more explicit)

lib/logflare/alerting.ex
```diff
@@ -192,6 +191,33 @@ defmodule Logflare.Alerting do
...
+    if alert_query do
+      upsert_alert_job(alert_query)
+    else
+      # alert query does not exist, maybe remove from scheduler
+      job = get_alert_job(alert_id)
+
+      if job do
+        delete_alert_job(alert_id)
+      end
+    end
```
jonatanklosko: This is doing multiple `:erpc` calls, passing data back and forth. For example, `get_alert_job` sends the whole job struct from the scheduler node to the caller, just to do another `:erpc` call, sending the alert id back.

Ideally you want a single `:erpc` call that does all the logic at once, on the scheduler node. In case of `sync_alert_job` we only need to send the `alert_id` to the scheduler node, and get `:ok | {:error, :not_found}` back, so there is little sent over the wire.

One way would be something like this:

```elixir
def sync_alert_job(alert_id) when is_integer(alert_id) do
  on_scheduler_node(fn ->
    # We can call AlertsScheduler here directly.
  end)
end
```

What do you think?

lib/logflare/alerting.ex
```diff
@@ -192,6 +191,33 @@ defmodule Logflare.Alerting do
     end)
   end
 
+  def sync_alert_jobs do
+    init_alert_jobs()
+    |> Enum.each(fn job ->
+      remote_call_scheduler(:add_job, [job])
+    end)
+  end
```
jonatanklosko: Might be worth noting that this is used for testing specifically.

Btw. should this call `AlertScheduler.delete_all_jobs` beforehand, otherwise removed jobs are kept around?

test/logflare_web/live_views/alerts_live_test.exs
```diff
@@ -17,6 +17,7 @@ defmodule LogflareWeb.AlertsLiveTest do
     insert(:plan, name: "Free")
     user = insert(:user)
     conn = login_user(conn, user)
+    start_supervised!(Logflare.Alerting.Supervisor)
```
jonatanklosko, line 20: Starting the global process in multiple test suites prevents them from running `async: true`. Currently they are not async anyway, but that's something you generally want to do as much as possible and designing for it is good.

I see you have a separate application supervision tree for test env and you don't include `Logflare.Alerting.Supervisor`. Is there a reason not to include it though? It's generally best to keep the application supervision tree as close as possible in test and prod.

jonatanklosko (commented):
Hey @Ziinc, I dropped a few comments :)

lib/logflare/alerting.ex
Ziinc: that's a nice one, i'll implement that 👍

lib/logflare/alerting.ex
```diff
@@ -143,22 +148,14 @@ defmodule Logflare.Alerting do
...
-    AlertsScheduler.jobs()
-    |> Enum.find_value(fn {_ref,
-                           %Quantum.Job{
-                             task: {_module, _function, [%AlertQuery{id: query_id}, _]}
-                           } = job} ->
-      if query_id == id do
-        job
-      end
-    end)
+    remote_call_scheduler(:find_job, [Integer.to_string(id)])
```
jonatanklosko: Now that you have `remote_call_scheduler`, I would just use that instead of `on_scheduler_node` for consistency, so it's a single approach. Also, this way, if someone needs to change those functions in the future, they don't end up doing multiple RPCs.

test/logflare_web/live_views/alerts_live_test.exs
Ziinc: would prefer to not start the alerting supervision tree during tests as the scheduler is stateful and the teardown of tests becomes quite important to ensure state is reset between tests, whereas starting it in the setup always ensures a clean slate. Can result in gotchas if one is not familiar with the necessary teardowns

Ziinc: we could rely on the macro to do it of course... as we do for some of the caches already
https://github.com/Logflare/logflare/blob/bd8e7e6fdd50128d95cac973df1affd95f28afc2/test/support/data_case.ex#L38

not fully sold on the benefits of adding it to the test tree though 😄

jonatanklosko: > not fully sold on the benefits of adding it to the test tree though 😄

Fair, ultimately it's up to you! The point I would stress is inability to `async: true`.

> would prefer to not start the alerting supervision tree during tests as the scheduler is stateful and the teardown of tests becomes quite important to ensure state is reset between tests

Statefulness is not necessarily a problem, it just needs to be accounted for in assertions. For example, instead of strict comparisons `list_alert_jobs() == [...]`, you would assert facts related to the given test, such as `new_job in list_alert_jobs()`.

PR #2499, opened 2025-07-04
test: improve coverage on Logflare.Sources.Rules

nelsonmestevao:
- Add missing tests for Logflare.Sources.Rules
- Add missing tests for Logflare.Sources.Rule

These modules didn't have the "Sources" prefix, I took the opportunity to make it consistent.

### DIFF
--- a/lib/logflare/backends/backend.ex
+++ b/lib/logflare/backends/backend.ex
@@ -8,7 +8,7 @@ defmodule Logflare.Backends.Backend do
   alias Logflare.Backends.Backend
   alias Logflare.Source
   alias Logflare.User
-  alias Logflare.Rule
+  alias Logflare.Sources.Rule
   alias Logflare.Alerting.AlertQuery
 
   @adaptor_mapping %{
--- a/lib/logflare/backends/source_sup.ex
+++ b/lib/logflare/backends/source_sup.ex
@@ -18,7 +18,7 @@ defmodule Logflare.Backends.SourceSup do
   alias Logflare.Logs.SearchQueryExecutor
   alias Logflare.Backends.RecentEventsTouch
   alias Logflare.Backends.RecentInsertsBroadcaster
-  alias Logflare.Rule
+  alias Logflare.Sources.Rule
   alias Logflare.Sources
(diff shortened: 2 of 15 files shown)


lib/logflare/rules/rule.ex
```diff
@@ -37,12 +37,12 @@ defmodule Logflare.Rule do
   end
 
   defp maybe_parse_lql(changeset) do
-    qs = get_field(changeset, :lql_string)
+    {:ok, rules} =
```
Ziinc, line 40: this would error if the lql string can't be parsed

test/logflare/rules_test.exs
```diff
@@ -34,17 +35,19 @@ defmodule Logflare.RulesTest do
                lql_string: "a:testing"
              })
 
-    # delete the rule
-    assert {:ok, _rule} = Rules.delete_rule(rule)
+    Rules.delete_rule!(rule.id)
```
Ziinc, line 38: would prefer to refactor away the bang version as it is only used in one location and is not really necessary.

lib/logflare/rules/rule.ex
Ziinc: would be the case if say we make a breaking change in the lql syntax.

test/logflare/rules_test.exs
Ziinc: can do in separate PR.

nelsonmestevao: Okay, I will do it then (in a follow up) :pray:

lib/logflare/rules/rule.ex
nelsonmestevao: I changed it because I could not make the Parser fail, at the moment, it will either raise or always return `{:ok,}`. Should I take into account future changes in the response type?

What do you think I should do?

Ziinc: 👍  if that's the case it should be fine

test/logflare/rules_test.exs
nelsonmestevao: For future reference, this is being done in https://github.com/Logflare/logflare/pull/2507.

PR #2500, opened 2025-07-05
feat: add relationship to backend in endpoint query for inferring language

amokan:
Minimal changes needed to setup relationship between an endpoint `Query` and a `Backend`. 

If the language is not set and a backend is when creating a new query - the system will infer the language, with some additional checks around the single tenant/supabase mode flags.

The intent is to pave the way for handling the query language in a more dynamic way. Future PRs will leverage this for supporting clickhouse SQL dialect for endpoint queries.


priv/repo/migrations/20250705172040_populate_backend_id_for_existing_queries.exs
```diff
@@ -0,0 +1,24 @@
+defmodule Logflare.Repo.Migrations.PopulateBackendIdForExistingQueries do
+  use Ecto.Migration
+
+  def up do
+    # Only update where a query.user has exactly one backend matching the query.language
+    # Others should probably be worked out manually
+    execute """
```
Ziinc: hmm just thinking if we even need to set the value via migration
if backend_id is nil, then we can assume that it is the system's default backend.

lib/logflare/endpoints/query.ex
```diff
@@ -27,7 +34,7 @@ defmodule Logflare.Endpoints.Query do
     field(:name, :string)
     field(:query, :string)
     field(:description, :string)
-    field(:language, Ecto.Enum, values: [:bq_sql, :pg_sql, :lql], default: :bq_sql)
+    field(:language, Ecto.Enum, values: [:bq_sql, :lql, :pg_sql], default: :bq_sql)
```
Ziinc, line 39: should add in ch_sql

lib/logflare/endpoints/query.ex
```diff
@@ -117,30 +131,55 @@ defmodule Logflare.Endpoints.Query do
...
+            put_change(changeset, :language, language)
+        end
+
+      _ ->
+        changeset
+    end
+  end
+
+  defp map_backend_to_language(%{type: :bigquery}), do: :bq_sql
+  defp map_backend_to_language(%{type: :postgres}), do: :bq_sql
```
Ziinc: this should only be the case when in single tenant mode with pg backend and with on supabase mode.
the BQ -> PG translation code only kicks in for this special case.

priv/repo/migrations/20250705171800_add_backend_id_to_endpoint_queries.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Logflare.Repo.Migrations.AddBackendIdToEndpointQueries do
+  use Ecto.Migration
+
+  def change do
+    alter table(:endpoint_queries) do
+      add :backend_id, references(:backends, on_delete: :delete_all), null: true
```
Ziinc: shouldn't delete the endpoint if the backend is deleted, should nilify or raise

priv/repo/migrations/20250705172040_populate_backend_id_for_existing_queries.exs
Ziinc: just wondering if this makes things extra rigid, esp with the seeding logic for supabase mode

lib/logflare/endpoints/query.ex
amokan: Thank you! This seems to be the gap that I was confused about.

priv/repo/migrations/20250705172040_populate_backend_id_for_existing_queries.exs
amokan: Yeah. Wasn't even sure I'd include this migration within the PR, TBH.

I'd prefer, long-term, that we could get to a place where there was no ambiguity to this relationship between a query and backend, but I'm sure there are bigger things to resolve before getting there.

lib/logflare/endpoints/query.ex
amokan: My thinking was to keep this PR tight around just the relationship between backend<->query and to add the CH enum on a subsequent PR.

priv/repo/migrations/20250705172040_populate_backend_id_for_existing_queries.exs
amokan: I've removed this migration for now.

lib/logflare/backends/backend.ex
```diff
@@ -39,6 +41,8 @@ defmodule Logflare.Backends.Backend do
       on_replace: :delete
     )
 
+    has_many(:queries, Query)
```
Ziinc:
```suggestion
    has_many(:endpoint_queries, Query)
```

PR #2504, opened 2025-07-08
feat: Use monaco editor in EndpointsLive

msmithstubbs:
* Add MonacoEditorComponent to be used as a LiveComponent
* Support providing list of strings to be used as completions in editor
* Remove some tests from EndpointsLive: change events are handled with a JS hook, so render_change no longer works for the query field

### DIFF
--- a/assets/js/monaco_editor_completion_provider.js
+++ b/assets/js/monaco_editor_completion_provider.js
@@ -1,0 +1,49 @@
+/*
+Listen for editor mount and register completion provider
+Supports MonacoEditorComponent
+*/
+const editorMountedHandler = (ev) => {
+  // build a list of completions for the editor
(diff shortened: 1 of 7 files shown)


assets/js/monaco_editor_completion_provider.js
```diff
@@ -0,0 +1,49 @@
...
+      return {
+        label: name,
+        kind: monaco.languages.CompletionItemKind.Module,
+        insertText: name,
+        range: range,
+      };
+    });
+  }
+
+  monaco.languages.registerCompletionItemProvider("sql", {
```
Ziinc: hmm we aren't able to add in completions via a <script> or similar within the component template? not a fan of separating out the js needed for a lv component, adds some additional complexity as well when we need to hunt for the js related to the monaco editor.

PR #2508, opened 2025-07-08
feat: managed service accounts for BYOB users

Ziinc:
This PR adds in adding of managed SAs to BYOB users' projects. This ensures that the IAM policy updating is non-destructive. It also requires the permissions to fetch and set the IAM policies (i.e. IAM admin), and thusly this is optional depending on whether the user has enabled it.

Queries are routed through partitioned Goth servers, each service account gets their own partitioned goth server to ensure that token refreshing does not bottleneck requests.
This is only true when:
1. system has managed SAs enabled and user is not byob
2. system has managed SAs enabled, and user is BYOB and has managed SAs enabled.
Otherwise, queries are all routed through Logflare.Goth (whcih is the main service account)

### DIFF
--- a/lib/logflare_web/controllers/user_controller.ex
+++ b/lib/logflare_web/controllers/user_controller.ex
@@ -6,6 +6,8 @@ defmodule LogflareWeb.UserController do
 
   alias Logflare.{User, Repo, Users, TeamUsers, Source.Supervisor, Billing.Stripe}
 
+  alias Logflare.Backends.Adaptor.BigQueryAdaptor
+
   defp env_service_account,
     do: Application.get_env(:logflare, Logflare.Google)[:service_account] || ""
 
@@ -48,6 +50,11 @@ defmodule LogflareWeb.UserController do
       {:ok, updated_user} ->
         if updated_user.bigquery_project_id != user.bigquery_project_id,
           do: Supervisor.reset_all_user_sources(user)
+
+        if updated_user.bigquery_enable_managed_service_accounts do
+          # update iam policy
+          BigQueryAdaptor.update_iam_policy(updated_user)
+        end
 
         conn
         |> put_flash(:info, "Account updated!")


Ziinc:
@Logflare/dashbit will need to merge this in first but a review would be appreciated 🙏

lib/logflare/google/resource_manager.ex
```diff
@@ -27,6 +28,80 @@ defmodule Logflare.Google.CloudResourceManager do
...
+         {:ok, policy} <- get_iam_policy(user),
+         {:contains?, _policy, false} <-
+           {:contains?, policy, contains_managed_service_accounts?(policy)} do
+      append_managed_service_accounts(user.bigquery_project_id, policy)
+    else
+      {:contains?, policy, _} -> {:ok, policy}
+      {:enabled?, false} -> {:error, :managed_service_accounts_disabled}
+      {:error, _err} = err -> err
+    end
+  end
```
wojtekmach, line 60: This is pretty subjective but in our experience, with/1 + "tagged tuples" decreases readability as one has to go back'n'forth between the "happy" path and the _corresponding_ else clause. For me, with/1 is ideal without else clause or with a _single_ catch-all else clause, anything other than that, to _me_, gives me a pause what's going on every single time.

I haven't tested it but I think this would be the equivalent:

```elixir
def append_managed_sa_to_iam_policy(user) do
  if BigQueryAdaptor.managed_service_accounts_enabled?() do
    with {:ok, policy} <- get_iam_policy(user) do
      if contains_managed_service_accounts?(policy) do
        {:ok, policy}
      else
        append_managed_service_accounts(user.bigquery_project_id, policy)
      end
    end
  else
    {:error, :managed_service_accounts_disabled}
  end
end
```

again, pretty subjective though!

priv/repo/migrations/20250630074142_add_users_bigquery_enable_managed_service_a_ccounts.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Logflare.Repo.Migrations.AddUsersBigqueryEnableManagedServiceACcounts do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :bigquery_enable_managed_service_accounts, :boolean, default: false
```
wojtekmach, line 6: consider adding `null: false` which would be safe since there's a default.

wojtekmach (commented):
Looks good to me!

PR #2512, opened 2025-07-09
feat: BQ tagging of queries by labels

Ziinc:
this adds in BQ label tagging of queries for billing processing.
It adds a `:labels` column to `endpoint_queries`, and uses the `LF-ENDPOINTS-LABELS` header for values.
It also allows referencing of params using `my_label=@my_param` syntax.
Static values can also be set.

### DIFF
--- a/lib/logflare/endpoints/query.ex
+++ b/lib/logflare/endpoints/query.ex
@@ -20,7 +20,8 @@ defmodule Logflare.Endpoints.Query do
              :cache_duration_seconds,
              :proactive_requerying_seconds,
              :max_limit,
-             :enable_auth
+             :enable_auth,
+             :labels
            ]}
   typed_schema "endpoint_queries" do
     field(:token, Ecto.UUID, autogenerate: true)
@@ -34,7 +35,8 @@ defmodule Logflare.Endpoints.Query do
     field(:proactive_requerying_seconds, :integer, default: 1_800)
     field(:max_limit, :integer, default: 1_000)
     field(:enable_auth, :boolean, default: true)
-
+    field(:labels, :string)
+    field(:parsed_labels, :map, virtual: true)
     field(:metrics, :map, virtual: true)
 
     belongs_to(:user, Logflare.User)
@@ -64,7 +66,8 @@ defmodule Logflare.Endpoints.Query do
       :max_limit,
       :enable_auth,
       :language,
-      :description
+      :description,
+      :labels
     ])
     |> validate_required([:name, :query, :language])
   end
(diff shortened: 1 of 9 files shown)


test/logflare_web/controllers/endpoints_controller_test.exs
```diff
@@ -109,6 +109,52 @@ defmodule LogflareWeb.EndpointsControllerTest do
...
+        |> put_req_header("lf-endpoint-labels", "other_value=1234,omit=334")
+        |> get(
+          ~p"/api/endpoints/query/#{endpoint.name}?#{%{"other_value" => 1234, "omit" => 333, "my_param" => "my_value", "sql" => "select 2"}}"
+        )
+
+      assert [_] = json_response(conn, 200)["result"]
+      assert conn.halted == false
+      assert_received labels = %{"my_label" => "my_value", "other_value" => "1234"}
+      refute labels["omit"]
+      refute labels["sql"]
```
wojtekmach: nitpick: maybe worth asserting on all the possible label keys, then we don't need to do refutations:

```suggestion
      assert_received labels
      assert labels == %{"my_label" => "my_value", "other_value" => "1234"}
```

but yeah, totally fine is!

test/logflare_web/controllers/endpoints_controller_test.exs
```diff
@@ -109,6 +109,52 @@ defmodule LogflareWeb.EndpointsControllerTest do
...
+        send(pid, opts[:body].labels)
+        {:ok, TestUtils.gen_bq_response()}
+      end)
+
+      endpoint =
+        insert(:endpoint,
+          user: user,
+          enable_auth: true,
+          query: "with a as (select 1 as b) select b from a",
+          labels: "my_label=@my_param,other_value"
```
wojtekmach: besides `field=@param` I think we also allow `field=value`, maybe worth testing it here too just in case?

lib/logflare/endpoints/query.ex
```diff
@@ -81,7 +84,8 @@ defmodule Logflare.Endpoints.Query do
       :max_limit,
       :enable_auth,
       :language,
-      :description
+      :description,
+      :labels
```
wojtekmach, line 88: maybe worth adding some basic validations? For example when setting `labels: "field1,"` I think the trailing comma could make it so there are two labels, `"field1"` and `""`. But yeah, if that's not gonna break anything I probably wouldn't bother with extra validations.

Ziinc: nice one, yea trailing commas would add an unnecessary empty label for it, added some filtering for that

PR #2524, opened 2025-07-12
feat: initial partial implementation of storage write api

WolfDan:
continuation of https://github.com/Logflare/logflare/pull/2407


lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -15,6 +17,31 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
   alias Logflare.Google.BigQuery.GenUtils
   alias Logflare.Google.CloudResourceManager
   alias Logflare.Google
+  alias Logflare.Google.BigQuery
+  alias Logflare.SourceSchemas
+  alias Logflare.Backends.Adaptor.BigQueryAdaptor.GoogleApiClient
+  require Record
+
+  Record.defrecord(
```
Ziinc: lets move these to its own StorageWrite module

native/arrow_ipc_ex/README.md
```diff
@@ -0,0 +1,20 @@
+# NIF for Logflare.Backends.Adaptor.BigQueryAdaptor.Native
```
Ziinc: can del

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
```diff
@@ -0,0 +1,66 @@
...
+
+    arrow_rows = %ArrowData{rows: arrow_record_batch, writer_schema: writer_schema}
+
+    request =
+      %AppendRowsRequest{
+        write_stream: "projects/#{project}/datasets/#{dataset}/tables/#{table}/streams/_default",
+        rows: {:arrow_rows, arrow_rows}
+      }
+
+    stream = BigQueryWrite.Stub.append_rows(channel)
```
Ziinc: just a note, we'll likely need to store the stream in ets (using cachex or otherwise) so that the streams can be used by other sources as well

lib/logflare/google/bigquery/event_utils/event_utils.ex
```diff
@@ -1,5 +1,20 @@
 defmodule Logflare.Google.BigQuery.EventUtils do
   @moduledoc false
+
```
Ziinc: just a note, we want to move this module into the BigQueryAdaptor namespace eventually, so using defdelegate in BigQueryAdaptor would be ideal. there should be some examples of this in the module already.

lib/logflare/backends/adaptor/bigquery_adaptor.ex
WolfDan: I've removed these code, since it's from arrow and we don't use it anymore

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
WolfDan: @Ziinc what do you mean with this one?

.tool-versions
```diff
@@ -1,5 +1,5 @@
 elixir 1.17.3-otp-27
 nodejs 22.9.0
-rust 1.84.1
+rust nightly
```
Ziinc: is this necessary?

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
```diff
@@ -0,0 +1,81 @@
...
+
+  def append_rows({:arrow, data_frame}, project, dataset, table) do
+    partition_count = System.schedulers_online()
+    partition = :erlang.phash2(self(), partition_count)
+
+    {:ok, goth_token} = Goth.fetch({Logflare.Goth, partition})
+
+    {:ok, channel} =
+      GRPC.Stub.connect("bigquerystorage.googleapis.com:443",
+        adapter: GRPC.Client.Adapters.Finch,
```
Ziinc, line 20: we should be able to specify the Finch pool that we want to use instead of relying on config.exs

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
```diff
@@ -0,0 +1,81 @@
...
+      request =
+        %AppendRowsRequest{
+          write_stream:
+            "projects/#{project}/datasets/#{dataset}/tables/#{table}/streams/_default",
+          rows: {:arrow_rows, arrow_rows}
+        }
+
+      stream = BigQueryWrite.Stub.append_rows(channel)
+
+      GRPC.Stub.send_request(stream, request, end_stream: true)
```
Ziinc: this means one handle_batch. = 1 stream with 1 request... would not be as efficient if say we sent multiple requests sequentially before ending the stream... just a thought.
potential area for improvement

lib/logflare/google/bigquery/event_utils/event_utils.ex
```diff
@@ -1,5 +1,20 @@
...
+  def log_event_to_struct(%Logflare.LogEvent{body: body}) do
```
Ziinc: just a note: we can rename this to some thing more precise, since it isn't actually a struct that is returned

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
Ziinc: @WolfDan since we can share the stream across multiple sources, we could technically avoid a situation where the stream goes into client half-closed state, to maximize the throughput.
currently, other pipelines would be initiating their own streams, but we might be able to get more throughput with stream sharing.
Just an intuition, i might be wrong.

test/logflare/backends/bigquery_adaptor_test.exs
```diff
@@ -182,6 +183,150 @@ defmodule Logflare.Backends.BigQueryAdaptorTest do
...
+        :timer.sleep(20)
+        {:ok, %GRPC.Channel{}}
+      end)
+      |> stub(:send_request, fn stream, request ->
+        # simulate latency
+        :timer.sleep(10)
+
+        %{rows: {:arrow_rows, %{rows: %{serialized_record_batch: ipc_msg}}}} = request
+
+        BencheeAsync.Reporter.record(byte_size(ipc_msg))
```
WolfDan, line 261: @Ziinc this is a bit weird as unit of work, while it "works" we are jsut counting bytes, but we don't really have a really stable way of knowing how many records were sent since the arrow format compress duplicates and such 

A solution I did was to create unique records, but even in that case it could be the case we serialize more or less than the ones mentioned

If you have any other ideas please let me know

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
WolfDan: @Ziinc I explored it a bit, but it will make thins a bit complicated from the point of when to close the stream, another issue with it is that in order to make it work we need bidirectional streaming on Finch to work correctly and it's a bit of slipery slope 

I keep it as it is, but if you have any input regarding this let me know

lib/logflare/application.ex
```diff
@@ -181,6 +181,21 @@ defmodule Logflare.Application do
...
+       pools: %{
+         "https://bigquerystorage.googleapis.com" => [
+           protocols: [:http2],
+           count: max(base, 20) * 2,
+           start_pool_metrics?: true,
+           conn_opts: [
+             client_settings: [
+               initial_window_size: 8_000_000,
+               max_frame_size: 8_000_000
+             ]
```
Ziinc, line 195: could you add in an inline comment for why this is needed for posterity

lib/logflare/application.ex
```diff
@@ -181,6 +181,21 @@ defmodule Logflare.Application do
            start_pool_metrics?: true
          ]
        }},
+      {Finch,
+       name: Logflare.FinchBQStorage,
```
Ziinc, line 185: FinchBQStorageWrite

lib/logflare/application.ex
```diff
@@ -181,6 +181,21 @@ defmodule Logflare.Application do
...
+       pools: %{
+         "https://bigquerystorage.googleapis.com" => [
+           protocols: [:http2],
+           count: max(base, 20) * 2,
```
Ziinc, line 189:
```suggestion
           count: max(base, 20), 
```
Can remove the x2 first and observe pool metrics before we bump this up more

lib/logflare/backends/adaptor/bigquery_adaptor/finch/bidirectional_stream.ex
```diff
@@ -0,0 +1,101 @@
+defmodule Grpc.Client.Adapters.Finch.BidirectionalStream do
+  alias GRPC.Client.Adapters.Finch.StreamState
+  alias Grpc.Client.Adapters.Finch.RequestUtils
+  use GenServer
+
+  @finch_instance_name Logflare.FinchBQStorage
```
Ziinc, line 6: If we're hardcoding it here then the application.ex should refer to this value instead through a public func

Ziinc: And can't We pass it throguh as an option instead? So that we can keep the grpc client adaptor free of any BQ logic

lib/logflare/backends/adaptor/bigquery_adaptor/finch/request_process.ex
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  @impl true
+  def init([stream_request_pid, path, client_headers, data, opts]) do
+    timeout = Keyword.get(opts, :timeout, :infinity)
+
+    req = Finch.build(:post, path, client_headers, data)
+
+    stream_ref =
+      Finch.async_request(req, @finch_instance_name, receive_timeout: timeout)
```
Ziinc, line 19: The finch instance name can be from the opt instead

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
```diff
@@ -0,0 +1,80 @@
...
+        headers: [
+          {"Authorization", "Bearer #{goth_token.token}"}
+        ]
+      )
+
+    {:ok, arrow_schema} = DataFrame.dump_ipc_schema(data_frame)
+
+    writer_schema = %Google.Cloud.Bigquery.Storage.V1.ArrowSchema{serialized_schema: arrow_schema}
+
+    {:ok, batch_msgs} = DataFrame.dump_ipc_record_batch(data_frame)
```
Ziinc, line 31: What is the chance that it would actually be split up into many batch messages? Given that Broadway will already batch it to a set number of events.
Using the Task to fire of the requests might be heavier than just firing it within the process due to copying needed

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -147,7 +149,20 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
     OpenTelemetry.Tracer.with_span :bigquery_pipeline, %{
       attributes: Map.new(attributes)
     } do
-      stream_batch(context, messages)
+      if context.bq_storage_write_api do
```
Ziinc, line 152: If user changes this at runtime then the value would become stale. Might be better to pull source value from cache.

Ziinc (approved):

Ziinc:
@WolfDan can you resolve the conflicts and the comments please?

lib/logflare/sources/source/bigquery/pipeline.ex
Ziinc: We previously avoided it for performance but in this case we need to pull it to avoid gotchas

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
WolfDan: @Ziinc it's not about the amount but rather the size of the message, if by some reason the data exceeds the 10mb size then we have to split it, while Nroadway can batch it, afaik it dosen't follow the concept of size which matters for this one 

It could be the case that we never split the msgs because there is a chance for it to be required to split the batch msgs due to it's size and not the amount of them

Ziinc: We already have size limits in the batching function of the Bigquery Pipeline up to 5MB, so that situation would not be possible.
We can safely assume everything would be around that max  ±1MB at extreme edge case.

Ziinc: In event that it does split it into multiple RecordBatch then we can emit a warning (since it shouldn't be happening) and can remove the Task overhead by doing both sequentially. Don't need to factor it in since it is an edge case

PR #2535, opened 2025-07-16
refactor: join endpoint for SourceController dashboard

nelsonmestevao:
Reuse the code in the dashboard controller when user has a team_user and when it doesn't in the http session.

Moved Routes helper to `sigil_p` in dashboard.html.heex.

### DIFF
--- a/lib/logflare_web/controllers/source_controller.ex
+++ b/lib/logflare_web/controllers/source_controller.ex
@@ -1,5 +1,6 @@
 defmodule LogflareWeb.SourceController do
   use LogflareWeb, :controller
+
   require Logger
 
   alias Logflare.Billing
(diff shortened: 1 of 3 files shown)


Ziinc (approved):

test/logflare_web/controllers/source_controller_test.exs
```diff
@@ -26,18 +26,37 @@ defmodule LogflareWeb.SourceControllerTest do
...
+
+      conn
+      |> put_session(:team_user_id, team_user.id)
+      |> visit(~p"/dashboard")
+      |> assert_has("li > a", text: "Dashboard", exact: true)
+      |> assert_has("h5", text: "Saved Searches", exact: true)
+      |> assert_has("a[href='#{~p"/sources/#{source}"}']", text: source.name)
+      |> assert_has("h5", text: "Teams")
+      |> assert_has(
+        "a[href='#{~p"/profile/switch?#{%{"user_id" => team_user.team.user_id, "team_user_id" => team_user.id}}"}']",
```
Ziinc, line 57: we can reduce the granularity of the selectors to account for potential template changes, so that there is less maintenance needed for tests

lib/logflare_web/controllers/source_controller.ex
```diff
@@ -28,44 +29,19 @@ defmodule LogflareWeb.SourceController do
...
   def dashboard(%{assigns: %{user: user, team: team}} = conn, _params) do
     user = Users.preload_sources(user)
     sources = Sources.preload_for_dashboard(user.sources)
 
-    home_team = team
-
-    team_users_with_teams = TeamUsers.list_team_users_by_and_preload(email: user.email)
+    {home_team, team_users_with_teams} =
+      if conn.assigns[:team_user] do
+        {Teams.get_home_team(conn.assigns.team_user),
```
Ziinc: Can we do a match to assign the team_user to a variable in the if arg?

Ziinc (changes requested):

lib/logflare_web/controllers/source_controller.ex
nelsonmestevao: Well spotted, thank you 🙏 Done in last push 💪

test/logflare_web/controllers/source_controller_test.exs
nelsonmestevao: Which ones are you talking exactly? In L57, for example, I was trying to make sure there's a link to switch teams, what do you think is overly coupled?

Ziinc: was thinking it could simply be an `a` for example, i think the rest is fine now that i look at it.
just a nit 😄 its fine to leave it in

PR #2538, opened 2025-07-16
fix: syn conflict resolution

Ziinc:
Stop the process if a conflict arises. Seems like syn does not handle process conflicts automatically if the handler is defined.

> If implemented, this method MUST return the pid() of the process that you wish to keep. **The other process will not be killed, so you will have to decide what to do with it**. If the custom conflict resolution method does not return one of the two Pids, or if the method crashes, none of the Pids will be killed and the conflicting name will be freed.

Likely the cause of the issue we are seeing with multiple GenSingleton servers getting created still.

### DIFF
--- a/lib/logflare/syn_event_handler.ex
+++ b/lib/logflare/syn_event_handler.ex
@@ -6,8 +6,6 @@ defmodule Logflare.SynEventHandler do
   """
   @behaviour :syn_event_handler
 
-  alias Logflare.Utils.Tasks
-
   require Logger
   @impl true
 
--- a/lib/logflare/syn_event_handler.ex
+++ b/lib/logflare/syn_event_handler.ex
@@ -86,14 +86,6 @@ defmodule Logflare.SynEventHandler do
   end
 
   defp try_to_stop(pid) do
-    Tasks.start_child(fn ->
-      node_to_kill = node(pid)
-      # Random sleep to prevent concurrent attempts to stop the same process
-      Process.sleep(:rand.uniform(1500))
-
-      if :erpc.call(node_to_kill, Process, :alive?, [pid], 5000) do
-        :erpc.call(node_to_kill, Process, :exit, [pid, :normal], 5000)
-      end
-    end)
+    Process.exit(pid, :syn_resolve_kill)
   end
 end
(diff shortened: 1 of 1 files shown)


lib/logflare/syn_event_handler.ex
```diff
@@ -54,26 +58,42 @@ defmodule Logflare.SynEventHandler do
...
+    Tasks.start_child(fn ->
+      node_to_kill = node(pid)
+      # Random sleep to prevent concurrent attempts to stop the same process
+      Process.sleep(:rand.uniform(1500))
+
+      if :erpc.call(node_to_kill, Process, :alive?, [pid], 5000) do
+        :erpc.call(node_to_kill, Process, :exit, [pid, :normal], 5000)
+      end
+    end)
+  end
```
jonatanklosko: Sending an exit signal with reason `:normal` typically does not stop the target process, see [`Process.exit/2`](https://hexdocs.pm/elixir/Process.html#exit/2).

Also, you don't really need the `Process.alive?` check, you can just send the exit signal and move on. If the process is not alive, it will be ignored. Is there a particular scenario you are trying to avoid with `Process.sleep`? I would expect sending the exit signal multiple times to be fine.

Finally, if you only send the signal, you may not need `Tasks.start_child` either. The only reason to do it would be if you are worried about [Blocking Signaling Over Distribution](https://www.erlang.org/doc/system/ref_man_processes#blocking-signaling-over-distribution), in which case I would leave a comment explaining it for clarity.

```suggestion
  defp try_to_stop(pid) do
    Process.exit(pid, :syn_resolve_kill)
  end
```

Ziinc: Using the Task was due to worries of blocking the syn process.
Yea also suspected that the alive check was unnecessary given that there is only one return value for Process.exit/2.

As for the normal vs kill, i was thinking that a kill might cause us to not be able to trap exits if needed... i'll switch it to kill and then revisit it if/when we need to trap exits

jonatanklosko: > As for the normal vs kill, i was thinking that a kill might cause us to not be able to trap exits if needed... i'll switch it to kill and then revisit it if/when we need to trap exits

If you use anything else than `:normal` or `:kill`, it will stop the process, but if it traps exits it will get converted to an `:EXIT` message instead. That's why I suggested `:syn_resolve_kill`. Or is there some other issue?

Ziinc: Ah indeed, the `:syn_resolve_kill` went over my head, no issues at all and def much cleaner

jonatanklosko:
Looks good to me!

PR #2546, opened 2025-07-18
Include log levels in search results

msmithstubbs:
This PR adds log levels to the Source and Search UI, and removes dependency on BigQuery UDFs.

CleanShot 2025-07-22 at 13 52 22@2x

* Log levels shown in sources and search views
  * There is some duplication of rendering between the live view, controller, and JS. I looked refactoring to consolidate these but it turned out to be a bigger job than was warranted for this PR. Will leave it for when we combine sources and search.
* Remove BigQuery UDF
  * `LF_TIMESTAMP_TRUNC` replaced with `where_timestamp_ago/4`
  * `LF_TIMESTAMP_TRUNC{_WITH_TIMEZONE}` replaced by SQL queries
  * `LF_GENERATE_TIMESTAMP_ARRAY` doesn’t appear to have been used. Removed.
* Removed `bigquery_udfs_hash` from user schema, and migration to remove from database
* SearchLVTest: enable and extend "run a query" to verify level in search results


assets/js/source.js
```diff
@@ -110,19 +110,20 @@ export function scrollBottom() {
...
-
-  const metadataElement = !_.isEmpty(body.metadata) ?
-    `
+  const { via_rule, origin_source_id, body } = e;
+  const metadata = JSON.stringify(body, null, 2);
+  const formatter = await userSelectedFormatter();
+  const formattedDatetime = formatter(body.timestamp);
+  const randomId = Math.random() * 10e16;
+  const metadataId = `metadata-${body.timestamp}-${randomId}`;
+  const log_level = _.get(body, ["metadata", "level"]);
```
Ziinc, line 119: small note: can refactor in future to import the specific module instead to reduce bundle size

lib/logflare/logs/search/logs_search_queries.ex
```diff
@@ -24,9 +36,25 @@ defmodule Logflare.Logs.SearchQueries do
     |> select([t], %{
       timestamp:
         fragment(
-          "(`$$__DEFAULT_DATASET__$$`.LF_TIMESTAMP_TRUNC_WITH_TIMEZONE(?, ?, ?)) as timestamp",
+          """
```
Ziinc, line 39: can abstract out to avoid duplication.

priv/repo/migrations/20250725034224_alter_users_remove_bigquery_udfs_hash.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Logflare.Repo.Migrations.AlterUsersRemoveBigqueryUdfsHash do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      remove :bigquery_udfs_hash, :string, default: "", null: false
```
Ziinc: lets leave the field in for now

lib/logflare_web/live/search_live/log_event_components.ex
```diff
@@ -0,0 +1,74 @@
...
+
+  @log_levels ~W(debug info warning error alert critical notice emergency)
+
+  attr :log, :map
+  attr :search_timezone, :string
+  attr :query_string, :string
+  attr :source, :map
+
+  def log_search_event(%{log: %{body: body}} = assigns) when is_map_key(body, "event_message") do
+    tz_part =
```
Ziinc, line 22: do we still need this given we don't show the tz part any more in the timestamp? can potentially eliminate this.

test/logflare_web/live/search_live/logs_search_lv_test.exs
```diff
@@ -607,27 +609,34 @@ defmodule LogflareWeb.Source.SearchLVTest do
...
+      Ecto.Adapters.SQL.Sandbox.allow(Logflare.Repo, self(), search_executor_pid)
+
       # post-init fetching
-      :timer.sleep(800)
 
       render_change(view, :start_search, %{
         "search" => %{@default_search_params | "querystring" => "somestring"}
       })
 
       # wait for async search task to complete
```
Ziinc:
```suggestion
```

Ziinc (changes requested):
we don't do destructive migrations except on major versions, which will not be any time soon.
We can add a section to development.md to note to note that this column is no longer used.
looks great for the rest! 🔥

assets/js/source.js
msmithstubbs: Sure - though I'm hoping we can just replace the bulk of this with LiveView.

lib/logflare/logs/search/logs_search_queries.ex
msmithstubbs: This is from the original PR so hadn't looked at it closely. Now that I do it seems we only ever call it without  a timestamp, so the second function is redundant. Can just remove it unless there's a reason we need it in future?

lib/logflare_web/live/search_live/log_event_components.ex
msmithstubbs: The timestamp is still shown in the content for copy to clipboard:

```heex
        <.link class="tw-text-[0.65rem]  group-hover:tw-visible tw-invisible" phx-click={JS.dispatch("logflare:copy-to-clipboard", detail: %{text: "#{@formatted_timestamp}    #{@message}"})} data-toggle="tooltip" data-placement="top" title="Copy to clipboard">
          copy
        </.link>
```

https://github.com/Logflare/logflare/pull/2546/files#diff-3805bb32c63be339e8f8f7bf33182f7ab3c2719c8e724ce968d3bb96081de869R51-R53

If we don't need the timestamp there then, yes, it can be eliminated.

assets/js/source.js
Ziinc: yea lets do the js slimming in a separate PR.

lib/logflare_web/live/search_live/log_event_components.ex
Ziinc: ah ok lets keep it then

lib/logflare/logs/search/logs_search_queries.ex
Ziinc: Yea we can remove the function entirely since we do the tz formatting client side.
let's do it in separate PR.

Ziinc (approved):

PR #2547, opened 2025-07-18
Fixing the logs to be shown in Studio / removing retention for self-hosted

activenode:
Currently no logs in Studio are shown. With this fallback, the logs will be able to be requested from Studio.

I also deleted the `functions.invocation-stats` because it isn't used in the `supabase/` repo. I would've loved to also delete the `usage.api-counts` file which isn't even working by no means (throws elixir parser errors) but at the same time I didn't want to delete a file that is still being used by an endpoint from Studio (https://github.com/supabase/supabase/blob/master/apps/studio/data/analytics/project-log-stats-query.ts#L36) -> maybe you wanna check if that makes sense because, as I said, it's not working.

Either ways, here's the reference issue: https://github.com/supabase/supabase/issues/37218


activenode:
Will do. For clarity for someone reading this ticket, here's some transparency:

The `@project` fallback is seen to be unnecessary because `Studio` should send this param along with the request.

My take was that, in self-hosted, this param is always going to be `default`, not to forget that this param, due to it's single-tenant nature, might not even be "parametrizable" and hence using a fallback value would be sensible.

The SB teams take was that sending along `default` (or more specifically the project slug which is `default` in self-hosted), will fix this. I don't necessarily agree that the fallback is wrong because it would still allow to configure it but I can see that, if used in any future case where multi-tenancy is allowed, this might hide-away an issue.

activenode:
(will run a test with Studio to ensure everything works as expected)

activenode:
As discussed in person, I also removed the `retention`. The filtering happens via Studio timestamp query params.

activenode:
Once merged, please comment in https://github.com/supabase/supabase/pull/37363

Ziinc (approved):

PR #2560, opened 2025-07-23
chore: add clickhouse image to docker-compose and ci

amokan:
Adds `clickhouse/clickhouse-server:25.4` to the docker compose and GitHub CI setup.

- [x] Add clickhouse-server to `docker-compose.yml`
- [x] Add clickhouse-server to GitHub CI yaml
- [x] Convert some mocked clickhouse tests to use server
- [x] Add clickhouse integration test showing full cycle during backend adaptor lifecycle


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -621,6 +633,15 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
...
   end
 
+  @spec get_engine_from_app_config() :: String.t()
+  defp get_engine_from_app_config() do
+    Application.get_env(:logflare, :clickhouse_backend_adapter, [])
+    |> Keyword.get(
+      :engine,
+      "SharedSummingMergeTree('/clickhouse/tables/{uuid}/{shard}', '{replica}')"
+    )
+  end
```
```suggestion
    Application.get_env(:logflare, :clickhouse_backend_adapter)[:engine] || @default_engine
  end
```

Ziinc: imo should remove the `@default_engine` fallback and let the caller decide.

Ziinc: would be more true to the function name

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
```diff
@@ -109,12 +111,14 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.QueryTemplates do
 
   - `:database` - (Optional) Will produce a fully qualified `<database>.<table>` string when provided with a value. Defaults to `nil`.
   - `:table` - (Optional) Defaults to `"key_type_counts_per_minute"` and should stay this way, but if you really want to change it - this is the way.
+  - `:engine` - (Optional) ClickHouse table engine. Defaults to `"SharedSummingMergeTree('/clickhouse/tables/{uuid}/{shard}', '{replica}')"` for replicated setups, use `"SummingMergeTree()"` for single-node setups.
```
Ziinc: not much docs in the wild on this, we could add in some for posterity in our CH backend docs

Ziinc: Also, looks like [docs](https://clickhouse.com/docs/engines/table-engines/mergetree-family/replication) is using ReplicatedSummingMergeTree

Also seems like we shouldn't specify the `Replicated` part?
https://clickhouse.com/docs/engines/table-engines/mergetree-family/replication#creating-replicated-tables

test/logflare/backends/adaptor/clickhouse_adaptor/provisioner_test.exs
```diff
@@ -0,0 +1,253 @@
...
+        assert test_backend.id == backend.id
+        :ok
+      end)
+
+      ClickhouseAdaptor
+      |> expect(:provision_all, fn {prov_source, prov_backend} ->
+        assert prov_source.id == source.id
+        assert prov_backend.id == backend.id
+        :ok
+      end)
```
Ziinc: should actually provision these, and verify that they're provisioned, no?

test/logflare/backends/adaptor/clickhouse_adaptor/provisioner_test.exs
```diff
@@ -0,0 +1,253 @@
...
+      user = insert(:user)
+      source = insert(:source, user: user)
+      backend = insert(:backend, type: :clickhouse)
+
+      [source: source, backend: backend]
+    end
+
+    test "fails initialization when connection test fails", %{source: source, backend: backend} do
+      ClickhouseAdaptor
+      |> expect(:test_connection, fn _, _ -> {:error, :connection_failed} end)
```
Ziinc: we can point it to a random port that doesn't have CH, that would let us avoid using mocks

test/logflare/backends/adaptor/clickhouse_adaptor/provisioner_test.exs
```diff
@@ -0,0 +1,253 @@
...
+
+      assert {:error, :table_creation_failed} = result
+    end
+
+    test "handles different provisioning error types", %{source: source, backend: backend} do
+      test_cases = [
+        {:error, :insufficient_permissions},
+        {:error, :database_not_found},
+        {:error, :timeout}
+      ]
```
Ziinc: we can run these against the db, then for the timeout we could mock the Ch response instead

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
amokan: Haha. I can't believe I missed this one 😄

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
amokan: Good catch.

lib/logflare/backends/adaptor/clickhouse_adaptor/provisioner.ex
```diff
@@ -47,6 +47,15 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Provisioner do
     with :ok <- ClickhouseAdaptor.test_connection(source, backend),
          :ok <- ClickhouseAdaptor.provision_all(args) do
       {:ok, state, {:continue, :close_process}}
+    else
+      {:error, reason} = error ->
+        Logger.error("Failed to provision ClickHouse resources",
+          source_token: source.token,
+          backend_id: backend.id,
+          reason: inspect(reason)
```
Ziinc:
```suggestion
          error_string: inspect(reason)
```

test/logflare/backends/adaptor/clickhouse_adaptor/pipeline_test.exs
```diff
@@ -1,74 +1,50 @@
...
-        assert producer_config[:module] ==
-                 {BufferProducer,
-                  [
-                    source_id: adaptor_state.source.id,
-                    backend_id: adaptor_state.backend.id
-                  ]}
+    on_exit(fn ->
+      if Process.alive?(supervisor_pid) do
+        Process.exit(supervisor_pid, :kill)
+      end
```
Ziinc:
```suggestion
      Process.exit(supervisor_pid, :kill)
```
don't need to check if proc is alive, if it is dead then it will be a noop

Ziinc (approved):

PR #2563, opened 2025-07-25
fix: ensures that the AlertsScheduler supervisor is properly stopped

Ziinc:
Seems like Process.exit does not shut down supervisors nicely, watcher now accounts for restarting Supervised procs (in the case of `:already_present`


lib/logflare/syn_event_handler.ex
```diff
@@ -84,6 +84,18 @@ defmodule Logflare.SynEventHandler do
   end
 
   defp try_to_stop(pid) do
-    Process.exit(pid, :syn_resolve_kill)
+    if is_supervisor?(pid) do
```
josevalim: Instead of checking this, could you ask its supervisor to shut it down using terminate_child? https://hexdocs.pm/elixir/Supervisor.html#terminate_child/2

This way it should be guaranteed and the supervisor will shut it down accordingly to its supervision strategy.

If this is running inside the GenSingleton, perhaps a new function in the GenSingleton API? 🤔

Ziinc: altered to use teminate_child/2, definitely a cleaner and more explicit way to handle it.
Though it wouldn't handle the situation where the GenSingleton process is a supervisor and would result in the same gotcha, though i guess that is another problem for another day.

PR #2570, opened 2025-07-29
Reduce queries in Search Results page load

msmithstubbs:
Following on from #2557

- [x] remove udf creation on search page load - fixed in #2546
- [x] remove all udf usage - fixed in #2546
- [x] 1 query made on page load
  - [x] Remove dry run query
  - [x] Configure per request cost limit
  - [x] Log warning if query cost above limit

⚠️ Query cost is checked and logged but not yet enforced:

https://github.com/Logflare/logflare/pull/2570/files#diff-bef332c925224bac9585297f3254655dd58c669688b52492e2233ba11f75fb0eR47-R48

UI discussion below has been split into a separate PR.

### DIFF
--- a/lib/logflare/ecto/bigquery/bq_repo.ex
+++ b/lib/logflare/ecto/bigquery/bq_repo.ex
@@ -44,7 +44,7 @@ defmodule Logflare.BqRepo do
         dryRun: false,
         jobTimeoutMs: @query_request_timeout,
         timeoutMs: @query_request_timeout,
-        # Not enforced for now.
+        # Not enforced for now
         # maximumBytesBilled: user.bigquery_processed_bytes_limit,
         labels: %{
           "managed_by" => "logflare",
@@ -59,6 +59,7 @@ defmodule Logflare.BqRepo do
       GenUtils.get_conn({:query, user})
       |> Api.Jobs.bigquery_jobs_query(project_id, body: query_request)
       |> GenUtils.maybe_parse_google_api_result()
+      |> warn_if_cost_above_limit(query_request.labels, user)
 
     with {:ok, response} <- result do
       response =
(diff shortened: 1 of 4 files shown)


msmithstubbs:
For the UI feedback we can trigger a dry run query after the editor parses the query and include it just below query editor:

CleanShot 2025-07-31 at 13 57 27@2x



When testing the query we can display the query cost above the results:

CleanShot 2025-07-31 at 13 54 23@2x

@Ziinc any feedback?

Ziinc:
- endpoints ui when editing query should not used caching, should always run query directly
- need to factor in params, it won't show when there is no params present in the field, right? Or would it still show even if params are empty/invalid?
- need formatting for units up to TB, not sure if BQ already returns it formatted.

Looks great!

Ziinc:
- for the testing ui let's disable caching
- we can show the estimated costs once the params are set or if no params are declared , but hide it if not set. 
- for alerts/query we should always show the bytes scanned

lib/logflare/ecto/bigquery/bq_repo.ex
```diff
@@ -95,4 +98,24 @@ defmodule Logflare.BqRepo do
...
+         user
+       )
+       when is_binary(total_bytes_processed) do
+    %{bigquery_processed_bytes_limit: limit} = user
+
+    if String.to_integer(total_bytes_processed) > limit do
+      Logger.warning("Query cost exceeded plan limit",
+        user: user.id,
+        cost: total_bytes_processed,
+        limit: limit
```
Ziinc:
```suggestion
      Logger.warning("Query bytes exceeded plan. limit: #{limit}, bytes: total_bytes_processed",
        user_id: user.id
```

Would be also good to add in any query label (I think the response should return the labels) for better monitoring. We can stringify the map to kv pairs.

msmithstubbs: Response doesn't return labels but they can be passed through from the request.

PR #2579, opened 2025-07-30
feat/perf: ipv6 support

Ziinc:
This adds in ipv6 support.

This does not inlcude ipv6 for postgres clustering strategy, that will be in separate PR once we move to the PostgresStrategy lib.


config/runtime.exs
```diff
@@ -225,14 +261,26 @@ if config_env() != :test do
   config :grpc, port: System.get_env("LOGFLARE_GRPC_PORT", "50051") |> String.to_integer()
 end
 
+socket_options_for_url = fn
+  url when is_binary(url) ->
+    case URI.parse(url) do
+      %URI{host: host} -> socket_options_for_host.(host)
+      _ -> raise "Failed to parse URL: #{inspect(url)}"
```
josevalim: URI.parse always succeeds but there is no guarantee the returned URL will have a host, so perhaps you want this?

```suggestion
      %URI{host: host} when is_binary(host) -> socket_options_for_host.(host)
      _ -> raise "Failed to parse URL with host: #{inspect(url)}"
```

config/runtime.exs
```diff
@@ -4,6 +4,27 @@ filter_nil_kv_pairs = fn pairs when is_list(pairs) ->
...
+socket_options_for_host = fn
+  host when is_binary(host) ->
+    case detect_ip_version.(host) do
+      {:ok, ip_version} -> [ip_version]
+      {:error, reason} -> raise "Failed to detect IP version: #{reason}"
+    end
+
+  _host ->
+    []
+end
```
josevalim: For what is worth, config/runtime.exs can access modules defined in your application (as long as it doesn't call any process, as the app did not start yet, so you can move these functions to a module if you want (especially if you want to write some tests).

config/runtime.exs
```diff
@@ -4,6 +4,27 @@ filter_nil_kv_pairs = fn pairs when is_list(pairs) ->
   Enum.filter(pairs, fn {_k, v} -> v !== nil end)
 end
 
+detect_ip_version = fn host ->
+  host = String.to_charlist(host)
+
+  cond do
+    match?({:ok, _}, :inet6_tcp.getaddr(host)) -> {:ok, :inet6}
```
josevalim, line 12: As far as I know this module is private. One option is to call `:inet.parse_address` and see if it returns `{:ok, {_, _, _, _, _, _, _, _}}` or `{:ok, {_, _, _, _}}`. What do you think?

PR #2580, opened 2025-07-30
perf: SourceSup auto-shutdown

Ziinc:
automatically shuts down SourceSup trees if there is no ingest activity detected for 60s. Performs shutdown locally only, not cluster-wide.

todos:
- [ ] test locally
- [ ] cleanup tests


config/config.exs
```diff
@@ -153,6 +153,15 @@ config :logflare, Logflare.Cluster.Utils, min_cluster_size: 1
 config :logflare, Logflare.Alerting.AlertsScheduler,
   init_task: {Logflare.Alerting, :init_alert_jobs, []}
 
+config :logflare, Logflare.Scheduler,
+  jobs: [
+    source_cleanup: [
+      schedule: "*/15 * * * *",
+      task: {Logflare.Sources, :shutdown_idle_sources, []},
+      run_strategy: Quantum.RunStrategy.All
```
SteffenDE: I don't know how Quantum works, so this question is mainly for my understanding: When configuring Quantum like this, does each node try to schedule a job on all nodes, or does it somehow ensure that only one node actually tries to schedule those jobs? I'd assume that if every node runs the scheduler, `Quantum.Strategy.Local` would be the strategy to choose, since each node would be responsible for scheduling only its own jobs.

lib/logflare/sources/source/supervisor.ex
```diff
@@ -207,7 +207,7 @@ defmodule Logflare.Sources.Source.Supervisor do
 
   defp do_lookup(source), do: Backends.lookup(Backends.SourceSup, source)
 
-  defp do_terminate_source_sup(%Source{} = source) do
+  def stop_source_local(%Source{} = source) do
     with {:ok, pid} <- do_lookup(source) do
       DynamicSupervisor.terminate_child(
```
v0idpwn: Potential issue: `DynamicSupervisor.terminate_child/2` is sync on the Dyn-sup side, so if the source sups take long to terminate, it will block the dynsup from spawning new children (and the terminates will happen one by one (per partition, I suppose))

josevalim: Good call. Ideally it is better to send a process a message to shutdown and let it terminate. But it will require setting it restart strategy to at least transient.

Ziinc: Transient should be fine... I'll do a follow up for this just to be conservative

config/config.exs
Ziinc: Yes, thanks for pointing it out, i switched to Quantum.Strategy.Local. using `{Quantum.RunStrategy.All, :cluster}` would result in job execution duplication (seems like i had misconfigured it too which might have caused it to slip through)

lib/logflare/sources/source/supervisor.ex
Ziinc: adjusted to use transient

PR #2542, opened 2025-07-17
feat: lql refactor / field selection additions

amokan:
Major refactoring effort around LQL in order to make it a bit more abstracted away from BQ logic.

Aside from refactoring the LQL modules, the functionality was extended to support:

- [x] selecting certain fields via `select:my_field.that.is.nested`  or `s:my.field`
- [x] selecting all fields with `s:*`

Later PRs will include:

- selecting from multiple tables (_via a union all_) via `from:my_source` (_decouple LQL from schemas_)
- ClickHouse support for LQL

PR #2541, opened 2025-07-17
feat: rework clickhouse connection pool handling

amokan:
This work prepares us for more robust ClickHouse reads by spinning up a `ClickhouseAdaptor.ConnectionManager` to handle connections for both ingest and query scenarios. Connection resources are released after predefined periods of inactivity (_currently set to 5 mins, based on prior work around BQ I bumped into_).

Unit testing around the overall `ClickHouseAdaptor` has also been improved.

PR #2585, opened 2025-07-31
feat: endpoint query backend abstractions / initial clickhouse endpoint query support

amokan:
While the primary goal is to handle clickhouse endpoint queries by following up on the work done within [PR#2541](https://github.com/Logflare/logflare/pull/2541) and [PR#2560](https://github.com/Logflare/logflare/pull/2560)... This PR also includes the refactoring of the general endpoint query handling by moving dialect-specific logic into the relevant backend adaptors as requested by @Ziinc.

This does not cover any aspects of LQL related to endpoint queries as that will occur after [PR#2542](https://github.com/Logflare/logflare/pull/2542) is reviewed/merged.


lib/logflare/backends/adaptor.ex
```diff
@@ -95,6 +100,69 @@ defmodule Logflare.Backends.Adaptor do
     |> mod.validate_config()
   end
 
+  @doc """
+  Returns the list of supported query languages for a given backend.
+  Returns an empty list if the adaptor does not support querying.
+  """
+  @spec get_supported_languages(Backend.t()) :: [atom()]
+  def get_supported_languages(%Backend{} = backend) do
```
Ziinc: i don't think there are any backends that will support beyond its dialect and lql, so this seems unecessary, we can rely on pattern matching in execute_query instead 🤔 
not opinionated, but seems like use case for it is minimal
and if backend doesn't support querying, then :not_implemented would be returned.

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -418,6 +515,56 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
   defdelegate append_managed_sa_to_iam_policy(user), to: CloudResourceManager
   defdelegate append_managed_service_accounts(project_id, policy), to: CloudResourceManager
   defdelegate patch_dataset_access(user), to: Google.BigQuery
-
   defdelegate get_conn(conn_type), to: GenUtils
+
+  @spec process_bq_error(map() | atom(), integer()) :: map()
```
Ziinc: we can move these to a specific bq utilitties module to avoid polluting the main module

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -688,4 +724,75 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
         []
     end
   end
+
+  @spec get_source_backend_for_query(Backend.t()) :: source_backend_tuple()
+  defp get_source_backend_for_query(%Backend{} = backend) do
```
Ziinc, line 729: we should be using the endpoint_queries.backend_id  to determine this? should default to system backend or return error when no backend is set for clickhouse.
can consider a default_query flag for backends for their language (eg. one clickhouse backend set as a defualt_query backend at any point in time)

lib/logflare/backends/adaptor.ex
amokan: Fair enough.

My thought when doing this was that not everything supports LQL out of the gates and we have some weird logic floating around for PG that supports BQ dialect and does that translation.

But I will change it.

amokan:
Putting the breaks on this PR for now to rework the connection manager/sup tree for CH.

amokan:
Closing this PR in favor of #2697

PR #2591, opened 2025-07-31
refactor: remove parens on zero arity functions and add parens for function calls

nelsonmestevao:


Ziinc:
@nelsonmestevao might be easier to break the PR up so that there would be less conflicts

PR #2600, opened 2025-08-05
Display query cost in UI

msmithstubbs:
This PR follows on from #2570 

- [x] query/endpoints/alerts uis to show the bytes/slots estimate (via dry run) and then actual run values when testing/running the query in the ui. These won't apply for CH backend.
- [x] Set disable cache when running a query from Endpoint, Alert, or Query UI.

image
image

closes ANL-980


lib/logflare/endpoints.ex
```diff
@@ -217,8 +217,8 @@ defmodule Logflare.Endpoints do
   @doc """
   Runs a an endpoint query
   """
-  @spec run_query(Query.t(), params :: map()) :: run_query_return()
-  def run_query(%Query{} = endpoint_query, params \\ %{}) do
+  @spec run_query(Query.t(), params :: map(), opts :: list()) :: run_query_return()
+  def run_query(%Query{} = endpoint_query, params \\ %{}, opts \\ []) do
```
Ziinc, line 223: would be good to add typespecs for the opts
https://hexdocs.pm/elixir/1.18.4/typespecs.html#keyword-lists

msmithstubbs: I've declared `run_query_options` and then refactored `run_query_string_opts` to include them.

`run_query_string_opts` is declared as a private type - wasn't sure if I should do same with `run_query_options`?

lib/logflare/endpoints.ex
```diff
@@ -279,12 +280,20 @@ defmodule Logflare.Endpoints do
...
+           | run_query_options()
   @typep language :: :bq_sql | :pg_sql | :lql
   @spec run_query_string(User.t(), {language(), String.t()}, run_query_string_opts()) ::
           run_query_return()
-  def run_query_string(user, {language, query_string}, opts \\ %{}) do
-    opts = Enum.into(opts, %{sandboxable: false, parsed_labels: %{}, params: %{}})
+  def run_query_string(user, {language, query_string}, opts \\ []) do
+    opts =
+      [sandboxable: false, use_query_cache: true, parsed_labels: %{}, params: %{}]
+      |> Keyword.merge(opts)
```
msmithstubbs, line 294: `opts` was declared as a keyword list in `run_query_string_opts` but used a map as the default in the function. I've switched to a keyword list for consistency.

lib/logflare/endpoints.ex
Ziinc: Yea this is fine 👍

lib/logflare_web/components/query_components.ex
```diff
@@ -0,0 +1,62 @@
...
+    {1.3, "MB"}
+
+    iex> humanize_bytes(50 * 1024 * 1024 * 1024)
+    {50.0, "GB"}
+
+    iex> humanize_bytes(2 * 1024 * 1024 * 1024 * 1024)
+    {2.0, "TB"}
+
+  """
+  @spec humanize_bytes(integer) :: {float(), String.t()}
```
Ziinc, line 45: Can see how this would be useful as a generic func in LogflareWeb.Utils,
Can we move it there 🙏

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -8,6 +8,7 @@ defmodule LogflareWeb.Source.SearchLV do
   import LogflareWeb.ModalLiveHelpers
   import LogflareWeb.SearchLV.Utils
   import LogflareWeb.SearchLive.TimezoneComponent
+  import LogflareWeb.QueryComponents, only: [humanize_bytes: 1]
```
Ziinc, line 11: Utils module alias would be better, imports make global find/replace refactoring more tricky

Ziinc (approved):

PR #2602, opened 2025-08-05
feat: Default ingest backend implementation

amokan:
### Default Ingest Backends

**Requirements**

- [x] User can set an ingesting backend (_BQ, CH, PG, S3, etc_) as a 'default ingest' backend using boolean flag on `backends` table
- [x] Default ingest backend is on top of the system-wide backend (_bq for the service_) that the application starts up with
- [x] Default ingesting backend needs client back-pressure (_429_) as well
- [x] Default ingesting backend needs to be enabled/disabled at a source-level for testing on prod
- [x] Add new attributes to the OpenApi schemas and JSON serialization for API
- [x] Add UX for toggling source flag
- [x] Add UX for toggling backend flag
- [x] Should be feature flagged with config cat for the 'multibackends' flag
- [x] sync SourceSup on change
- [x] cross cluster syncing


lib/logflare/backends.ex
```diff
@@ -197,8 +208,16 @@ defmodule Logflare.Backends do
             :update ->
               SourceSup.start_backend_child(source, backend)
 
+              if source.v2_pipeline do
+                Cluster.Utils.rpc_multicall(SourceSup, :start_backend_child, [source, backend])
+              end
```
Ziinc, line 213:
```suggestion
              Cluster.Utils.rpc_multicall(SourceSup, :start_backend_child, [source, backend])
```

lib/logflare/backends.ex
```diff
@@ -197,8 +208,16 @@ defmodule Logflare.Backends do
...
+
             :replace ->
               SourceSup.stop_backend_child(source, backend)
+
+              if source.v2_pipeline do
+                Cluster.Utils.rpc_multicall(SourceSup, :stop_backend_child, [source, backend])
+              end
```
Ziinc, line 220:
```suggestion
              Cluster.Utils.rpc_multicall(SourceSup, :stop_backend_child, [source, backend])
```

lib/logflare/backends.ex
```diff
@@ -263,6 +282,53 @@ defmodule Logflare.Backends do
...
+        sources
+
+      %Backend{} = backend ->
+        backend = Repo.preload(backend, :sources)
+        backend.sources
+
+      nil ->
+        []
+    end
+  end
```
Ziinc, line 301: would prefer to have listing function equivalent to `Backends.list_backends` but for sources, so:
`Sources.list_sources(backend_id: backend_id)`. 

`Backends.get_backend/1` would also not preload sources so it would never have any sources set there, in general context functions shouldn't hit the context's cache, otherwise it gets hard to tell what is potentially stale and what is fresh.

lib/logflare/backends.ex
```diff
@@ -263,6 +282,53 @@ defmodule Logflare.Backends do
...
+  @doc """
+  Syncs backend across all cluster nodes for v2 pipeline sources.
+
+  This ensures that when a backend's configuration changes (e.g., `default_ingest?` flag),
+  all running `SourceSup` instances are updated immediately without waiting for the
+  periodic `SourceSupWorker` check.
+  """
+  @spec sync_backend_across_cluster(integer()) :: :ok
+  def sync_backend_across_cluster(backend_id) when is_integer(backend_id) do
+    case __MODULE__.Cache.get_backend(backend_id) do
```
Ziinc, line 312: likewise, doing a `Sources.list_sources` and working from fresh data would be much cleaner.

Also, anything in the Backends.* namespace is always for v2 sources, so we don't need to check for v2 pipeline.

lib/logflare/backends.ex
```diff
@@ -483,22 +549,63 @@ defmodule Logflare.Backends do
...
+
+    case default_backends do
+      [] ->
+        # fall back to checking all queues
+        check_all_queues(source)
+
+      backends ->
+        # Always check system default queue (BQ/PG)
+        system_default_full? =
+          case PubSubRates.Cache.get_local_buffer(id, nil) do
```
Ziinc, line 578: the local buffer count is made up of multiple queues for BQ (via DynamicPipeline), and a single queue for PG. In this case, it is also a single queue for CH (until we add in DynamicPipeline for CH)
so we will need to check ALL queues and see if any of the queues are.

Fetching all backends would also not be fast enough (as this is on each api hit) and in event of db hit, all api requests are blocked and can cause memory issues, hence why the queue counts are cached inside PubSubRates.Cache under queues and queues publish their counts

we can merge this in first and reiterate on it, as PR getting large.

Ziinc (approved):

Ziinc:
@amokan i think we'll need clickhouse added to the ci

PR #2608, opened 2025-08-06
fix: don't recreate LogEvent from cached LogEvent

msmithstubbs:
This PR fixes an issue I came across when working on #2606:

1. The first time an event is loaded it's map which is transformed into a `%Loglare.LogEvent{}` struct and cached
2. If the same event is loaded again the cached structure is fetched. This threw an exception as `make_from_db` doesn't accept a struct

This PR only caches the `LogEvent` when it is built. I couldn't see a reason to cache every time but don't know too much about the caching infrastructure. Maybe it was intentional to cache it every time?


Ziinc:
can we add a  test please so that we don't have this regression again

Ziinc (approved):

PR #2614, opened 2025-08-08
fix: prevent Tesla.Env authorization headers leaking

Ziinc:


lib/logflare/utils.ex
```diff
@@ -205,4 +210,45 @@ defmodule Logflare.Utils do
...
+      if String.downcase(key) in ["authorization", "x-api-key"] do
+        {key, "REDACTED"}
+      else
+        {key, value}
+      end
+    end
+  end
+end
+
+defimpl String.Chars, for: Tesla.Env do
```
wojtekmach: Tesla does not implement String.Chars for Tesla.Env by default, `String.Chars.impl_for(%Tesla.Env{}) #=> nil` and `to_string(%Tesla.Env{})` crashes on undefined protocol error. Perhaps we don't need to worry about this at all? Or some third party dependency is adding this protocol implementation?

lib/logflare/utils.ex
```diff
@@ -205,4 +210,45 @@ defmodule Logflare.Utils do
...
+      apply(Logflare.Utils.tesla_env_to_string(), :to_string, [env])
+    else
+      redacted_headers = Logflare.Utils.redact_sensitive_headers(headers)
+
+      apply(Logflare.Utils.tesla_env_to_string(), :to_string, [%{env | headers: redacted_headers}])
+    end
+  end
+end
+
+defimpl Inspect, for: Tesla.Env do
```
wojtekmach: At the moment, Tesla.Env does not have custom Inspect implementation (it uses Inspect.Any) so instead of `@original_inspect Inspect.impl_for(%Tesla.Env{})` and `Logflare.Utils.tesla_env_inspect()` wrapper, I think you can call Inspect.Any.inspect directly:

```elixir
defimpl Inspect, for: Tesla.Env do
  def inspect(env, options) do
    env = %{env | headers: ...}
    Inspect.Any.inspect(env, options)
  end
end
```

That said, if Tesla, or any other third party package, starts implementing Inspect for Tesla.Env, you'll get a warning on protocol redefinition so it's really helpful to compile with `--warnings-as-errors` to catch things like that. That said, if there's a possibility of that ever happening and it slipping through then you could get into weird scenarios where the order of modules being compiled decides which protocol implementation is the one.

Our recommendation is to use https://hexdocs.pm/elixir/main/Inspect.Opts.html#default_inspect_fun/1 in your application start/2, that should be the most future proof as it works completely at runtime:

```elixir
previous_fun = Inspect.Opts.default_inspect_fun()

Inspect.Opts.default_inspect_fun(fn
  %Tesla.Env{} = env, opts ->
    env = %{env | headers: ...}
    previous_fun.(env, opts)

  value, opts ->
    previous_fun.(value, opts)
end)
```

Btw, if you're setting headers via Tesla.Middleware.Headers middleware, they could leak when you print the client:

```elixir
client =
  Tesla.client([
    {Tesla.Middleware.Headers, [{"authorization", "secret"}]}
  ])

dbg(client)
```

```elixir
client #=> %Tesla.Client{
  fun: nil,
  pre: [{Tesla.Middleware.Headers, :call, [[{"authorization", "secret"}]]}],
  post: [],
  adapter: nil
}
```

And just for completeness, if you use Tesla.Middleware.KeepRequest, they could leak too:


```elixir
client =
  Tesla.client([
    {Tesla.Middleware.Headers, [{"authorization", "secret"}]},
    Tesla.Middleware.KeepRequest
  ])

env = Tesla.get!(client, "https://httpbin.org/anything")
dbg(env.opts[:req_headers])
#=> [{"authorization", "secret"}]
```

wojtekmach (commented):
I left some comments below, let us know what you think!

lib/logflare/utils.ex
Ziinc: yea since this is a security concern, I'll lean on the conservative side and cover all bases even though upstream doesn't implement it.

chasers (approved):

lib/logflare/utils.ex
Ziinc: I went with using the default_inspect_fun as suggested, as well as adding in a comprehensive struct redaction for Tesla.Client and Tesla.Env 😄

PR #2618, opened 2025-08-08
refactor: extract nested code in lib/logflare_web/live/backends/backends_live.ex

nelsonmestevao:

### DIFF
--- a/lib/logflare_web/live/backends/backends_live.ex
+++ b/lib/logflare_web/live/backends/backends_live.ex
@@ -63,15 +63,7 @@ defmodule LogflareWeb.BackendsLive do
         Map.put(config, "headers", %{key => value})
         |> case do
           %{"metadata" => metadata_str} = config when is_binary(metadata_str) ->
-            metadata =
-              metadata_str
-              |> String.split(",")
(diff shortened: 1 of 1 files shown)


lib/logflare_web/live/backends/backends_live.ex
```diff
@@ -289,4 +263,33 @@ defmodule LogflareWeb.BackendsLive do
...
+    |> String.split(",")
+    |> Enum.reduce(%{}, fn pair, acc ->
+      case String.split(pair, "=", parts: 2) do
+        [key, value] -> Map.put(acc, String.trim(key), String.trim(value))
+        _ -> acc
+      end
+    end)
+  end
+
+  defp changeset_to_flash_message(changeset) do
```
Ziinc, line 278: Would be good to extract this out to it's own utils function as this is used in quite a lot of places.

Ziinc (approved):

lib/logflare_web/live/backends/backends_live.ex
nelsonmestevao: Done in https://github.com/Logflare/logflare/pull/2680 :pray:

PR #2632, opened 2025-08-11
fix: support older tables partitioned with _PARTITIONTIME

msmithstubbs:
This PR fixes a regression where the log event viewer could not fetch an event from an older source using the `_PARTITONTIME`  field.

`fetch_event_by_id/3` now checks the source partition type and uses `timestamp` or `_PARTITIONTIME` depending on the source type.

### DIFF
--- a/lib/logflare/logs/log_events.ex
+++ b/lib/logflare/logs/log_events.ex
@@ -65,7 +65,6 @@ defmodule Logflare.Logs.LogEvents do
   end
 
   def partition_query(query, [min, max], :pseudo) do
-
     where(
       query,
       [t],


lib/logflare/logs/log_events.ex
```diff
@@ -56,4 +57,24 @@ defmodule Logflare.Logs.LogEvents do
...
   end
+
+  def partition_query(query, [min, max], :timestamp) do
+    query
+    |> where([t], t.timestamp >= ^min)
+    |> where([t], t.timestamp <= ^max)
+  end
+
+  def partition_query(query, [min, max], :pseudo) do
+    {min, Timex.to_date(min)}
```
rogerleite: I think you forgot to remove this line.

chasers:
```suggestion
```

chasers (changes requested):

lib/logflare/logs/log_events.ex
```diff
@@ -56,4 +57,23 @@ defmodule Logflare.Logs.LogEvents do
...
   end
+
+  def partition_query(query, [min, max], :timestamp) do
+    query
+    |> where([t], t.timestamp >= ^min)
+    |> where([t], t.timestamp <= ^max)
+  end
+
+  def partition_query(query, [min, max], :pseudo) do
+
```
chasers:
```suggestion
```

chasers (approved):

PR #2659, opened 2025-08-17
refactor: rework of the default backend logic

amokan:
Relies on the `sources_backends` join table as this allows for more efficient/targeted lookups when ingesting logs through existing functions such as `Logflare.Backends.list_backends/1` and its cached implementation. While this does go against the idea of moving away from the join table, it seems to be the most sane way to handle this situation.

- UX updated to behave similarly to the drain rules form shown when viewing a backend (_see screenshot below_)
- Only allows the selection of a source that has `default_ingest_backend_enabled?` set to true
- Handles log ingest and routing logic
- Spins up backend supervision trees when added as a default ingest target
- Load tests show no breakdown or connection issues when working with Clickhouse Cloud and BigQuery for backends

Screenshot 2025-08-17 at 2 09 53 PM

### DIFF
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -2,26 +2,29 @@ defmodule Logflare.Backends do
   @moduledoc false
 
   import Ecto.Query
+  import Logflare.Utils.Guards
 
+  alias Ecto.Changeset
   alias Logflare.Backends.Adaptor
   alias Logflare.Backends.Backend
+  alias Logflare.Backends.IngestEventQueue
+  alias Logflare.Backends.RecentEventsTouch
   alias Logflare.Backends.SourceRegistry
   alias Logflare.Backends.SourcesSup
   alias Logflare.Backends.SourceSup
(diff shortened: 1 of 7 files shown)


lib/logflare/backends.ex
```diff
@@ -287,8 +419,11 @@ defmodule Logflare.Backends do
   @spec sync_backend_across_cluster(integer()) :: :ok
   def sync_backend_across_cluster(backend_id) when is_integer(backend_id) do
     with %Backend{} = backend <- get_backend(backend_id) do
-      for source <- Sources.list_sources(backend_id: backend_id) do
+      sources = Sources.list_sources(backend_id: backend_id)
+
+      for source <- sources do
         Cluster.Utils.rpc_multicall(SourceSup, :start_backend_child, [source, backend])
+        Cluster.Utils.rpc_multicall(__MODULE__, :clear_list_backends_cache, [source.id])
```
Ziinc, line 426: this should already get cleared through the CacheBuster, so might result in double fetching

PR #2663, opened 2025-08-18
refactor: extract nested code in lib/logflare/sql/dialect_translation.ex

nelsonmestevao:

### DIFF
--- a/lib/logflare/sql/dialect_translation.ex
+++ b/lib/logflare/sql/dialect_translation.ex
@@ -670,67 +670,70 @@ defmodule Logflare.Sql.DialectTranslation do
         get_in(from, ["relation", "Table", "alias", "name", "value"])
       end)
 
-    for from <- from_list do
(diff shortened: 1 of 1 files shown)


lib/logflare/sql/dialect_translation.ex
```diff
@@ -670,67 +670,70 @@ defmodule Logflare.Sql.DialectTranslation do
...
-      Enum.reduce(from["joins"] || [], %{}, fn
-        %{
-          "relation" => %{
-            "UNNEST" => %{
-              "array_expr" => %{"Identifier" => %{"value" => identifier_val}},
-              "alias" => %{"name" => %{"value" => alias_name}}
-            }
+    from_list
+    |> Enum.map(&build_mappings(&1, table_aliases))
+    |> Enum.reduce(%{}, fn mappings, acc -> Map.merge(acc, mappings) end)
```
Ziinc, line 675: This would iterate over the list twice, combining both into a reduce would be more efficient.

PR #2668, opened 2025-08-18
refactor: extract nested code in lib/logflare/lql/parser.ex

nelsonmestevao:


lib/logflare/lql/parser.ex
```diff
@@ -182,6 +174,17 @@ defmodule Logflare.Lql.Parser do
       {:error, err}
   end
 
+  defp build_chart_rule(chart_rule_tokens, typemap, querystring) do
+    if not Enum.empty?(chart_rule_tokens) do
```
Ziinc, line 178: We can remove this with an pattern match function head

PR #2671, opened 2025-08-18
refactor: extract nested code in lib/logflare/logs/vercel/nimble_vercel_parser.ex

nelsonmestevao:

### DIFF
--- a/lib/logflare/logs/vercel/nimble_vercel_parser.ex
+++ b/lib/logflare/logs/vercel/nimble_vercel_parser.ex
@@ -180,16 +180,15 @@ defmodule Logflare.Logs.Vercel.NimbleLambdaMessageParser do
   defp build_map(maybe_json, message, tokens) do
     cleaned_json = String.replace(maybe_json, "\n", "")
 
-    with {:error, _} <- JSON.decode(cleaned_json),
-         {"lines", lines} <- maybe_parse_multiline_json_body(maybe_json) do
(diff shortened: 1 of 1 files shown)


lib/logflare/logs/vercel/nimble_vercel_parser.ex
```diff
@@ -168,30 +171,28 @@ defmodule Logflare.Logs.Vercel.NimbleLambdaMessageParser do
...
 
+  defp build_map(maybe_json, message, tokens) do
+    cleaned_json = String.replace(maybe_json, "\n", "")
+
+    with {:error, _} <- JSON.decode(cleaned_json),
+         {"lines", lines} <- maybe_parse_multiline_json_body(maybe_json) do
+      %{"multiline" => lines}
+    else
+      {:ok, json} ->
+        %{"data" => json, "message" => message}
```
Ziinc: would prefer the happy path to be within the with/do 🙏

Ziinc (changes requested):

lib/logflare/logs/vercel/nimble_vercel_parser.ex
nelsonmestevao: Will change it :pray:

nelsonmestevao: Done in last push @Ziinc :pray:

PR #2674, opened 2025-08-18
fix: cleaner label keys and values for BigQuery

chasers:
- cleans query labels according to the BigQuery docs

Query job history in BigQuery:

```
SELECT
  job_id,
  creation_time,
  user_email,
  query,
  ARRAY_AGG(STRUCT(l.key AS key, l.value AS value)) AS labels
FROM
  `region-us`.INFORMATION_SCHEMA.JOBS_BY_PROJECT j
LEFT JOIN
  UNNEST(j.labels) AS l
WHERE
  creation_time >= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 HOUR)
  AND statement_type = "SELECT" -- remove this if you want all statements
GROUP BY
  job_id,
  creation_time,
  user_email,
  query
ORDER BY
  creation_time DESC
LIMIT 100;
```

### DIFF
--- a/assets/tailwind.config.js
+++ b/assets/tailwind.config.js
@@ -1,7 +1,7 @@
 /** @type {import('tailwindcss').Config} */
 module.exports = {
   content: [
-    "./**/*.{js,jsx,ts,tsx}",
+    "./js/*.{js,jsx,ts,tsx}",
     "../lib/logflare_web/*.{eex,leex,heex,ex}",
     "../lib/logflare_web/**/*.{eex,leex,heex,ex}",
   ],
--- a/lib/logflare/logs/source_parsers/syslog_parser.ex
+++ b/lib/logflare/logs/source_parsers/syslog_parser.ex
@@ -133,17 +133,20 @@ defmodule Logflare.Logs.SyslogParser do
     if Enum.empty?(sd_element_values) do
       tokens
     else
-      sd =
-        sd_element_values
-        |> Enum.map(fn sd_element ->
-          Enum.reduce(sd_element, %{}, fn
-            {:sd_name, sd_name}, acc -> Map.put(acc, "id", sd_name)
-            [param_name: k, param_value: v], acc -> Map.put(acc, k, v)
-          end)
-        end)
+      sd = build_sd(sd_element_values)
 
       new_tokens ++ [sd: sd]
     end
+  end
+
+  defp build_sd(sd_element_values) do
+    sd_element_values
+    |> Enum.map(fn sd_element ->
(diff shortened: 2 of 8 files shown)


lib/logflare/ecto/bigquery/bq_repo.ex
```diff
@@ -33,6 +33,21 @@ defmodule Logflare.BqRepo do
...
+      %{
+        "managed_by" => "logflare",
+        "logflare_plan" => plan,
+        "logflare_account" => user.id
+      }
+      |> Map.merge(Map.get(override, :labels, %{}))
+      |> Enum.map(fn {k, v} ->
+        {GenUtils.format_key(k), GenUtils.format_key(v)}
+      end)
+      |> Enum.into(%{})
```
Ziinc: This will iterate over the enum twice, can be done in one pass with Enum.reduce.

test/logflare_web/controllers/endpoints_controller_test.exs
```diff
@@ -280,8 +280,8 @@ defmodule LogflareWeb.EndpointsControllerTest do
       assert labels == %{
                "my_label" => "my_value",
                "other_value" => "1234",
-               "endpoint_id" => endpoint.id,
-               "logflare_account" => user.id,
+               "endpoint_id" => Integer.to_string(endpoint.id),
+               "logflare_account" => Integer.to_string(user.id),
```
Ziinc: Letting the GenUtil.format_key/1 do the stringification would have less gotchas

Ziinc (approved):
Might want to check if BQ is able to handle the stripped ISO timestamps, otherwise a different strategy such as converting the timestamp to a unix timestamp in the label declaration might be better.

For example something like this might work better if BQ can't directly read the cleaned timestamp labels 

```
my_timestamp=@iso_timestamp|>unix,...
```

lib/logflare/ecto/bigquery/bq_repo.ex
chasers: Used Map.new

chasers:
> Might want to check if BQ is able to handle the stripped ISO timestamps, otherwise a different strategy such as converting the timestamp to a unix timestamp in the label declaration might be better.
> 
> For example something like this might work better if BQ can't directly read the cleaned timestamp labels
> 
> ```
> my_timestamp=@iso_timestamp|>unix,...
> ```

Yeah okay it's handling iso timestamps now and converting to unix. If we want to do anything programmatically with these param values from the query logs it will be a lot easier now.

PR #2684, opened 2025-08-21
feat: event context viewer

msmithstubbs:
Depends on #2662 and #2711 

* upto 50 events before and 50 after a selected event can be viewed in a modal
* on opening the modal we ensure the selected event is visible. 
* if their are 50 before or after a message is displayed to communicate that events have been truncated to 50.
* search tailing, if enabled, is paused while the modal is open, and resumed when closed.
* I'd welcome any suggestions on optimising context search query

https://github.com/user-attachments/assets/bd117bb5-9412-4c85-98dc-b794ece323b7

Searches that include a suggested key will be used in the context view:

CleanShot 2025-08-29 at 15 07 20@2x


assets/js/source_lv_hooks.js
```diff
@@ -224,6 +224,22 @@ hooks.SourceLogsSearch = {
...
+    ...
+  </div>
+*/
+hooks.ScrollIntoView = {
+  mounted() {
+    this.el.addEventListener("scrollIntoView", (event) => {
+      event.target.scrollIntoView({ behavior: "instant" });
+    });
+  },
+};
```
msmithstubbs, line 243: Attempted to avoid a phx-hook here but couldn't make it work. Can revisit, otherwise I'm keen to update LiveView to 1.1 and use colocated hooks.

test/logflare_web/live/search_live/event_context_component_test.exs
```diff
@@ -0,0 +1,233 @@
...
+          legacy_source.id,
+          lql_rules
+        )
+
+        assert_receive {^ref, query}
+
+        case partition_type do
+          :pseudo ->
+            assert query =~ "_PARTITIONTIME"
+            assert query =~ "_PARTITIONDATE IS NULL"
```
msmithstubbs, line 166: Does this cover the requirements for legacy partitioning?

lib/logflare/logs/logs_search.ex
```diff
@@ -116,6 +116,75 @@ defmodule Logflare.Logs.Search do
...
+      |> order_by([t], asc: t.timestamp, asc: t.id)
+
+    with so <- %{so | query: query},
+         so = %{error: nil} <- do_query(so) do
+      {:ok, so}
+    else
+      so -> {:error, so}
+    end
+  end
+
```
msmithstubbs, line 173: Not sure what to do with this. Could break it up into separate functions but it will just be hiding the complexity rather than reducing it. Suggestions appreciated 🙏

Ziinc: union all doesn't seem very efficent, it can be done in a singular query:

```
WITH numbered_rows AS (
    SELECT *,
           ROW_NUMBER() OVER (ORDER BY id) as row_num
    FROM source_table
    WHERE .... -- clustering/suggested filters
),
target_position AS (
    SELECT row_num as target_row_num
    FROM numbered_rows
    WHERE   ....  -- clustering/suggested filters
      AND id = ... -- replace with target ID
      
)
SELECT nr.*
FROM numbered_rows nr
CROSS JOIN target_position tp
WHERE nr.row_num BETWEEN (tp.target_row_num - 5) AND (tp.target_row_num + 5)
ORDER BY nr.id;
```

lib/logflare_web/live/search_live/event_context_component.ex
```diff
@@ -0,0 +1,210 @@
...
+      </div>
+      <div class="tw-h-[calc(100vh-200px)] tw-overflow-y-auto tw-pr-2 tw-pl-5 -tw-ml-5" id="context_log_events" phx-hook="ScrollIntoView" phx-value-scroll-target={@target_event_id}>
+        <.async_result assign={@logs}>
+          <:loading><%= live_react_component("Components.Loader", %{}, id: "shared-loader") %></:loading>
+          <:failed>
+            <div id="context_log_events_error">An error occurred.</div>
+          </:failed>
+
+          <div :if={@is_truncated_before} class="tw-text-center tw-py-2 tw-uppercase tw-text-sm">
+            Limit 50 events before selection
```
Ziinc:
```suggestion
            Showing 50 events before selected event
```

lib/logflare/logs/logs_search.ex
msmithstubbs: I did attempt to do it with one query but figured it would require select *all* rows within the min/max timestamps, and then trimming down to 50 before and after. If you had high frequency event in that range it would mean discarding a bunch of rows, as opposed to a couple of queries with LIMIT. 

Will give this another try.

msmithstubbs:
Refactored the query and added the feature flag.

test/logflare_web/live/search_live/event_context_component_test.exs
Ziinc: yea this will cover situations where events are still in the streaming buffer

Ziinc (approved):

PR #2692, opened 2025-08-25
refactor: extract nested code in lib/logflare/sources/source_routing.ex

nelsonmestevao:

### DIFF
--- a/lib/logflare/sources/source_routing.ex
+++ b/lib/logflare/sources/source_routing.ex
@@ -61,66 +61,48 @@ defmodule Logflare.Logs.SourceRouting do
   @spec route_with_lql_rules?(LE.t(), Rule.t()) :: boolean()
   def route_with_lql_rules?(%LE{body: le_body}, %Rule{lql_filters: lql_filters})
       when lql_filters != [] do
-    lql_rules_match? =
-      Enum.reduce_while(lql_filters, true, fn lql_filter, _acc ->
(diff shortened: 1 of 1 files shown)


lib/logflare/sources/source_routing.ex
```diff
@@ -61,66 +61,48 @@ defmodule Logflare.Logs.SourceRouting do
...
+      |> Enum.any?(fn le_value ->
+        evaluate_filter_condition(lql_filter, le_value)
+      end)
+    end)
+  end
 
-                operator == :list_includes ->
-                  le_value == value
+  defp evaluate_filter_condition(lql_filter, le_value) do
+    %FilterRule{value: value, operator: operator, modifiers: mds} = lql_filter
```
Ziinc:
```suggestion
    %FilterRule{value: value, operator: operator, modifiers: modifiers} = lql_filter
```

lib/logflare/sources/source_routing.ex
```diff
@@ -61,66 +61,48 @@ defmodule Logflare.Logs.SourceRouting do
...
-        else
-          {:halt, false}
-        end
-      end)
+        operator in [:<=, :<, :>=, :>] ->
+          apply(Kernel, operator, [le_value, value])
+      end
 
-    lql_rules_match?
+    if mds[:negate], do: not matches?, else: matches?
```
Ziinc:
```suggestion
    if modifiers[:negate], do: not matches?, else: matches?
```

Ziinc (approved):

PR #2694, opened 2025-08-25
refactor: improve nested code in lib/logflare/lql/backend_transformer/bigquery.ex

nelsonmestevao:


lib/logflare/lql/backend_transformer/bigquery.ex
```diff
@@ -201,25 +201,18 @@ defmodule Logflare.Lql.BackendTransformer.BigQuery do
     path
     |> split_by_dots()
     |> Enum.slice(0..-2//1)
-    |> case do
-      [] ->
-        query
```
nelsonmestevao, line 206: This was a redundant case since Enum module does nothing when enumerable is empty 🙏

PR #2697, opened 2025-08-26
feat: endpoint backend abstractions / clickhouse endpoint query support

amokan:
Includes the following:

- Initial implementation of query support for Clickhouse backends
- Refactoring of the general endpoint query handling by moving dialect-specific logic into the relevant backend adaptors


lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -421,6 +417,119 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
...
+    {method, options} = Keyword.pop!(options, :method)
+    {url, options} = Keyword.pop!(options, :url)
+    {headers, options} = Keyword.pop!(options, :headers)
+    {body, options} = Keyword.pop!(options, :body)
+
+    Finch.build(method, url, headers, body)
+    |> Finch.request(Logflare.FinchGoth, options)
+  end
+
+  defp execute_query_with_context(
```
Ziinc, line 472: not a fan this fat function where it is trying to do so much and cater for so many querying situations, would be cleaner to have a  function head catering to each query situation (e.g. endpoints/alerts/backend+string/backend+ecto) that finally call to the base `execute_query` that essentially only wraps  `BqRepo.query_with_sql_and_params/5`

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -421,6 +417,119 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
   defdelegate append_managed_sa_to_iam_policy(user), to: CloudResourceManager
   defdelegate append_managed_service_accounts(project_id, policy), to: CloudResourceManager
   defdelegate patch_dataset_access(user), to: Google.BigQuery
-
   defdelegate get_conn(conn_type), to: GenUtils
+
+  # handles pagination for the IAM api
+  defp get_next_page(project_id, page_token) do
```
Ziinc, line 423: lets do out of scope refactoring/reordering in separate smaller PRs in the future to make the review process faster

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -624,4 +655,43 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
...
+  defp convert_query_params(sql_statement, allowed_params)
+       when is_non_empty_binary(sql_statement) and is_list(allowed_params) do
+    allowed_set = MapSet.new(allowed_params)
+
+    # Convert `@param` syntax to ClickHouse `{param:String}` syntax
+    Regex.replace(~r/@(\w+)/, sql_statement, fn match, param ->
+      if MapSet.member?(allowed_set, param) do
+        "{#{param}:String}"
+      else
+        Logger.warning("SQL parameter `@#{param}` not in allowed list")
```
Ziinc, line 692:
```suggestion
```
don't think we need this since it is dependent on user input

test/logflare/backends/adaptor/clickhouse_adaptor_test.exs
```diff
@@ -173,6 +175,42 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptorTest do
...
+      result = ClickhouseAdaptor.execute_query(backend, "SELECT 1 as test_value", [])
+
+      assert {:ok, [%{"test_value" => 1}]} = result
+    end
+
+    test "converts `@param` syntax to ClickHouse `{param:String}` format", %{backend: backend} do
+      result =
+        ClickhouseAdaptor.execute_query(
+          backend,
+          {"SELECT @test_value as param_result", ["test_value"], %{"test_value" => "hello"}},
```
Ziinc, line 200: perhaps to cut down on duplication of the declared_params and the input_params, perhaps the params map could always be a map of all declared params (with non-declared ones filtered out), such that we don't need to do any validation logic at the Adaptor layer. Validation logic for the params then comes in at the Endpoints context layer instead.
The queryable adaptors are only concerned with the sql string and the named/positional params in the string which is the bare minimum to run the query, so bringing down the concept of "declared" and  "input" to the adaptor layer seems unnecessary

amokan: Makes sense.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
amokan: PR for this - https://github.com/Logflare/logflare/pull/2713

lib/logflare/backends/adaptor/bigquery_adaptor.ex
amokan: I address this function in PR #2714  - will follow that up with some work in `Logflare.Alerting` to remove the hardcoded hit on the BqRepo there.

PR #2701, opened 2025-08-27
feat: add cluster information to otel metrics exporter

andremaker:
Adds cluster information set on runtime.exs to the OtelMetricsExporter's resource information


lib/telemetry.ex
```diff
@@ -40,7 +40,8 @@ defmodule Logflare.Telemetry do
               name: "Logflare",
               version: Application.spec(:logflare, :vsn) |> to_string()
             },
-            instance: inspect(Node.self())
+            instance: inspect(Node.self()),
```
```suggestion
            node: inspect(Node.self()),
```

Ziinc (changes requested):
seems like env var might be a much better and more consistent way of approaching this:
https://opentelemetry.io/docs/languages/js/resources/#setup
https://github.com/search?q=repo%3Aelectric-sql%2Felixir-otel-metric-exporter%20OTEL_RESOURCE_ATTRIBUTES&type=code

then we can keep the OtelMetricsExporter `:resources` in sync with our exported traces from the official clients.

lib/telemetry.ex
andremaker: i was thinking... that was the name already, before my changes. Would not changing the key cause issues with backward compatibility for the clients?

config/runtime.exs
```diff
@@ -388,6 +388,7 @@ if System.get_env("LOGFLARE_OTEL_ENDPOINT") do
   config :opentelemetry,
     sdk_disabled: false,
     traces_exporter: :otlp,
+    resource: logflare_metadata,
```
Ziinc: please verify that the resource format for traces and metrics match on dev. as is, the `:resource` field from metrics differs greatly from the traces. tracing format should take precedence.

Ziinc: trace: 
Screenshot 2025-09-05 at 3 48 54 AM

metric
image

Ziinc (changes requested):

config/runtime.exs
andremaker: should then i create a "cluster" object and attribute the metadata to a "name" key on that object?

Ziinc: having a path of `resource.service.cluster` is good.

Ziinc: don't need separate object

andremaker: done

Ziinc (approved):

PR #2707, opened 2025-08-28
perf: replication slot optimizations and termination cleanup

### LINKED ISSUE #2375: Self-hosted Logflare in Supabase mode has orphaned replication slot
While setting up an instrumented Postgresql cluster for self-hosted supabase, I am running into replication lag alerts due to a replication slot being created by logflare.

Image

It appears that running with a postgres backend with LOGFLARE_SUPABASE_MODE=true enables the v2 pipeline:

```
   defp do_start_source_sup(%{v2_pipeline: true} = source) do
     with :ok <- Backends.start_source_sup(source) do
       do_lookup(source)
     end
   end
   
   defp do_start_source_sup(source) do
     DynamicSupervisor.start_child(V1SourceDynSup, {V1SourceSup, source: source})
   end
```

... which uses a direct connection to Postgres via the `PostgresAdaptor` that does not consume the replication slot that is still created during startup:

```
  @doc """
  Attempts to start a cainophile child in the ContextCache.Supervisor.
  If it already exists, it will return with an error tuple.
  """
  def maybe_start_cainophile do
    spec = cainophile_child_spec()
    Supervisor.start_child(__MODULE__, spec)
  end

  def remove_cainophile do
    Supervisor.terminate_child(__MODULE__, Cainophile.Adapters.Postgres)
    Supervisor.delete_child(__MODULE__, Cainophile.Adapters.Postgres)
    :ok
  end

  defp cainophile_child_spec do
    hostname = ~c"#{Application.get_env(:logflare, Repo)[:hostname]}"
    username = Application.get_env(:logflare, Repo)[:username]
    password = Application.get_env(:logflare, Repo)[:password]
    database = Application.get_env(:logflare, Repo)[:database]
    port = Application.get_env(:logflare, Repo)[:port]

    slot = Application.get_env(:logflare, CacheBuster)[:replication_slot]
    publications = Application.get_env(:logflare, CacheBuster)[:publications]

    %{
      id: Cainophile.Adapters.Postgres,
      restart: :permanent,
      type: :worker,
      start:
        {Cainophile.Adapters.Postgres, :start_link,
         [
           [
             register: publisher_name(),
             epgsql: %{
               host: hostname,
               port: port,
               username: username,
               database: database,
               password: password
             },
             slot: slot,
             wal_position: {"0", "0"},
             publications: publications
           ]
         ]}
    }
  end

```

This behavior appears ubiquitous in current docker compose setups for Supabase, and the  current local development version of Supabase as well.

Ziinc:
This PR adds in the following:
- cainophile backend pid explicit shutdown - for edge cases where backend pid in replication mode gets orphaned from cainophile process somehow.
- cainophile wal flushing, to ensure that wal retained on postgres is kept to a minimum in periods of inactivity in published tables

This also moves cluster syncing logic to the CacheBusterWorker, and also cleans up the CacheBuster module by separating out the worker module.

cainophile changes are in our fork:
- These changes will trap exits and do an explciit cleanup of the backend pid https://github.com/cainophile/cainophile/compare/267999b3142f491f2b57f9f892bc9c0ddcadb3a5...Logflare:cainophile:master
- This one will flush the wal periodically https://github.com/Logflare/cainophile/pull/2

closes #2375


lib/logflare/context_cache/supervisor.ex
```diff
@@ -28,12 +32,19 @@ defmodule Logflare.ContextCache.Supervisor do
...
+        [
+          {GenSingleton, child_spec: cainophile_child_spec()}
+        ]
 
   defp get_children(_) do
     list_caches() ++
       [
         ContextCache.TransactionBroadcaster,
+        {GenSingleton, child_spec: cainophile_child_spec()},
+        {PartitionSupervisor, child_spec: CacheBusterWorker, name: CacheBusterWorker.Supervisor},
```
chasers, line 47: I've done 100_000 changes in a second locally testing some stuff and Elixir can handle it.

I don't really see why we would need to partition this one.

lib/logflare/context_cache/cache_buster.ex
```diff
@@ -99,41 +67,16 @@ defmodule Logflare.ContextCache.CacheBuster do
...
-  end
 
-  defp maybe_do_cross_cluster_syncing({Alerting, alert_id}) do
-    # sync alert job
-    Utils.Tasks.start_child(fn ->
-      Alerting.sync_alert_job(alert_id)
+        GenServer.cast(
+          {:via, PartitionSupervisor, {__MODULE__.Supervisor, records}},
+          {:to_bust, records}
+        )
```
chasers, line 74: I guess this is already partitioned. If we are going to partition it the key should probably be the table the change is for, so that those changes are at least serialized by table.

lib/logflare/context_cache/cache_buster_worker.ex
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  defp maybe_do_cross_cluster_syncing({Rules, rule_id}) when is_integer(rule_id) do
+    # sync rule
+    Utils.Tasks.start_child(fn ->
+      Rules.sync_rule(rule_id)
+    end)
+  end
+
+  defp maybe_do_cross_cluster_syncing(_), do: :noop
```
chasers, line 56: Why do we have a separate thing for all of these in this module? 

Can't we do this on each node where they receive the wal record from PubSub? 

Seems odd to call another node from the node receiving the wal record when the wal record is already getting broadcast everywhere.

Ziinc: What you're describing is the correct behaviour, the syncing only occurs locally since it gets globally broadcasted and doesn't call another node.

Ziinc: I think the only one that does call another node is the backend syncing which slipped through in CH work.

lib/logflare/context_cache/cache_buster.ex
Ziinc: Keeping workers partitioned adds some cheap future proofing and allows more heavy state syncing functions or warming functions to be done.
We could remove the partitioning, it wouldn't affect performance much just increases risk of busting/syncing being the bottleneck

josevalim:
Hi folks, apologies as I am most likely missing some context here, but why do you have a transaction broadcaster that subscribes to cainophile and then broadcast, instead of directly broadcasting from the cainophile process through the adapter API?

Also, regarding the cainophile PR, what if instead of trapping exits by default and calling cleanup, you instead redirected all `handle_info` to the adapter?

```
  @impl true
  def handle_info(message, state) do
    adapter_impl(state.config).handle_info(message, state.connection)
  end
```

Then you could implement the logic in the adapter and be a bit more precise when checking the EXIT message. I am worried about the current `{:EXIT, _, _}` because even something simple like calling a Task in the adapter will spawn a separate process that will eventually lead to an EXIT message. And now you will call cleanup in the adapter while it is a totally expected workflow.

josevalim:
By the way, I'm sorry if I'm repeating myself, you can implement the replication logic + broadcast on top of Postgrex. Here's an example of using Phoenix PubSub plus Postgres for emitting broadcasts: https://github.com/josevalim/sync/blob/main/lib/sync/replication.ex

Ziinc:
@josevalim previously, all nodes spun up its own cainophile process. now that cainophile is a global singular process, keeping the subscription pids list in the cainophile genserver small is desirable due to autoscaler resulting in some stale dead pids over time. we haven't got around to implementing a cleanup for that yet. 

👍 for the adaptor callback forwarding

will be reworking this PR to include only the CacheBusterWorker changes and cainophile cleanup + logging changes.

@chasers i think i found the cause for the stale data behaviour we're seeing. the local broadcast here means that only the node with the TransactionBroadcaster is busting its cache, and all other nodes are lagging behind until cachex ttl kicks in.

https://github.com/Logflare/logflare/commit/a76552eb6823c5360c7e3f4939fc5fae35396e78#diff-e4e1bf588757459ad42368f78ef87b32d22430eb1af7f9afe240b05ba80c9cdeL50-R50

chasers:
@Ziinc good for now to get this out, but probably want to switch to Postgrex.ReplicationConnection and drop Cainophile completely.

chasers:
Yeah @josevalim we are killing Cainophile procs with our syn conflict handler and for some reason it's leaving replication slots open and wal is building up in them.

PR #2710, opened 2025-08-28
feat: allow optional selection of backend in the UI when configuring a new query endpoint

amokan:
Some minor UI changes to allow the selection of a backend when setting up a new query endpoint to make testing easier.

- Adds a new 'Backend' dropdown with a default option of BQ.
- When selecting a backend in the UI, the `Query` changeset is handled so the query language can be inferred based on the selected backend
- The `LogflareWeb.QueryComponents.query_cost` has been adjusted to only display when `@total_bytes_processed` is not nil. Clickhouse queries do not return metadata like this at this point.
- Because of the way we're handling validation fairly loose in the `/endpoints/new` route, I added an additional clause for `Sql.transform` that returns the query when the `User` is nil.
- This feature requires a ConfigCat flag of `"endpointBackendSelection"` to be enabled and a configured `Backend` for the dropdown to appear in the UI.

Screenshot 2025-08-28 at 3 00 15 PM


lib/logflare_web/live/endpoints/endpoints_live.ex
```diff
@@ -204,15 +218,39 @@ defmodule LogflareWeb.EndpointsLive do
...
+
   def handle_event("validate", _params, socket) do
-    # noop
+    # noop for other validation events
     {:noreply, socket}
   end
 
   def handle_info({:query_string_updated, query_string}, socket) do
+    endpoint_language =
+      Ecto.Changeset.get_field(socket.assigns.endpoint_changeset, :language, :bq_sql)
```
Ziinc: using the changeset to store and determine the language will likely lead to state issues between what is seen in the inputs vs what is submitted. would prefer to avoid changesets manipulation in the LV directly.
This has already bitten us just a week ago with dealing with the monaco editor state

Ziinc: listening to params changes and dynamically changing languages in background and merging in values with params on submission would be easier to test as well...
alternative is setting the value of hidden inputs dynamically but these won't get covered well with the LV form submission tests.

test/logflare_web/live_views/endpoints_live_test.exs
```diff
@@ -360,4 +360,92 @@ defmodule LogflareWeb.EndpointsLiveTest do
...
+        endpoint: %{
+          backend_id: backend.id,
+          name: "test endpoint",
+          query: "SELECT 1"
+        }
+      })
+
+      assigns = get_view_assigns(view)
+      assert assigns.selected_backend_id == to_string(backend.id)
+      assert Logflare.Endpoints.derive_language_from_backend_id(backend.id) == :ch_sql
```
Ziinc, line 444: we can render the determined language in the form so that we don't need to check the internals for this

amokan: Created #2733 to improve this

PR #2660, opened 2025-08-18
perf: add top processes and ETS tables metrics to telemetry

andremaker:
Adds the following OTel metrics:

- Top processes by memory consumption
- Top processes by message queue size
- Top ETS tables

Adds cluster information to otel metrics exporter

PR #2715, opened 2025-09-01
test: process and ets table metrics

andremaker:
Adds automated tests for the new metrics add at [PR 2660](https://github.com/Logflare/logflare/pull/2660)

### DIFF
--- a/test/logflare/telemetry_test.exs
+++ b/test/logflare/telemetry_test.exs
@@ -1,0 +1,94 @@
+defmodule Logflare.TelemetryTest do
+  use Logflare.DataCase
+  alias Logflare.Telemetry
+
+  describe "process metrics" do
+    test "retrieves and emits top 10 by memory" do
+      event = [:logflare, :system, :top_processes, :memory]
+      attach_forwarder(event)
+
+      Telemetry.process_memory_metrics()
+
+      1..10
+      |> Enum.each(fn _ ->
+        assert_receive {:telemetry_event, ^event, metrics, meta}
+
+        assert match?(%{size: _}, metrics)
+        assert match?(%{name: _}, meta)
+      end)
+
+      refute_receive {:telemetry_event, ^event, _, _}
+    end
+
+    test "retrieves and emits top 10 by message queue" do
+      event = [:logflare, :system, :top_processes, :message_queue]
+      attach_forwarder(event)
+
+      Telemetry.process_message_queue_metrics()
+
+      1..10
+      |> Enum.each(fn _ ->
+        assert_receive {:telemetry_event, ^event, metrics, meta}
+
+        assert match?(%{length: _}, metrics)
+        assert match?(%{name: _}, meta)
+      end)
+
+      refute_receive {:telemetry_event, ^event, _, _}
+    end
+  end
+
+  describe "ets_table_metrics/1" do
+    test "retrieves and emits top 10 by table size" do
+      event = [:logflare, :system, :top_ets_tables, :individual]
+      attach_forwarder(event)
+
+      Telemetry.ets_table_metrics()
+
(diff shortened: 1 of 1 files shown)


test/logflare/telemetry_test.exs
```diff
@@ -0,0 +1,94 @@
+defmodule Logflare.TelemetryTest do
+  use Logflare.DataCase
```
Ziinc: can directly use [ExUnit.Case](https://hexdocs.pm/ex_unit/1.0.4/ExUnit.Case.html) with `async: false`

Ziinc: DataCase has heavier setup which these tests don't need

test/logflare/telemetry_test.exs
```diff
@@ -0,0 +1,94 @@
...
+      :telemetry.attach(
+        id,
+        event_name,
+        fn ^event_name, measurements, metadata, _config ->
+          send(test_pid, {:telemetry_event, event_name, measurements, metadata})
+        end,
+        nil
+      )
+
+    ExUnit.Callbacks.on_exit(fn -> :telemetry.detach(id) end)
```
Ziinc:
```suggestion
    on_exit(fn -> :telemetry.detach(id) end)
```
should already be imported

test/logflare/telemetry_test.exs
```diff
@@ -0,0 +1,94 @@
...
+
+      1..10
+      |> Enum.each(fn _ ->
+        assert_receive {:telemetry_event, ^event, metrics, meta}
+
+        assert match?(%{size: _}, metrics)
+        assert match?(%{name: _}, meta)
+      end)
+
+      refute_receive {:telemetry_event, ^event, _, _}
```
Ziinc:
```suggestion
      assert_receive {:telemetry_event, ^event, %{size: _}, %{name: _}}
      refute_receive {:telemetry_event, ^event, _, _}
```
iterating 10 times will not remove it from the inbox so the Enum.each is redundant.
Doing a simple assertion that it is there is sufficient.

test/logflare/telemetry_test.exs
```diff
@@ -0,0 +1,94 @@
...
+
+      1..100
+      |> Enum.each(fn _ ->
+        assert_receive {:telemetry_event, ^event, metrics, meta}
+
+        assert match?(%{size: _}, metrics)
+        assert match?(%{name: _}, meta)
+      end)
+
+      refute_receive {:telemetry_event, ^event, _, _}
```
Ziinc: as above.

test/logflare/telemetry_test.exs
```diff
@@ -0,0 +1,94 @@
...
+
+      1..10
+      |> Enum.each(fn _ ->
+        assert_receive {:telemetry_event, ^event, metrics, meta}
+
+        assert match?(%{length: _}, metrics)
+        assert match?(%{name: _}, meta)
+      end)
+
+      refute_receive {:telemetry_event, ^event, _, _}
```
Ziinc: see below

test/logflare/telemetry_test.exs
```diff
@@ -0,0 +1,94 @@
...
+
+      1..10
+      |> Enum.each(fn _ ->
+        assert_receive {:telemetry_event, ^event, metrics, meta}
+
+        assert match?(%{size: _}, metrics)
+        assert match?(%{name: _}, meta)
+      end)
+
+      refute_receive {:telemetry_event, ^event, _, _}
```
Ziinc: see below.

Ziinc (changes requested):

Ziinc (approved):

PR #2717, opened 2025-09-01
feat: adjust chart period when performing search

msmithstubbs:
Adjusts the search results chart period if longer than 1000 ticks.

* If the user selects a date range that would halt the search due to an insufficient chart period, we increase the chart period as required
* if the user edits the query directly we don't change the chart period

https://github.com/user-attachments/assets/ffe1b21f-2d0d-4f57-b86c-5e8f8b9ab1cd


test/logflare_web/live/search_live/logs_search_lv_test.exs
```diff
@@ -458,6 +458,35 @@ defmodule LogflareWeb.Source.SearchLVTest do
...
+             |> has_element?("#search_chart_period option[selected]", "hour")
+
+      # don't override explicit user choice of period
+      render_change(view, :form_update, %{
+        "search" => %{
+          @default_search_params
+          | "querystring" => query,
+            "chart_period" => "second"
+        }
+      })
```
msmithstubbs, line 493: This wasn't specified in the requirements but I think it makes sense: if the user has specified a period we should respect it and halt the search.

Ziinc: This means that if the user makes a change to the chart period manually, then there no longer is any auto-adjustment?

Ziinc (commented):
Since the calculate_minimum_required_period/3 always works its way up from min period needed, i think it factors in the situation where the search range drops from days to minutes/seconds, right? Just want to check that this is was considered apart from demo.

I think should not completely stop auto-adjustment if user makes a change to the input (or at least have a reset mechanism), otherwise a full page reload is needed to reset the state which can be more frustrating

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -940,6 +963,36 @@ defmodule LogflareWeb.Source.SearchLV do
     )
   end
 
+  defp put_halt_flash_message(socket, search_op) do
+    {:halted, message} = search_op.status
+
+    msg =
+      if message =~ ~r/longer chart aggregation period.$/ or message =~ ~r/shorter chart period.$/ do
```
msmithstubbs, line 970: Would prefer just have an atom (`:chart_period_too_short`) to match on but it seemed overkill to rework `SearchOperation` to achieve that.

Ziinc (approved):

PR #2729, opened 2025-09-03
feat: endpoints PII redaction

Ziinc:
This PR adds in config flag for ipv4 and ipv6 redaction from endpoints query results.

This PR also fixes some bugs relating to displaying of logflare endpoints configuration in the show page.

### DIFF
--- a/lib/logflare/endpoints/pii_redactor.ex
+++ b/lib/logflare/endpoints/pii_redactor.ex
@@ -1,0 +1,93 @@
+defmodule Logflare.Endpoints.PiiRedactor do
+  @moduledoc """
+  Handles PII redaction for endpoint query results.
+
+  Currently supports redaction of IP addresses in query result values.
+  """
+
+  @doc """
+  Redacts PII from query results based on the query's redact_pii flag.
+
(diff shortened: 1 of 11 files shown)


lib/logflare/endpoints/pii_redactor.ex
```diff
@@ -0,0 +1,93 @@
...
+      "User REDACTED logged in"
+
+      iex> Logflare.Endpoints.PiiRedactor.redact_ip_addresses("IPv6: 2001:0db8:85a3::8a2e:0370:7334")
+      "IPv6: REDACTED"
+  """
+  @spec redact_ip_addresses(String.t()) :: String.t()
+  def redact_ip_addresses(value) when is_binary(value) do
+    value
+    |> redact_ipv4_addresses()
+    |> redact_ipv6_addresses()
```
chasers, line 75: Fine for now but if we want to add more things to "PII" I wonder if we can do these in one regex

Ziinc: probably can build the regex from parts from a massive string 😆

PR #2735, opened 2025-09-03
fix: allow multiple sources to be linked to a backend as default ingest sources

amokan:
Fixes an issue where the system was not allowing more than one source to be linked to a backend as a 'default ingest source'.

This issue was caused by a couple different things going on:
- The logic in `Backends.update_backend/2` calls `handle_default_ingest_associations/4` to handle making the associations between source/backend and was not handling the situation where a default ingest source was already linked due to pattern matching.
- Once that issue ^^^ was resolved, I realized the clickhouse adaptor then would try to spin up a query/read connection pool using a key of `backend_id`. Because we already had one, it was stopping that supervisor from returning successfully within its `init` callback. Now it handles this particular pool situation in a more idempotent way.

image

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
@@ -513,10 +513,23 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
 
     children = [
       ConnectionManager.child_spec({source, backend, ingest_ch_opts}),
-      ConnectionManager.child_spec({backend, query_ch_opts}),
       Provisioner.child_spec(args),
       Pipeline.child_spec(pipeline_state)
     ]
+
+    # Only add the backend-based query pool if it's not already running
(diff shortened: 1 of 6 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -513,11 +513,24 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
...
       ConnectionManager.child_spec({source, backend, ingest_ch_opts}),
-      ConnectionManager.child_spec({backend, query_ch_opts}),
       Provisioner.child_spec(args),
       Pipeline.child_spec(pipeline_state)
     ]
 
+    # Only add the backend-based query pool if it's not already running
+    # This is to cover for adding multiple default ingest sources to the same backend
+    children =
+      case ConnectionManager.connection_pool_via(backend) do
```
Ziinc, line 523: with/do would be easier to read

PR #2738, opened 2025-09-04
feat: configure additional ingestion-related telemetry events

amokan:
- Adds the 3 new telemetry events related to ingestion (_see below_)
- Adds `attach_forwarder/2` to `Logflare.TestUtils` to allow for reuse (_this function was originally just in `test/logflare/telemetry_test.exs`_)
- Adds basic tests to validate events are emitted

----

### New Events

```elixir
sum("logflare.backends.ingest.count",
  tags: [:backend_type],
  description: "Ingest counts by backend type"
)
```

```elixir
sum("logflare.backends.clickhouse.ingest.count",
  tags: [],
  description: "ClickHouse ingest counts by source id"
)
```

```elixir
distribution("logflare.backends.ingest.dispatch.stop.duration",
  tags: [:backend_type],
  unit: {:native, :millisecond},
  description: "Ingest dispatch latency by backend type"
)
```


lib/telemetry.ex
```diff
@@ -217,6 +217,19 @@ defmodule Logflare.Telemetry do
...
         description: "Top ETS tables by size, grouped by name"
+      ),
+      counter("logflare.backends.ingest.count",
+        tags: [:backend_type, :source_id],
+        description: "Ingest counts by backend type"
+      ),
+      counter("logflare.backends.clickhouse.ingest.count",
+        tags: [:source_id, :backend_id],
+        description: "ClickHouse ingest counts by source id"
+      ),
```
v0idpwn: Notice that the `counter` in telemetry metrics are "number of times this event was emitted". You probably want `sum` here to consider the measurement value.

v0idpwn: ref: https://hexdocs.pm/telemetry_metrics/Telemetry.Metrics.html#counter/2

amokan: Good catch @v0idpwn 

Thank you!

Ziinc: lets also remove the source_id and backend_it breakdown, it would result in too much cardinality

lib/telemetry.ex
```diff
@@ -217,6 +217,19 @@ defmodule Logflare.Telemetry do
...
+      sum("logflare.backends.ingest.count",
+        tags: [:backend_type, :source_id],
+        description: "Ingest counts by backend type"
+      ),
+      sum("logflare.backends.clickhouse.ingest.count",
+        tags: [:source_id, :backend_id],
+        description: "ClickHouse ingest counts by source id"
+      ),
+      distribution("logflare.backends.ingest.dispatch.stop.duration",
+        tags: [:backend_type, :source_id, :backend_id],
```
Ziinc: just backend_type would be fine

amokan: done

PR #2742, opened 2025-09-05
feat: cache clear button

Ziinc:
adds a cache clear button to the ui
image


lib/logflare/endpoints.ex
```diff
@@ -124,30 +124,12 @@ defmodule Logflare.Endpoints do
...
-          key in [
-            :query,
-            :sandboxable,
-            :cache_duration_seconds,
-            :proactive_requerying_seconds,
-            :max_limit,
-            :enable_auth,
-            :labels
-          ]
+      for pid <- Resolver.list_caches(endpoint) do
```
SteffenDE, line 127: Just to make sure since the PR title or description does not mention this: you also removed the should_kill_caches? check, so now it will always invalidate.

lib/logflare_web/live/endpoints/actions/show.html.heex
```diff
@@ -3,6 +3,9 @@
     ~/<.subheader_path_link live_patch to={~p"/endpoints"}>endpoints</.subheader_path_link>/<%= @show_endpoint.name %>
   </:path>
   <.subheader_link to={~p"/access-tokens"} text="access tokens" fa_icon="key" />
+  <button phx-click="clear-endpoint-cache" phx-value-endpoint-id={@show_endpoint.id} class="tw-text-black tw-p-1 tw-flex tw-gap-1 tw-items-center tw-justify-center tw-border-0 tw-bg-transparent">
+    <i class="inline-block h-3 w-3 fas fa-trash"></i><span> clear cache</span>
+  </button>
```
SteffenDE, line 8: Just as an idea: maybe you add a `<.subheader_button>` component to prevent copying the classes here? Otherwise you need to remember to update this file if you change how the subheader links render.

lib/logflare_web/live/endpoints/endpoints_live.ex
```diff
@@ -227,6 +230,29 @@ defmodule LogflareWeb.EndpointsLive do
     end
   end
 
+  def handle_event(
+        "clear-endpoint-cache",
+        %{"endpoint-id" => endpoint_id},
+        socket
+      ) do
+    endpoint = Endpoints.get_endpoint_query(endpoint_id)
+    :ok = Endpoints.clear_all_endpoint_caches(endpoint)
```
SteffenDE, line 239: you need to be careful in case not every user should be able to clear arbitrary endpoints. Since you expect this event to only be invoked from the button, you could skip the `phx-value-endpoint-id` and directly access the `show_endpoint` assign here.

Always remember that `handle_event` params come directly from the client and can never be trusted.

Also, I don't know how long this typically takes, but the fact that you wait for 30 seconds in `clear_all_endpoint_caches` suggests that it might take a bit. If that's the case I would recommend to do this asynchronously using `start_async` and show a loading indicator while it is processing. Otherwise the whole LiveView will be unresponsive while it is processing this event.

test/logflare_web/live_views/endpoints_live_test.exs
```diff
@@ -117,6 +117,26 @@ defmodule LogflareWeb.EndpointsLiveTest do
...
+    assert render(view) =~ ~r/active caches:.+\d+/
+
+    TestUtils.retry_assert([sleep: 250], fn ->
+      view
+      |> element(".subhead button", "clear cache")
+      |> render_click()
+
+      assert render(view) =~ "Cache cleared successfully"
+      assert render(view) =~ ~r/active caches:.+0/
+    end)
```
SteffenDE, line 137: I don't think that you need this retry logic. `render_click` waits for the handle_event to succeed, so you should always see the success message afterwards.

SteffenDE (commented):
A couple of comments from my side :)

test/logflare_web/live_views/endpoints_live_test.exs
Ziinc: shifting the render_click out of the retry_assert actually results in the test becoming very flaky, so would need to leave it in for now. it is likely due to the Task.async_await but haven't really figured out why it is slow 🤔

Ziinc:
closing as now obsolete

PR #2747, opened 2025-09-05
fix: CH connection manager bug with multiple sources

amokan:
My prior 'fix' was incorrect and evaluating the wrong process (the connection pool itself) and instead should have been evaluating the connection manager process as that is what is directly managed by the `ClickhouseAdaptor`.

Aside from this problem, I was also attempting to pattern match `{:ok, _pid}` for the results of `GenServer.whereis/1` which returns a pid or nil value.

A test has been added to ensure the logic operates as expected going forward.

#### Validation

Within the CH adaptor's `init` callback, you can add `ConnectionManager.child_spec({backend, query_ch_opts})` to the children variable and comment out the dynamic check on line 522 and run `mix test test/logflare/backends/adaptor/clickhouse_adaptor_test.exs` to see the difference.

### DIFF
--- a/test/logflare/backends/adaptor/clickhouse_adaptor_test.exs
+++ b/test/logflare/backends/adaptor/clickhouse_adaptor_test.exs
@@ -1,7 +1,9 @@
 defmodule Logflare.Backends.Adaptor.ClickhouseAdaptorTest do
   use Logflare.DataCase, async: false
 
+  alias Logflare.Backends
   alias Logflare.Backends.Adaptor.ClickhouseAdaptor
+  alias Logflare.Backends.Adaptor.ClickhouseAdaptor.ConnectionManager
   alias Logflare.Sources.Source
 
   doctest ClickhouseAdaptor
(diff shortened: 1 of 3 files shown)


Ziinc (approved):

test/logflare/backends/adaptor/clickhouse_adaptor_test.exs
```diff
@@ -211,6 +213,56 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptorTest do
...
+
+      assert is_pid(initial_manager_pid)
+      assert Process.alive?(initial_manager_pid)
+      assert ConnectionManager.pool_active?(backend)
+
+      # start a second clickhouse adaptor with the same backend, but a different source
+      {:ok, source2} = Backends.update_source_backends(source2, [backend])
+      {:ok, ch_pid2} = ClickhouseAdaptor.start_link({source2, backend})
+
+      Process.sleep(200)
```
Ziinc, line 254: nit: might have race condition on ci

PR #2752, opened 2025-09-09
refactor: Dashboard LiveView

msmithstubbs:
Convert the Dashboard to a LiveView

I'd like to move deleting a saved search and deleting a team user into liveview actions, but thought better to follow up with separate PRs to keep this manageable and merged sooner.

Closes ANL-1028

## todo

- [x] finish up team switching and tests 

CleanShot 2025-09-10 at 16 17 00@2x

https://github.com/user-attachments/assets/b908e92a-8385-4fcb-8362-3e2b68fb7023


lib/logflare_web/live/dashboard_live.ex
```diff
@@ -0,0 +1,227 @@
...
+            inserts_string: Number.Delimit.number_to_delimited(payload.log_count)
+        }
+      end)
+      |> assign(fade_in: true)
+
+    {:noreply, socket}
+  end
+
+  def update_source_metrics(socket, id, update_fn) when is_integer(id) do
+    case Enum.find(socket.assigns.sources, fn source -> source.id == id end) do
```
Ziinc: Would be more efficient to store a map of the metrics in the assigns and updating it as the rates come in

Ziinc: Would need to also factor in staleness though

lib/logflare_web/live/dashboard_live.ex
```diff
@@ -0,0 +1,227 @@
+defmodule LogflareWeb.DashboardLive do
+  alias Logflare.Repo
+  use LogflareWeb, :live_view
+
+  alias Logflare.{Billing, Sources, Teams, TeamUsers, Users}
```
Ziinc: separate lines for module aliases preferable

lib/logflare_web/live/dashboard_live.ex
msmithstubbs: I've changed source metrics to a map of assigns as suggested.

Regarding stale updates - we could add a timestamp to each payload and then compare them when received in the dashboard. Don't know if the increase in traffic is a concern?

Ziinc (approved):

PR #2754, opened 2025-09-09
test: telemetry correctly skips deleted ets tables

andremaker:
Adds test for the fix from the [PR 2746](https://github.com/Logflare/logflare/pull/2746), assuring that  ETS tables that were deleted between the listing and the call of :ets.info don't crashes the process


lib/logflare/utils.ex
```diff
@@ -295,4 +295,11 @@ defmodule Logflare.Utils do
       Logflare.Utils.stringify_tesla_struct(env)
     end
   end
+
+  @doc """
+  Returns information about the ets table as a list of tuples.
+  This is just a wrapper of :ets.info/1 to enable mocking without having issues
+  from overwritting :ets module
```
Ziinc:
```suggestion
  from mocking the :ets module
```

PR #2757, opened 2025-09-10
refactor: fix linting error and delete old templates

nelsonmestevao:


Ziinc:
templates are used for vercel oauth, please remove the deletions

Ziinc (changes requested):
need to keep templates. can refactor to heex

PR #2759, opened 2025-09-10
feat: initial Clickhouse support for transforming LQL -> Ecto queries

amokan:
This change adds a ClickHouse implementation of `Logflare.Lql.BackendTransformer` to handle common transformations from LQL that use CH-specific functions.

While this change is fairly useless on its own - it is a foundational item to complete the full LQL -> Ecto -> SQL for Clickhouse.

_Subsequent PRs will handle the remaining tasks as well as documentation updates related to LQL_

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,5 +1,5 @@
 elixir 1.17.3-otp-27
 nodejs 22.9.0
 rust 1.84.1
-erlang 27.1.2
+erlang 27.3.4
 protoc 28.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ARG ELIXIR_VERSION=1.17.3
-ARG OTP_VERSION=27.1.2
-ARG DEBIAN_VERSION=bullseye-20241016-slim
+ARG OTP_VERSION=27.3.4.3
+ARG DEBIAN_VERSION=trixie-20250908-slim
 
 ARG BUILDER_IMAGE="hexpm/elixir:${ELIXIR_VERSION}-erlang-${OTP_VERSION}-debian-${DEBIAN_VERSION}"
 ARG RUNNER_IMAGE="debian:${DEBIAN_VERSION}"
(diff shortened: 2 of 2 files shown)


lib/logflare/lql/backend_transformer.ex
```diff
@@ -112,11 +112,13 @@ defmodule Logflare.Lql.BackendTransformer do
...
   def for_dialect("bigquery"), do: __MODULE__.BigQuery
+  def for_dialect("clickhouse"), do: __MODULE__.ClickHouse
   def for_dialect(value) when is_atom(value), do: value |> to_dialect() |> for_dialect()
 
   @doc """
   Converts a dialect atom to its corresponding dialect string.
   """
   @spec to_dialect(dialects()) :: String.t()
   def to_dialect(:bigquery), do: "bigquery"
+  def to_dialect(:clickhouse), do: "clickhouse"
```
Ziinc, line 123: just a thought: perhaps we can standardize over time to a singular format, because right now there is `bq_sql` and `ch_sql` and string/atom versions of those as well.

lib/logflare/lql/backend_transformer/clickhouse.ex
```diff
@@ -0,0 +1,258 @@
...
+    end
+  end
+
+  @spec negated?(map()) :: boolean()
+  defp negated?(modifiers), do: Map.get(modifiers, :negate)
+
+  @spec build_combined_select(Query.t(), select_rules :: [map()]) :: Query.t()
+  defp build_combined_select(query, select_rules) do
+    Enum.reduce(select_rules, query, fn %{path: path}, acc_query ->
+      field_atom = String.to_atom(path)
```
Ziinc, line 254: hmm we can't keep it as a string? since this is based on user input, it isn't super safe

Ziinc: keeping the field to be accessed as a string would be much safer

PR #2763, opened 2025-09-12
UserMetricsPoller for dashboard

msmithstubbs:
UserMetricPoller provides periodic source metrics updates to the Dashboard LiveView. 

The UserMetricsPoller is started when a dashboard mounts and subscribes.

The UI feedback hasn't been implemented yet since it's not clear it's ever visible to the user: If that the browser tab is visible then updates are being polled. 

## To do
- [x] Dashboard should untrack when browser tab is inactive
- [x] Add buffers to broadcast metrics
- [ ] Provide UI feedback to user when metrics are not being received 
- [x] Creating a source should refresh the UserMetricsPoller state (ie. source_ids )


lib/logflare/sources/user_metrics_poller.ex
```diff
@@ -0,0 +1,148 @@
...
+    GenServer.call(via_tuple(user_id), {:subscribe, dashboard_pid})
+  end
+
+  @doc "Unsubscribe a dashboard from metrics updates"
+  def unsubscribe(user_id, dashboard_pid \\ self()) do
+    GenServer.call(via_tuple(user_id), {:unsubscribe, dashboard_pid})
+  end
+
+  defp via_tuple(user_id) do
+    {:via, :syn, {:core, user_id}}
```
Ziinc: Would need to add the module to the name

lib/logflare/gen_singleton/watcher.ex
```diff
@@ -49,6 +49,11 @@ defmodule Logflare.GenSingleton.Watcher do
   end
 
   @impl true
+  def handle_info({:DOWN, ref, :process, _pid, reason}, state)
+      when ref == state.monitor_ref and reason in [:normal, :shutdown] do
+    {:stop, :normal, state}
+  end
+
```
msmithstubbs, line 64: Any issues with this change? 

I couldn't see any reason we'd need Watcher to restart a process that has exited normally (and don't want UserMetricPoller to be restarted if it wants to shutdown).

msmithstubbs: Had to make another change so opened #2778

lib/logflare/sources/user_metrics_poller.ex
```diff
@@ -0,0 +1,172 @@
...
+  defp fetch_cluster_metrics(sources) do
+    Enum.map(sources, fn %{id: id, token: token} ->
+      {rate_results, _bad_nodes} =
+        Cluster.Utils.rpc_multicall(Logflare.PubSubRates.Cache, :get_local_rates, [token])
+
+      {buffer_results, _bad_nodes} =
+        Cluster.Utils.rpc_multicall(Logflare.PubSubRates.Cache, :get_local_buffer, [id, nil])
+
+      {[{:ok, inserts_results}], _bad_nodes} =
+        Cluster.Utils.rpc_multicall(Logflare.PubSubRates.Cache, :get_inserts, [token])
```
Ziinc: better to do this all in one multicall instead of 1 multicall for each source.
so init the multicall -> function that runs on the node then fetches the list of sources from Sources.Cache and gets the rates for each, building a map -> sends map of rates back to caller node.
caller node then aggregates the map before broadcasting again

lib/logflare/sources/user_metrics_poller.ex
```diff
@@ -0,0 +1,172 @@
...
+
+      aggregated_metrics =
+        aggregate_rates(rate_results)
```
Ziinc: jsut a note, aggregated map size might get large due to large number of sources 200-300, would make sense to batch the broadcasting at like 50 sources for each batch to reduce size of message and reduce time of genserver block when handling the message

lib/logflare/sources/user_metrics_poller.ex
msmithstubbs: Fixed 👍

lib/logflare/sources/user_metrics_poller.ex
```diff
@@ -0,0 +1,172 @@
...
+    else
+      {:stop, :normal, state}
+    end
+  end
+
+  def handle_info(:refresh_sources, state) do
+    sources =
+      Sources.list_sources_by_user(state.user_id) |> Enum.map(&Map.take(&1, [:id, :token]))
+
+    schedule_sources_refresh()
```
Ziinc: i think we can rely on the Sources.Cache in this case, it will get busted if there is a new source

lib/logflare/sources/user_metrics_poller.ex
```diff
@@ -0,0 +1,172 @@
...
+    {:noreply, %{state | sources: sources}}
+  end
+
+  defp schedule_sources_refresh do
+    Process.send_after(self(), :refresh_sources, @refresh_interval)
```
Ziinc: perhaps on top of doing this, we can be reactive by hooking into  CacheBusterWorker
https://github.com/Logflare/logflare/blob/5ab4ec8f0ced0c4d8eaac60d36901e5ad2b78041/lib/logflare/context_cache/cache_buster_worker.ex
and sending an refresh notification message to sources in the syn group so that they can re-fetch on source change.
this would avoid a stale state where someone makes a new source but doesn't see any metrics for new ingestion

would need a test for the above situation too

lib/logflare/sources/user_metrics_poller.ex
msmithstubbs: Sources.Cache doesn't have a list cache. Users does: Users.Cache.preload_sources/1 so we could use that. It's more likely to be cached and avoids creating keys in Sources.Cache with non-primary keys. 

I notice that Users.Cache.preload_sources/1 uses the whole user struct as the cache key. Is that intentional?

lib/logflare/sources/user_metrics_poller.ex
msmithstubbs: Added `get_all_local_metrics/2` so we can do this in one call for each source.

msmithstubbs: Ah, you wanted one call for *all* sources. Will do.

Ziinc: Yup one call for all will make things a lot faster due to network io

Ziinc: Also need to consider that some nodes might be in a shutting down state so we might not get an answer

lib/logflare/sources/user_metrics_poller.ex
Ziinc: Should be using the  id as cache key for the cache buster to work.

msmithstubbs: Will prepare a fix. 👍

Ziinc: https://github.com/Logflare/logflare/blob/main/lib%2Flogflare%2Fusers%2Fcache.ex#L40
You're referring to this?
This is fine actually, the ContextCache will use the args as the mapping key for the reverse index.

Ziinc: I guess a more efficient way to store the reverse index is to hash the args, but that would be beyond scope of this PR

lib/logflare/sources/user_metrics_poller.ex
msmithstubbs: Now calls `get_all_local_metrics/1` which takes the user id and returns all metrics for all users' sources.

> Also need to consider that some nodes might be in a shutting down state so we might not get an answer 

Any that don't respond will be returned as bad nodes, which are ignored. Think that's fine but let me know if we need to do more for this case.

lib/logflare/sources/user_metrics_poller.ex
msmithstubbs: These are now broadcast in batches of 50.

lib/logflare/sources/user_metrics_poller.ex
msmithstubbs: As discussed, using `Users.Cache.preload_sources/1`.

lib/logflare/sources/user_metrics_poller.ex
msmithstubbs: Each time the poller request metrics `get_all_local_metrics/1` calls `Users.Cache.preload_sources/1`. Since that cache is busted when a source is created do we still need to hook into CacheBusterWorker?

Ziinc: i think we can defer that, doesn't seem critical.

assets/js/source_lv_hooks.js
```diff
@@ -254,4 +254,22 @@ hooks.FormatTimestamps = {
   },
 };
 
+hooks.DocumentVisibility = {
+  mounted() {
+    this.handleVisibilityChange = () => {
+      this.pushEvent("visibility_change", {
+        visibility: document.visibilityState,
```
Ziinc, line 261: i think we will also need to consider document blur/focus, as i notice that when the window goes to the background, the polling still continues as the visibilityState api does not consider unminimized unfocused windows as "non-visible".

can be done in separate PR.

Ziinc (approved):
💯 will be merged after i cut a release.

assets/js/source_lv_hooks.js
msmithstubbs: This is true. One issue I considered was a situation where the dashboard is in an unfocused window but still being viewed. For example, I might open the dashboard in the browser then switch to the terminal and try sending the some requests to a source. I would probably expect to see live updates, and it would be useful for the user to have that feedback. 

Perhaps we wait until this is merged and some of the other rate publishing is removed and then consider if it is worth optimising this further (by trying to detect unfocused windows, etc)?

PR #2764, opened 2025-09-12
bq writer storage improvements

WolfDan:


test/logflare/backends/bigquery_adaptor_test.exs
```diff
@@ -207,6 +207,37 @@ defmodule Logflare.Backends.BigQueryAdaptorTest do
...
+      log_event_body_updated =
+        Map.put(log_event.body, "new_field", %{"new" => "value"})
+
+      log_event = Map.put(log_event, :body, log_event_body_updated)
+
+      log_event_2 = build(:log_event, source: source)
+
+      pid = self()
+
+      Logflare.Backends.Adaptor.BigQueryAdaptor.GoogleApiClient
```
Ziinc, line 225: would be better to mock the GRPC client  instead so that this test will fail if GoogleApiClient implementation changes.

PR #2765, opened 2025-09-12
refactor: DashboardLive saved searches

msmithstubbs:
Move SavedSearch#delete to DashboardLive

Follows on from #2752


test/support/factory.ex
```diff
@@ -382,4 +382,15 @@ defmodule Logflare.Factory do
       language: :bq_sql
     }
   end
+
+  def saved_search_factory do
+    %Logflare.SavedSearch{
+      lql_filters: [],
+      lql_charts: [],
```
Ziinc, line 389: these probably need to be generated from the lql string

PR #2772, opened 2025-09-15
fix: endpoint query / sql ast bugs

amokan:
Some improvements to the logic around handling AST for sandboxed queries. At least to get us a better understanding of the source queries causing issues.

### DIFF
--- a/lib/logflare/sql.ex
+++ b/lib/logflare/sql.ex
@@ -357,7 +357,8 @@ defmodule Logflare.Sql do
   # applies only to the sandboed query
   defp maybe_validate_sandboxed_query_ast({cte_ast, ast}, data) when is_list(ast) do
     with :ok <- validate_query(ast, data),
-         :ok <- has_restricted_sources(cte_ast, ast) do
+         :ok <- has_restricted_sources(cte_ast, ast),
+         :ok <- validate_sandboxed_query_ast(ast, data) do
       :ok
     end
   end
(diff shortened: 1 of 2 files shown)


lib/logflare/sql.ex
```diff
@@ -530,6 +539,22 @@ defmodule Logflare.Sql do
...
+  defp validate_sandboxed_query_ast(ast, %{sandboxed_query: sandboxed_query} = data) do
+    case extract_replacement_query_from_ast(ast, data) do
+      {:ok, _replacement_query} ->
+        :ok
+
+      {:error, transformed_statements} ->
+        Logger.warning("Sandboxed query validation: would produce nil replacement_query",
+          error_string: sandboxed_query,
+          ast_statements_count: length(transformed_statements),
+          first_statement: List.first(transformed_statements)
```
Ziinc: instead of adding more situational fields, lets just add it to the error message instead. also easier to search.
fine to leave the query on error_string

lib/logflare/sql.ex
```diff
@@ -632,17 +674,21 @@ defmodule Logflare.Sql do
...
+        {k, Map.merge(sandbox_query, %{"body" => %{"Query" => replacement_query}})}
+
+      {:ok, replacement_query} ->
+        {k, Map.merge(sandbox_query, Map.drop(replacement_query, ["with"]))}
+
+      {:error, transformed_statements} ->
+        Logger.warning("Sandboxed query fallback triggered - validation should have caught this",
+          sandboxed_query: sandboxed_query,
+          ast_statements_count: length(transformed_statements),
+          first_statement: List.first(transformed_statements)
```
Ziinc: as above

lib/logflare/sql.ex
```diff
@@ -111,6 +111,10 @@ defmodule Logflare.Sql do
         ast: statements
       })
       |> Parser.to_string()
+    else
+      error ->
+        Logger.warning("SQL transformation failed", error_string: query)
```
Ziinc: these errors shouldn't be logged as user error would be high (parsing on each endpoint query change). we would want to log the query on an error and reraise it, something like that.

Ziinc: might be even better to set the Logger.metadata for a `:query_string` field so that any related error log raised gets it

lib/logflare/sql.ex
Ziinc: also, as above, logging on each sql parse would lead to a lot of log lines. probably better to add as metadata for the raise.

PR #2776, opened 2025-09-17
test: remove flakyness of the test for delete_alert_query/1

nelsonmestevao:
The `AlertsScheduler` needs time to start up before we can safely delete the alert query.


lib/logflare/alerting.ex
```diff
@@ -367,10 +367,11 @@ defmodule Logflare.Alerting do
...
   end
 
-  defp on_scheduler_node(func) do
+  def on_scheduler_node(func) do
     with pid when is_pid(pid) <- GenServer.whereis(scheduler_name()) do
-      node = node(pid)
-      :erpc.call(node, func, 5000)
+      pid
+      |> node()
+      |> :erpc.call(func, 5000)
```
nelsonmestevao: I had to do this because Mimic wouldn't mock `on_scheduler_node/1` if we didn't prepend the module name call. (I couldn't use Mimic to mock `:erpc` directly.)


@Ziinc Should I move this function and the `scheduler_name/0` function to some specific module where I would organize all the `ERPC` calls?

Ziinc: we probably can shift it to https://github.com/Logflare/logflare/blob/925c124d51d6fe7efdfae7b07385414e2f8a0994/lib/logflare/cluster/utils.ex

nelsonmestevao: Done on last pushed commit 23f3e78c355a16d01ceac4ba49fe50c3e1ecaff1 🙏

lib/logflare/cluster/utils.ex
```diff
@@ -24,4 +25,9 @@ defmodule Logflare.Cluster.Utils do
   def rpc_multicall(mod, func, args, timeout \\ 5_000) do
     :rpc.multicall(node_list_all(), mod, func, args, timeout)
   end
+
+  @spec erpc_call(node(), function(), non_neg_integer()) :: term()
+  def erpc_call(node, func, timeout \\ 5_000) do
```
Ziinc, line 30: We can just call it rpc_call, since we don't need to distinguish between them from caller pov

Ziinc (approved):

lib/logflare/cluster/utils.ex
nelsonmestevao: Yes, done in https://github.com/Logflare/logflare/pull/2785 :pray:

PR #2778, opened 2025-09-18
feat: allow restart to be specified for GenSingleton.Watcher

msmithstubbs:
In preparation for #2763 
UserMetricsPoller needs to be a singleton GenServer but allowed to exit.

### DIFF
--- a/test/logflare/gen_singleton_test.exs
+++ b/test/logflare/gen_singleton_test.exs
@@ -77,4 +77,21 @@ defmodule Logflare.GenSingletonTest do
       end)
     end
   end
+
+  describe "Watcher child_spec/1" do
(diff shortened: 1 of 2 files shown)


test/logflare/gen_singleton_test.exs
```diff
@@ -77,4 +77,21 @@ defmodule Logflare.GenSingletonTest do
       end)
     end
   end
+
+  describe "Watcher child_spec/1" do
+    test "defaults restart to :permanent" do
```
Ziinc, line 82: This kinda tests implementation, as opposed to starting up an instance of GenSingleton and then sending the exit signal

Ziinc (approved):

PR #2780, opened 2025-09-18
Feat: user system monitoring

andremaker:
This enables users to create system monitoring sources, that ingests only their user specific signals, using our Backends module


lib/logflare/user.ex
```diff
@@ -117,7 +119,8 @@ defmodule Logflare.User do
     :valid_google_account,
     :provider_uid,
     :company,
-    :partner_upgraded
+    :partner_upgraded,
+    :system_monitoring
```
Ziinc, line 129: should be a user_allowed field, as user can opt into this feature.

lib/logflare_web/templates/user/edit.html.eex
```diff
@@ -92,6 +92,13 @@
   <%= section_header("BigQuery Backend") %>
   <p>Set your Google Cloud Platform project id to setup your own BigQuery tables. Make sure to add our service account (<code><%= @service_account %></code>) to IAM with BigQuery Data Owner and BigQuery Job User permissions. See <%= link "this guide", to: Routes.marketing_path(@conn, :big_query_setup) %> for more info.</p>
   <%= render(LogflareWeb.UserView, "edit_bq_form.html", conn: @conn, changeset: @changeset) %>
+  <%= section_header("System monitoring") %>
```
Ziinc:
```suggestion
  <%= section_header("System Monitoring") %>
```

mix.exs
```diff
@@ -238,7 +238,7 @@ defmodule Logflare.Mixfile do
       {:opentelemetry_exporter, "~> 1.6"},
       {:opentelemetry_phoenix, "~> 2.0.0-rc.2"},
       {:opentelemetry_bandit, "~> 0.2.0-rc.1"},
-      {:otel_metric_exporter, "~> 0.3.11"},
+      {:otel_metric_exporter, git: "https://github.com/supabase/elixir-otel-metric-exporter"},
```
Ziinc: we can move this to a separate PR so that it isn't blocked.

lib/logflare/user.ex
andremaker: i believe i already did that way. Did i miss something in this sense?

lib/logflare/sources.ex
```diff
@@ -116,6 +119,36 @@ defmodule Logflare.Sources do
...
+      end
+
+    Repo.insert_all(Source, entries,
+      returning: true,
+      conflict_target: [:user_id, :system_source_type],
+      on_conflict: :nothing
+    )
+    |> case do
+      {_, []} -> {:ok, list_sources_by_user(user_id)}
+      {_, result} -> {:ok, result}
```
Ziinc: Might be less than expected, should always list the system sources for the user

lib/logflare/sources.ex
```diff
@@ -116,6 +119,36 @@ defmodule Logflare.Sources do
...
+        }
+      end
+
+    Repo.insert_all(Source, entries,
+      returning: true,
+      conflict_target: [:user_id, :system_source_type],
+      on_conflict: :nothing
+    )
+    |> case do
+      {_, []} -> {:ok, list_sources_by_user(user_id)}
```
Ziinc: This would list all sources, no?
Should be filtered down to system sources only

lib/logflare/sources.ex
andremaker: do you mean when some have failed in the past and thus the insert list wont return empty?
I will fix that making it always load the full list, but please tell me if you meant something else so i can solve it as well

Ziinc: Yes that is the case I am referring to, so if it is less than the full amount, it should load it from the table.

lib/logflare/backends/source_sup.ex
```diff
@@ -55,11 +63,77 @@ defmodule Logflare.Backends.SourceSup do
...
+          name: "Logflare",
+          service: %{
+            name: "Logflare",
+            version: Application.spec(:logflare, :vsn) |> to_string()
+          },
+          node: inspect(Node.self()),
+          cluster: Application.get_env(:logflare, :metadata)[:cluster]
+        },
+        export_callback: generate_exporter_callback(source),
+        name: :"#{source.name}-#{user.id}",
```
Ziinc: lets use a :via tuple like `Backends.via_source(source,Monitoring)`

lib/logflare/backends/source_sup.ex
```diff
@@ -44,6 +50,8 @@ defmodule Logflare.Backends.SourceSup do
       |> Enum.map(&Backend.child_spec(source, &1))
       |> Enum.uniq()
 
+    otel_exporter = maybe_get_otel_exporter(source, user)
```
Ziinc, line 50: would be worth to move this to a `Monitoring` context module to keep the SourceSup clean.

lib/telemetry.ex
```diff
@@ -248,6 +260,55 @@ defmodule Logflare.Telemetry do
...
+
+  defp get_entity_from_metadata(%{backend_id: backend_id}),
+    do: Backends.Cache.get_backend(backend_id)
+
+  defp get_entity_from_metadata(_), do: nil
+
+  defp user_monitoring_metrics?(user_id),
+    do: Users.Cache.get(user_id).system_monitoring and metric_monitor_sourcesup_up?(user_id)
+
+  defp metric_monitor_sourcesup_up?(user_id) do
```
Ziinc: should be moved to a context module like `Backends.Monitoring`

lib/telemetry.ex
```diff
@@ -248,6 +260,55 @@ defmodule Logflare.Telemetry do
...
+    Sources.Cache.get_by(user_id: user_id, system_source_type: :metrics)
+    |> case do
+      nil -> true
+      source -> Backends.source_sup_started?(source.id)
+    end
```
Ziinc: would be better to use a :via tuple for the name and then do a GenServer.whereis for faster proc lookup. Backends.lookup/2 will work as well.

lib/logflare_web/router.ex
```diff
@@ -146,6 +146,10 @@ defmodule LogflareWeb.Router do
     plug(LogflareWeb.Plugs.VerifyApiAccess, scopes: ~w(partner))
   end
 
+  pipeline :block_system_source do
+    plug(LogflareWeb.Plugs.BlockSystemSource)
```
Ziinc: we could add this plug to `:otlp_api` and `:api` without needing a separate pipeline

lib/logflare_web/templates/user/edit.html.eex
```diff
@@ -9,6 +9,17 @@
     </div>
   </div>
 </div>
+<div class="content container mx-auto">
+  <%= section_header("System monitoring") %>
+  <div class="sub-form">
+    <%= form_for @changeset, Routes.user_path(@conn, :update), fn owner -> %>
+      <div class="form-group">
+        <%= checkbox owner, :system_monitoring, onclick: "this.form.submit();" %>
```
Ziinc: separate button to follow the ux of the other forms would be better since changing ux of form submission is not in scope (even though not a bad change -- autosaving would need full page refactor to not make things look disjointed.)

lib/logflare_web/router.ex
```diff
@@ -532,7 +536,7 @@ defmodule LogflareWeb.Router do
 
     # logpush
     scope "#{path}/cloudflare", LogflareWeb, assigns: %{resource_type: :source} do
-      pipe_through([:logpush, :api, :require_ingest_api_auth])
+      pipe_through([:logpush, :api, :require_ingest_api_auth, :block_system_source])
```
Ziinc: As above, can consolidate to existing pipelines.

lib/telemetry.ex
```diff
@@ -29,13 +33,15 @@ defmodule Logflare.Telemetry do
...
+  @user_specific_tags [:source_id, :source_token, :backend_id]
+
   @impl true
   def init(_arg) do
     otel_exporter =
       if Application.get_env(:logflare, :opentelemetry_enabled?) do
         otel_exporter_opts =
           Application.get_all_env(:opentelemetry_exporter)
-          |> Keyword.put(:metrics, metrics())
+          |> Keyword.put(:metrics, metrics() |> add_filters())
```
Ziinc: not a bad idea, but we want to keep the subset of metrics exposed to users small and intentional for now. a separate function for user-specific metrics would be better so that there is less noise for users.

we only want one specific metric for a start:
- count of events ingested by a backend for a source

lib/telemetry.ex
```diff
@@ -60,7 +66,13 @@ defmodule Logflare.Telemetry do
     Supervisor.init(children, strategy: :one_for_one)
   end
 
-  def metrics do
+  if Mix.env() == :test do
```
```suggestion
  if Application.get_env(:logflare, :env) == :test do
```
shouldn't reference Mix at runtime.

Ziinc: https://hexdocs.pm/mix/Mix.html#env/0
> This function should not be used at runtime in application code (as opposed to infrastructure and build code like Mix tasks). Mix is a build tool and may not be available after the code is compiled (for example in a release).

lib/telemetry.ex
```diff
@@ -248,6 +260,55 @@ defmodule Logflare.Telemetry do
...
+  defp get_entity_from_metadata(%{source_token: token}),
+    do: Sources.Cache.get_source_by_token(token)
+
+  defp get_entity_from_metadata(%{backend_id: backend_id}),
+    do: Backends.Cache.get_backend(backend_id)
+
+  defp get_entity_from_metadata(_), do: nil
+
+  defp user_monitoring_metrics?(user_id),
+    do: Users.Cache.get(user_id).system_monitoring and metric_monitor_sourcesup_up?(user_id)
```
```suggestion
    do: Users.Cache.get(user_id).system_monitoring
```
don't think we need to check if the process is up, if it is up then it will pick up the metric. if it isn't then it gets lost to the void (which is fine)
making this call on every event would add up.

test/logflare/telemetry_test.exs
```diff
@@ -61,4 +63,254 @@ defmodule Logflare.TelemetryTest do
...
+    test "are routed to user's system source when flag is true", %{
+      user_1: user,
+      backend_1: %{id: backend_id}
+    } do
+      user |> Users.update_user_allowed(%{system_monitoring: true})
+
+      system_source =
+        Sources.get_by(user_id: user.id, system_source_type: :metrics)
+
+      SourceSup.start_link(system_source)
```
Ziinc: start_supervised!/2 preferred.

test/logflare/telemetry_test.exs
```diff
@@ -1,9 +1,11 @@
 defmodule Logflare.TelemetryTest do
-  use ExUnit.Case, async: false
+  use Logflare.DataCase, async: false
   use Mimic
```
Ziinc:
```suggestion
```

test/logflare/telemetry_test.exs
```diff
@@ -1,9 +1,11 @@
...
 
   alias Logflare.Telemetry
   alias Logflare.TestUtils
+  alias Logflare.{Users, Sources}
```
Ziinc:
```suggestion
  alias Logflare.Users
  alias Logflare.Sources
```

test/logflare/telemetry_test.exs
```diff
@@ -61,4 +63,254 @@ defmodule Logflare.TelemetryTest do
...
+               %{
+                 {:last_value, "logflare.test.user_specific.value"} => %{
+                   %{backend_id: ^backend_1_id} => _
+                 }
+               },
+               user_2_exporter_metrics
+             )
+    end
+
+    setup do
```
Ziinc: setup should be at top of file before describe blocks.

test/logflare/logs/logs_test.exs
```diff
@@ -187,4 +192,52 @@ defmodule Logflare.LogsTest do
...
+
+      refute Enum.any?(
+               Backends.list_recent_logs(system_source),
+               &match?(%{body: %{"event_message" => "user not monitoring"}}, &1)
+             )
+    end
+
+    setup do
+      :ok =
+        :logger.add_primary_filter(:user_log_intercetor, {&Logflare.UserLogInterceptor.run/2, []})
```
Ziinc: Nice approach.
We should move the UserLogInterceptor under `Backends.Monitoring` namespace.

lib/telemetry.ex
```diff
@@ -29,13 +33,15 @@ defmodule Logflare.Telemetry do
 
   @metrics_interval 30_000
 
+  @user_specific_tags [:source_id, :source_token, :backend_id]
```
```suggestion
  @user_specific_tags [:user_id, :source_id, :source_token, :backend_id]
```

test/logflare/telemetry_test.exs
```diff
@@ -61,4 +63,261 @@ defmodule Logflare.TelemetryTest do
...
+          config: %{url: "http://test.com"},
+          default_ingest?: false
+        )
+
+      {:ok, user_1: user_1, user_2: user_2, backend_1: backend_1, backend_2: backend_2}
+    end
+  end
+
+  defp setup_otel do
+    original_logflare_env = Application.get_all_env(:logflare)
```
Ziinc: we're testing the library at this point. we can remove the otel-specific stuff and test the configuration that is passed to OtelMetricsExporter instead.

test/logflare/telemetry_test.exs
Ziinc: should also scope this to the specific describe block since it isn't needed for all tests.

test/logflare/telemetry_test.exs
```diff
@@ -61,4 +63,261 @@ defmodule Logflare.TelemetryTest do
...
+      SourceSup.start_link(system_source_1)
+
+      # user 2 setup
+
+      user_2 |> Users.update_user_allowed(%{system_monitoring: true})
+
+      system_source_2 =
+        Sources.get_by(user_id: user_2.id, system_source_type: :metrics)
+
+      SourceSup.start_link(system_source_2)
```
Ziinc: we can start the exporter specs in isolation with start_supervised!/2 to make this test much simpler. note that different `:id` option would need to be provided as 2rd arg.

test/logflare/telemetry_test.exs
```diff
@@ -61,4 +63,261 @@ defmodule Logflare.TelemetryTest do
...
+             )
+
+      refute match?(
+               %{
+                 {:last_value, "logflare.test.user_specific.value"} => %{
+                   %{backend_id: ^backend_1_id} => _
+                 }
+               },
+               user_2_exporter_metrics
+             )
```
Ziinc: can we condense the assertions?

lib/logflare/backends/source_sup.ex
andremaker: i tried that now, but OtelMetricExporter only accepts atoms as name, so it prevents the exporter child from spawning when using `:via`. Should we make the OtelMetricExporter fork accept the tuple, or just keep using atom?

lib/telemetry.ex
andremaker: yes, i have committed it by distraction, but fixed right after. Unfortunately you caught it during the very brief time when it was up 😅

lib/telemetry.ex
andremaker: i see. I thought it would be bad to drop those metrics

test/logflare/telemetry_test.exs
andremaker: how exactly?

lib/logflare/backends/source_sup.ex
andremaker: maybe `UserMonitoring` to decrease ambiguity?

lib/telemetry.ex
andremaker: is this a new metric? If not, what is the name of that event?

test/logflare/telemetry_test.exs
andremaker: you mean just checking if the config is right?
I believe its important to check if things are working as a whole, instead of just parts of it, but i will do as you prefer

Ziinc: Yes I think that this is not necessary since we aren't interested in the export portion and only interested in ensuring that the callback for the lib gets triggered when passed in config.

test/logflare/telemetry_test.exs
Ziinc: Hmm we can keep it as is, it's fine

lib/telemetry.ex
Ziinc: Yea this would be a new metric, we can emit telemetry in the BufferProducer so that it automatically covers all backends

test/logflare/telemetry_test.exs
andremaker: this part is already ensured through the otelMetricExporter test for that feature, i believe. In the logflare side, we can ensure just that the callback is doing what its supposed to do, and that the metric keep function itself is returning true or false in the correct scenarios. This would be way simpler than starting the whole telemetry module

lib/logflare/backends/user_monitoring.ex
```diff
@@ -0,0 +1,118 @@
...
+          name: "Logflare",
+          service: %{
+            name: "Logflare",
+            version: Application.spec(:logflare, :vsn) |> to_string()
+          },
+          node: inspect(Node.self()),
+          cluster: Application.get_env(:logflare, :metadata)[:cluster]
+        },
+        export_callback: generate_exporter_callback(source),
+        name: :"#{source.name}-#{user.id}",
```
Ziinc, line 29: @andremaker we will need it to accept a tuple as a name. can you do a PR to our fork for it please

test/logflare/backends/user_monitoring_test.exs
```diff
@@ -0,0 +1,258 @@
...
+             )
+
+      refute match?(
+               %{
+                 {:last_value, "logflare.test.user_specific.value"} => %{
+                   %{backend_id: ^backend_1_id} => _
+                 }
+               },
+               user_2_exporter_metrics
+             )
```
Ziinc, line 251: instead of testing this, shouldn't we instead assert that the metrics ingested into the system sources?
reaching into OtelMetricsExporter storage doesn't cover the callback areas.

test/logflare/backends/user_monitoring_test.exs
```diff
@@ -0,0 +1,258 @@
...
+
+      assert match?(
+               %{
+                 {:last_value, "logflare.test.user_specific.value"} => %{
+                   %{backend_id: ^backend_id} => 456
+                 }
+               },
+               user_exporter_metrics
+             )
```
Ziinc, line 142: as below, would prefer a more e2e test where we assert that the callback is called and the metrics are ingested.
for the mocked telemetry we can maybe export the values at a much faster interval with a higher `:export_period`

lib/logflare/backends/user_monitoring.ex
```diff
@@ -0,0 +1,118 @@
...
+    if Application.get_env(:logflare, :env) == :test do
+      [
+        last_value("logflare.test.user_specific.value",
+          description: "To test how user specific metrics are handled by exporter",
+          tags: [:backend_id],
+          keep: keep_function
+        )
+      ]
+    else
+      []
```
Ziinc: missing the backend count metric

lib/telemetry.ex
```diff
@@ -61,6 +63,12 @@ defmodule Logflare.Telemetry do
   end
 
   def metrics do
+    if Application.get_env(:logflare, :env) == :test,
+      do: metrics(:mocked),
+      else: metrics(:normal)
+  end
+
+  defp metrics(:normal) do
```
Ziinc: don't need this anymore since we are using `system_metrics(..)`

Ziinc (changes requested):

lib/logflare/backends/user_monitoring.ex
andremaker: yes, the metric still need to be implemented. Will be the next thing

Ziinc (approved):

PR #2784, opened 2025-09-19
feat: optional datadog pools

Ziinc:
Adds optional datadog pools, pools will no longer be created for self-hosting unless explicitly set
 closes ANL-811


lib/logflare/application.ex
```diff
@@ -201,42 +201,67 @@ defmodule Logflare.Application do
...
+         %{
+           # default pool uses finch defaults
+           :default => [protocols: [:http1]],
+           #  explicitly set http2 for other pools for multiplexing
+           "https://bigquery.googleapis.com" => [
+             protocols: [:http1],
+             size: 115,
+             count: http1_count,
+             start_pool_metrics?: true
+           ]
```
v0idpwn, line 214: Should this one also be conditional?

Ziinc: yea can be separate PR as that one is more involved and dependent on the selected system default backend.

PR #2787, opened 2025-09-20
feat: Add Sentry Adaptor

AbhiPrasad:
This PR adds a backend for [Sentry](https://sentry.io/about/), specifically for Sentry's new [structured logging product](https://docs.sentry.io/product/explore/logs/).

resolves https://github.com/getsentry/sentry/issues/95914

image

To configure the Sentry backend, you need to add a DSN (data source name), which indicates which sentry instance to send logs to. This was chosen over just a regular URL as most users are familiar with getting a DSN from sentry.

To parse the DSN into an endpoint, we use the same logic as we do with Sentry's Elixir SDK: https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/dsn.ex

The Sentry adapter itself is based on the webhook backend. The only thing that is different is that to send data to Sentry, it needs to be sent as an envelope instead of plain JSON.

Docs for envelopes can be found here: https://develop.sentry.dev/sdk/data-model/envelopes/
And the docs for log envelope items can be found here: https://develop.sentry.dev/sdk/data-model/envelope-items/#log

### DIFF
--- a/lib/logflare/backends/adaptor/sentry_adaptor.ex
+++ b/lib/logflare/backends/adaptor/sentry_adaptor.ex
@@ -1,0 +1,250 @@
+defmodule Logflare.Backends.Adaptor.SentryAdaptor do
+  @sdk_name "sentry.logflare"
+  @sdk_version "0.1.0"
+  @sentry_envelope_content_type "application/x-sentry-envelope"
+
+  @moduledoc """
+  Sentry adaptor for sending logs to Sentry's logging API.
+
+  This adaptor wraps the WebhookAdaptor to provide specific functionality
+  for sending logs to Sentry in the expected envelope format.
+
+  ## Configuration
+
+  The adaptor requires a single configuration parameter:
+
+  - `dsn` - The Sentry DSN string in the format:
+    `{PROTOCOL}://{PUBLIC_KEY}:{SECRET_KEY}@{HOST}{PATH}/{PROJECT_ID}`
+
+  ## Example DSN
+
+      https://abc123@o123456.ingest.sentry.io/123456
+  """
+
+  alias Logflare.Backends.Adaptor.WebhookAdaptor
+  alias Logflare.Backends.Adaptor.SentryAdaptor.DSN
+
+  @behaviour Logflare.Backends.Adaptor
+
+  def child_spec(arg) do
+    %{
+      id: __MODULE__,
+      start: {__MODULE__, :start_link, [arg]}
+    }
+  end
+
+  @impl Logflare.Backends.Adaptor
+  def start_link({source, backend}) do
+    backend = %{backend | config: transform_config(backend)}
+    WebhookAdaptor.start_link({source, backend})
+  end
+
+  @impl Logflare.Backends.Adaptor
+  def transform_config(%{config: config}) do
+    case DSN.parse(config.dsn) do
+      {:ok, parsed_dsn} ->
+        %{
+          url: parsed_dsn.endpoint_uri,
+          headers: %{"content-type" => @sentry_envelope_content_type},
+          http: "http2",
+          format_batch: fn log_events ->
+            build_envelope(log_events, parsed_dsn.original_dsn)
+          end
+        }
+
+      {:error, reason} ->
+        raise ArgumentError, "Invalid Sentry DSN: #{reason}"
+    end
+  end
+
+
+  @impl Logflare.Backends.Adaptor
+  def execute_query(_ident, _query, _opts), do: {:error, :not_implemented}
+
+  @impl Logflare.Backends.Adaptor
+  def cast_config(params) do
+    {%{}, %{dsn: :string}}
+    |> Ecto.Changeset.cast(params, [:dsn])
+  end
+
+  @impl Logflare.Backends.Adaptor
+  def validate_config(changeset) do
+    changeset
+    |> Ecto.Changeset.validate_required([:dsn])
+    |> validate_dsn()
+  end
+
+  defp validate_dsn(changeset) do
+    case Ecto.Changeset.get_field(changeset, :dsn) do
+      nil ->
+        changeset
+
+      dsn ->
+        case DSN.parse(dsn) do
+          {:ok, _parsed_dsn} ->
+            changeset
+
+          {:error, reason} ->
+            Ecto.Changeset.add_error(changeset, :dsn, "Invalid DSN: #{reason}")
+        end
+    end
+  end
+
+  defp build_envelope(log_events, original_dsn) do
+    sentry_logs = Enum.map(log_events, &translate_log_event/1)
+
+    header = %{
+      "dsn" => original_dsn,
+      "sent_at" => DateTime.utc_now() |> DateTime.to_iso8601(:extended)
+    }
+
+    item_header = %{
+      "type" => "log",
+      "item_count" => length(sentry_logs),
+      "content_type" => "application/vnd.sentry.items.log+json"
+    }
+
+    item_payload = %{"items" => sentry_logs}
+
+    Enum.join(
+      [
+        Jason.encode!(header),
+        Jason.encode!(item_header),
+        Jason.encode!(item_payload)
+      ],
+      "\n"
+    )
+  end
+
+  defp translate_log_event(%Logflare.LogEvent{} = le) do
+    # Convert microsecond timestamp to seconds (with fractional part)
+    # Use timestamp from outer body if available, fallback to nested body
+    timestamp_seconds = (le.body["timestamp"] || le.body["body"]["timestamp"]) / 1_000_000
+
+    # Extract message from the log event
+    message = le.body["event_message"] || le.body["body"]["message"] || inspect(le.body)
+
+    # Determine log level - check both body and nested body
+    level =
+      case le.body["body"]["level"] || le.body["level"] do
+        nil -> "info"
+        level when is_binary(level) -> normalize_level(level)
+        level when is_atom(level) -> normalize_level(Atom.to_string(level))
+        _ -> "info"
+      end
+
+    # Extract or generate trace_id
+    trace_id = extract_trace_id(le)
+
+    # Build attributes from log event metadata
+    attributes = build_attributes(le)
+
+    %{
+      "timestamp" => timestamp_seconds,
+      "level" => level,
+      "body" => message,
+      "trace_id" => trace_id,
+      "attributes" => attributes
+    }
+  end
+
+  defp build_attributes(%Logflare.LogEvent{} = le) do
+    base_attrs = %{
+      "sentry.sdk.name" => to_sentry_value(@sdk_name),
+      "sentry.sdk.version" => to_sentry_value(@sdk_version),
+      "logflare.source.name" => to_sentry_value(le.source.name),
+      "logflare.source.id" => to_sentry_value(le.source.id)
+    }
+
+    # Get the nested body data for metadata
+    nested_body = le.body["body"] || %{}
+
+    # Get direct metadata field as well
+    direct_metadata = le.body["metadata"] || %{}
+
+    # Add all metadata from the nested log event body, excluding standard fields
+    nested_attrs =
+      nested_body
+      |> Map.drop(["timestamp", "message", "event_message", "level", "trace_id", "trace.id"])
+      |> Map.new(fn {k, v} -> {k, to_sentry_value(v)} end)
+
+    # Add direct metadata attributes
+    metadata_attrs =
+      direct_metadata
+      |> Map.new(fn {k, v} -> {k, to_sentry_value(v)} end)
+
+    # Also process fields from the top-level body, excluding standard fields and LogEvent management fields
+    # This handles cases where fields are placed directly in le.body
+    top_level_attrs =
+      le.body
+      |> Map.drop(["timestamp", "message", "event_message", "level", "trace_id", "trace.id", "body", "metadata", "id", "valid", "drop", "sys_uint", "params", "ingested_at"])
+      |> Map.new(fn {k, v} -> {k, to_sentry_value(v)} end)
+
+    base_attrs
+    |> Map.merge(nested_attrs)
+    |> Map.merge(metadata_attrs)
+    |> Map.merge(top_level_attrs)
+  end
+
+  defp valid_trace_id?(trace_id) when is_binary(trace_id) do
+    # Valid trace ID should be 32 characters (128 bits) hex string
+    # and not be all zeros
+    with true <- String.length(trace_id) == 32,
+         true <- String.match?(trace_id, ~r/^[0-9a-fA-F]+$/),
+         false <- trace_id == String.duplicate("0", 32) do
+      true
+    else
+      _ -> false
+    end
+  end
+
+  defp valid_trace_id?(_), do: false
+
+  defp extract_trace_id(%Logflare.LogEvent{} = le) do
+    # Look for trace_id in nested body first, then outer body
+    nested_body = le.body["body"] || %{}
+
+    case nested_body["trace_id"] || nested_body["trace.id"] || le.body["trace_id"] || le.body["trace.id"] do
+      nil ->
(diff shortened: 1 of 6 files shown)


test/logflare/backends/adaptor/sentry_adaptor/dsn_test.exs
```diff
@@ -0,0 +1,132 @@
+defmodule Logflare.Backends.Adaptor.SentryAdaptor.DSNTest do
+  use ExUnit.Case, async: true
+
+  alias Logflare.Backends.Adaptor.SentryAdaptor.DSN
+
+  doctest DSN
```
Ziinc, line 6: please condense into doctests and parameterized tests. Most of the assertions are duplicated and can be part of the match assert

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -0,0 +1,250 @@
+defmodule Logflare.Backends.Adaptor.SentryAdaptor do
+  @sdk_name "sentry.logflare"
+  @sdk_version "0.1.0"
```
Ziinc: would make more sense to take the logflare app version number, would make it prone to human error if we rely on this version number to be changed each time manually

lib/logflare/backends/adaptor/sentry_adaptor/dsn.ex
```diff
@@ -0,0 +1,86 @@
...
+      throw("DSN with query parameters is not supported. Please remove query parameters from the DSN.")
+    end
+
+    unless is_binary(uri.path) do
+      throw("missing project ID at the end of the DSN URI: #{inspect(dsn)}")
+    end
+
+    unless is_binary(uri.userinfo) do
+      throw("missing user info in the DSN URI: #{inspect(dsn)}")
+    end
```
Ziinc, line 45: this might leak the dsn, even if it is malformed. since it is getting returned as an ecto changeset error.

test/logflare/backends/adaptor/sentry_adaptor_test.exs
```diff
@@ -0,0 +1,568 @@
...
+          "dsn" => invalid_dsn
+        })
+
+      refute changeset.valid?
+      assert %{dsn: [error]} = errors_on(changeset)
+      assert error =~ "Invalid DSN"
+    end
+  end
+
+  describe "DSN parsing" do
```
Ziinc: don't need to retest this

test/logflare/backends/adaptor/sentry_adaptor_test.exs
```diff
@@ -0,0 +1,568 @@
...
+
+  doctest @subject
+
+  setup do
+    start_supervised!(AllLogsLogged)
+    insert(:plan)
+    :ok
+  end
+
+  describe "cast and validate" do
```
Ziinc, line 21: too verbose, can be condensed to 1-2 tests.

test/logflare/backends/adaptor/sentry_adaptor_test.exs
```diff
@@ -0,0 +1,568 @@
...
+          config: %{dsn: "https://abc123@o123456.ingest.sentry.io/123456"}
+        )
+
+      start_supervised!({AdaptorSupervisor, {source, backend}})
+      :timer.sleep(500)
+      [backend: backend, source: source]
+    end
+
+    # Helper function to normalize envelope for snapshot comparison
+    defp normalize_envelope_for_snapshot(envelope_body) do
```
Ziinc: testing implementation, the assertion can simply check that values are in envelope format by splitting the received request body and doing some simple match assertions.
also, priv functions should be at bottom of file.

test/logflare/backends/adaptor/sentry_adaptor_test.exs
```diff
@@ -0,0 +1,568 @@
...
+          }
+        )
+
+      assert {:ok, _} = Backends.ingest_logs([le], source)
+      assert_receive {^ref, envelope_body}, 2000
+
+      # Normalize envelope and compare against snapshot
+      normalized_envelope = normalize_envelope_for_snapshot(envelope_body)
+
+      expected_envelope = """
```
Ziinc: as above.

test/logflare/backends/adaptor/sentry_adaptor_test.exs
```diff
@@ -0,0 +1,568 @@
...
```
Ziinc: as above.

test/logflare/backends/adaptor/sentry_adaptor_test.exs
```diff
@@ -0,0 +1,568 @@
...
+        {"notice", "info"},
+        {"warning", "warn"},
+        {"warn", "warn"},
+        {"error", "error"},
+        {"critical", "fatal"},
+        {"alert", "fatal"},
+        {"emergency", "fatal"},
+        {"unknown", "info"}
+      ]
+
```
Ziinc, line 160: testing implementation, can just check 1-2 levels. can simply check that the envelop contains the level.

test/logflare/backends/adaptor/sentry_adaptor_test.exs
```diff
@@ -0,0 +1,568 @@
...
+        end)
+
+        assert {:ok, _} = Backends.ingest_logs([le], source)
+        assert_receive {^ref, envelope_body}, 2000
+
+        # Normalize envelope and verify level mapping via snapshot
+        normalized_envelope = normalize_envelope_for_snapshot(envelope_body)
+
+        # Build expected envelope dynamically to ensure proper interpolation
+        expected_envelope = [
```
Ziinc: as above

test/logflare/backends/adaptor/sentry_adaptor_test.exs
```diff
@@ -0,0 +1,568 @@
...
+        assert item["level"] == "info"
+        assert item["timestamp"] == 1.7040672e9 # Normalized timestamp
+        assert is_nil(item["trace_id"]) # trace_id removed for normalization
+
+        attributes = item["attributes"]
+        assert attributes["logflare.source.id"]["value"] == source.id
+        assert attributes["logflare.source.name"]["value"] == source.name
+        assert attributes["sentry.sdk.name"]["value"] == "sentry.logflare"
+        assert attributes["sentry.sdk.version"]["value"] == "0.1.0"
+      end
```
Ziinc: if the goal of the test is to check if multiple log events get sent in a batch, then can just check that the body contains all 3 messages. 
this overlaps a lot with other tests and can be trimmed down to be more succinct.

test/logflare/backends/adaptor/sentry_adaptor_test.exs
```diff
@@ -0,0 +1,568 @@
...
+          %{
+            "message" => "Test message",
+            "timestamp" => 1704067200_000_000
+          }
+        }
+      ]
+
+      for {description, body_data} <- test_cases do
+        le = build(:log_event, body: body_data)
+        result = transformed_config.format_batch.([le])
```
Ziinc: would prefer not to test anon functions and instead verify the trace_id is inside of the sent envelope in the request body.

test/logflare/backends/adaptor/sentry_adaptor_test.exs
```diff
@@ -0,0 +1,568 @@
...
+          "no trace_id present",
+          %{
+            "message" => "Test message",
+            "timestamp" => 1704067200_000_000
+          }
+        }
+      ]
+
+      for {description, body_data} <- test_cases do
+        le = build(:log_event, body: body_data)
```
Ziinc:
```suggestion
      for {description, body_data} <- test_cases do
        le = build(:log_event, message: body_data["message"], timestamp: body_data["timestamp"], trace_id: body_data["trace_id"])
```
the fields are passed as keyword lists and not on the `:body` key

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -0,0 +1,250 @@
...
+    base_attrs = %{
+      "sentry.sdk.name" => to_sentry_value(@sdk_name),
+      "sentry.sdk.version" => to_sentry_value(@sdk_version),
+      "logflare.source.name" => to_sentry_value(le.source.name),
+      "logflare.source.id" => to_sentry_value(le.source.id)
+    }
+
+    # Get the nested body data for metadata
+    nested_body = le.body["body"] || %{}
+
```
Ziinc: event data is not nested in the `body` key and can be in any json format with the exception of `event_message` and `timestamp` at the top level of the event json.

Ziinc (changes requested):
Thanks for the contribution! ❤️ would need some adjustments before we merge it in but core logic looks good

Ziinc:
I'm also on the fence on generating a "fake" trace_id, as it would not a true otel trace id from the official SDKs. 
does sentry's logging API require this field to be set? otherwise i would lean towards omission to avoid confusion from users on where this trace_id is coming from.

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -0,0 +1,250 @@
...
+      "attributes" => attributes
+    }
+  end
+
+  defp build_attributes(%Logflare.LogEvent{} = le) do
+    base_attrs = %{
+      "sentry.sdk.name" => to_sentry_value(@sdk_name),
+      "sentry.sdk.version" => to_sentry_value(@sdk_version),
+      "logflare.source.name" => to_sentry_value(le.source.name),
+      "logflare.source.id" => to_sentry_value(le.source.id)
```
Ziinc:
```suggestion
      "logflare.source.name" => to_sentry_value(le.source.name),
      "logflare.source.service_name" => to_sentry_value(le.source.service_name),
      "logflare.source.uuid" => to_sentry_value(le.source.token)
```

lib/logflare/backends/adaptor/sentry_adaptor.ex
AbhiPrasad: yeah good call, switched to using `Application.spec(:logflare, :vsn)` (but not sure if that's the best way to do it)

lib/logflare/backends/adaptor/sentry_adaptor.ex
AbhiPrasad: good call, done with 5c0b741019e100ac9906b3cb3ac9d298fb180746

lib/logflare/backends/adaptor/sentry_adaptor/dsn.ex
AbhiPrasad: Done with efdab64f0734d61f94d43905ebceaeb1ef4ff492

test/logflare/backends/adaptor/sentry_adaptor/dsn_test.exs
AbhiPrasad: done with 150558ca4fbea9069e9efb641f7967e6f8e375ea

test/logflare/backends/adaptor/sentry_adaptor_test.exs
AbhiPrasad: I've refactored all the tests in this file in f516e13131641e97dce0ceeff6e10b5d745722e5

AbhiPrasad (commented):
Thanks for the review @Ziinc! I've made some changes based on your PR feedback, especially to the tests.

re: trace id - it is required by the sentry platform so we are setting it. This is specced in https://develop.sentry.dev/sdk/telemetry/logs/

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -0,0 +1,238 @@
...
+          url: parsed_dsn.endpoint_uri,
+          headers: %{"content-type" => @sentry_envelope_content_type},
+          http: "http2",
+          format_batch: fn log_events ->
+            case log_events do
+              # If there are no log events, don't build an envelope
+              [] -> nil
+              _ -> build_envelope(log_events, parsed_dsn.original_dsn)
+            end
+          end
```
Ziinc, line 51:
```suggestion
          format_batch: fn log_events ->
             build_envelope(log_events, parsed_dsn.original_dsn)
          end
```
log_events will never be empty as it is batched by broadway.

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -0,0 +1,238 @@
...
+    changeset
+    |> Ecto.Changeset.validate_required([:dsn])
+    |> validate_dsn()
+  end
+
+  defp validate_dsn(changeset) do
+    case Ecto.Changeset.get_field(changeset, :dsn) do
+      nil ->
+        changeset
+
```
Ziinc: this will never occur as the `:dsn` is already validated as required and there is only one field.

Ziinc: we can remove the Ecto.Changeset.get_field and just have the `case DSN.parse(...`

Ziinc: can also do a match on the changeset to avoid a get_change call.

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -0,0 +1,238 @@
...
+    # Build attributes from log event metadata
+    attributes = build_attributes(log_event)
+
+    %{
+      "timestamp" => timestamp_seconds,
+      "level" => level,
+      "body" => message,
+      "trace_id" => trace_id,
+      "attributes" => attributes
+    }
```
Ziinc:
```suggestion
    %{
      "timestamp" => timestamp_seconds,
      "level" => level,
      "body" => message,
      "trace_id" => extract_trace_id(log_event),
      "attributes" => build_attributes(log_event)
    }
```

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -0,0 +1,238 @@
...
+        Jason.encode!(item_header),
+        Jason.encode!(item_payload)
+      ],
+      "\n"
+    )
+  end
+
+  defp translate_log_event(%Logflare.LogEvent{} = log_event) do
+    # Convert microsecond timestamp to seconds (with fractional part)
+    # Use timestamp from outer body if available, fallback to nested body
```
Ziinc:
```suggestion
```

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    top_level_attrs =
+      log_event.body
+      |> Map.drop([
+        "timestamp",
+        "event_message",
+        "level",
+        "trace_id",
+        "trace.id"
```
Ziinc, line 153: is this a known case that needs to be supported? it isn't in otel spec or sentry docs

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -0,0 +1,238 @@
...
+    # Valid trace ID should be 32 characters (128 bits) hex string
+    # and not be all zeros
+    with true <- String.length(trace_id) == 32,
+         true <- String.match?(trace_id, ~r/^[0-9a-fA-F]+$/),
+         false <- trace_id == String.duplicate("0", 32) do
+      true
+    else
+      _ -> false
+    end
+  end
```
Ziinc: i don't think validation of the trace id should be handled by logflare,  event should be rejected by sentry downstream

lib/logflare/backends/adaptor/sentry_adaptor.ex
AbhiPrasad: Some logging services emit this as `trace.id` (Vercel's Log Drain is an example: https://vercel.com/docs/drains/reference/logs)

image

This isn't part of the spec, but we can probably get this written down somewhere. I can take care of that.

PR #2788, opened 2025-09-22
feat: add configurable reservations for alerts/search queries

Ziinc:
This adds in reservation assignment for different query workloads. User can set a BQ reservation for different query workloads, and these query workloads will get the user-level reservation set on the BQ job.

Endpoints is intentionally left out and will be tackled in a separate PR.

Pre-merge checklist:
- [ ] verify on dev


chasers (approved):

test/logflare_web/live/search_live/logs_search_lv_test.exs
```diff
@@ -1080,6 +1080,71 @@ defmodule LogflareWeb.Source.SearchLVTest do
...
+
+    test "when bigquery reservation search is not set, do not set reservation option", ctx do
+      user = insert(:user, bigquery_reservation_search: nil)
+      %{conn: conn} = setup_user_session(Map.put(ctx, :user, user)) |> Map.new()
+      source = insert(:source, user: user)
+      pid = self()
+
+      expect(GoogleApi.BigQuery.V2.Api.Jobs, :bigquery_jobs_query, fn _conn, _proj_id, opts ->
+        reservation = opts[:body].reservation
+        send(pid, {:reservation, reservation})
```
SteffenDE, line 1097: Just as a note: you could also assert in here instead of sending a message back to the test process. This is of course also fine as is!

PR #2795, opened 2025-09-25
fix: metrics fixes

Ziinc:
- bumps bandit to have the correct version for http options, [495f27b](https://github.com/Logflare/logflare/pull/2795/commits/495f27bfee5d9876f9ef51d215a186655228c7af) 
- adds two metrics needed
- fixes ingestion monitoring for backends by type fyi @amokan


SteffenDE (commented):
The changes look good to me! Probably no need to say this, but don't forget to change any reports or alerts wherever you send your metrics to as well :)

PR #2798, opened 2025-09-26
fix open api generation

WolfDan:
add ymlr dependency for open api generation


Ziinc:
@WolfDan could you also add a ci check that will verify that the openapi specs can be generated? 🙏

PR #2800, opened 2025-09-29
perf: remove autoconnection between nodes on same network

Ziinc:
This removes the auto-connection behaviour when nodes detect each other via epmd. This does not prevent libcluster from managing the connections, it only prevents auto-connection attempts. this occurs when there are multiple clusters on the same network but with different cookies for each cluster.

image

closes ANL-1113

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,9 +142,17 @@ start.orange: __start__
 start.pink: ERL_NAME = pink
 start.pink: PORT = 4001
 start.pink: ENV_FILE = .dev.env
-# start.pink: ERL_COOKIE = pinkmonster
 start.pink: LOGFLARE_GRPC_PORT = 50052
 start.pink: __start__
+
+
+start.green: ERL_NAME = green
+start.green: PORT = 4002
+start.green: ERL_COOKIE = greenmonster
+start.green: ENV_FILE = .dev.env
+start.green: LOGFLARE_GRPC_PORT = 50053
+start.green: __start__
+
 
 # temp alias
 
@@ -181,7 +189,7 @@ migrate:
 	@env $$(cat .dev.env | xargs) mix ecto.migrate
 
 
-.PHONY: __start__ migrate start.sb.pg start.sb.bq start.st.pg start.st.bq start.orange start.pink
+.PHONY: __start__ migrate start.sb.pg start.sb.bq start.st.pg start.st.bq start.orange start.pink start.green
 
 # Encryption and decryption of secrets
 # Usage:
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,7 @@ start.orange: __start__
 start.pink: ERL_NAME = pink
 start.pink: PORT = 4001
 start.pink: ENV_FILE = .dev.env
+# start.pink: ERL_COOKIE = pinkmonster
 start.pink: LOGFLARE_GRPC_PORT = 50052
 start.pink: __start__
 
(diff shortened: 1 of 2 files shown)


josevalim:
I like that idea a lot because it means you can have fixed node names (which certainly helps while debugging).

Ziinc:
>Normally, connections are established automatically when another node is referenced. This functionality can be disabled by setting the Kernel configuration parameter dist_auto_connect to false, see [kernel](https://linux.die.net/man/6/kernel)(6). In this case, connections must be established explicitly by calling net_kernel:connect_node/1. 
> 
> https://linux.die.net/man/3/net_kernel

>  these changes will only have have an effect if A can connect to B across clusters but they should not connect as they have different cookies
yes this is the case, we're noticing that the existing nodes are attempting to connect to new nodes that join the network even if they have different cookies.

The postgres strategy is not triggering these connections, however, as the pub/sub channel is based on the cookie, so libcluster is actually not triggering the connections.

I have done some more testing and have been able to replicate this locally:

1. start orange, pink, and green nodes, with orange and pink with the same cookie and green having a different cookie.
2. kill pink node once it is connected to the orange node. Orange will have the pink nodename in `Node.list(:known)`
3. change the pink node's cookie to match green node, and start pink. The pink node now connects to green, and orange will attempt to connect to pink even though the Postgres strategy does not receive a heartbeat from the pink or green nodes.

However, this is fixed once i set `dist_auto_connect` to `never`. 

My current hypothesis is that in our situation, IPs may get reused for internal networking when nodes are brought up. some of these nodes might have been on a different cluster, resulting in the `Node.list(:known)` getting stale. I have yet to figure out the actual trigger for this: working theory is that `global` is trying to re-establish the connection when it notices that the nodename is active once again, but the connection attempt fails due to mismatched cookie.
This occurs across the entire cluster, which results in log error spikes, and occurs for every node that joins the cluster.

Ziinc:
another note for this behaviour is that it only starts once there is a prior connection and the nodename has been added to the `Node.list(:known)`

josevalim:
> My current hypothesis is that in our situation, IPs may get reused for internal networking when nodes are brought up. some of these nodes might have been on a different cluster, resulting in the Node.list(:known) getting stale.

So from your description, using unique node names per deployment could also address this issue?

In any case, given everything said, my only remaining concern about the current PR is that it may be prone to race conditions? For example, imagine you have nodes A and B that periodically exchange information about available nodes and send new nodes messages for set up. Once C is up, if A notices C before B, and then exchange that information to B, B may attempt to connect to C before it receives the libcluster broadcast, and since autoconnection is disabled, this may fail.

Perhaps this particular pattern doesn't happen in this app and the approach would be fine, but I thought I would mention it. :)

Ziinc:
yes unique node names would be possible, the race condition is not ideal either for this solution even if works... i'll try more unique node names and see if that can resolve the issue.

@chasers this error is from [dist_utils](https://github.com/erlang/otp/blob/master/lib%2Fkernel%2Fsrc%2Fdist_util.erl#L992) and is part of the handshake that can arise from [many potential areas](https://github.com/search?q=repo%3Aerlang%2Fotp+handshake_we_started&type=code). Syn only monitors the nodes and [doesn't trigger connects](https://github.com/search?q=repo%3Aostinelli%2Fsyn%20net_kernel&type=code), so we can eliminate syn as the trigger. likely normal otp behaviour, which also makes sense as a netsplit resolution strategy.

PR #2802, opened 2025-09-30
fix: faster backend syncing

Ziinc:
This PR adds in faster backend supervisor syncing when adding sources to backends.
it uses multicast instead as we don't really care about the results

Closes ANL-1100

### DIFF
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -428,10 +428,23 @@ defmodule Logflare.Backends do
     with %Backend{} = backend <- get_backend(backend_id) do
       sources = Sources.list_sources(backend_id: backend_id)
 
-      for source <- sources do
-        Cluster.Utils.rpc_multicall(SourceSup, :start_backend_child, [source, backend])
-        Cluster.Utils.rpc_multicall(__MODULE__, :clear_list_backends_cache, [source.id])
+      if length(sources) > 0 do
+        Cluster.Utils.rpc_multicast(__MODULE__, :sync_backends_local, [backend, sources])
       end
+    end
+
+    :ok
+  end
+
+  @doc """
+  Syncs a backend for local node for v2 pipeline sources.
+  expects the backend and sources to be loaded from the database.
+  """
+  @spec sync_backends_local(Backend.t(), [Source.t()]) :: :ok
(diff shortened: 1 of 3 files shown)


lib/logflare/backends.ex
```diff
@@ -428,15 +428,28 @@ defmodule Logflare.Backends do
     with %Backend{} = backend <- get_backend(backend_id) do
       sources = Sources.list_sources(backend_id: backend_id)
 
-      for source <- sources do
-        Cluster.Utils.rpc_multicall(SourceSup, :start_backend_child, [source, backend])
-        Cluster.Utils.rpc_multicall(__MODULE__, :clear_list_backends_cache, [source.id])
+      if length(sources) > 0 do
```
wojtekmach, line 431: doesn't matter for small lists but instead of calculating the length of the list just to compare with zero, it's faster to compare the list with empty one:

```suggestion
      if sources != [] do
```

test/logflare/backends_test.exs
```diff
@@ -1058,4 +1058,88 @@ defmodule Logflare.BackendsTest do
...
+      {:ok, user: user, source: source}
+    end
+
+    test "syncs backend across cluster for all associated sources", %{
+      source: source,
+      user: user
+    } do
+      # Start actual SourceSup process
+      start_supervised!({SourceSup, source})
+      :timer.sleep(500)
```
wojtekmach, line 1077: I noticed there's quite a few sleep calls in this test case, perhaps at some point worth reconsider testing strategy to improve performance when that becomes an issue. For example, we can add a lifecycle hook, something like

```elixir
pid = self()
start_supervised!({SourceSup, source: source, ready: fn -> send(pid, :ready) end})
assert_receive :ready
```

Another common idea is to emit Telemetry events and listen for those in tests.

wojtekmach (commented):
Looks good to me!

lib/logflare/backends.ex
Ziinc: addressed in https://github.com/Logflare/logflare/pull/2815

test/logflare/backends_test.exs
Ziinc: will keep this in mind, yes that would be a much nicer pattern

PR #2807, opened 2025-10-01
perf: cluster node name suffix

Ziinc:
Closes ANL-1113
This makes the node name unique cluster wide, resolving the connection issue of stale hostnames as per #2800 

image

This has been verified to fix the issue as per this replication steps:
https://github.com/Logflare/logflare/pull/2800#issuecomment-3348030464


josevalim (commented):
Looks good to me! :)

PR #2808, opened 2025-10-01
feat: comprehensive secrets redaction for backends

Ziinc:
Adds in secrets redaction for backends. ensures that secrets are not leaked even in Backend GET.

Each backend adaptor is responsible for ensuring the redaction

### DIFF
--- a/lib/logflare/backends/adaptor/elastic_adaptor.ex
+++ b/lib/logflare/backends/adaptor/elastic_adaptor.ex
@@ -46,11 +46,7 @@ defmodule Logflare.Backends.Adaptor.ElasticAdaptor do
 
   @impl Logflare.Backends.Adaptor
   def redact_config(config) do
-    if Map.get(config, :password) do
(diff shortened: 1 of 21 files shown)


lib/logflare/backends/adaptor/elastic_adaptor.ex
```diff
@@ -44,6 +44,15 @@ defmodule Logflare.Backends.Adaptor.ElasticAdaptor do
     }
   end
 
+  @impl Logflare.Backends.Adaptor
+  def redact_config(config) do
+    if Map.get(config, :password) do
```
wojtekmach: For completeness, you can use Map.replace_lazy, https://hexdocs.pm/elixir/Map.html#replace_lazy/3

wojtekmach: Or rather simply Map.replace in this particular case.

PR #2809, opened 2025-10-01
perf: reduce schema sample rate to at most 1 per sec

Ziinc:
This reduces the schema sample rate to at most 1 per second on a per-node level, ensuring moderate triggering of the Schema diffing logic.


SteffenDE (commented):
Looks good to me!

PR #2813, opened 2025-10-01
perf: local-only metrics and rate limiting

Ziinc:
This removes all global metrics broadcasting which should improve the performance greatly for large clusters.


wojtekmach (commented):
Looks good to me

PR #2814, opened 2025-10-01
perf: only load source socket conditionally

Ziinc:
this loads the source socket for receiving broadcasted log events conditionally. With new UserMetricsPoller, we no longer need the dashboard to connect to the socket.
this removes ws connection on the search page, speeding it up and removes it from the app js bundle, speeding up execution as well.

On a separate note: There is still an occasional noticeable "lag" when the ws upgrades from the initial liveview mount, but nothing is standing out to me at the moment as culprit. This lag time in the ws connection upgrade does not always happen consistently though, and usually only occurs on first page load of the server. It could be that the Endpoint is not ready to accept the connections, but would be rather odd...

### DIFF
--- a/assets/build.mjs
+++ b/assets/build.mjs
@@ -57,12 +57,12 @@ let sassPostcssPlugin = sassPlugin({
 
 const options = {
   logLevel: "info",
-  entryPoints: ["js/app.js"],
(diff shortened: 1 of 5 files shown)


josevalim:
@Ziinc could the slow down be on the time taken to execute mount itself? Could assign_async help?

PR #2816, opened 2025-10-02
refactor: improve dialyzer types in source_controller

nelsonmestevao:


lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -446,6 +446,12 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
   end
 
   defdelegate get_iam_policy(user), to: CloudResourceManager
+
+  @spec append_managed_sa_to_iam_policy(User.t()) ::
```
Ziinc: Shouldn't this be at the delegated module instead?

lib/logflare_web/controllers/source_controller.ex
```diff
@@ -222,47 +222,60 @@ defmodule LogflareWeb.SourceController do
...
-
-      {:error, response} ->
+    with source = %Source{} <- Sources.get_by_and_preload(id: source_id) do
+      case WebhookNotificationServer.test_post(source) do
+        {:ok, %Tesla.Env{} = _response} ->
+          conn
+          |> put_flash(:info, "Webhook test successful!")
+          |> redirect(to: Routes.source_path(conn, :edit, source.id))
+
+        {:error, %Tesla.Env{} = response} ->
```
Ziinc: We can put these unhappy paths in the else as well

lib/logflare_web/controllers/source_controller.ex
```diff
@@ -222,47 +222,60 @@ defmodule LogflareWeb.SourceController do
...
-
-      {:error, response} ->
+    with source = %Source{} <- Sources.get_by_and_preload(id: source_id) do
+      case SlackHookServer.test_post(source) do
+        {:ok, %Tesla.Env{} = _response} ->
+          conn
+          |> put_flash(:info, "Slack hook test successful!")
+          |> redirect(to: Routes.source_path(conn, :edit, source.id))
+
+        {:error, %Tesla.Env{} = response} ->
```
Ziinc: These matches can be in the else instead

Ziinc (changes requested):

lib/logflare/backends/adaptor/bigquery_adaptor.ex
nelsonmestevao: Agree :+1:  Moved in last push :rocket:

lib/logflare_web/controllers/source_controller.ex
nelsonmestevao: Done in last refactor :bow:

lib/logflare_web/controllers/source_controller.ex
nelsonmestevao: Done in last refactor 🙇

lib/logflare_web/controllers/source_controller.ex
nelsonmestevao: I didn't do that first time because in the else statements I was using `source.id`. However, I refactored it to use `source_id` from params in 659e5753962adf15218130d75f93932c4c0a82f1.

lib/logflare/utils.ex
```diff
@@ -60,6 +60,7 @@ defmodule Logflare.Utils do
     %{"test" => "data"}
   """
   @spec stringify_keys(map()) :: map()
+  @spec stringify_keys(list(map())) :: list(map())
```
Ziinc, line 63: Spec can be combined, and also. I think the list version is never used in codebase so should be safe to make it private

nelsonmestevao: Did it here: https://github.com/nelsonmestevao/logflare/blob/45ae239756c3e4090d1ba8b9b630a461318b8c97/lib/logflare/utils.ex#L63

PR #2819, opened 2025-10-06
refactor: improve dialyzer types in user module related problems

nelsonmestevao:


lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -238,18 +238,6 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
-      {:error, :emfile = response} ->
-        Logger.error("Stream batch emfile error!", tesla_response: response)
-
-      {:error, :timeout = response} ->
-        Logger.warning("Stream batch timeout error!", tesla_response: response)
-
-      {:error, :checkout_timeout = response} ->
-        Logger.warning("Stream batch checkout_timeout error!", tesla_response: response)
-
-      {:error, response} ->
```
Ziinc, line 251: Test coverage for this error handling is not high so would need some tests around these logs that are needed for debugging

Ziinc (changes requested):

PR #2820, opened 2025-10-07
feat: add format button for formatting SQL queries

msmithstubbs:
Adds format button to 
1. query page
2. endpoints edit page
3. alerts edit page

* We use `SqlFmt.format_query/1` for formatting which always returns `{:ok, result}` so no any UI is required.
* Have not included unit tests since most of the actual effects here are in the JS layer.
* Event listeners are now setup before rendering the editor 

https://github.com/user-attachments/assets/46247bd7-1d28-43fd-9cbe-0665ada6d2b4

https://github.com/user-attachments/assets/ec113daa-a9a6-420a-ae55-2c3158854248

### DIFF
--- a/lib/logflare_web/live/monaco_editor_component.ex
+++ b/lib/logflare_web/live/monaco_editor_component.ex
@@ -50,7 +50,14 @@ defmodule LogflareWeb.MonacoEditorComponent do
 
     ~H"""
     <div id={@id}>
-      <LiveMonacoEditor.code_editor value={@field.value} target={@myself} change="parse-query" path="query_string" id="query" opts={@editor_opts} />
+      <.support_scripts completions={@completions} id={@field.id <> "_completions"} />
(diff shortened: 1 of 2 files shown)


lib/logflare_web/live/monaco_editor_component.ex
```diff
@@ -107,6 +119,16 @@ defmodule LogflareWeb.MonacoEditorComponent do
     """
   end
 
+  def handle_event("format-query", %{"value" => value}, socket) do
+    case SqlFmt.format_query(value) do
+      {:ok, formatted} ->
```
Ziinc, line 124: Hmm if an ok tuple is always expected then we can just do a match and can remove the case safely

lib/logflare_web/live/query_live.ex
```diff
@@ -242,6 +245,16 @@ defmodule LogflareWeb.QueryLive do
     {:noreply, socket}
   end
 
+  def handle_event("format-query", _params, socket) do
+    case SqlFmt.format_query(socket.assigns.query_string) do
```
Ziinc, line 249: As above

PR #2821, opened 2025-10-07
feat: add api for membership management

staaldraad:
Add and delete users from a team

### DIFF
--- a/lib/logflare_web/controllers/api/team_controller.ex
+++ b/lib/logflare_web/controllers/api/team_controller.ex
@@ -2,21 +2,20 @@ defmodule LogflareWeb.Api.TeamController do
   use LogflareWeb, :controller
   use OpenApiSpex.ControllerSpecs
 
+  require Logger
+
+  alias Logflare.Backends.Adaptor.BigQueryAdaptor
   alias Logflare.Teams
   alias Logflare.Teams.Team
   alias Logflare.TeamUsers
   alias Logflare.TeamUsers.TeamUser
   alias Logflare.User
   alias Logflare.Users
-  alias Logflare.Backends.Adaptor.BigQueryAdaptor
   alias LogflareWeb.OpenApi.Accepted
   alias LogflareWeb.OpenApi.Created
   alias LogflareWeb.OpenApi.List
   alias LogflareWeb.OpenApi.NotFound
   alias LogflareWeb.OpenApi.UnprocessableEntity
-
-  require Logger
-
   alias LogflareWeb.OpenApiSchemas.Team, as: TeamSchema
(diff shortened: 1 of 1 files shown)


test/logflare_web/controllers/api/team_controller_test.exs
```diff
@@ -201,4 +203,100 @@ defmodule LogflareWeb.Api.TeamControllerTest do
...
+      user: user,
+      main_team: main_team
+    } do
+      new_member = insert(:user)
+
+      assert conn
+             |> add_access_token(user, "private")
+             |> post(~p"/api/teams/#{main_team.token}/members", %{email: new_member.email})
+             |> response(204)
+             |> assert_schema("AcceptedResponse") == ""
```
Ziinc, line 219: should return created user, should also be a 201 since we're creating a team_user

test/logflare_web/controllers/api/team_controller_test.exs
```diff
@@ -201,4 +203,100 @@ defmodule LogflareWeb.Api.TeamControllerTest do
...
+    test "returns not found if doesn't own the team", %{conn: conn, main_team: main_team} do
+      invalid_user = insert(:user)
+      new_member = insert(:user)
+
+      assert conn
+             |> add_access_token(invalid_user, "private")
+             |> post(~p"/api/teams/#{main_team.token}/members", %{email: new_member.email})
+             |> json_response(404)
+             |> assert_schema("NotFoundResponse") == %{error: "Not Found"}
+    end
```
Ziinc, line 254: 401 would be more appropriate.

would also need a 400 test for bad team_user attrs

test/logflare_web/controllers/api/team_controller_test.exs
```diff
@@ -201,4 +203,100 @@ defmodule LogflareWeb.Api.TeamControllerTest do
...
+  end
+
+  describe "remove_member/2" do
```
Ziinc, line 257:
```suggestion
  describe "remove_member/2 for team_user with user record" do
```

Also needs a test for team_user with when there is no user record, which is when they are invited without having a prior account.

test/logflare_web/controllers/api/team_controller_test.exs
```diff
@@ -201,4 +203,100 @@ defmodule LogflareWeb.Api.TeamControllerTest do
              |> assert_schema("NotFoundResponse") == %{error: "Not Found"}
     end
   end
+
+  describe "add_member/2" do
+    test "adds an existing user to a team", %{
```
Ziinc, line 208: also needs case where there is no existing user record and we only create the record on `team_users` table

Ziinc (changes requested):
as discussed, would also need `private:admin` scope for pseudo rbac

PR #2829, opened 2025-10-09
fix: remove backends that cannot be queried from the list of selectable backends

Ziinc:
Removes non-querying backends from the list of selectable backends by making the callback optional.

### DIFF
--- a/lib/logflare/backends/adaptor.ex
+++ b/lib/logflare/backends/adaptor.ex
@@ -97,6 +97,16 @@ defmodule Logflare.Backends.Adaptor do
     |> function_exported?(:map_query_parameters, 4)
   end
 
+  @doc """
+  Returns true if a provided `Backend` supports executing queries.
+  """
+  @spec can_query?(Backend.t()) :: boolean()
+  def can_query?(%Backend{} = backend) do
+    backend
+    |> get_adaptor()
+    |> function_exported?(:execute_query, 3)
+  end
+
   @callback start_link(source_backend()) ::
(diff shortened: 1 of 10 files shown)


lib/logflare/backends/adaptor.ex
```diff
@@ -97,6 +97,16 @@ defmodule Logflare.Backends.Adaptor do
...
   end
 
+  @doc """
+  Returns true if a provided `Backend` supports executing queries.
+  """
+  @spec can_query?(Backend.t()) :: boolean()
+  def can_query?(%Backend{} = backend) do
+    backend
+    |> get_adaptor()
+    |> function_exported?(:execute_query, 3)
```
wojtekmach, line 107: Just a friendly reminder, I believe nothing to worry about today, that `function_exported?` checks if module _is_ loaded and if function exists but it does _not_ lazily load the module. In releases modules are eagerly loaded but in dev/test they're lazily loaded. So this could be giving unexpected results, seemingly random ones, see https://hexdocs.pm/elixir/main/Kernel.html#function_exported?/3. Per those docs, you could add a `Code.ensure_loaded?/1` which nowadays is fast when the module is in fact already loaded, but one always needs to be careful in hot paths.

That being said, you're using this pattern in other places in this module so I assume it's working well and by the time we hit this code, module is indeed already loaded. Thought I'd drop a note anyway!

Ziinc: will tidy this up in subsequent future PR

PR #2832, opened 2025-10-09
Interlinking:  create Query, Endpoint, or Alert from search view

msmithstubbs:
Adds a **Create New** button with a menu to link to new query, alerts, or endpoints with the current query.

The building the link relies on the SQL from a completed search operation, so the button is disabled until the first search execution completes. 

This has the benefit of ensuring the search operation has succeed. One drawback is that if the search field is changed the "Create new" links won't be updated until the search is executed. We could disable the button on search form change, or build the SQL from the LQL search query. I think to build it  from LQL will require a larger refactor of `LogsSearch` but it is an option. 

## Todo

- [x] Link to `/query` from search 
- [x] Link to `/alerts/new` from search 
- [x] Link to `/endpoints/new` from search 
- [x] Link to `/query` from debug modal
- [x] alerts and endpoints links behind feature flags 

https://github.com/user-attachments/assets/487dc58f-002d-49f2-ad9e-f27f8756bece

https://github.com/user-attachments/assets/1c5bb54a-650c-48fe-b786-e5461d9e572e

### DIFF
--- a/lib/logflare_web/live/search_live/logs_search_lv.ex
+++ b/lib/logflare_web/live/search_live/logs_search_lv.ex
@@ -441,7 +441,7 @@ defmodule LogflareWeb.Source.SearchLV do
 
     sql =
       Utils.sql_params_to_sql(search_op.sql_string, search_op.sql_params)
-      |> replace_table_with_source_name(source)
+      |> Utils.replace_table_with_source_name(source)
 
     destination =
       case resource do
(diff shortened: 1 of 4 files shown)


lib/logflare_web/live/query_live.ex
```diff
@@ -172,8 +172,15 @@ defmodule LogflareWeb.QueryLive do
...
     q =
       case params["q"] do
-        "" -> nil
-        v -> v
+        "" ->
+          nil
+
+        nil ->
+          nil
+
```
Ziinc, line 180:
```suggestion
        v when v in ["", nil] ->
          nil
```

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -427,6 +427,37 @@ defmodule LogflareWeb.Source.SearchLV do
     {:noreply, reset_search(socket)}
   end
 
+  def handle_event(
+        "create_new",
+        %{"kind" => kind, "resource" => resource},
+        %{assigns: assigns} = socket
+      ) do
+    %{source: source} = socket.assigns
```
Ziinc, line 435:
```suggestion
        %{assigns: %{source: source} = assigns} = socket
      ) do
```

lib/logflare_web/utils.ex
```diff
@@ -153,4 +153,41 @@ defmodule LogflareWeb.Utils do
 
     {Float.round(size, 2), unit}
   end
+
+  @spec sql_params_to_sql(String.t(), list()) :: String.t()
+  def sql_params_to_sql(sql, params) when is_binary(sql) and is_list(params) do
```
Ziinc, line 158: on a side note, storing of sql params in the BQ param structure is legacy behaviour and we can look towards refactoring it in future, i can imagine that this will be quite cumbersome when implementing search for postgres+clickhouse.

Ziinc: we haven't decided on a final schema format yet, but likely arrow

PR #2834, opened 2025-10-09
feat: sandbox query/cte support for clickhouse

amokan:
Follows the BQ approach to add ClickHouse-specific handling for sandboxed queries/cte support using ClickHouse SQL.

- Adds ClickHouse-specific function restrictions (`currentuser` and other functions that are meant to hit external systems outside of clickhouse)
- Updates documentation (`docs/docs.logflare.com/docs`) with relevant details about query support for ClickHouse backends and sandboxed queries.
- Paves a foundation to implement LQL-based sandboxed queries.


docs/docs.logflare.com/docs/backends/clickhouse.mdx
```diff
@@ -3,7 +3,7 @@ sidebar_position: 7
 ---
 # ClickHouse
 
-The ClickHouse backend is **ingest-only** to a ClickHouse [HTTP](https://clickhouse.com/docs/interfaces/http) endpoint.
+The ClickHouse backend supports both **ingestion** and **querying** via a ClickHouse [HTTP](https://clickhouse.com/docs/interfaces/http) endpoint.
```
Ziinc, line 6: just a heads up, we also need to eventually add in docs on the body flattening logic that is being used.

lib/logflare/sql.ex
```diff
@@ -95,8 +125,44 @@ defmodule Logflare.Sql do
...
 
-  # clickhouse and postgres
-  def transform(language, query, %User{} = user) when language in ~w(ch_sql pg_sql)a do
+  # clickhouse with sandboxed query support
+  def transform(:ch_sql = language, input, %User{} = user) do
+    {query, sandboxed_query} =
+      case input do
+        q when is_non_empty_binary(q) -> {q, nil}
+        other when is_tuple(other) -> other
+      end
```
Ziinc, line 134: can use a separate function head for this normalization.

lib/logflare/sql.ex
```diff
@@ -481,33 +553,58 @@ defmodule Logflare.Sql do
...
-  defp has_restricted_functions({_k, v}, :ok = acc) when is_list(v) or is_map(v) do
-    has_restricted_functions(v, acc)
+  defp has_restricted_functions({_k, v}, :ok = acc, data) when is_list(v) or is_map(v) do
+    has_restricted_functions(v, acc, data)
   end
 
-  defp has_restricted_functions(_kv, acc), do: acc
+  defp has_restricted_functions(_kv, acc, _data), do: acc
+
+  @spec get_restricted_functions_for_dialect(String.t() | nil) :: [String.t()]
```
Ziinc, line 604: list_restricted_functions_for_dialect

Ziinc: can remove `nil` arg.

lib/logflare/sql.ex
```diff
@@ -23,6 +23,9 @@ defmodule Logflare.Sql do
 
   @typep query_language :: :bq_sql | :ch_sql | :pg_sql
 
+  @bq_restricted_functions ~w(external_query session_user)
+  @ch_restricted_functions ~w(azureblobstorage cluster currentuser deltalake file gcs hdfs hudi iceberg jdbc mongodb mysql odbc postgresql redis remote remotesecure s3 sqlite url)
```
Ziinc, line 27:
```suggestion
  @ch_restricted_functions ~w(azureblobstorage cluster currentuser deltalake file gcs hdfs hudi iceberg jdbc mongodb mysql odbc postgresql redis remote remotesecure s3 sqlite url hostname getmacro fqdn sleep sleepeachrow currentdatabase currentschemas hasthreadfuzzer uptime version buildid filesystemavailable filesystemunreserved filesystemcapacity getsetting getsettingordefault tcpport currentprofiles enabledprofiles defaultprofiles currentroles enabledroles defaultroles getserverport getoskernelversion zookeepersessionuptime connectionid getclienthttpheader showcertificate displayname getserversetting getmergetreesetting)
```
other server admin functions

lib/logflare/sql.ex
Ziinc, line 604:
```suggestion
  @spec get_restricted_functions_for_dialect(String.t()) :: [String.t()]
```

test/logflare/sql_test.exs
```diff
@@ -439,6 +439,118 @@ defmodule Logflare.SqlTest do
 
       assert {:ok, %{1 => "company"}} = Sql.parameter_positions(ch_query, dialect: "clickhouse")
     end
+
+    test "sandboxed queries work with simple CTEs" do
```
Ziinc, line 443: should also include 1 e2e test that runs an actual sandboxed cte in the ClickHouseAdaptor test file and verifies that it cannot query sources outside of the CTE.
Also, there is some logic in the transformation that allows for BQ fully qualified names to do a passthrough if it matches the user's BYOB bigquery project id and dataset id, should secure and completely prevent the scenario in case for clickhouse as data security measure.
A bad actor could set their byob project id and dataset id to a source name together with the generated table name and could freely query it.

PR #2839, opened 2025-10-10
feat: account level default ingestion

Ziinc:
Closes ANL-1138

if backends.default_ingest=true → ingest all sources to default backend ( BigQuery) and all backends with this true flag

if backends.default_ingest=false + source_backends record → ingest only for that Source to default backend (BigQuery) and that specified backend 

Added a pill indicator for default ingestion.
Also removed the source backend list from sources page, as that is no longer relevant.

this greatly simplifies the ingestion model.
the old source column is left in as per non-destructive migration policy. will be removed when we eventually do a v2.


lib/logflare_web/live/backends/actions/show.html.heex
```diff
@@ -29,47 +29,57 @@
...
 
-  <.form :let={f} :if={@show_default_ingest_form?} id="default_ingest" for={%{}} as={:default_ingest} action="#" phx-submit="save_default_ingest">
-    <div class="form-group">
-      <%= label(f, :source_id, "Source") %>
-      <%= select(f, :source_id, Enum.map(@available_sources || [], fn s -> {s.name, s.id} end),
-        class: "form-control",
-        prompt: [key: "Choose a source"]
-      ) %>
-      <small id="source_id" class="form-text text-muted">This source will use this backend as its default ingest destination. Only sources with default ingest backend support enabled are shown.</small>
+    <div :if={length(@available_sources || []) > 0}>
```
josevalim, line 46:
```suggestion
    <div :if={@available_sources != []}>
```

lib/logflare_web/live/backends/actions/show.html.heex
```diff
@@ -29,47 +29,57 @@
...
+      <%= hidden_input(f, :backend_id, value: @backend.id) %>
+      <%= submit("Save", class: "btn btn-primary") %>
+    </.form>
 
-      <.button phx-click="remove_default_ingest" phx-value-source_id={source.id} variant="danger">
-        Remove
-      </.button>
-    </li>
-  </ul>
+    <p :if={Enum.empty?(@backend.sources || [])}>No sources configured yet</p>
```
josevalim, line 70: It seems this is a pattern from before this pull request, but it would be great to remove the conditionals such as `|| []`, by making sure they are always initialized to a list.

lib/logflare_web/live/backends/actions/show.html.heex
```diff
@@ -29,47 +29,57 @@
...
-        Add a Source
-      <% end %>
-    </.button>
-  </div>
+  <%= if @backend.default_ingest? do %>
+    <div class="alert alert-info">
+      <i class="fas fa-info-circle"></i>
+      <strong>Ingest from all sources:</strong> This backend is configured to receive logs from ALL your sources automatically.
+      To limit this backend to specific sources, disable the "Ingest from all sources" option in the backend settings.
+    </div>
```
josevalim, line 39: I am not sure if you folks decide the UI but I wonder if you should move the toggle here, so someone who wants to manage sources does not need to go to another page, then come back here, to control them.

josevalim (commented):
Looks great to me! Just some minor nitpicks on the UI bits :)

Ziinc:
Closing this as we are no longer going the route of account-level ingestion for now.

PR #2842, opened 2025-10-10
fix: reraise correctly and fix bug on pattern matching

nelsonmestevao:


lib/logflare_web/templates/marketing/index.html.eex
```diff
@@ -1,6 +1,6 @@
 <div class="homepage">
 
-  <%# hero %>
+  <% # hero %>
```
Ziinc, line 3: is there a reason for these changes??

nelsonmestevao: The `<%#` will be deprecated in next version of elixir (1.18), so I followed the advice to add a space, but I can revert it.

lib/logflare/logs/processor.ex
```diff
@@ -46,7 +46,7 @@ defmodule Logflare.Logs.Processor do
         Backends.ensure_source_sup_started(source)
         result = Backends.ingest_logs(batch, source)
 
-        new_meta = Map.merge(metadata, %{success: result == :ok})
+        new_meta = Map.merge(metadata, %{success: elem(result, 0) == :ok})
```
nelsonmestevao, line 49: Previously this was never true.

lib/logflare/context_cache/supervisor.ex
```diff
@@ -25,11 +25,11 @@ defmodule Logflare.ContextCache.Supervisor do
     Supervisor.start_link(__MODULE__, [], name: __MODULE__)
   end
 
-  @env Application.compile_env(:logflare, :env)
-
   @impl Supervisor
   def init(_) do
-    Supervisor.init(get_children(@env), strategy: :one_for_one)
+    Application.get_env(:logflare, :env)
```
nelsonmestevao, line 30: The compile time attribute was an issue for dialyzer to other envs other than the current one.

PR #2848, opened 2025-10-14
feat: include recommended query fields when fetching single event

msmithstubbs:
Ensure that event view modal and permalink include any source `bigquery_clustering_fields` and `suggested_keys` when fetching the individual log event. 

 

https://github.com/user-attachments/assets/290d627a-0ecc-4089-9bb4-dd46bc465772

This demo shows:

1. a source with a suggested search key and a clustering field
2. forcing a search without the suggested key
3. clicking the permalink opens a new tab with the suggested key and clustering field included in the query string.  

Closes ANL-1031

### DIFF
--- a/lib/logflare/logs/search_operations.ex
+++ b/lib/logflare/logs/search_operations.ex
@@ -14,6 +14,7 @@ defmodule Logflare.Logs.SearchOperations do
   alias Logflare.Logs.SearchUtils
   alias Logflare.Lql
   alias Logflare.Lql.BackendTransformer.BigQuery, as: BigQueryTransformer
+  alias Logflare.Lql.Rules
   alias Logflare.Lql.Rules.ChartRule
   alias Logflare.Lql.Rules.FilterRule
   alias Logflare.SourceSchemas
@@ -23,6 +24,9 @@ defmodule Logflare.Logs.SearchOperations do
 
   @type chart_period :: :day | :hour | :minute | :second
   @type dt_or_ndt :: DateTime.t() | NaiveDateTime.t()
+
+  @default_select_rules "s:timestamp s:id s:event_message"
+  @log_level_select_rule "s:metadata.level@level"
 
   @default_limit 100
   @default_max_n_chart_ticks 1_000
@@ -78,77 +82,6 @@ defmodule Logflare.Logs.SearchOperations do
       |> limit(@default_limit)
 
     %{so | query: query}
-  end
-
-  @doc """
-  Unnests and selects:
-
-  * recommended query fields (`bigquery_clustering_fields`, `suggested_keys`)
-  * `metadata.level`
-  """
-  @spec unnest_recommended_fields(SO.t()) :: SO.t()
-  def unnest_recommended_fields(%SO{source: source} = so) do
-    source_schema = SourceSchemas.Cache.get_source_schema_by(source_id: source.id)
-    flatmap = Map.get(source_schema || %{}, :schema_flat_map)
-
-    recommended_fields =
-      (Logflare.Sources.Source.recommended_query_fields(source) ++ ["metadata.level"])
-      |> Enum.uniq()
-
-    if Enum.empty?(recommended_fields) or is_nil(flatmap) do
-      so
-    else
-      {metadata_fields, top_level_fields} =
-        recommended_fields
-        |> Enum.filter(&Map.has_key?(flatmap, &1))
-        |> Enum.split_with(&String.starts_with?(&1, "metadata."))
-
-      metadata_field_atoms =
-        metadata_fields
-        |> Enum.map(fn "metadata." <> field -> String.to_atom(field) end)
-
-      top_level_field_atoms =
-        top_level_fields
-        |> Enum.map(&String.to_atom/1)
-
-      query =
-        so.query
-        |> maybe_unnest_metadata(metadata_field_atoms)
-        |> maybe_select_top_level_fields(top_level_field_atoms)
-
-      %{so | query: query}
-    end
-  end
-
-  defp maybe_unnest_metadata(query, []), do: query
-
-  defp maybe_unnest_metadata(query, metadata_field_atoms) do
-    select_map = build_select_map(metadata_field_atoms, :metadata)
-
-    query
-    |> join(:inner, [t], m in fragment("UNNEST(?)", t.metadata), on: true)
-    |> select_merge([..., m], ^select_map)
-  end
-
-  defp maybe_select_top_level_fields(query, []), do: query
-
-  defp maybe_select_top_level_fields(query, top_level_field_atoms) do
-    select_map = build_select_map(top_level_field_atoms, :top_level)
-
-    query
-    |> select_merge([t], ^select_map)
-  end
-
-  defp build_select_map(field_atoms, :metadata) do
-    Enum.reduce(field_atoms, %{}, fn field, acc ->
-      Map.put(acc, field, dynamic([..., m], field(m, ^field)))
-    end)
-  end
-
-  defp build_select_map(field_atoms, :top_level) do
-    Enum.reduce(field_atoms, %{}, fn field, acc ->
-      Map.put(acc, field, dynamic([t], field(t, ^field)))
-    end)
   end
 
   @spec apply_halt_conditions(SO.t()) :: SO.t()
@@ -411,19 +344,22 @@ defmodule Logflare.Logs.SearchOperations do
   defp to_value_unit(_average), do: {1, "MINUTE"}
 
   @spec apply_select_rules(SO.t()) :: SO.t()
+  def apply_select_rules(%SO{type: :events, query: _q, lql_rules: nil} = so) do
+    %{so | lql_rules: []}
+    |> apply_select_rules()
+  end
+
   def apply_select_rules(%SO{type: :events, query: q} = so) do
-    default_rules = [
-      Lql.Rules.SelectRule.build(path: "timestamp"),
-      Lql.Rules.SelectRule.build(path: "id"),
-      Lql.Rules.SelectRule.build(path: "event_message")
-    ]
+    default_rules = system_select_rules(so)
 
     select_rules =
       so.lql_rules
       |> Lql.Rules.get_select_rules()
       |> Kernel.++(default_rules)
+      |> Rules.SelectRule.normalize()
 
     q = Lql.apply_select_rules(q, select_rules)
+
     %{so | query: q}
   end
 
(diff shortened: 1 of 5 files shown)


lib/logflare_web/live/search_live/log_event_components.ex
```diff
@@ -5,11 +5,141 @@ defmodule LogflareWeb.SearchLive.LogEventComponents do
...
+          fields
+          |> Enum.reject(&MapSet.member?(existing_filter_paths, &1))
+          |> Enum.filter(&Map.has_key?(event_message, strip_meta(&1)))
+          |> Enum.map(fn field_name ->
+            Lql.Rules.FilterRule.build(
+              path: field_name,
+              operator: :=,
+              value: Map.get(event_message, strip_meta(field_name))
+            )
+          end)
```
msmithstubbs: Is it correct to only check `event.body["event_message"]` here? Do we need to check keys on `event.body` as well?

Ziinc: no this is incorrect. the event_message is irrelevant for building the lql.

If recommended fields are `my_field,other_field` and the user's current lql query is `some_string`, and the log event is as follows:
```
%LogEvent{
  body: %{
    "my_field" => 123,
    "other_field"=> "some value"
  }
}
``` 
then the the generated lql should be:
```
my_field:123 other_field:"some value" some_string
```

Ziinc: you can think of this as just adding additional LQL filters using the values on the event body.

lib/logflare_web/live/search_live/log_event_components.ex
```diff
@@ -5,11 +5,141 @@ defmodule LogflareWeb.SearchLive.LogEventComponents do
...
+  def lql_with_recommended_fields(lql_rules, event, source) do
+    fields = Source.recommended_query_fields(source)
+
+    existing_filter_paths =
+      lql_rules
+      |> Lql.Rules.get_filter_rules()
+      |> Enum.map(& &1.path)
+      |> MapSet.new()
+
+    case Jason.decode(event.body["event_message"]) do
```
Ziinc: we don't need to decode the event_message, fields should be based on body only.

test/logflare_web/live/search_live/log_events_components_test.exs
```diff
@@ -0,0 +1,138 @@
...
+          user: user,
+          suggested_keys: "m.user_id",
+          bigquery_clustering_fields: "session_id"
+        )
+
+      search_op_log_events = %{
+        rows: [
+          build(:log_event, message: "Log message 1", metadata: %{user_id: 123}, source: source),
+          build(:log_event,
+            message: Jason.encode!(%{session_id: "abc123"}),
```
Ziinc: we only care about the body

Ziinc: can be removed.

test/logflare_web/live/search_live/log_events_components_test.exs
```diff
@@ -0,0 +1,138 @@
...
+      html =
+        render_component(&LogEventComponents.logs_list/1, %{
+          search_op_log_events: nil,
+          last_query_completed_at: nil,
+          loading: true,
+          search_timezone: "America/New_York",
+          source: source,
+          tailing?: false,
+          querystring: "",
+          lql_rules: lql_rules
```
Ziinc: quite verbose, could we move most of these to default values:
```
          search_op_log_events: nil,
          last_query_completed_at: nil,
          loading: false,
          search_timezone: "Etc/UTC",
          tailing?: false,
          querystring: "",
```

test/logflare_web/live/search_live/log_events_components_test.exs
```diff
@@ -0,0 +1,138 @@
...
+        })
+
+      # Assert loading state is rendered
+      assert html =~ ~s|class="blurred list-unstyled console-text-list"|
```
Ziinc: can we use a dataid or element id instead? styling would be less stable in case of refactors.

test/logflare_web/live/search_live/log_events_components_test.exs
```diff
@@ -0,0 +1,138 @@
...
+          loading: false,
+          search_timezone: "America/New_York",
+          source: source,
+          tailing?: false,
+          querystring: "",
+          lql_rules: lql_rules
+        })
+
+      # Assert logs list ul is NOT rendered when search_op_log_events is nil
+      refute html =~ ~s|id="logs-list"|
```
Ziinc, line 129:
```suggestion
      refute html =~ ~s|id="logs-list"|
      refute html =~ ~s|<ul|
```
assertions across tests for similar checks should be the same

test/logflare_web/live/search_live/log_events_components_test.exs
```diff
@@ -0,0 +1,138 @@
...
+          tailing?: false,
+          querystring: "",
+          lql_rules: lql_rules
+        })
+
+      # Assert loading state is rendered
+      assert html =~ ~s|class="blurred list-unstyled console-text-list"|
+
+      # Assert logs list is NOT rendered
+      refute html =~ ~s|<ul|
```
Ziinc:
```suggestion
      refute html =~ ~s|id="logs-list"|
      refute html =~ ~s|<ul|
```

test/logflare_web/live/search_live/log_events_components_test.exs
```diff
@@ -0,0 +1,138 @@
...
+          loading: false,
+          search_timezone: "America/New_York",
+          source: source,
+          tailing?: false,
+          querystring: "",
+          lql_rules: lql_rules
+        })
+
+      assert html =~ "Log message 1"
+      assert html =~ Plug.HTML.html_escape(~s|{"session_id":"abc123"}|)
```
Ziinc:
```suggestion
      assert html =~ Plug.HTML.html_escape(~s|{"session_id":"abc123"}|)
      assert html =~ ~s|id="logs-list"|
      assert html =~ ~s|<ul|
```

lib/logflare_web/live/search_live/log_event_components.ex
msmithstubbs: uses event.body now

lib/logflare_web/live/search_live/log_event_components.ex
msmithstubbs: uses event.body now

test/logflare_web/live/search_live/log_events_components_test.exs
msmithstubbs: removed

test/logflare_web/live/search_live/logs_search_lv_test.exs
```diff
@@ -18,7 +18,11 @@ defmodule LogflareWeb.Source.SearchLVTest do
 
   defp setup_mocks(_ctx) do
     stub(GoogleApi.BigQuery.V2.Api.Jobs, :bigquery_jobs_query, fn _conn, _proj_id, _opts ->
-      {:ok, TestUtils.gen_bq_response()}
+      {:ok,
```
Ziinc: I think we're missing the changes needed for fetching the recommended fields in the search query, this assumes that the fields will be present in the fetched data already.
Some assertions into the mock function to conditionally add in the user_id field if the query selects it would be good.

Ziinc (changes requested):

test/logflare_web/live/search_live/logs_search_lv_test.exs
msmithstubbs: So I don't think the query does select `user_id` here though. That's why the links were timing out: when users performed a search that omitted search fields the permalink generated a query that timed out.  

My understanding is that `user_id` will **always** be returned here because it's just part of the row returned from the BQ query.

Can do this change though, just wanted to confirm.

msmithstubbs: Ignore above comment. Understand now -  need to change the search query to pull in these fields.

msmithstubbs: Added unnest_recommended_fields/1 to include those fields on event.body

lib/logflare/logs/search_operations.ex
```diff
@@ -80,24 +80,77 @@ defmodule Logflare.Logs.SearchOperations do
...
+      (Logflare.Sources.Source.recommended_query_fields(source) ++ ["metadata.level"])
+      |> Enum.uniq()
+
+    if Enum.empty?(recommended_fields) or is_nil(flatmap) do
+      so
+    else
+      {metadata_fields, top_level_fields} =
+        recommended_fields
+        |> Enum.filter(&Map.has_key?(flatmap, &1))
+        |> Enum.split_with(&String.starts_with?(&1, "metadata."))
```
Ziinc: need to consider `m.` as well

lib/logflare/logs/search_operations.ex
```diff
@@ -80,24 +80,77 @@ defmodule Logflare.Logs.SearchOperations do
...
 
       %{so | query: query}
-    else
-      so
     end
   end
 
+  defp maybe_unnest_metadata(query, []), do: query
+
+  defp maybe_unnest_metadata(query, metadata_field_atoms) do
```
Ziinc: have we considered adding to the LQL string using the select rules like `s:metadata.level s:my_field s:my.nested_field` ? it could greatly simplify this as it looks like we're duplicating the LQL select rules logic.

Ziinc (changes requested):

lib/logflare/logs/search_operations.ex
Ziinc: and technically we could use lql to generate the query fully, since with the select rules we're in control of what gets selected. execption is that we don't have any aliasing implemented for the select rules
Screenshot 2025-10-30 at 1 40 48 PM

msmithstubbs: Hadn't considered this. Yes, it would be simpler. 

They won't be unnested, so are you thinking we add support to Lql for that?
Or we could leave them as flattened keys in the search results and handle it when building the links.

Ziinc: hmm yea unnesting for the log level would be necessary to work with the current ui components....

thinking that we can get add in aliases like so:
```
select:my.nested.field@new_field_name
```
this syntax will follow the same timestamp shorthand pattern as `t:last@5days`, and using `@` as opposed to another `:` would be more readable.
 
 we can split this out to a separate PR if this is getting a a little too much changes.

Leaving it as nested (unflattened) in search results would require some updating to ui, which i am not opposed to either.

note that the future goal is to let users select fields into the log list view, so we need to have the separation between system-generated select rules and user-created select rules.

For example, selecting an additional field into the log list view:
```
s:m.error_code
```
would show somethign like this below the event message:
```
<timestamp> info My very long event message
                 metadata.error_code  404
```

PR #2850, opened 2025-10-14
feat: LQL sandboxed queries

amokan:
Introduces LQL support for sandboxed queries, enabling safe execution of user-defined LQL queries against CTE-based result sets in both BigQuery and ClickHouse backends.

- Adds `Lql.to_sandboxed_sql/3` to convert LQL query args to backend-specific SQL
- Adds full support for chart queries (time aggregations, etc) as well as select and filter rule support
- Note that if both `?lql` and `?sql` args are provided to an endpoint - the `?sql` arg will be used
- Adds simple liveview UX for testing sandbox queries (_screenshot below_)
- Removes `:ecto_ch` package in favor of our own Ecto -> Clickhouse SQL conversion pipeline for edge cases we have coming from LQL (`Logflare.Ecto.ClickHouse`).

Screenshot 2025-10-16 at 1 44 27 PM

----

**Update:** After building the test UI - it became clear we had some gaps on Ecto -> Clickhouse SQL generation. Ended up rolling our own conversion pipeline to better get a handle on that. Was able to drop `:ecto_ch` as a package in doing so.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
@@ -70,8 +70,8 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
   end
 
   @impl Logflare.Backends.Adaptor
-  def ecto_to_sql(%Ecto.Query{} = query, _opts) do
-    case Logflare.Ecto.ClickHouse.to_sql(query) do
+  def ecto_to_sql(%Ecto.Query{} = query, opts) do
+    case Logflare.Ecto.ClickHouse.to_sql(query, opts) do
       {:ok, {ch_sql, ch_params}} ->
         ch_params = Enum.map(ch_params, &SqlUtils.normalize_datetime_param/1)
         {:ok, {ch_sql, ch_params}}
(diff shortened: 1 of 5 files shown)


chasers (approved):

test/logflare/lql/backend_transformer/clickhouse_test.exs
```diff
@@ -9,22 +9,37 @@ defmodule Logflare.Lql.BackendTransformer.ClickHouseTest do
   alias Logflare.Lql.Rules.SelectRule
 
   describe "behaviour implementation" do
-    test "dialect/0 returns 'clickhouse'" do
-      assert ClickHouse.dialect() == "clickhouse"
+    test "implements all required callbacks" do
+      assert function_exported?(ClickHouse, :transform_filter_rule, 2)
```
edgurgel, line 13: BTW you can make this a bit more automated using `module.behaviour_info(:callbacks)` like this:

```elixir
iex> GenServer.behaviour_info(:callbacks)
[
  terminate: 2,
  init: 1,
  handle_info: 2,
  handle_continue: 2,
  handle_cast: 2,
  handle_call: 3,
  format_status: 2,
  format_status: 1,
  code_change: 3
]
```

Also using `@impl` makes the compiler check for missing callbacks AFAIK

PR #2865, opened 2025-10-27
feat: adds new LQL `FromRule`

amokan:
Adds support for `f:sandboxed_table_or_source` or `from:` LQL rule

- [x] only allows 1 table to be selected
- [x] search ui should set the `f` rule automatically
- [x] when used in the context of a sandbox query endpoint, if more than one cte is available - require the from rule to be supplied. If only a single cte - default to use that if a from rule is not supplied.

### DIFF
--- a/test/logflare_web/live/search_live/logs_search_lv_test.exs
+++ b/test/logflare_web/live/search_live/logs_search_lv_test.exs
@@ -333,7 +333,6 @@ defmodule LogflareWeb.Source.SearchLVTest do
 
       querystring = find_querystring(html)
 
-      assert querystring =~ ~r/^f:[a-f0-9\-]+ /
       assert querystring =~ "c:count(*) c:group_by(t::minute)"
     end
 


lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -120,7 +122,8 @@ defmodule LogflareWeb.Source.SearchLV do
     socket =
       with {:ok, lql_rules} <-
              Lql.decode(qs, get_bigquery_schema(source)),
-           lql_rules = Lql.Rules.put_new_chart_rule(lql_rules, Lql.Rules.default_chart_rule()),
+           lql_rules = Rules.put_new_chart_rule(lql_rules, Rules.default_chart_rule()),
+           lql_rules = ensure_from_rule(lql_rules, source),
```
Ziinc: don't think we need `ensure_from_rule/2` in this situation, since search ui is already restricted to 1 source so having a from rule would be redundant. 
also think that adding in a `from:` rule to user LQL as of now might break queries when we do a gradual rollout as clusters on previous versions would be lagging behind and would not know of what a `from` rule is.

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -713,6 +716,19 @@ defmodule LogflareWeb.Source.SearchLV do
...
+      lql_rules
+    else
+      from_rule = %FromRule{
+        table: source.token,
+        table_type: :source
+      }
+
+      Rules.put_from_rule(lql_rules, from_rule)
+    end
+  end
```
Ziinc:
```suggestion
```
as above.

PR #2868, opened 2025-10-28
fix: error response for plugs

Ziinc:
This properly fixes error responses for plugs by delegating to the `LogflareWeb.API.FallbackController`

The ErrorView changes do not fix underlying responses as view is not set.


SteffenDE (commented):
Looks great to me! I like the extra fetch functions :)

(It looks like some tests are still failing because there are cases where the FallbackController is called with nil, but I'm sure you're already on it)

PR #2879, opened 2025-10-30
chore: simplify clickhouse schema for raw ingest logs

amokan:
This PR simplifies the Clickhouse source ingest table schema to:

- `id` UUID
- `payload` String
- `timestamp` DateTime64(6)

With `payload` representing the stringified `LogEvent.body` attribute without flattening or pruning of attributes.

This gives us more flexibility with our intentions around materialized views and maintains the canonical data shape that we received.

_Note that this will **NOT** migrate any existing ingest tables - but simply provision new ingest table(s) with the updated schema._

### Additional Changes Worth Calling Out

- Clickhouse source ingest tables are now prefixed with `ingest_logs_` rather than `log_events_`. This was to avoid collisions on existing tables we have in the wild as we work through this transition. We can go back to `log_events` if we want to at a later time.
- The default ingest table TTL has been changed from 3 days to 5 days
- Removed the key count materialized view creation and target table creation for key counts. This will be reimplemented later along with additional mat views.

### DIFF
--- a/test/logflare/sql_test.exs
+++ b/test/logflare/sql_test.exs
@@ -514,9 +514,9 @@ defmodule Logflare.SqlTest do
       Process.sleep(200)
 
       cte_query =
-        "with src as (select event_message from #{source.name}) select event_message from src"
+        "with src as (select payload from #{source.name}) select payload from src"
 
-      consumer_query = "select event_message from src"
+      consumer_query = "select payload from src"
 
       assert {:ok, transformed} = Sql.transform(:ch_sql, {cte_query, consumer_query}, user)
       assert {:ok, results} = ClickhouseAdaptor.execute_query(backend, transformed, [])


lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
```diff
@@ -6,22 +6,12 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.QueryTemplates do
   import Logflare.Utils.Guards
 
   @default_table_engine Application.compile_env(:logflare, :clickhouse_backend_adaptor)[:engine]
-  @default_ttl_days 3
+  @default_ttl_days 5
```
Ziinc, line 9: our default for BQ is 3d though, why the change?

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -379,22 +357,16 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
...
-          Jason.encode!(flattened_body),
+          Jason.encode!(log_event.body),
           DateTime.from_unix!(log_event.body["timestamp"], :microsecond)
         ]
       end)
 
     opts = [
-      names: ["id", "event_message", "body", "timestamp"],
-      types: ["UUID", "String", "String", "DateTime64(6)"]
+      names: ["id", "payload", "timestamp"],
```
Ziinc, line 368: why not preserve the `body` column name? 
it would match the LogEvent's `body` as well as postgres backend's use of `body` column.

amokan: I can change it back to `body` - not a big deal.

PR #2882, opened 2025-10-31
Axiom.co adaptor

bblaszkow06:
The adaptor implementation for axiom.co based on the WebhookAdaptor

### DIFF
--- a/lib/logflare/backends/adaptor/axiom_adaptor.ex
+++ b/lib/logflare/backends/adaptor/axiom_adaptor.ex
@@ -53,7 +53,6 @@ defmodule Logflare.Backends.Adaptor.AxiomAdaptor do
     url =
       config
       |> dataset_uri()
-      |> URI.append_path("/ingest")
       |> URI.append_query(URI.encode_query(query))
       |> URI.to_string()
 
@@ -119,14 +118,13 @@ defmodule Logflare.Backends.Adaptor.AxiomAdaptor do
         ],
         {Tesla.Adapter.Finch, name: Logflare.FinchDefault, receive_timeout: 5_000}
       )
-      |> Tesla.get(url)
+      |> Tesla.post(url, [])
 
     case result do
-      {:error, _reason} = err -> err
-      {:ok, %Tesla.Env{status: 403}} -> {:error, "Unauthorized: possibly invalid auth token"}
-      {:ok, %Tesla.Env{status: 404}} -> {:error, "Dataset #{config.dataset_name} doesn't exist"}
       {:ok, %Tesla.Env{status: 200}} -> :ok
+      {:ok, %Tesla.Env{body: %{"message" => message}}} -> {:error, message}
       {:ok, env} -> {:error, "Unexpected response: #{env.status} #{inspect(env.body)}"}
+      {:error, reason} -> {:error, "Request error: #{reason}"}
     end
   end
 
@@ -134,7 +132,7 @@ defmodule Logflare.Backends.Adaptor.AxiomAdaptor do
     %URI{
       scheme: "https",
       host: domain,
-      path: "/v1/datasets/" <> dataset_name
+      path: "/v1/datasets/#{dataset_name}/ingest"
     }
   end
 end
--- a/test/logflare/backends/adaptor/axiom_adaptor_test.exs
+++ b/test/logflare/backends/adaptor/axiom_adaptor_test.exs
@@ -87,25 +87,15 @@ defmodule Logflare.Backends.Adaptor.AxiomAdaptorTest do
     test "succceeds on 200 response", ctx do
       @tesla_adapter
       |> expect(:call, 2, fn env, _opts ->
-        assert env.method == :get
-        assert env.url == "https://api.axiom.co/v1/datasets/#{@valid_config.dataset_name}"
+        assert env.method == :post
+        assert env.url == "https://api.axiom.co/v1/datasets/#{@valid_config.dataset_name}/ingest"
         assert Tesla.get_header(env, "authorization") == "Bearer #{@valid_config.api_token}"
 
         {:ok,
          %Tesla.Env{
            status: 200,
            body:
-             """
-             {
-               "created": "2022-07-20T02:35:14.137Z",
-               "description": "string",
-               "id": "string",
-               "kind": "axiom:events:v1",
-               "name": "string",
-               "who": "string"
-             }
-             """
-             |> Jason.decode!()
+             ~s({"ingested":0,"failed":0,"failures":[],"processedBytes":2,"blocksCreated":0,"walLength":0})
          }}
       end)
 
@@ -113,65 +103,29 @@ defmodule Logflare.Backends.Adaptor.AxiomAdaptorTest do
       assert :ok = @subject.test_connection({ctx.source, ctx.backend})
     end
 
-    test "returns error on 403 response", ctx do
-      @tesla_adapter
-      |> expect(:call, fn _env, _opts ->
+    test "returns error on failure", ctx do
+      error_responses = [
+        {:ok,
+         %Tesla.Env{status: 401, body: ~s({"code":401,"message":"forbidden"})}
+         |> Tesla.put_header("content-type", "application/json")},
         {:ok,
          %Tesla.Env{
            status: 403,
-           body:
-             """
-             {
-               "code": 403,
-               "message": "Forbidden"
-             }
-             """
-             |> Jason.decode!()
-         }}
-      end)
-
-      assert {:error, reason} = @subject.test_connection(ctx.backend)
-      assert reason =~ "auth"
-    end
-
-    test "returns error on 404 response", ctx do
-      @tesla_adapter
-      |> expect(:call, fn _env, _opts ->
-        {:ok,
-         %Tesla.Env{
-           status: 404,
-           body:
-             """
-             {
-               "code": 404,
-               "message": "Not found"
-             }
-             """
-             |> Jason.decode!()
-         }}
-      end)
-
-      assert {:error, reason} = @subject.test_connection(ctx.backend)
-      assert reason =~ @valid_config.dataset_name
-    end
-
-    test "returns error on 500 response", ctx do
-      @tesla_adapter
-      |> expect(:call, fn _env, _opts ->
-        {:ok, %Tesla.Env{status: 500, body: ""}}
-      end)
-
-      assert {:error, reason} = @subject.test_connection(ctx.backend)
-      assert is_binary(reason)
-    end
-
-    test "returns error on request error", ctx do
-      @tesla_adapter
-      |> expect(:call, fn _env, _opts ->
+           body: ~s({"code":403,"message":"not allowed to ingest into dataset"})
+         }
+         |> Tesla.put_header("content-type", "application/json")},
         {:error, :nxdomain}
-      end)
+      ]
 
-      assert {:error, :nxdomain} = @subject.test_connection(ctx.backend)
+      for response <- error_responses do
+        @tesla_adapter
+        |> expect(:call, fn _env, _opts ->
+          response
+        end)
+
+        assert {:error, reason} = @subject.test_connection(ctx.backend)
+        assert is_binary(reason)
+      end
     end
   end
 


docs/docs.logflare.com/docs/backends/axiom.mdx
```diff
@@ -0,0 +1,28 @@
+---
+sidebar_position: 9
+---
+
+# Axiom
+
+The Axiom backend is **ingest-only**, and sends the logs to Axiom.co via REST API
+
+It has support for connection testing.
```
Ziinc:
```suggestion
```

docs/docs.logflare.com/docs/backends/axiom.mdx
```diff
@@ -0,0 +1,28 @@
+---
+sidebar_position: 9
+---
+
+# Axiom
+
+The Axiom backend is **ingest-only**, and sends the logs to Axiom.co via REST API
```
Ziinc:
```suggestion
The Axiom backend is **ingest-only**, and sends the logs to Axiom.co via their [HTTP API](https://axiom.co/docs/restapi/ingest)
```

docs/docs.logflare.com/docs/backends/axiom.mdx
```diff
@@ -0,0 +1,28 @@
...
+  The dataset must be created in Axiom before starting the ingestion from Logflare.
+- `:domain` (string) - The domain specific to the region your account is using,
+    Defaults to the US domain (`api.axiom.co`)
+
+### Implementation Details
+
+Implementation is based on the [webhook backend](/backends/webhook).
+
+The timestamp of each log event is transformed into format recognized by axiom,
+otherwise the log events are forwarded without modifications.
```
Ziinc:
```suggestion
The timestamp of each log event is transformed into format recognized by Axiom.
```

lib/logflare/backends/adaptor/axiom_adaptor.ex
```diff
@@ -0,0 +1,140 @@
...
+      |> Tesla.get(url)
+
+    case result do
+      {:error, _reason} = err -> err
+      {:ok, %Tesla.Env{status: 403}} -> {:error, "Unauthorized: possibly invalid auth token"}
+      {:ok, %Tesla.Env{status: 404}} -> {:error, "Dataset #{config.dataset_name} doesn't exist"}
+      {:ok, %Tesla.Env{status: 200}} -> :ok
+      {:ok, env} -> {:error, "Unexpected response: #{env.status} #{inspect(env.body)}"}
+    end
+  end
```
Ziinc, line 129: Would be better to make a POST request with an empty payload (i.e. no log events in the post)
Would better test that the http request would actually go through.

[axiom's API](https://axiom.co/docs/restapi/endpoints/createToken) doesn't have a nice GET equivalent for the ingest endpoint that lets us do a HEAD.
The api key might also not have any dataset permissions beyond ingest, so cannot rely on this uri to verify connection

Ziinc: should also try to get the tesla client as close to the actual client used in webhook adaptor, or reuse it. that way we can verify that it would actually be close to actual ingestion request.

lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -297,6 +297,28 @@
               The batch timeout in milliseconds.
             </small>
           </div>
+        <% "axiom" -> %>
+          <div class="form-group">
+            {label(f_config, :dataset_name, "Dataset name")}
+            {text_input(f_config, :dataset_name, class: "form-control")}
+            <small class="form-text text-muted">
+              The name of the Axiom's dataset used by the backend
```
Ziinc:
```suggestion
              The name of the Axiom's dataset to ingest into
```

lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -297,6 +297,28 @@
...
+            </small>
+          </div>
+          <div class="form-group">
+            {label(f_config, :api_token, "API Token")}
+            {text_input(f_config, :api_token, class: "form-control")}
+            <small class="form-text text-muted">
+              Axiom API token used for authorization
```
Ziinc:
```suggestion
              Axiom API token used for authorization. Requires the dataset ingest capability.
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+    api_token = "THE-API-KEY"
+    dataset_name = "logflare"
+    domain = "api.axiom.co"
+
+    config = %{api_token: api_token, dataset_name: dataset_name, domain: domain}
+
+    backend =
+      insert(:backend,
+        type: :axiom,
+        sources: [source],
```
Ziinc, line 22:
```suggestion
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+      assert {:ok, _} = Backends.ingest_logs(log_events, source)
+      assert_receive {^ref, body}, 2000
+
+      assert length(body) == 3
+
+      # Check that all expected messages are present
+      messages = Enum.map(body, fn item -> item["event_message"] end)
+      assert "Log 1" in messages
+      assert "Log 2" in messages
+      assert "Log 3" in messages
```
Ziinc:
```suggestion
      assert_receive {^ref, [_, _, _]}, 2000
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+          source: source,
+          event_message: "Log 2",
+          timestamp: 1_704_067_200_000_000
+        ),
+        build(:log_event,
+          source: source,
+          event_message: "Log 3",
+          timestamp: 1_704_067_200_000_000
+        )
+      ]
```
Ziinc:
```suggestion
      log_events = build_list(:log_event, 3, source: source)
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+  use Logflare.DataCase, async: false
+
+  alias Logflare.Backends
+  alias Logflare.Backends.Adaptor
+  alias Logflare.Backends.AdaptorSupervisor
+  alias Logflare.Backends.Adaptor.WebhookAdaptor.Client
+  alias Logflare.SystemMetrics.AllLogsLogged
+
+  @subject Adaptor.AxiomAdaptor
+
```
Ziinc, line 11:
```suggestion
  @valid_config %{...}
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+    dataset_name = "logflare"
+    domain = "api.axiom.co"
+
+    config = %{api_token: api_token, dataset_name: dataset_name, domain: domain}
+
+    backend =
+      insert(:backend,
+        type: :axiom,
+        sources: [source],
+        config: config
```
Ziinc:
```suggestion
        config: @valid_config
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+
+  defp backend_data(_ctx) do
+    user = insert(:user)
+    source = insert(:source, user: user)
+
+    api_token = "THE-API-KEY"
+    dataset_name = "logflare"
+    domain = "api.axiom.co"
+
+    config = %{api_token: api_token, dataset_name: dataset_name, domain: domain}
```
Ziinc:
```suggestion
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+
+    test "sets default options" do
+      api_token = "THE-API-KEY"
+      dataset_name = "logflare"
+
+      changeset =
+        Adaptor.cast_and_validate_config(@subject, %{
+          "api_token" => api_token,
+          "dataset_name" => dataset_name
+        })
```
Ziinc:
```suggestion
      changeset =
        Adaptor.cast_and_validate_config(@subject,  Map.drop(@valid_config, ["domain"])
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+
+      assert changeset.valid?
+
+      data = Ecto.Changeset.apply_changes(changeset)
+      assert map_size(data) == 3
+      assert data.api_token == api_token
+      assert data.dataset_name == dataset_name
+      assert data.domain == "api.axiom.co"
```
Ziinc:
```suggestion
```
testing implementation.

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+      api_token = "THE-API-KEY"
+      dataset_name = "logflare"
+      eu_domain = "api.eu.axiom.co"
+
+      changeset =
+        Adaptor.cast_and_validate_config(@subject, %{
+          "api_token" => api_token,
+          "dataset_name" => dataset_name,
+          "domain" => eu_domain
+        })
```
Ziinc:
```suggestion
      changeset =
        Adaptor.cast_and_validate_config(@subject, %{
          @valid_config | "domain" => eu_domain
        })
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+
+      assert changeset.valid?
+
+      data = Ecto.Changeset.apply_changes(changeset)
+      assert map_size(data) == 3
+      assert data.api_token == api_token
+      assert data.dataset_name == dataset_name
+      assert data.domain == eu_domain
```
Ziinc:
```suggestion
      assert %{domain: ^eu_domain} = Ecto.Changeset.apply_changes(changeset)
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+
+  describe "transform_config/1" do
+    test "converts config to WebhookAdaptor format" do
+      api_token = "THE-API-KEY"
+      dataset_name = "logflare"
+      domain = "api.axiom.co"
+
+      backend = %{
+        config: %{api_token: api_token, dataset_name: dataset_name, domain: domain}
+      }
```
Ziinc:
```suggestion
    test "converts config to WebhookAdaptor format" do
      backend = %{config: @valid_config}
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+
+      assert config.headers["content-type"] == "application/json"
+      assert config.headers["authorization"] == "Bearer #{api_token}"
+      assert config.http == "http2"
+      assert config.gzip == true
+      assert is_function(config.format_batch)
+    end
+  end
+
+  describe "test_connection/1" do
```
Ziinc, line 82: 1 valid test case and 1 invalid test case is enough

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+      assert reason =~ "auth"
+    end
+
+    test "returns error on 404 response", ctx do
+      @tesla_adapter
+      |> expect(:call, fn _env, _opts ->
+        {:ok,
+         %Tesla.Env{
+           status: 404,
+           body:
```
Ziinc: mocking body is not needed for this assertion

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+      assert :ok = @subject.test_connection({ctx.source, ctx.backend})
+    end
+
+    test "returns error on 403 response", ctx do
+      @tesla_adapter
+      |> expect(:call, fn _env, _opts ->
+        {:ok,
+         %Tesla.Env{
+           status: 403,
+           body:
```
Ziinc: mocking body is not needed for this assertion

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+    end
+
+    test "returns error on request error", ctx do
+      @tesla_adapter
+      |> expect(:call, fn _env, _opts ->
+        {:error, :nxdomain}
+      end)
+
+      assert {:error, :nxdomain} = @subject.test_connection(ctx.backend)
+    end
```
Ziinc: testing implementation at this point.

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+      message = "Test log message"
+
+      log_events = [
+        build(:log_event,
+          source: source,
+          event_message: message,
+          random_attribute: "nothing",
+          timestamp: 1_704_067_200_000_000
+        )
+      ]
```
Ziinc:
```suggestion
      log_event = build(:log_event, source: source, random_attribute: "nothing")
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+
+      assert {:ok, _} = Backends.ingest_logs(log_events, source)
```
Ziinc:
```suggestion
      assert {:ok, _} = Backends.ingest_logs([log_event], source)
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+      assert_receive {^ref, body}, 2000
+      assert [log] = body
+      assert log["event_message"] == message
+      assert log["timestamp"] == "2024-01-01T00:00:00.000000Z"
+      assert log["random_attribute"] == "nothing"
```
Ziinc:
```suggestion
      assert_receive {^ref, [log]}, 2000
      assert log["event_message"] == log_event.body["event_message"]
      assert log["timestamp"] != log_event.body["timestamp"]
      assert log["random_attribute"] == "nothing"
```

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -0,0 +1,302 @@
...
+      |> expect(:call, 2, fn env, _opts ->
+        assert env.method == :get
+        assert env.url == "https://api.axiom.co/v1/datasets/#{ctx.config.dataset_name}"
+        assert Tesla.get_header(env, "authorization") == "Bearer #{ctx.config.api_token}"
+
+        {:ok,
+         %Tesla.Env{
+           status: 200,
+           body:
+             """
```
Ziinc: do we need to mock the body here? seems inconsequential.

Ziinc (changes requested):
looks good ❤️ , mostly issues with testing conciseness and the connection testing method. 


We'll also need an equivalent UI PR for this, [see](https://github.com/Logflare/logflare/pull/2787)

lib/logflare/backends/adaptor/axiom_adaptor.ex
bblaszkow06: > Would be better to make a POST request with an empty payload (i.e. no log events in the post)
Would better test that the http request would actually go through.

The ingest-only token seems to allow GET on the dataset, but that's a good idea anyway! While a completely empty body doesn't work, an empty array is accepted.

>should also try to get the tesla client as close to the actual client used in webhook adaptor, or reuse it. that way we can verify that it would actually be close to actual ingestion request.

Sharing or reusing the client seems like the way to go. I didn't want to bloat the scope of this PR, but IMO it would be best if instead of passing all the args for `WebhookAdaptor.Client` via `WebhookAdaptor`, the caller could pass `Tesla.Client.t` with all the middleware in place. My idea for that would require:
- Abstracting away `TeslaAdaptor/HttpAdaptor`
  - defining Broadway pipeline
  - accepting Tesla.Client
  - meant to be only used by other adaptors, not directly (no Adaptor behaviour)
- Migrating the WebhookAdaptor to use that
- Possibly migrating other Adaptors to use it or ensure the WebhookAdaptor keeps a compatible API

This would get rid of the deprecated module-based Tesla client, allow to share the client definition for connection testing and log draining. Tesla's middleware could be used for body transformation, replacing the `format_batch` function, and it would be easier to manipulate them (e.g. no need to pass a flag adding gzip compression).

WDYT about going this way in next PR (and keeping this one as-is)?

test/logflare/backends/adaptor/axiom_adaptor_test.exs
bblaszkow06: 1. `log_event.body["event_message"]` strongly couples this test to the log_event shape. Can I assume this won't change any time soon?
2. This is the only test verifying the correctness of timestamp conversion. Why do you suggest to remove it?

test/logflare/backends/adaptor/axiom_adaptor_test.exs
bblaszkow06: This actually breaks the test cases (`Backends.ingest_logs` won't forward them). This was based on Sentry adaptor, if there's a better way to enable forwarding in tests, let me know

lib/logflare/backends/adaptor/axiom_adaptor.ex
Ziinc: Not opposed to reworking it to be more middleware oriented.

Yes we can work on it next as it would affect the two other new adaptors.

Also, something else that we need to refactor is the way that we pull config for the WebhookAdaptor, it needs to pull from Backends.Cache on init and in the Broadway pipeline, to avoid stale configs.

Let's do them together in the next PR

test/logflare/backends/adaptor/axiom_adaptor_test.exs
Ziinc: Ah OK it's fine then

test/logflare/backends/adaptor/axiom_adaptor_test.exs
Ziinc: - event_message won't change and is a special field and will always be present. 
- removes need for specific timestamp being set above and reducing loc, and asserts the same thing (that the unix timestamp gets formatted). Testing that it's an ISO timestamp doesn't matter as it's an implementation detail.

test/logflare/backends/adaptor/axiom_adaptor_test.exs
bblaszkow06: The inconsequence comes from the fact, that for test_connection I can mock Finch adaptor i.e. I mock the response at quite low level, so it goes through the middlewares. Ingest tests mock the whole client, bypassing Tesla, while the returned value is ignored. It also assumes the client is tested somewhere else.

While it's not necessary, I've added the bodies as they are part of the contract - these are the responses we might get and I wanted to ensure they all are handled, as I can imagine a scenario where a middleware crashes for a real response with body while test passes with a mock, empty body. And it has more black-box approach ignoring the fact that only status is used by the implementation.

Anyway, these more thorough tests, added with coming refactor in mind, are something I'm more used to.

If that doesn't convince you, I'll remove the bodies, of course 😉 
EDIT: It will change anyway after switching to POST `/ingest` but I'm still curious for the answer so I can adapt to the testing guidelines here

test/logflare/backends/adaptor/axiom_adaptor_test.exs
bblaszkow06: > Testing that it's an ISO timestamp doesn't matter as it's an implementation detail.

On that I wouldn't agree, the timestamp is parsed by Axiom, so it has to conform to the format passed as a query parameter, the fact it has ISO format with particular precision is part of the contract I'd like to test here

Ziinc (approved):

test/logflare/backends/adaptor/axiom_adaptor_test.exs
Ziinc: Hmm fair point on comprehensiveness, though it does add substantial additional test code to maintain and makes the test less readable. 
it isn't a big deal to include it, but can be misleading if the actual API responses change and results in mock drifting from reality. Would prefer in future that tests include the minimum amount of logic to reach the desired test assertion criteria, because any addition just adds overhead to the already large codebase. Regressions in such scenarios can be handled in separate additioanl cases when they actually do arise, as opposed to pre-eptively trying to cover all areas.

PR #2884, opened 2025-10-31
fix(ch): when selecting a list of fields, don't interpret it as an array

v0idpwn:
`select(q, t, [t.a, t.b, t.c])` refers to selecting `a`, `b` and `c` columns, but it was being treated as an `[a, b, c]` array.


amokan (approved):
Looks good!

PR #2893, opened 2025-11-06
chore: rework of clickhouse ingestion pipeline to use a simplified flow for our fixed schema

amokan:
Wasn't happy with the behavior of the old ingest flow as there were a lot of abstractions around connections. Given we have a very simple ingest schema, it made more sense to really dial that in and leverage `Finch` for connections.

Connection pools for ingest are now configured in the `Logflare.Application` module.

Read queries, grant checks, and provisioning are still handled via the `Ch` client and it's `DbConnection` based flow.

### DIFF
--- a/lib/logflare/application.ex
+++ b/lib/logflare/application.ex
@@ -188,7 +188,7 @@ defmodule Logflare.Application do
        pools: %{
          default: [
            protocols: [:http1],
-           size: max(base, 20),
+           size: max(base * 125, 150),
            count: http1_count,
            start_pool_metrics?: true
          ]


lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -0,0 +1,201 @@
...
+      encode_datetime64(timestamp)::binary
+    >>
+  end
+
+  @doc false
+  @spec encode_batch([__MODULE__.Row.t()]) :: binary()
+  def encode_batch([%__MODULE__.Row{} | _] = rows) do
+    rows
+    |> Enum.map(&encode_row/1)
+    |> :erlang.iolist_to_binary()
```
v0idpwn: Feels like you could return iodata here?

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -0,0 +1,201 @@
...
+  end
+
+  @doc false
+  @spec encode_row(__MODULE__.Row.t()) :: binary()
+  def encode_row(%__MODULE__.Row{id: id, body: body, timestamp: timestamp}) do
+    <<
+      encode_uuid(id)::binary,
+      encode_string(body)::binary,
+      encode_datetime64(timestamp)::binary
+    >>
```
v0idpwn: and here

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
v0idpwn: I don't think it makes much of a difference in this case (since the binary isn't moved anywhere, just gzipped afterwards, it will be kept as refs) but probably a good practice

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -0,0 +1,201 @@
...
+  import Logflare.Utils.Guards
+
+  alias Logflare.Backends.Backend
+  alias Logflare.LogEvent
+
+  typedstruct module: Row do
+    field :id, Ecto.UUID.t()
+    field :body, String.t()
+    field :timestamp, DateTime.t()
+  end
```
v0idpwn: In my understanding this is ported to map the schema that existed previously, but with the new approach doesn't it make more sense to encode LogEvents directly?

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -0,0 +1,181 @@
...
+        {:error, reason}
+    end
+  end
+
+  @doc false
+  @spec encode_row(LogEvent.t()) :: iodata()
+  def encode_row(%LogEvent{body: body}) do
+    [
+      encode_uuid(body["id"]),
+      encode_string(Jason.encode!(body)),
```
v0idpwn:
```suggestion
      encode_string(Jason.encode_to_iodata!(body)),
```

v0idpwn (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
amokan: Great idea!

lib/logflare/application.ex
```diff
@@ -183,6 +183,16 @@ defmodule Logflare.Application do
            ]
          ]
        }},
+      {Finch,
+       name: Logflare.FinchClickhouseIngest,
+       pools: %{
+         default: [
+           protocols: [:http1],
+           size: max(base, 20),
```
amokan: @Ziinc would love your insight on this pool size for these

Ziinc: If we keep it centralised then we would need to have much more, on par with BQ http1 size.

lib/logflare/backends/adaptor/clickhouse_adaptor/connection_manager.ex
```diff
@@ -148,27 +90,13 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.ConnectionManager do
   end
 
   @impl true
-  def init(args) do
+  def init(%Backend{id: backend_id}) do
```
Ziinc: if we could just set this to backend_id if thats the case.
less terms stored in the process dict.

amokan: Done

lib/logflare/application.ex
amokan: I've matched this with the google pool. Once we validate this approach is smooth, I can knock out an alternative approach with smaller, ephemeral pools per source/backend if we want to go that route.

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -0,0 +1,180 @@
...
+  def insert(%Backend{} = backend, table, log_events, opts)
+      when is_list(log_events) and is_list(opts) do
+    with {:ok, connection_opts} <- build_connection_opts(backend) do
+      insert(connection_opts, table, log_events, opts)
+    end
+  end
+
+  def insert(connection_opts, table, [%LogEvent{} | _] = log_events, opts)
+      when is_list(connection_opts) and is_non_empty_binary(table) and is_list(opts) do
+    compress = Keyword.get(opts, :compress, true)
```
Ziinc, line 44: when would we want to **not** compress? we can remove if it is a redundant option, unless its for testing/repl only? though the option isn't being used at all in tests and we could remove the `compression: false` tests since.

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -0,0 +1,180 @@
...
+    async = Keyword.get(opts, :async, true)
```
Ziinc, line 45: likewise, we could remove it unless its necessary for repl, and even if not, i think it would not have much of an impact to remove.

test/logflare/backends/adaptor/clickhouse_adaptor/ingester_test.exs
```diff
@@ -0,0 +1,293 @@
+defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.IngesterTest do
+  use Logflare.DataCase, async: false
+
+  alias Logflare.Backends.Adaptor.ClickhouseAdaptor
+  alias Logflare.Backends.Adaptor.ClickhouseAdaptor.Ingester
+
+  @sleep_time_after_insert 100
```
Ziinc, line 7: TestUtils.retry_assert would counteract potential ci flakiness.

lib/logflare/application.ex
```diff
@@ -183,6 +183,16 @@ defmodule Logflare.Application do
...
          ]
        }},
+      {Finch,
+       name: Logflare.FinchClickhouseIngest,
+       pools: %{
+         default: [
+           protocols: [:http1],
+           size: max(base * 125, 150),
+           count: http1_count,
+           start_pool_metrics?: true
```
ruslandoga, line 193: 👋 

It might also help to close the idle connections before ClickHouse does it with `conn_max_idle_time: to_timeout(seconds: 9)` to avoid (probably quite rare) `%Mint.TransportError{reason: :closed}` "races".


<details><summary>Details</summary>

ClickHouse closes idle HTTP connections after [`keep_alive_timeout`](https://clickhouse.com/docs/operations/server-configuration-parameters/settings#keep_alive_timeout) seconds passes. It's set to 10s in DockerHub images and ClickHouse cloud by default. If the request is sent at around that time, Finch would return `:closed` error.

Example:

```elixir
Mix.install [:finch]

Finch.start_link(pools: %{default: [size: 1]}, name: Ch1)

defmodule Tel do
  require Logger

  def handle_event(event, measurements, _meta, _config) do
    Logger.debug("""
    EVENT: #{inspect(event)}
    MEASUREMENTS: #{inspect(measurements)}
    """)
  end
end

:telemetry.attach_many("ch", [[:finch, :connect, :start], [:finch, :connect, :stop], [:finch, :reused_connection]], &Tel.handle_event/4, [])

q = fn sql ->
  url = "https://b98p3r3rsd.eu-central-1.aws.clickhouse.cloud:8443"
  # I'll delete this user after a few days :)
  headers = [{"x-clickhouse-user", "default"}, {"x-clickhouse-key", "FwZABB5Ayf_fl"}]
  req = Finch.build(:post, url, headers, sql)
  Finch.request(req, Ch1)
end

[q.("select 1"), Process.sleep(10000), q.("select 2")]
```

Output:

```
15:40:16.649 [debug] EVENT: [:finch, :connect, :start]
MEASUREMENTS: %{system_time: 1763642416648201201}


15:40:16.968 [debug] EVENT: [:finch, :connect, :stop]
MEASUREMENTS: %{duration: 320611218}


15:40:27.034 [debug] EVENT: [:finch, :reused_connection]
MEASUREMENTS: %{}
```

Result:

```
[
  {:ok,
   %Finch.Response{
     status: 200,
     body: "1\n",
     headers: [
       {"x-clickhouse-summary",
        "{\"read_rows\":\"1\",\"read_bytes\":\"1\",\"written_rows\":\"0\",\"written_bytes\":\"0\",\"total_rows_to_read\":\"1\",\"result_rows\":\"0\",\"result_bytes\":\"0\",\"elapsed_ns\":\"848242\"}"},
       {"date", "Thu, 20 Nov 2025 12:40:17 GMT"},
       {"connection", "Keep-Alive"},
       {"content-type", "text/tab-separated-values; charset=UTF-8"},
       {"access-control-expose-headers",
        "X-ClickHouse-Query-Id,X-ClickHouse-Summary,X-ClickHouse-Server-Display-Name,X-ClickHouse-Format,X-ClickHouse-Timezone,X-ClickHouse-Exception-Code"},
       {"x-clickhouse-server-display-name", "clickhouse-cloud"},
       {"transfer-encoding", "chunked"},
       {"x-clickhouse-query-id", "295f8231-d82b-4316-b6e0-8103eae1259e"},
       {"x-clickhouse-format", "TabSeparated"},
       {"x-clickhouse-timezone", "UTC"},
       {"keep-alive", "timeout=10, max=9999"}
     ],
     trailers: []
   }},
  :ok,
  {:error, %Mint.TransportError{reason: :closed}}
]
```

</details>

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -0,0 +1,180 @@
...
+  def encode_batch([%LogEvent{} | _] = rows) do
+    Enum.map(rows, &encode_row/1)
+  end
+
+  @doc false
+  @spec encode_as_uuid(Ecto.UUID.t() | String.t()) :: binary()
+  def encode_as_uuid(uuid_string) when is_non_empty_binary(uuid_string) do
+    uuid_string
+    |> String.replace("-", "")
+    |> Base.decode16!(case: :mixed)
```
ruslandoga, line 98: Context: https://github.com/Logflare/logflare/pull/2964

`encode_as_uuid` might also need to:
1. verify the size of the binary to be 16 bytes -- otherwise it can compromise the rest of the payload
2. encode the uuid as two little endian 64-bit uints

Maybe something like this could work:

```elixir
uuid_raw = 
  uuid_string
  |> String.replace("-", "")
  |> Base.decode16!(case: :mixed)

case uuid_raw do
  <<u1::64, u2::64>> -> <<u1::64-little, u2::64-little>>
  _other -> raise "invalid uuid string: #{inspect(uuid_string)}"
end
```

amokan: Thanks again and perfect timing on this find @ruslandoga

I have a PR up with the changes - https://github.com/Logflare/logflare/pull/2965

lib/logflare/application.ex
amokan: Thank you yet again @ruslandoga

Connection issues have been driving me crazy on CH and I bet this was the issue.

PR up for that here - https://github.com/Logflare/logflare/pull/2966

PR #2895, opened 2025-11-07
perf: ch pipeline tweaks

Ziinc:
minor performance tweaks for clickhouse pipeline:
- add in more frequent gc-ing
- add in process hibernation
- increase batcher concurrency - mimicking BQ pipeline's ratio 
- increase max batch size to 1.5k events
- increase batch timeout to 1.5s
- set processor max demand to 100 events.


lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -36,16 +36,20 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Pipeline do
 
     Broadway.start_link(__MODULE__,
       name: name,
+        hibernate_after: 5_000,
+        spawn_opt: [
```
josevalim: Sorry for the nit:

```suggestion
      hibernate_after: 5_000,
      spawn_opt: [
```

josevalim (commented):
I assume the reason why you are toggling `fullswap_after` is because Broadway (`gen_stage`) reuses the same processes for the duration of the pipeline. So it will keep on accumulating garbage. Perhaps it would even be worth having a GenStage Dispatcher where processes are spawned and then immediately terminated, so we don't worry about GC, but I guess that would be separate conversation.

In any case, this looks good to me. Just bear in mind the default is 65k, so this may be too low, but running actual workloads will give the right answer. :) Overall, LGTM!

amokan (approved):

PR #2896, opened 2025-11-07
perf: fully truncate non-default backend queues

Ziinc:
truncate non-default backend queues with the queue janitor.
reduces number of events remaining in the queue on idle/slow ingest, improving memory footprint

### DIFF
--- a/lib/logflare/sources/source/bigquery/pipeline.ex
+++ b/lib/logflare/sources/source/bigquery/pipeline.ex
@@ -109,8 +109,10 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
   def ack({queue, source_token}, successful, _failed) do
     # TODO: re-queue failed
     metrics = Sources.get_source_metrics_for_ingest(source_token)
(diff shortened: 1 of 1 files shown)


lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -109,8 +109,10 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
   def ack({queue, source_token}, successful, _failed) do
     # TODO: re-queue failed
     metrics = Sources.get_source_metrics_for_ingest(source_token)
+    {_sid, bid, _tid} = queue
 
-    if metrics.avg > 100 do
+    # delete immediately if not default backend or if avg rate is above 100
+    if metrics.avg > 100 or bid != nil do
```
amokan, line 115: We should probably make a `is_not_nil` guard in `Logflare.Utils.Guards` because I know we have a handful of these floating around now.

amokan (approved):

PR #2913, opened 2025-11-12
feat: ingestion metrics for system monitoring

Ziinc:
System monitoring for users:
- provisions system-managed sources for each User
- Each system source is specific to a type of event
- otel-format metrics are collected and then exported into a `:metrics` source pre-aggregated.
- a centralized OtelMetricExporter is used to avoid running `:keep` callback multiple times for each each SourceSup
- export_callback/2 overrides default library behaviour
- we rely on string metadata keys as a declarative way of deciding what metadata keys to keep and set as otel metric attribute.

for user-specific logs, these are routed to the source based on whether specific identifying metadata keys are present. If the log is pertaining to a specific source/backend belonging to the user, then it will get ingested into that source.

4 specific user metrics are implemented
- egress on http adaptors
- total ingested events
- total ingested bytes
- total bytes processed for queries

Refactors tests to avoid testing implementation.

- [x] docs
- [x] test
- [x] log drains egress metric
- [x] endpoints query bytes metric

Closes ANL-1202

### DIFF
--- a/test/logflare/alerting_test.exs
+++ b/test/logflare/alerting_test.exs
@@ -208,7 +208,7 @@ defmodule Logflare.AlertingTest do
         {:ok, TestUtils.gen_bq_response([%{"testing" => "123"}])}
       end)
 
-      assert {:ok, %{rows: [%{"testing" => "123"}], total_bytes_processed: 1}} =
+      assert {:ok, %{rows: [%{"testing" => "123"}], total_bytes_processed: _}} =
                Alerting.execute_alert_query(alert_query)
 
       #  no reservation set by user
@@ -227,7 +227,7 @@ defmodule Logflare.AlertingTest do
         {:ok, TestUtils.gen_bq_response([%{"testing" => "123"}])}
       end)
 
-      assert {:ok, %{rows: [%{"testing" => "123"}], total_bytes_processed: 1}} =
+      assert {:ok, %{rows: [%{"testing" => "123"}], total_bytes_processed: _}} =
                Alerting.execute_alert_query(alert_query)
 
       assert_receive {:reservation, reservation}
--- a/test/logflare/bigquery/pipeline_test.exs
+++ b/test/logflare/bigquery/pipeline_test.exs
@@ -13,6 +13,7 @@ defmodule Logflare.BigQuery.PipelineTest do
   @pipeline_name :test_pipeline
   describe "pipeline" do
     setup do
+      insert(:plan)
       user = insert(:user)
       source = insert(:source, user_id: user.id)
       {:ok, source: source}


lib/logflare/backends/user_monitoring.ex
```diff
@@ -26,49 +32,54 @@ defmodule Logflare.Backends.UserMonitoring do
...
-    fn metadata ->
-      case Users.get_related_user_id(metadata) do
-        nil -> false
-        user_id -> user_id == source.user_id && user_monitoring?(user_id)
-      end
-    end
-  end
+  def keep_metric_function(%Source{id: source_id} = source) do
+    fn
+      %{"source_id" => ^source_id} = metadata ->
```
andremaker: this won't go well with my implementation of `Users.get_related_user_id`, as it uses atoms. I believe metric metadata always use atom keys, is not it?

lib/logflare/backends/user_monitoring.ex
```diff
@@ -26,49 +32,54 @@ defmodule Logflare.Backends.UserMonitoring do
...
+      _ ->
+        false
     end
   end
 
   defp user_monitoring?(user_id),
     do: Users.Cache.get(user_id).system_monitoring
 
   defp generate_exporter_callback(source) do
+    metrics_source = get_system_source_metrics(source.user_id)
```
andremaker: i believe this is not needed, as this function is only called for the metrics sources. So the source sent as arg to this function is always the one we need

lib/logflare/backends/user_monitoring.ex
andremaker: this will make the metrics source only monitor metrics related to itself, which i think defeats the purpose

docs/docs.logflare.com/docs/concepts/system-monitoring.md
```diff
@@ -0,0 +1,82 @@
...
+sidebar_position: 6
+---
+
+# System Monitoring
+
+System Monitoring collects metrics and logs about your sources, backends, and endpoints. Three dedicated system sources capture operational data:
+
+- **`system.metrics`** - OpenTelemetry metrics for ingestion, queries, and egress
+- **`system.logs`** - Application logs for your sources and backends
+- **`system.rejected_events`** - Rejected events from validation failures
```
chasers: I'm thinking we should just have two; logs and metrics. Rejected events is a log with rejected logs in it?

chasers (dismissed):

docs/docs.logflare.com/docs/concepts/system-monitoring.md
Ziinc: yes, haven't wired up the storage of the events yet, but would essentially be storing the entire event body inside as stringified json.

Ziinc: removed it from docs since it isn't working fully yet.

lib/logflare/sources.ex
```diff
@@ -587,6 +587,37 @@ defmodule Logflare.Sources do
...
+    |> String.split(",")
+    |> Enum.map(fn label ->
+      String.split(label, "=")
+      |> then(fn
+        [label, "m." <> path] -> {label, ["metadata" | String.split(path, ".")]}
+        [label, path] -> {label, String.split(path, ".")}
+        [label] -> {label, [label]}
+      end)
+    end)
+    |> Map.new()
```
josevalim, line 627: For completeness, you could use custom Ecto.Types to store those aleady as lists in the database: https://hexdocs.pm/ecto/Ecto.Type.html - although I am not expecting this code to be performance sensitive enough to matter.

lib/logflare_web/templates/source/edit.html.eex
```diff
@@ -309,6 +309,22 @@
...
+
+  <%= if @user.system_monitoring do %>
+    <%= section_header("Monitoring Labels") %>
+    <p>Set labels for this source, to be used for monitoring. Labels are used to group metrics together when system monitoring is enabled. Values of the labels will be determined at ingest time.</p>
+    <%= form_for @changeset, Routes.source_path(@conn, :update, @source),fn f -> %>
+      <div class="form-group">
+        <%= text_input f, :labels , class: "form-control form-control-margin" %>
+        <%= error_tag f, :labels %>
+      <small class="form-text text-muted">
+        Comma-separated. For example: <code>status=m.level,my_project=project.name</code>.
```
josevalim, line 324: It may be worth doing either some validation or normalization? For example, to make sure there are no spaces betwen commas? Or that there is a single `=`? See also the comment above about already storing the parsed data.

lib/logflare/sources.ex
```diff
@@ -587,6 +587,37 @@ defmodule Logflare.Sources do
...
+  end
+
+  def get_labels_mapping(%{labels: nil}), do: %{}
+  def get_labels_mapping(%{labels: ""}), do: %{}
+
+  def get_labels_mapping(source) do
+    (source.labels || "")
+    |> String.split(",")
+    |> Enum.map(fn label ->
+      String.split(label, "=")
```
josevalim: This code will currently crash if the user writes more than a single `=`:

```suggestion
      String.split(label, "=", parts: 2)
```

lib/logflare/backends/user_monitoring.ex
```diff
@@ -25,60 +30,78 @@ defmodule Logflare.Backends.UserMonitoring do
...
-      :ok
-    end
+  defp exporter_callback({:metrics, metrics}, config) do
+    metrics
+    |> OtelMetricExporter.Protocol.build_metric_service_request(config.resource)
+    |> Protobuf.encode()
+    |> Protobuf.decode(ExportMetricsServiceRequest)
+    |> Map.get(:resource_metrics)
+    |> Logs.OtelMetric.handle_batch(%{})
+    |> List.flatten()
```
josevalim: `handle_batch` already seems to return a flattened list:

```suggestion
```

Perhaps handle_batch could also use `Enum.flat_map`, but that's a separate conversation. :)

lib/logflare/users/users.ex
```diff
@@ -186,10 +186,15 @@ defmodule Logflare.Users do
   def get_related_user_id(map) do
     case map do
```
josevalim, line 187: This is a separate conversation but I wonder if something could be done to remove the number of "sources" here.

Also, it is worth noting that you are caching whole structs, but you may only need the `user_id` field. For example, `Cache.get_endpoint_query` seems to be used exclusively here and you don 't care about the other fields, so there is no need to store or read them. You may benefit from adding a `get_user_id_by_endpoint_query`. This may also apply to other operations, but please double check!

Ziinc: yes, i intend to refactor that off as it isn't as performant, and we should be able to derive the user_id from wherever the telemetry is getting emitted and set it on the metadata, thereby bypassing the whole need to fetch from Cachex

lib/logflare_web/templates/source/edit.html.eex
Ziinc: added

test/logflare/backends/user_monitoring_test.exs
```diff
@@ -66,189 +86,251 @@ defmodule Logflare.Backends.UserMonitoringTest do
...
 
-      refute match?(
-               %{
-                 {:sum, "logflare.backends.ingest.event_count"} => %{
-                   %{backend_id: ^backend_2_id} => _
-                 }
-               },
-               user_1_exporter_metrics
-             )
+    test "endpoints.query.total_processed_bytes" do
```
amokan:
```suggestion
    test "endpoints.query.total_bytes_processed" do
```

Minor nit - actual metric name is `total_bytes_processed`

lib/logflare/backends/user_monitoring.ex
```diff
@@ -25,60 +30,77 @@ defmodule Logflare.Backends.UserMonitoring do
...
+    |> Protobuf.encode()
+    |> Protobuf.decode(ExportMetricsServiceRequest)
+    |> Map.get(:resource_metrics)
+    |> Logs.OtelMetric.handle_batch(%{})
+    |> Enum.group_by(fn event ->
+      event
+      |> Map.get("attributes")
+      |> Users.get_related_user_id()
+    end)
+    |> Enum.each(fn {user_id, user_events} ->
```
amokan, line 94: Is it possible that a user can delete the system source?

If so, maybe something like this could be safer 🤷 

```
|> Enum.each(fn {user_id, user_events} ->
  with %{} = source <- Sources.Cache.get_by(user_id: user_id, system_source_type: :metrics) do
    source
    |> Sources.Cache.preload_rules()
    |> Sources.refresh_source_metrics()
    |> then(&Processor.ingest(user_events, Logs.Raw, &1))
  end
end)
```

docs/docs.logflare.com/docs/concepts/system-monitoring.md
```diff
@@ -0,0 +1,77 @@
+---
+sidebar_position: 6
+---
+
+# System Monitoring
+
+System Monitoring collects metrics and logs about your sources, backends, and endpoints. Three dedicated system sources capture operational data:
```
amokan:
```suggestion
System Monitoring collects metrics and logs about your sources, backends, and endpoints. Two dedicated system sources capture operational data:
```

lib/logflare/logs/otel_metric.ex
```diff
@@ -52,7 +52,7 @@ defmodule Logflare.Logs.OtelMetric do
     handle_metric_data(metric.data, base)
   end
 
-  defp handle_metric_data({:gauge, %{data_points: data_points}}, base) do
+  def handle_metric_data({:gauge, %{data_points: data_points}}, base) do
```
amokan: Were the changes from `defp` to `def` intentional or just for local testing?

Ziinc: ah it was used in the previous iteration. reverted.

PR #2919, opened 2025-11-13
perf: refactor LogEvent.make/2 

Ziinc:
- refactored LogEvent
-  removed Logflare.Logs context (from v1)
- removed DeepEqFields validator (intra list validation is expensive, overlaps with schema validation, and backend should handle this)
- removed unused LogEvent struct fields for better memory management
  - removed `:source` struct storage and instead uses `:source_id` to avoid storing Source term in :ets
- pre-compiles regex (theoretically faster)
- removes redundant list iterations in cleaner
- remove deprecated `log_entry` field that was previously used for setting event_message

Overall goals of this refactor:
- reduce `:ets` storage by removing Source and `:params` storage
- improve ips from reduced mapper/struct operations (~5% ips improvement)
- reduce map iterations (eliminate Validator, additional ~9% ips improvement)

### With struct fields (start of refactor)
```
Operating System: macOS
CPU Information: Apple M4 Pro
Number of Available Cores: 14
Available memory: 48 GB
Elixir 1.17.3
Erlang 27.3.4
JIT enabled: true

Benchmark suite executing with the following configuration:
warmup: 1 s
time: 4 s
memory time: 3 s
reduction time: 3 s
parallel: 1
inputs: none specified
Estimated total run time: 1 min 6 s

Name                                                             ips        average  deviation         median         99th %
100 fields with 5 level nesting (no_cast_embeds: true)         68.02       14.70 ms     ±5.12%       14.72 ms       18.42 ms
100 fields with 5 level nesting (no_sys_uint: true)            67.94       14.72 ms     ±5.66%       14.69 ms       19.80 ms
100 fields with 5 level nesting (no_source: true)              67.80       14.75 ms     ±5.61%       14.75 ms       19.64 ms
100 fields with 5 level nesting (all opts)                     67.75       14.76 ms     ±5.39%       14.75 ms       19.33 ms
100 fields with 5 level nesting (no_params: true)              65.46       15.28 ms    ±13.95%       14.78 ms       27.07 ms
100 fields with 5 level nesting                                64.84       15.42 ms    ±14.19%       14.85 ms       24.93 ms

Comparison: 
100 fields with 5 level nesting (no_cast_embeds: true)         68.02
100 fields with 5 level nesting (no_sys_uint: true)            67.94 - 1.00x slower +0.0182 ms
100 fields with 5 level nesting (no_source: true)              67.80 - 1.00x slower +0.0492 ms
100 fields with 5 level nesting (all opts)                     67.75 - 1.00x slower +0.0585 ms
100 fields with 5 level nesting (no_params: true)              65.46 - 1.04x slower +0.58 ms
100 fields with 5 level nesting                                64.84 - 1.05x slower +0.72 ms
```

### With Validator removed (with above struct fields refactoring)
```

Name                                      ips        average  deviation         median         99th %
100 fields with 5 level nesting         74.35       13.45 ms     ±9.31%       12.94 ms       18.86 ms

Memory usage statistics:

Name                                    average  deviation         median         99th %
100 fields with 5 level nesting        32.75 MB     ±0.00%       32.75 MB       32.75 MB

Reduction count statistics:

Name                                    average  deviation         median         99th %
100 fields with 5 level nesting          3.67 M     ±0.01%         3.67 M         3.67 M
```

### DIFF
--- a/lib/logflare/logs/log_event.ex
+++ b/lib/logflare/logs/log_event.ex
@@ -307,7 +307,9 @@ defmodule Logflare.LogEvent do
     determine_timestamp(%{"timestamp" => round(x)})
   end
 
-  defp default_timestamp do
+  defp determine_timestamp(_), do: default_timestamp()
+
+  defp default_timestamp() do
     System.system_time(:microsecond)
   end
 end
--- a/lib/logflare/logs/cleaner.ex
+++ b/lib/logflare/logs/cleaner.ex
@@ -12,7 +12,8 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
         acc
 
       x, acc when is_map(x) or is_list(x) ->
-        [deep_reject_nil_and_empty(x) | acc]
+        cleaned = deep_reject_nil_and_empty(x)
+        if nil_or_empty?(cleaned), do: acc, else: [cleaned | acc]
 
       x, acc ->
         [x | acc]
--- a/test/logflare_web/live/search_live/logs_search_lv_test.exs
+++ b/test/logflare_web/live/search_live/logs_search_lv_test.exs
@@ -403,7 +403,7 @@ defmodule LogflareWeb.Source.SearchLVTest do
       :timer.sleep(100)
 
       # Backends.via_source(source, Schema, nil)
-      Schema.handle_cast({:update, le}, %{
+      Schema.handle_cast({:update, le, source}, %{
         source_id: source.id,
         source_token: source.token,
         bigquery_project_id: nil,


lib/logflare/logs/log_event.ex
```diff
@@ -316,4 +272,36 @@ defmodule Logflare.LogEvent do
...
+  defp determine_timestamp(%{"timestamp" => x}) when is_integer(x) do
+    case Integer.digits(x) |> Enum.count() do
+      19 -> Kernel.round(x / 1_000)
+      16 -> x
+      13 -> x * 1_000
+      10 -> x * 1_000_000
+      7 -> x * 1_000_000_000
+      _ -> x
+    end
+  end
```
v0idpwn, line 304: Unsure how much this is called but its a bit inefficient.

I did some tests and the fastest I came up with is:
```
  def optimized_v2(0), do: 0
  def optimized_v2(x) when is_integer(x) do
    digit_count = trunc(:math.log10(x)) + 1

    case digit_count do
      19 -> Kernel.round(x / 1_000)
      16 -> x
      13 -> x * 1_000
      10 -> x * 1_000_000
      7 -> x * 1_000_000_000
      _ -> x
    end
  end
```

Which should be ~3 times faster than this one and hog **way less** memory

Another alternative in case we are afraid of some floating point imprecision hurting us (which seems unlikely, but :X):
```
  def optimized_v1(x) when is_integer(x) do
    cond do
      x >= 1_000_000_000_000_000_000 and x < 10_000_000_000_000_000_000 ->
        Kernel.round(x / 1_000)
      x >= 1_000_000_000_000_000 and x < 10_000_000_000_000_000 ->
        x
      x >= 1_000_000_000_000 and x < 10_000_000_000_000 ->
        x * 1_000
      x >= 1_000_000_000 and x < 10_000_000_000 ->
        x * 1_000_000
      x >= 1_000_000 and x < 10_000_000 ->
        x * 1_000_000_000
      true -> x
    end
  end
```

Still uses way less memory, slightly slower than the one above, and uglier.

amokan: Couldn't resist this one 🤓 

If floating point is a concern, it seems like you could replace `Kernel.round(x / 1_000)` on those 19-digit values with `div(x, 1_000)`.

image


And a spin off of Felipe's last example (_without the lt checks_) - and doesn't seem to read too bad 🤷

```elixir
defp determine_timestamp(%{"timestamp" => 0}), do: 0
defp determine_timestamp(%{"timestamp" => x}) when is_integer(x) do
  cond do
    x >= 1_000_000_000_000_000_000 -> div(x, 1_000)
    x >= 1_000_000_000_000_000 -> x
    x >= 1_000_000_000_000 -> x * 1_000
    x >= 1_000_000_000 -> x * 1_000_000
    x >= 1_000_000 -> x * 1_000_000_000
    true -> x
  end
end
```

amokan: Ignoring the floating point (_which is probably fine_), it seems like the first answer from @v0idpwn flies on my laptop though. The swap of `Kernel.round/1` to `div/2` does add about ~8.5ns

v0idpwn: 👀 I've resisted removing the `lt` checks to maintain the exact same behaviour, but I agree it looks nicer (and should be a bit faster)

Ziinc: will do a follow up

PR #2921, opened 2025-11-13
fix: stop connection pools related to ClickHouse and BigQuery when in Single Tenant Mode in Postgres Mode

nelsonmestevao:
Closes ANL-1194.

### DIFF
--- a/lib/logflare/networking.ex
+++ b/lib/logflare/networking.ex
@@ -18,6 +18,7 @@ defmodule Logflare.Networking do
            :default => [protocols: [:http1]]
          }
          |> Map.merge(datadog_connection_pools())}
+      | base_finch_pools()
     ]
   end
 
@@ -45,16 +46,6 @@ defmodule Logflare.Networking do
                max_frame_size: 8_000_000
              ]
            ]
-         ]
-       }},
-      {Finch,
-       name: Logflare.FinchClickhouseIngest,
-       pools: %{
-         default: [
-           protocols: [:http1],
-           size: max(base * 125, 150),
-           count: http1_count,
-           start_pool_metrics?: true
          ]
        }},
       {Finch,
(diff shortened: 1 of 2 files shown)


lib/logflare/networking.ex
```diff
@@ -0,0 +1,129 @@
...
+           :default => [protocols: [:http1]],
+           #  explicitly set http2 for other pools for multiplexing
+           "https://bigquery.googleapis.com" => [
+             protocols: [:http1],
+             size: 115,
+             count: http1_count,
+             start_pool_metrics?: true
+           ]
+         }
+         |> Map.merge(datadog_connection_pools())}
```
nelsonmestevao, line 85: @Ziinc datadog connection pools should always happen right? So I should always return this process from line 69, right?

nelsonmestevao: Did it in af53f167e18fa5a4832bd9554692469cb84c4cae. I am assuming things correctly right?

Ziinc: Yes the conditional addition of Datadog pools should be done, seems like it is there correctly.

lib/logflare/networking.ex
```diff
@@ -0,0 +1,140 @@
...
+             # a larger default window size ensures that the number of packages exchanges is smaller, thus speeding up the requests
+             # by reducing the amount of networks round trip, with the cost of having larger packages reaching the server per connection.
+             client_settings: [
+               initial_window_size: 8_000_000,
+               max_frame_size: 8_000_000
+             ]
+           ]
+         ]
+       }},
+      {Finch,
```
Ziinc: Let's shift this to be always included even in single tenant mode.
It uses default so it won't have the same errors as the github issue.

nelsonmestevao: Done in last push 🙏

PR #2922, opened 2025-11-14
perf: memory tweaks (acceptors, ErlSysMon, bandit, Broadway, Schema)

Ziinc:
- reduces number of acceptors to 1.25k (a little higher than initial of 1k)
- reduce fullsweep_after to 10 for all broadway pipelines
- add in bandit option for gc-ing between each keepalive requests, and drops it to 30. this means max time between GCing is 30 mins, since each keepalive max is 10mins.
- refactor ErlSysMon to use a Task for better memory management
- add fullsweep_after to Schema module

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -37,7 +37,7 @@ config :logflare, LogflareWeb.Endpoint,
   adapter: Bandit.PhoenixAdapter,
   http: [
     http_options: [log_protocol_errors: :short, log_client_closures: false],
-    http1_options: [gc_every_n_keepalive_requests: 3],
+    http_1_options: [gc_every_n_keepalive_requests: 3],
     thousand_island_options: [
       num_acceptors: 1250,
       # default backend keepalive timeout is fixed at 600 seconds


lib/logflare/erl_sys_mon.ex
```diff
@@ -48,20 +49,22 @@ defmodule Logflare.ErlSysMon do
...
-          Keyword.take(dict, [:"$ancestors", :"$initial_call"])
-
-        other ->
-          other
-      end
-
-    Logger.warning(
-      "#{__MODULE__} message: " <> inspect(msg) <> "|\n process info: #{inspect(pid_info)}"
-    )
+    Tasks.async(fn ->
```
chasers: Doesn't seem right to me to create a task just to log something

And if we're really getting this many messages then really not sure about creating tasks for each one

chasers (dismissed):

lib/logflare/erl_sys_mon.ex
ruslandoga: Sorry for unsolicited feedback, was just passing through :)

`Task.async` would send a message back to this process, maybe [start](https://hexdocs.pm/elixir/Task.html#start/1) is a better fit.

Ziinc: Removed the task, and just doing :hibernate and :fullsweep_after.
also adjusted the string interpolation in case it is a binary memory issue.

PR #2924, opened 2025-11-14
feat: LQL support for postgres

amokan:
Adds LQL support for Postgres - now completing LQL for all current SQL dialects

Closes ANL-1110


lib/logflare/lql/backend_transformer/postgres.ex
```diff
@@ -0,0 +1,674 @@
...
+  import Logflare.Utils.Guards
+
+  alias Ecto.Query
+  alias Logflare.Lql.Rules.ChartRule
+
+  @special_top_level ~w(event_message timestamp id)
+
+  # PostgreSQL-specific macros for time truncation
+  defmacrop pg_trunc_second(ts_field) do
+    quote do: fragment("date_trunc('second', ?)", unquote(ts_field))
```
Ziinc, line 22: perhaps we can interpolate the interval value into the fragment? then we might be able to remove a lot of function heads and condense this file quite a bit.
Like `pg_trunc(:second, field)`

lib/logflare/lql/backend_transformer/postgres.ex
```diff
@@ -0,0 +1,674 @@
...
+    query
+    |> select([t], %{
+      timestamp: pg_trunc_second(field(t, ^timestamp_field)),
+      count: count(field(t, ^timestamp_field))
+    })
+    |> group_by([t], pg_trunc_second(field(t, ^timestamp_field)))
+    |> order_by([t], pg_trunc_second(field(t, ^timestamp_field)))
+  end
+
+  def transform_chart_rule(query, :count, _field_path, :minute, timestamp_field) do
```
Ziinc, line 141: given that the operations for chart rules are quite similar across all agg functions, we can likely condense all of them into one function head

Ziinc:
merging in first.

PR #2925, opened 2025-11-14
feat: Adds clickhouse `wait_for_async_insert` config option

amokan:
This adds a new clickhouse adaptor config option for `wait_for_async_insert` that defaults to true.

We will remove this from the UI once we confirm the pipelines are behaving better with Clickhouse inserts flushing consistently.

### DIFF
--- a/test/logflare/backends/adaptor/clickhouse_adaptor/ingester_test.exs
+++ b/test/logflare/backends/adaptor/clickhouse_adaptor/ingester_test.exs
@@ -1,10 +1,12 @@
 defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.IngesterTest do
   use Logflare.DataCase, async: false
 
+  import Mimic
+
   alias Logflare.Backends.Adaptor.ClickhouseAdaptor
   alias Logflare.Backends.Adaptor.ClickhouseAdaptor.Ingester
 
-  @sleep_time_after_insert 100
+  setup :verify_on_exit!
 
   describe "encode_as_varint/1" do
     test "encodes zero" do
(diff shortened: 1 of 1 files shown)


test/logflare/backends/adaptor/clickhouse_adaptor/ingester_test.exs
```diff
@@ -289,5 +289,74 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.IngesterTest do
 
       assert [%{"count" => 1}] = result
     end
+
+    test "respects `wait_for_async_insert: true` config", %{
```
Ziinc: We can condense the tests into 1 since essentially they are just validating that insert works, and not the actual setting behaviour

test/logflare/backends/adaptor/clickhouse_adaptor_test.exs
```diff
@@ -108,6 +177,30 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptorTest do
       [source: source, backend: backend]
     end
 
+    test "respects `wait_for_async_insert` config setting", %{source: source, backend: backend} do
```
Ziinc: Can remove this, not quite sure what this is verifying since the above test does the same.

test/logflare/backends/adaptor/clickhouse_adaptor_test.exs
```diff
@@ -95,6 +95,75 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptorTest do
...
+        |> ClickhouseAdaptor.cast_config()
+        |> ClickhouseAdaptor.validate_config()
+
+      refute changeset.valid?
+      assert changeset.errors[:url]
+      assert changeset.errors[:database]
+      assert changeset.errors[:port]
+    end
+
+    test "validates URL format" do
```
Ziinc: Seems outside of scope

test/logflare/backends/adaptor/clickhouse_adaptor/ingester_test.exs
amokan: I've improved these to now check the actual settings in the finch requests

PR #2931, opened 2025-11-18
Extract common code for HttpBased adaptors

bblaszkow06:
The idea of this refactor is to get rid of adaptor code reuse by calling `WebhookAdaptor` and transforming the config for it, but instead have another layer of abstraction (lower than `Adaptor`).

The PR adds:
1. `HttpBased.Pipeline` - a pipeline extracted from `WebhookAdaptor`, but using backend/source ids and cache, and expecting a client module
2. `HttpBased.Client` - a behaviour with helpers for defining Tesla-based clients
3. `HttpBased.EgressTracer` - middleware to replace `WebhookAdaptor.EgressMiddleware` (takes metadata from request opts instead of client definition)
4. `HttpBased.LogEventTransformer` - middleware transforming LogEvents into a JSON-serializable list of maps. Can be used with an additional mapper function or replaced entirely in Tesla's client pipeline.

If the PR is accepted, it will require a follow-up PR, which:
- Migrates `WebhookAdaptor` to the above
- Migrates all other HTTP-based adaptors
- removes `transform_config` callback, possibly along with `format_batch`


lib/logflare/backends/adaptor/http_based/client.ex
```diff
@@ -0,0 +1,78 @@
...
+    Tesla.client(
+      [
+        Tesla.Middleware.Telemetry,
+        opts[:url] && {Tesla.Middleware.BaseUrl, opts[:url]},
+        opts[:query] && {Tesla.Middleware.Query, opts[:query]},
+        opts[:token] && {Tesla.Middleware.BearerAuth, token: opts[:token]},
+        opts[:basic_auth] && {Tesla.Middleware.BasicAuth, opts[:basic_auth]},
+        opts[:headers] && {Tesla.Middleware.Headers, opts[:headers]},
+        Access.get(opts, :formatter, LogEventTransformer),
+        Tesla.Middleware.JSON,
```
Ziinc: this would be changed with http protobuf right?

lib/logflare/backends/adaptor/http_based/egress_tracer.ex
```diff
@@ -0,0 +1,47 @@
+defmodule Logflare.Backends.Adaptor.HttpBased.EgressTracer do
```
Ziinc, line 1: instead of copying the file, lets just move the file so that we don't have duplicates, as there is concurrent work gong on the egress telemetry.

lib/logflare/backends/adaptor/http_based/client.ex
```diff
@@ -0,0 +1,78 @@
...
+  Sends log events via the client. The metadata is meant for `Logflare.Backends.Adaptor.HttpBased.EgressTracker`,
+  should be passed as `opts: [metadata: metadata]` when making a request via Tesla.
+  """
+  @callback send_logs(config :: map(), log_events :: [Logflare.LogEvent.t()], metadata :: map()) ::
+              :ok
+
+  @doc """
+  Tests the connectivity via adapter. Should make a request resembling one in `c:send/3` as close as possible.
+  """
+  @callback test_connection(config :: map()) :: :ok | {:error, term()}
```
Ziinc: this duplicates the Adaptor callback, we can remove the client callback

lib/logflare/backends/adaptor/http_based/log_event_transformer.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Logflare.Backends.Adaptor.HttpBased.LogEventTransformer do
+  @moduledoc """
+  Middleware extracting body from `Logflare.LogEvent`s.
+
+  Assumes that following middlewares (e.g. `Tesla.Middleware.JSON`) can handle conversion of maps to `t:iodata/0`.
+
+  ## Options
+
+  - `:map` - a mapping function, transforming the LogEvent's body
```
Ziinc: map is not a good self-documenting name for this argument. something like `transform_func` would be better.

also, would not refer to it as an option in the moduledoc, would be better to refer to it an "optional argument"

lib/logflare/backends/adaptor/webhook_adaptor/egress_middleware.ex
Ziinc: as above, would rather that this module calls out to a function in the new egress tracer module.

lib/logflare/backends/adaptor/http_based/client.ex
Ziinc: also not too clear on why it should implemented client level, since the adaptor can use the client for a generic request as well

lib/logflare/backends/adaptor/http_based/client.ex
```diff
@@ -0,0 +1,78 @@
...
+      true ->
+        {Tesla.Adapter.Finch, name: Logflare.FinchDefaultHttp1, receive_timeout: 5_000}
+    end
+  end
+
+  @doc """
+  Sends log events via the client. The metadata is meant for `Logflare.Backends.Adaptor.HttpBased.EgressTracker`,
+  should be passed as `opts: [metadata: metadata]` when making a request via Tesla.
+  """
+  @callback send_logs(config :: map(), log_events :: [Logflare.LogEvent.t()], metadata :: map()) ::
```
Ziinc:
```suggestion
  @callback send(config :: map(), log_events :: [Logflare.LogEvent.t()], metadata :: map()) ::
```
generic `send_request/3` is better as we want to refer to generic events as much as possible and not refer to logs only

lib/logflare/backends/adaptor/axiom_adaptor.ex
```diff
@@ -105,34 +68,68 @@ defmodule Logflare.Backends.Adaptor.AxiomAdaptor do
...
-          Tesla.Middleware.JSON,
-          {Tesla.Middleware.CompressRequest, format: "gzip"},
-          Tesla.Middleware.Telemetry
-        ],
-        {Tesla.Adapter.Finch, name: Logflare.FinchDefault, receive_timeout: 5_000}
-      )
-      |> Tesla.post(url, [])
+  def test_connection(%Backend{config: config}) do
+    __MODULE__.Client.test_connection(config)
+  end
```
Ziinc: i can foresee potential  situations where we would need to refer to the backend fields.
and as mentioned, if it is just delegating then we can just implement it here instead of a layer of indirection.

lib/logflare/backends/adaptor/axiom_adaptor.ex
```diff
@@ -105,34 +68,68 @@ defmodule Logflare.Backends.Adaptor.AxiomAdaptor do
...
 
-    case result do
-      {:ok, %Tesla.Env{status: 200}} -> :ok
-      {:ok, %Tesla.Env{body: %{"message" => message}}} -> {:error, message}
-      {:ok, env} -> {:error, "Unexpected response: #{env.status} #{inspect(env.body)}"}
-      {:error, reason} -> {:error, "Request error: #{reason}"}
-    end
+  @doc false
+  @spec transform_timestamp(map()) :: map()
+  def transform_timestamp(log_event_body) do
```
Ziinc: should be private function if not being used elsewhere.

lib/logflare/backends/adaptor/axiom_adaptor.ex
```diff
@@ -105,34 +68,68 @@ defmodule Logflare.Backends.Adaptor.AxiomAdaptor do
...
+    @behaviour HttpBased.Client
+
+    @impl HttpBased.Client
+    def send_logs(config, log_events, metadata) do
+      # TODO: Log failures
+      config
+      |> new()
+      |> Tesla.post("", log_events, opts: [metadata: metadata])
+
+      :ok
```
Ziinc: should return direct return values with Tesla.Env for broadway pipeline consumption

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -94,140 +72,33 @@ defmodule Logflare.Backends.Adaptor.SentryAdaptor do
...
-      "body" => message,
-      "trace_id" => extract_trace_id(log_event),
-      "attributes" => build_attributes(log_event)
-    }
-  end
+    @impl HttpBased.Client
+    def send_logs(config, log_events, metadata) do
+      config
+      |> new()
+      |> Tesla.post("", log_events, opts: [metadata: metadata])
```
Ziinc: feels like this is getting repeated across the adaptors.
would there be a case where the post needs to be customized? i would think that keeping this in the pipeline would be better.

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -94,140 +72,33 @@ defmodule Logflare.Backends.Adaptor.SentryAdaptor do
...
-    case log_event.body["trace_id"] || log_event.body["trace.id"] do
-      nil ->
-        generate_trace_id()
-
-      trace_id when is_binary(trace_id) ->
-        trace_id
-
-      trace_id ->
-        to_string(trace_id)
+      :ok
```
Ziinc: should return tesla return value for pipeline to handle retries in the future.

lib/logflare/backends/adaptor/sentry_adaptor.ex
```diff
@@ -94,140 +72,33 @@ defmodule Logflare.Backends.Adaptor.SentryAdaptor do
...
-
-      v when is_float(v) ->
-        %{"value" => v, "type" => "double"}
-
-      v when is_map(v) or is_list(v) ->
-        %{"value" => Jason.encode!(v), "type" => "string"}
 
-      v ->
-        %{"value" => inspect(v), "type" => "string"}
+    defp new(config) do
```
Ziinc: feels like we could have a `client_opts` callback on HttpBased.Client and let the Adaptor modules implement it as an additional behaviour instead. then the pipeline just calls that behaviour and pipes it into the `HttpBased.Client.new/1` and thereafter `Tesla.request/4`

Ziinc (changes requested):
not quite sold on the HttpBased.Client behaviour and the way that the callbacks are being called, seems like there is not enough varioations between calling patterns to justify the additional Client module that each http adaptor must implement.

lib/logflare/backends/adaptor/sentry_adaptor.ex
bblaszkow06: I had 2 cases in mind:
- URL customization - if test_connection happens to use different endpoint, it might be needed to pass the path here
- method customization - with no additional wrapper there's no nice way to change the HTTP method (maybe some custom middleware could do it, but this seems hacky)

Ziinc: - url option would handle pt1 when creating the client 
- method could be another option

lib/logflare/backends/adaptor/sentry_adaptor.ex
bblaszkow06: Did just that, but I didn't like Pipeline calling Tesla directly - moved that to `HttpBased.Client`

lib/logflare/backends/adaptor/http_based/client.ex
bblaszkow06: went with `send_events` - `send_request` sounds too vague considering the typespec here

lib/logflare/backends/adaptor/sentry_adaptor.ex
bblaszkow06: Since no adaptor seems to need it atm, I got rid of that. However, I was considering solving it with an optional callback `request_opts`, allowing overrides to the request opts:

```elixir
# in HttpBased.Client
  def send_events(module, backend, events, metadata \\ %{}) do
    metadata =
      (backend.metadata || %{})
      |> Map.new(fn {k, v} -> {"backend.#{k}", v} end)
      |> Map.merge(metadata)

    overrides =
      if function_exported?(module, :request_opts, 1) do
        module.request_opts(backend)
      else
        config
      end

    request_opts =
      [url: "", body: events, opts: [metadata: metadata]]
      |> Keyword.merge(overrides, fn
        :opts, opts, new_opts -> Keyword.merge(opts, new_opts)
        _key, _value, override -> override
      end)

    backend
    |> module.client_opts()
    |> new()
    |> Tesla.request(request_opts)
  end
```

WDYT? Is it going to be needed?

PR #2932, opened 2025-11-18
feat: Add logging and lock in settings for Clickhouse ingestion

amokan:
- [x] Logging and telemetry around Clickhouse ingestion
- [x] Remove options for Clickhouse ingestion and lock to `async_insert => 1` and `wait_for_async_insert => 1` (_along with gzip compression_)

Primary goal here is to get additional logging/telemetry around Clickhouse inserts and to simplify the internal API for ingestion by eliminating some extra options.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
@@ -10,9 +10,6 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Ingester do
   alias Logflare.LogEvent
 
   @finch_pool Logflare.FinchClickhouseIngest
-  @async_insert true
-  @wait_for_async_insert true
-  @gzip_compression true
 
   @doc """
   Inserts a list of `LogEvent` structs into ClickHouse.
@@ -35,12 +32,7 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Ingester do
     request_body = encode_batch(log_events)
 
     {request_body, headers} =
-      if @gzip_compression do
-        compressed = :zlib.gzip(request_body)
-        {compressed, [{"content-encoding", "gzip"}]}
-      else
-        {request_body, []}
-      end
+      {:zlib.gzip(request_body), [{"content-encoding", "gzip"}]}
 
     headers = [
       {"content-type", "application/octet-stream"},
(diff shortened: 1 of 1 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -330,9 +328,19 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor do
   def insert_log_events({%Source{}, %Backend{}}, []), do: :ok
 
   def insert_log_events({%Source{} = source, %Backend{} = backend}, [%LogEvent{} | _] = events) do
+    Logger.metadata(source_id: source.id, source_token: source.token, backend_id: backend.id)
```
Ziinc, line 331:
```suggestion
    Logger.metadata(source_id: source.token, source_token: source.token, backend_id: backend.id)
```
This will get rejected by BQ due to type difference

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -163,8 +139,8 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Ingester do
     params =
       URI.encode_query(%{
         "query" => query,
-        "async_insert" => if(async, do: "1", else: "0"),
-        "wait_for_async_insert" => if(wait_for_async_insert, do: "1", else: "0")
+        "async_insert" => "1",
+        "wait_for_async_insert" => "1"
```
Ziinc, line 143: Can add docs to CH backend docs on the ingestion flags that we use

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -77,11 +84,26 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Pipeline do
       }
     )
 
-    source = Sources.Cache.get_by_id(source_id)
-    backend = Backends.Cache.get_backend(backend_id)
-    events = for %{data: le} <- messages, do: le
+    attributes =
+      for {k, v} <- [
```
Ziinc, line 88: Setting the map directly will be faster, nil values will get ignored and removed from attributes on ingestion.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
amokan: I've made a change for this in #2933

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
amokan: Made this change in #2933

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
amokan: I will create a ticket for this and handle those changes when the dust settles on ingest

PR #2933, opened 2025-11-19
feat: adds retries to clickhouse ingest pipeline

amokan:
It makes sense to get some sensible retries going in the Clickhouse ingest pipeline. Lots of ways to do this, but I felt the most practical way to handle it was leveraging the Broadway `ack` callback.

To avoid indefinite looping of messages or some new ETS table, I added a `retries` attribute to the `LogEvent` struct so we can just increment that - shove the events back into the buffer and move on :shrug:


lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
Ziinc: we can avoid the comprehension and create the attribute map directly since `nil` values get ignored.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -95,27 +99,61 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Pipeline do
...
+    |> Enum.each(fn {%{source_id: source_id, backend_id: backend_id}, messages} ->
+      {retriable, exhausted} =
+        Enum.split_with(messages, fn %{data: event} ->
+          (event.retries || 0) < @max_retries
+        end)
+
+      if exhausted != [] do
+        Logger.warning(
+          "Dropping #{length(exhausted)} ClickHouse events after #{@max_retries} retries",
+          source_id: source_id,
```
Ziinc: needs to be source_token, can pull from cache.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -95,27 +99,61 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Pipeline do
...
+        )
+      end
+
+      if retriable != [] do
+        events =
+          for %{data: %LogEvent{} = event} <- retriable do
+            %{event | retries: (event.retries || 0) + 1}
+          end
+
+        IngestEventQueue.add_to_table({source_id, backend_id}, events)
```
Ziinc: need to also remove the old event, otherwise we will have duplicates in the recent logs.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -95,27 +99,61 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Pipeline do
...
+        Enum.split_with(messages, fn %{data: event} ->
+          (event.retries || 0) < @max_retries
+        end)
+
+      if exhausted != [] do
+        Logger.warning(
+          "Dropping #{length(exhausted)} ClickHouse events after #{@max_retries} retries",
+          source_id: source_id,
+          backend_id: backend_id
+        )
```
Ziinc: should remove the old event if its dropped.
JFYI we'll ingest it into the rejected logs  as a follow up.

PR #2943, opened 2025-11-20
OTLP HTTP adaptor

bblaszkow06:
This adds support for log export using OTLP - "http/protobuf" for now

### DIFF
--- a/test/logflare/backends/adaptor/otlp_adaptor_test.exs
+++ b/test/logflare/backends/adaptor/otlp_adaptor_test.exs
@@ -3,6 +3,7 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptorTest do
 
   alias Logflare.Backends
   alias Logflare.Backends.Adaptor
+  alias Logflare.Backends.Adaptor.HttpBased
   alias Logflare.Backends.AdaptorSupervisor
   alias Logflare.SystemMetrics.AllLogsLogged
   alias Opentelemetry.Proto.Collector.Logs.V1.ExportLogsPartialSuccess
@@ -86,8 +87,7 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptorTest do
         |> Enum.map(&Protobuf.encode/1)
 
       for response_body <- response_bodies do
-        @tesla_adapter
-        |> expect(:call, 2, fn env, _opts ->
+        mock_adapter(2, fn env ->
           assert env.method == :post
           assert env.url == "http://localhost:4318/v1/logs"
 
@@ -111,11 +111,7 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptorTest do
       ]
 
       for response <- error_responses do
-        @tesla_adapter
-        |> expect(:call, fn _env, _opts ->
-          response
-        end)
-
+        mock_adapter(fn _env -> response end)
         assert {:error, reason} = @subject.test_connection(ctx.backend)
         assert reason != nil
       end
@@ -135,8 +131,7 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptorTest do
       this = self()
       ref = make_ref()
 
-      @tesla_adapter
-      |> expect(:call, fn env, _opts ->
+      mock_adapter(fn env ->
         assert Tesla.build_url(env) == "http://localhost:4318/v1/logs"
 
         assert env.method == :post
@@ -171,8 +166,7 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptorTest do
       this = self()
       ref = make_ref()
 
-      @tesla_adapter
-      |> expect(:call, fn env, _opts ->
+      mock_adapter(fn env ->
         send(this, {ref, IO.iodata_to_binary(env.body)})
         {:ok, %Tesla.Env{status: 200, body: ""}}
       end)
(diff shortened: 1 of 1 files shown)


docs/docs.logflare.com/docs/backends/otlp.mdx
```diff
@@ -0,0 +1,18 @@
...
+The OTLP backend is **ingest-only**, and exports the logs via [OTLP](https://opentelemetry.io/docs/specs/otlp/) to any tool or service that supports it.
+
+## Behaviour and configurations
+
+### Configuration
+
+The backend can be configured with the following options:
+
+- `:endpoint` (string, required) - The URL of the OTLP endpoint.
+- `:protocol` (string) - The protocol to use for sending logs: `"grpc"` or `"http/protobuf"`. Defaults to `"http/protobuf"`.
```
Ziinc: should only say that http/protobuf is supported for now

lib/logflare/backends/adaptor/otlp_adaptor.ex
```diff
@@ -0,0 +1,111 @@
+defmodule Logflare.Backends.Adaptor.OtlpAdaptor do
+  @moduledoc """
+  Adaptor sending logs to ingest compatible with [OTLP](https://opentelemetry.io/docs/specs/otlp/#protocol-details)
+  This adaptor is **ingest-only**
+
+  ## Configuration
+  - `:endpoint` - URL of OTLP Endpoint
+  - `:protocol` - Protocol used for sending logs. Currently supported is "http/protobuf", with "grpc" coming soon.
+  - `:gzip` - Enables gzip compression of request
+  - `:headers` - A map of additional headers to set when making an HTTP request
```
Ziinc: don't need to duplicate the docs

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
```diff
@@ -0,0 +1,124 @@
...
+
+    response =
+      env
+      |> Tesla.put_header("content-type", @protobuf_content_type)
+      |> Tesla.put_body(transform_batch(env.body, metadata))
+      |> Tesla.run(next)
+
+    with {:ok, %{status: 200} = env} <- response do
+      if Tesla.get_header(env, "content-type") == @protobuf_content_type do
+        response = Protobuf.decode(env.body, ExportLogsServiceResponse)
```
Ziinc, line 32: note that in the future we might need to factor in the Metrics/Traces response when we export traces/metrics.

Ziinc: this adaptor needs to be generic enough to handle all 3 types of responses.

lib/logflare/backends/adaptor/otlp_adaptor.ex
```diff
@@ -0,0 +1,111 @@
...
+      {:ok, %Tesla.Env{status: 200, body: %{partial_success: %{error_message: ""}}}} -> :ok
+      {:ok, env} -> {:error, env}
+      {:error, _reason} = err -> err
+    end
+  end
+
+  @impl HttpBased.Client
+  def client_opts(%Backend{config: config}) do
+    [
+      url: Path.join(config.endpoint, "/v1/logs"),
```
Ziinc: usually for such otel configurations, it is common for user to explicitly provide the path (since they might have it behind a proxy or otherwise), so let's force the user to provide the full url with path, and add it as a changeset validation.

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
```diff
@@ -0,0 +1,124 @@
...
+    end
+  end
+
+  defp transform_batch(events, metadata) do
+    attributes = for {k, v} <- metadata, do: {"logflare.#{k}", v}
+
+    %ExportLogsServiceRequest{
+      resource_logs: [
+        %ResourceLogs{
+          resource: build_resource(attributes),
```
Ziinc: can populate the resource with more information as per [here](https://github.com/Logflare/logflare/blob/5c9617da411d757f488368d8969954f332d619db/lib/logflare/backends/user_monitoring.ex#L19)
i think we don't need to set the attribute metadata as of now.

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
```diff
@@ -0,0 +1,124 @@
...
+  end
+
+  defp build_resource(kv) do
+    %Resource{attributes: Enum.map(kv, &make_key_value/1)}
+  end
+
+  defp build_scope_logs(logs) do
+    [
+      %ScopeLogs{
+        scope: %InstrumentationScope{name: "logflare"},
```
Ziinc: we can use the `Source.service_name || Source.name` for this

lib/logflare/backends/adaptor/otlp_adaptor.ex
bblaszkow06: >and add it as a changeset validation.

If we assume the endpoint can be behind a proxy, there's no way to add stronger validation. In theory, the user may use a subdomain for the log ingest endpoint

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
bblaszkow06: It would require more changes, so leaving this as-is for now

lib/logflare/backends/adaptor/otlp_adaptor.ex
Ziinc: What I mean is that it is possible for them to have a different path for the logs ingestion, so this would be less flexible in that respect

Ziinc: What I mean is that it is possible for them to have a different path for the logs ingestion, so this would be less flexible in that respect

PR #2945, opened 2025-11-20
chore: force explorer to compile in CI with legacy artifacts option

amokan:


amokan:
false alarm. closing this.

PR #2955, opened 2025-11-24
chore: use http connection retries for clickhouse pipeline

amokan:
This change uses Tesla for http connections to Clickhouse and leverages its retry mechanism to retry up to 3 times when facing connection drops or issues.

I have also set the Broadway-level retries for Clickhouse to `0`, effectively disabling them for the time being. Rather than remove the tests for that logic, I have change them to check the retry count in the pipeline and only run those tests when enabled.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
@@ -63,7 +63,8 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Ingester do
        should_retry: &retriable?/1}
     ]
 
-    adapter = {Tesla.Adapter.Finch, name: @finch_pool, receive_timeout: 60_000}
+    adapter =
+      {Tesla.Adapter.Finch, name: @finch_pool, pool_timeout: 4_000, receive_timeout: 8_000}
 
     Tesla.client(middleware, adapter)
   end


lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -28,32 +31,49 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Ingester do
 
   def insert(connection_opts, table, [%LogEvent{} | _] = log_events)
       when is_list(connection_opts) and is_non_empty_binary(table) do
+    client = build_client(connection_opts)
     url = build_request_url(connection_opts, table)
-    request_body = encode_batch(log_events)
+    request_body = encode_batch(log_events) |> :zlib.gzip()
```
Ziinc, line 36: https://hexdocs.pm/tesla/Tesla.Middleware.Compression.html

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
Ziinc, line 171: https://hexdocs.pm/tesla/Tesla.Middleware.BasicAuth.html

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -28,32 +31,49 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Ingester do
...
+         build_auth_header(connection_opts)
+       ]},
+      {Tesla.Middleware.Retry,
+       delay: @initial_delay,
+       max_retries: @max_retries,
+       max_delay: @max_delay,
+       should_retry: &retriable?/1}
+    ]
+
+    adapter = {Tesla.Adapter.Finch, name: @finch_pool, receive_timeout: 60_000}
```
Ziinc:
```suggestion
    adapter = {Tesla.Adapter.Finch, name: @finch_pool, pool_timeout: 2_500, receive_timeout: 5_000}
```
safer to keep it lower, can match the ingest options
https://github.com/Logflare/logflare/blob/c98f89f763b1f9cd703a791186fbdac00e8c32df/lib/logflare/google/bigquery/gen_utils.ex#L325

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -28,32 +31,49 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Ingester do
...
   end
 
+  @spec build_client(Keyword.t()) :: Tesla.Client.t()
+  defp build_client(connection_opts) do
+    middleware = [
+      {Tesla.Middleware.Headers,
+       [
+         {"content-type", "application/octet-stream"},
+         {"content-encoding", "gzip"},
+         build_auth_header(connection_opts)
```
Ziinc, line 57: could be handled by middlewares

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -28,32 +31,49 @@ defmodule Logflare.Backends.Adaptor.ClickhouseAdaptor.Ingester do
...
-      | headers
-    ]
-
-    request = Finch.build(:post, url, headers, request_body)
-
-    case Finch.request(request, @finch_pool) do
-      {:ok, %{status: 200}} ->
+    case Tesla.post(client, url, request_body) do
+      {:ok, %Tesla.Env{status: 200}} ->
         :ok
```
Ziinc, line 40: can return the env so that Adaptor callback can emit telemetry in future, need to collect metrics from x-clickhouse-summary header

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
amokan: PR created for this - https://github.com/Logflare/logflare/pull/2984

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
amokan: @Ziinc I think my only concern here is that the Tesla middleware for compression requires [the body be a binary](https://github.com/elixir-tesla/tesla/blob/v1.15.3/lib/tesla/middleware/compression.ex#L40) so we'd have to perform an intermediate binary allocation if going that route.

PR #2956, opened 2025-11-25
feat: LQL supports select aliases

msmithstubbs:
* add support for an alias with select: `s:my_field@my_alias`
* extract default select fields into LQL SelectRules

This PR doesn't include any support for displaying aliases in the UI. That will be covered in a separate PR.

Part of #2848

### DIFF
--- a/lib/logflare/lql/backend_transformer/bigquery.ex
+++ b/lib/logflare/lql/backend_transformer/bigquery.ex
@@ -548,35 +548,36 @@ defmodule Logflare.Lql.BackendTransformer.BigQuery do
 
   @spec build_combined_select(Query.t(), [Logflare.Lql.Rules.SelectRule.t()]) :: Query.t()
   defp build_combined_select(query, select_rules) do
-    select_rules
-    |> Enum.reduce(query, fn %{path: path, alias: alias}, acc_query ->
+    Enum.reduce(select_rules, query, fn %{path: path, alias: alias}, acc_query ->
       is_nested = path not in @special_top_level and String.contains?(path, ".")
-      field_name = path |> split_by_dots |> List.last()
-
-      cond do
-        is_nested ->
-          name = if is_binary(alias), do: alias, else: String.replace(path, ".", "_")
-
-          acc_query
-          |> handle_nested_field_access(path)
-          |> select_merge([..., t], %{
-            ^name => fragment("? AS ?", field(t, ^field_name), identifier(^name))
-          })
-
-        is_binary(alias) ->
-          acc_query
-          |> select_merge([t], %{
-            ^alias => fragment("? AS ?", field(t, ^field_name), identifier(^alias))
-          })
-
-        is_nil(alias) ->
-          acc_query
-          |> select_merge([t], %{^path => field(t, ^field_name)})
-      end
+      field_name = path |> split_by_dots() |> List.last()
+      add_select_for_field(acc_query, path, field_name, alias, is_nested)
     end)
   end
 
-  @spec unnest_paths_for_select([String.t()]) :: [String.t()]
+  @spec add_select_for_field(Query.t(), String.t(), String.t(), String.t() | nil, boolean()) ::
+          Query.t()
+  defp add_select_for_field(query, path, field_name, alias, true = _is_nested) do
+    name = if is_binary(alias), do: alias, else: String.replace(path, ".", "_")
+
+    query
+    |> handle_nested_field_access(path)
+    |> select_merge([..., t], %{
+      ^name => fragment("? AS ?", field(t, ^field_name), identifier(^name))
+    })
+  end
+
+  defp add_select_for_field(query, _path, field_name, alias, false = _is_nested)
+       when is_binary(alias) do
+    select_merge(query, [t], %{
+      ^alias => fragment("? AS ?", field(t, ^field_name), identifier(^alias))
+    })
+  end
+
+  defp add_select_for_field(query, path, field_name, nil = _alias, false = _is_nested) do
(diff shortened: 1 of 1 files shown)


lib/logflare/lql/backend_transformer/bigquery.ex
```diff
@@ -546,21 +546,38 @@ defmodule Logflare.Lql.BackendTransformer.BigQuery do
...
+      field_name = path |> split_by_dots() |> List.last()
+      add_select_for_field(acc_query, path, field_name, alias, is_nested)
     end)
   end
 
-  @spec unnest_paths_for_select([String.t()]) :: [String.t()]
+  @spec add_select_for_field(Query.t(), String.t(), String.t(), String.t() | nil, boolean()) ::
+          Query.t()
+  defp add_select_for_field(query, path, field_name, alias, true = _is_nested) do
+    name = if is_binary(alias), do: alias, else: String.replace(path, ".", "_")
```
Ziinc, line 561: i think if the alias has an invalid character that is not supported by the given backend, we should just send an error back to the user instead of silently transforming it

Ziinc: could lead to a mismatch of expectations

lib/logflare/lql/backend_transformer/clickhouse.ex
```diff
@@ -485,8 +485,9 @@ defmodule Logflare.Lql.BackendTransformer.ClickHouse do
 
   @spec build_combined_select(Query.t(), select_rules :: [map()]) :: Query.t()
   defp build_combined_select(query, select_rules) do
-    Enum.reduce(select_rules, query, fn %{path: path}, acc_query ->
-      select_merge(acc_query, [l], %{^path => field(l, ^path)})
+    Enum.reduce(select_rules, query, fn %{path: path, alias: alias}, acc_query ->
+      path_or_alias = if is_binary(alias), do: alias, else: String.replace(path, ".", "_")
```
Ziinc, line 489: as above, should validate not transform

lib/logflare/lql/backend_transformer/postgres.ex
```diff
@@ -667,8 +667,19 @@ defmodule Logflare.Lql.BackendTransformer.Postgres do
 
   @spec build_combined_select(Query.t(), select_rules :: [map()]) :: Query.t()
   defp build_combined_select(query, select_rules) do
-    Enum.reduce(select_rules, query, fn %{path: path}, acc_query ->
-      select_merge(acc_query, [l], %{^path => field(l, ^path)})
+    Enum.reduce(select_rules, query, fn %{path: path, alias: alias_name}, acc_query ->
+      path_or_alias =
+        if is_binary(alias_name), do: alias_name, else: String.replace(path, ".", "_")
```
Ziinc, line 672: as above, should validate not transform.

lib/logflare/lql/parser/helpers.ex
```diff
@@ -328,4 +328,14 @@ defmodule Logflare.Lql.Parser.Helpers do
   end
 
   def check_for_no_invalid_metadata_field_values(rule, _), do: rule
+
+  @spec check_for_invalid_blank_select_alias(Keyword.t() | {atom(), any()}) ::
+          Keyword.t() | {atom(), any()} | no_return()
+  def check_for_invalid_blank_select_alias({:invalid_blank_select_alias, "@"}) do
+    throw(
+      "Error while parsing select clause: select alias cannot be blank, expected alias name after @"
```
Ziinc, line 336:
```suggestion
      "Error while parsing select clause: select alias cannot be blank, expected alias name after @. For example, s:my_field@my_alias"
```

test/logflare/lql/backend_transformer/postgres_test.exs
```diff
@@ -425,4 +425,26 @@ defmodule Logflare.Lql.BackendTransformer.PostgresTest do
...
+
+  describe "apply_select_rules_to_query/3 with aliases" do
+    test "applies top-level field with alias" do
+      query = from(l in "logs")
+      select_rule = %SelectRule{path: "event_message", alias: "msg"}
+
+      result = Postgres.apply_select_rules_to_query(query, [select_rule], [])
+
+      assert %Ecto.Query{select: %{expr: expr}} = result
+      assert expr |> Macro.to_string() =~ "msg"
```
Ziinc, line 437:
```suggestion
      assert expr |> Macro.to_string() |> String.downcase() =~ "as msg"
```

test/logflare/lql/backend_transformer/postgres_test.exs
```diff
@@ -425,4 +425,26 @@ defmodule Logflare.Lql.BackendTransformer.PostgresTest do
...
+    end
+
+    test "applies nested field with alias" do
+      query = from(l in "logs")
+      select_rule = %SelectRule{path: "metadata.user.id", alias: "user_id"}
+
+      result = Postgres.apply_select_rules_to_query(query, [select_rule], [])
+
+      assert %Ecto.Query{select: %{expr: expr}} = result
+      assert expr |> Macro.to_string() =~ "user_id"
```
Ziinc, line 447:
```suggestion
      assert expr |> Macro.to_string() |> String.downcase() =~ "as user_id"
```

Ziinc (approved):

lib/logflare/lql/backend_transformer/bigquery.ex
msmithstubbs: This isn't actually a validation but setting the top level field name. 

If an alias is set, use that. If not, use the flattened structure (eg. `some.nested.path` becomes `some_nested_path`)

Can remove the transform if you like?

PR #2958, opened 2025-11-25
perf: syn gen scope

Ziinc:
- reduces messages queue buildup for syn_gen_scope by partitioning the processes
- increase startup delay for RecentTouch to up to 10s range.
- increase cleanup interval for idle sources.

Closes ANL-1230


chasers (approved):

SteffenDE (commented):
Looks good to me! :)

PR #2959, opened 2025-11-25
test(e2e): checks if logs from API-Gateway are received

laraujo7:


test/e2e/supabase/features/index.spec.ts
```diff
@@ -5,3 +5,18 @@ test('has title', async ({ page }) => {
 
   await expect(page).toHaveTitle(/Default Project | Default Organization | Supabase/);
 });
+
+test('receives logs from API Gateway', async ({ request, page }) => {
+  await request.get('http://localhost:8000/api/platform/organizations');
+
+  await page.waitForTimeout(5000);
```
nelsonmestevao: I think this should be moved to a `test.beforeAll` so it is called before all tests and then we can reuse this and multiple other request calls in our tests 🙏

nelsonmestevao:
```js
test.beforeAll(async ({ request }) => {
  await request.get('http://localhost:8000/api/platform/organizations');
  await new Promise(r => setTimeout(r, 5000)); // Since we cannot use the `page.waitForTimeout` in beforeAll statments
});
```

test/e2e/supabase/features/index.spec.ts
```diff
@@ -5,3 +5,18 @@ test('has title', async ({ page }) => {
 
   await expect(page).toHaveTitle(/Default Project | Default Organization | Supabase/);
 });
+
+test('receives logs from API Gateway', async ({ request, page }) => {
+  await request.get('http://localhost:8000/api/platform/organizations');
```
nelsonmestevao:
```suggestion
  await request.get('/api/platform/organizations');
```

Does this work? That way we avoid using hard coded values 🙏

laraujo7: Yeap it works :muscle:

Ziinc (approved):

PR #2960, opened 2025-11-25
chore: screenshot playwright tests on failure

nelsonmestevao:


Ziinc (approved):
Can also adjust ci artifacts to display as part of the action output in future

PR #2968, opened 2025-11-26
fix: add team query param everywhere

Ziinc:
Add team query param everywhere to more links

### DIFF
--- a/lib/logflare_web/live/search_live/query_debug_component.ex
+++ b/lib/logflare_web/live/search_live/query_debug_component.ex
@@ -15,14 +15,14 @@ defmodule LogflareWeb.Search.QueryDebugComponent do
   def render(assigns) do
     assigns =
       assign_new(assigns, :team, fn
-        %{team_user: team_user} ->
+        %{team_user: team_user} when not is_nil(team_user) ->
           Teams.get_team!(team_user.team_id)
 
         %{team: team} ->
           team
 
-        %{user: user} ->
-          Teams.get_team!(user.team_id)
+        %{user: %{team: team}} when not is_nil(team) ->
+          team
 
         _ ->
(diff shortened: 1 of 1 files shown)


lib/logflare_web/live/search_live/query_debug_component.ex
```diff
@@ -6,14 +6,19 @@ defmodule LogflareWeb.Search.QueryDebugComponent do
...
 
   def update(assigns, socket) do
     {:ok, assign(socket, assigns)}
   end
 
   def render(assigns) do
     assigns =
-      assign_new(assigns, :search_op, fn
+      assign_new(assigns, :team, fn
+        %{team_user: team_user} ->
```
amokan: Prob want a guard on this to handle cases where `team_user` is nil

lib/logflare_web/live/search_live/templates/logs_search.html.heex
```diff
@@ -18,10 +18,7 @@
 <div id="source-logs-search-control" class="subhead " phx-hook="SourceLogsSearch">
   <div class="container mx-auto">
     <h5>
-      ~/logs/{link(@source.name,
-        to: Routes.source_path(@socket, :show, @source),
-        class: "text-primary"
-      )}/search
+      ~/logs/<.team_link team={@team || @team_user} href={~p"/sources/#{@source}"} class="text-primary">{@source.name}</.team_link>/search
```
amokan: I think if this falls back to `@team_user` it would end up with `?t=<team_user.id>` instead of `?t=<team_user.team_id>` right?

Ziinc: fixed

PR #2970, opened 2025-11-27
feat: arrow rs implementation

WolfDan:


lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
```diff
@@ -24,18 +27,17 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor.GoogleApiClient do
         ]
       )
 
-    {:ok, arrow_schema} = DataFrame.dump_ipc_schema(data_frame)
+    {arrow_schema, batch_msgs} =
+      data_frame
```
Ziinc, line 31: this isn't a data frame anymore correct?

native/arrowipc_ex/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+#[rustler::nif(schedule = "DirtyIo")]
+fn get_ipc_bytes(env: Env, dataframe_json: String) -> NifResult<(Binary, Vec<Binary>)> {
+    // convert to Newline delimited JSON (the reader only accepts this format)
+    // remove the [] brackets
+    let mut json_chars = dataframe_json.chars();
+    json_chars.next();
+    json_chars.next_back();
+    // replace comma separators to new line
+    let dataframe_json = json_chars.as_str().replace("},{", "}\n{");
+    let byte_data = dataframe_json.as_bytes();
```
Ziinc: this seems quite fragile, we could pass it as ndjson from elixir side instead of having to do this conversion

native/arrowipc_ex/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    let inferred_schema = {
+        let (inferred_schema, _) = infer_json_schema(byte_data, None).unwrap();
+
+        let new_fields: Vec<Arc<Field>> = inferred_schema
+            .fields()
+            .iter()
+            .map(|field| {
+                if field.name() == "timestamp" {
+                    // this is a hardcoded field and the only one that we know is timestamp
```
Ziinc, line 22: we technically have 3 fields, `timestamp`, `event_message` and `id` that are always persent

native/arrowipc_ex/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        ipc_schema_bytes
+    };
+
+    // Determine max_chunksize of the record batches. Because max size of
+    // AppendRowsRequest is 10 MB, we need to split the table if it's too big.
+    // See: https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#appendrowsrequest
+    const max_request_bytes: usize = {
+        let base: usize = 2;
+        10 * base.pow(20) // 10 MB
```
Ziinc: might want to reduce the batch size slightly down to 8MB to be conservative

Ziinc (approved):
@WolfDan could you paste in your benchee results to the description? 🙏 looks 🔥

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
WolfDan: Yes that's correct! But still an appropriate name for what the data is

native/arrowipc_ex/src/lib.rs
WolfDan: Sure, the only reason that I did it here is because less processing compared to elixir

native/arrowipc_ex/src/lib.rs
WolfDan: Yes, but timestamp is the only one that actually is a timestamp the others are strings and they don't need any changes

PR #2978, opened 2025-11-28
feat: team dropdown in navbar

msmithstubbs:
This is a WIP of a team switcher dropdown for the navbar.

## todo 
- [x] remove teams list from dashboard
- [x] use existing styling for dropdown (no checkmark) 
- [x] test on mobile layout
- [x] keep team name if only one team
- [x] include current query params in team link
- [x] test coverage
- [x] don't switch teams if a `t=` param is present. 

CleanShot 2025-12-03 at 14 49 40@2x

### User with single team

A user with only one team will see the team name in the navbar but no dropdown.

CleanShot 2025-12-03 at 14 46 56@2x

## Mobile layout

The switcher can be used when on a small screen such as a mobile device.

https://github.com/user-attachments/assets/236f43f5-0245-4432-baa5-b5f22270b4a8

### DIFF
--- a/lib/logflare_web/live/endpoints/endpoints_live.ex
+++ b/lib/logflare_web/live/endpoints/endpoints_live.ex
@@ -96,7 +96,7 @@ defmodule LogflareWeb.EndpointsLive do
     socket =
       socket
       |> assign(:show_endpoint, endpoint)
-      |> LogflareWeb.AuthLive.assign_context_by_resource(endpoint, user.email)
+      |> maybe_assign_team_context(params, endpoint)
       |> then(fn
         socket when endpoint != nil ->
           {:ok, parsed_result} =
@@ -482,4 +482,10 @@ defmodule LogflareWeb.EndpointsLive do
   end
 
   defp maybe_redact_query(query, _redact_pii), do: query
+
+  defp maybe_assign_team_context(socket, %{"t" => _team_id}, _endpoint), do: socket
+
+  defp maybe_assign_team_context(socket, _params, endpoint) do
+    LogflareWeb.AuthLive.assign_context_by_resource(socket, endpoint, socket.assigns.user.email)
+  end
 end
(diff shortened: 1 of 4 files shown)


Ziinc:
Styling wise, let's keep the selection dropdown similar to the current styling, bold and greyed for the current selection, that obviates the need for the checkmark icon

lib/logflare_web/templates/layout/navbar.html.heex
```diff
@@ -77,6 +77,7 @@
     </ul>
     <%= if assigns[:user] do %>
       <ul class="navbar-nav ml-auto">
+        <.team_switcher :if={assigns[:teams] && length(assigns[:teams]) > 1} team_context={Logflare.Teams.TeamContext.build(assigns)} teams={assigns[:teams]} current_path={@conn.request_path} />
```
Ziinc: So this hides the switcher if there is only one team? I think preserving the dropdown even if it is only one item is fine, user needs to know what their team name is, even if the switcher is useless.

Ziinc: Might be better to include the current query params as well, in case user is switching on the query page.

msmithstubbs: all makes sense. will do.

msmithstubbs: Kept the team name for single team user

Ziinc:
👍 for the mobile ux

lib/logflare_web/live/endpoints/endpoints_live.ex
```diff
@@ -482,4 +482,10 @@ defmodule LogflareWeb.EndpointsLive do
   end
 
   defp maybe_redact_query(query, _redact_pii), do: query
+
+  defp maybe_assign_team_context(socket, %{"t" => _team_id}, endpoint), do: socket
+
+  defp maybe_assign_team_context(socket, params, endpoint) do
+    LogflareWeb.AuthLive.assign_context_by_resource(socket, endpoint, socket.assigns.user.email)
+  end
```
msmithstubbs, line 490: if a team is set with the `t=` param it will be respected. This is required otherwise using the team switcher to change teams will be overridden.

PR #2985, opened 2025-12-01
feat: Syslog backend

ruslandoga:
Continues from https://github.com/Logflare/logflare/pull/2067

---

### TODOs

- [x] Enable payload encryption with user-uploaded cipher keys
- [x] Format log messages according to syslog protocol [RFC5424](https://tools.ietf.org/html/rfc5424)
- [x] Deliver over TCP as described in [RFC6587](https://tools.ietf.org/html/rfc6587) (implemented with octet counting)
- [x] TLS support based on [RFC5425](https://datatracker.ietf.org/doc/html/rfc5425)
- [x] Move telegraf usage to a docker-compose workflow with write-to-file assertions
- [x] CI workflow should also be updated
- [x] Create TCP backend via /backends liveview form
- [x] clean up tests
- [x] add Syslog Structured Data to config
- [x] add failed batch retries
- [x] update ui
- [x] add docs
- [x] more tests for retries
- [x] try it out irl with some hosted syslog sink service

---

### Questions

- [x] How should LogEvent data be mapped to the Syslog hostname, app-name, procid, and msgid fields?
- [x] How should Syslog Structured Data be handled, should it be populated from LogEvent? Some platforms like Better Stack [rely](https://betterstack.com/docs/logs/syslog-setup/) on SD for authentication. Maybe SD can be provided via config.
- [x] ~~Should this backend expose the Broadway topology parameters in the backend / application configuration?~~
- [x] ~~Instead of requiring separate `host`, `port`, and `tls` fields, could it be a connection URI (e.g., `tcp://host:port`, or `ssl://host:port` or `tls://host:port`)? It might even contain cipher keys/certs/... like in psql-style urls.~~


lib/logflare/backends/adaptor/tcp_adaptor/pipeline.ex
```diff
@@ -0,0 +1,74 @@
...
+      name: name,
+      producer: [
+        module: {Backends.BufferProducer, backend_id: backend.id, source_id: source.id},
+        transformer: {__MODULE__, :transform, []}
+      ],
+      processors: [
+        default: [min_demand: 1]
+      ],
+      batchers: [
+        tcp: [concurrency: 1, batch_size: 50]
```
Ziinc:
```suggestion
        tcp: [concurrency: 5, batch_size: 50]
```

lib/logflare/backends/adaptor/syslog_adaptor/pool.ex
```diff
@@ -0,0 +1,173 @@
+defmodule Logflare.Backends.Adaptor.TCPAdaptor.Pool do
+  @moduledoc false
+  import Kernel, except: [send: 2]
+  @behaviour NimblePool
+
+  def start_link(opts) do
+    config = Keyword.fetch!(opts, :config)
+    name = Keyword.fetch!(opts, :name)
+    NimblePool.start_link(worker: {__MODULE__, config}, lazy: true, name: name)
```
Ziinc: if configuration changes at runtime, say the user updates their cert files, we would need to accommodate for the configuration changes, either by restarting the pool (and all workers) or triggering a reconnect at the worker level  if the underlying backend config changes'
would need test coverage for this case as well.

lib/logflare/backends/adaptor/syslog_adaptor/pool.ex
```diff
@@ -0,0 +1,173 @@
...
+  def init_pool(config) do
+    host = Map.fetch!(config, :host)
+    port = Map.fetch!(config, :port)
+    transport = if Map.get(config, :tls), do: :ssl, else: :gen_tcp
+
+    connect_opts =
+      [mode: :binary, active: false, nodelay: true]
+      |> maybe_configure_ssl(transport, config, host)
+
+    pool_state = %{
```
Ziinc, line 42: as mentioned, holding config state in worker might lead to stale configuration, especially in a multi-cluster setup.
should pull config from backend cache

lib/logflare/backends/adaptor/tcp_adaptor/syslog.ex
```diff
@@ -0,0 +1,109 @@
...
+  [RFC5424]: https://www.rfc-editor.org/rfc/rfc5424
+  [octet-counting framing]: https://www.rfc-editor.org/rfc/rfc6587#section-3.4.1
+  """
+
+  alias Logflare.LogEvent
+
+  def format(log_event, cipher_key \\ nil) do
+    %LogEvent{id: id, body: body} = log_event
+
+    level = get_in(body["body"]["level"]) || body["level"] || "info"
```
Ziinc:
```suggestion
    level = get_in(body["metadata"]["level"]) || body["level"] || "info"
```

lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex
```diff
@@ -0,0 +1,109 @@
...
+      body
+      |> Map.fetch!("timestamp")
+      |> DateTime.from_unix!(:microsecond)
+      |> DateTime.truncate(:millisecond)
+      |> DateTime.to_iso8601()
+
+    metadata = body["metadata"]
+    hostname = get_in(metadata["host"]) || body["host"]
+    app_name = get_in(metadata["app_name"]) || body["app_name"] || "logflare"
+    procid = get_in(metadata["pid"]) || body["pid"]
```
Ziinc: should factor in OpenTelemetry resource as well, which would be at the top-level `resource` field.
scope is also sometimes used but we can ignore that for now.

CleanShot 2025-12-10 at 19 04 54@2x

lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex
```diff
@@ -0,0 +1,109 @@
...
+      # SP
+      ?\s
+      # MSG
+      | msg
+    ]
+
+    [Integer.to_string(IO.iodata_length(syslog_msg)), ?\s | syslog_msg]
+  end
+
+  @levels %{
```
Ziinc: top of module

lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex
```diff
@@ -0,0 +1,109 @@
...
+    "informational" => 6,
+    "info" => 6,
+    "debug" => 7
+  }
+
+  defp severity_code(level) when level in 0..7, do: level
+
+  defp severity_code(str) when is_binary(str) do
+    str = String.downcase(str)
+    Map.fetch!(@levels, str)
```
Ziinc: would raise if not present or non-conventional level is used. should default to info if level is unknown,

Ziinc: it is already defaulting to "level" if the key is not present, but this would account for unrecognized levels.

lib/logflare/backends/adaptor/tcp_adaptor.ex
```diff
@@ -0,0 +1,126 @@
+defmodule Logflare.Backends.Adaptor.TCPAdaptor do
```
Ziinc: would be more accurate to call this SyslogAdaptor, since TCPAdaptor would be more generic and wouldnt care about  the message format.
we can refactor this into something like HttpBased in the future

lib/logflare/backends/backend.ex
```diff
@@ -26,7 +26,8 @@ defmodule Logflare.Backends.Backend do
     s3: Adaptor.S3Adaptor,
     axiom: Adaptor.AxiomAdaptor,
     otlp: Adaptor.OtlpAdaptor,
-    last9: Adaptor.Last9Adaptor
+    last9: Adaptor.Last9Adaptor,
+    tcp: Adaptor.TCPAdaptor
```
Ziinc:
```suggestion
    syslog: Adaptor.SyslogAdaptor
```

lib/logflare_web/live/backends/backends_live.ex
```diff
@@ -310,7 +310,8 @@ defmodule LogflareWeb.BackendsLive do
       {"S3", :s3},
       {"Axiom", :axiom},
       {"OTLP", :otlp},
-      {"Last9", :last9}
+      {"Last9", :last9},
+      {"TCP", :tcp}
```
Ziinc:
```suggestion
      {"Syslog", :syslog
```

lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex
ruslandoga: ~~This is for procid, right?~~

lib/logflare/backends/adaptor/syslog_adaptor.ex
```diff
@@ -1,12 +1,12 @@
...
+defmodule Logflare.Backends.Adaptor.Syslog do
   @moduledoc """
   TCP backend adaptor that sends out Syslog-formatted messages.
   """
 
   use Supervisor
   use TypedStruct
   import Ecto.Changeset
-  alias Logflare.Backends.Adaptor.TCPAdaptor.{Pool, Pipeline}
+  alias Logflare.Backends.Adaptor.Syslog.{Pool, Pipeline}
```
ruslandoga: Fixed in https://github.com/Logflare/logflare/pull/2985/commits/1bdde714555f87d218226ae128c6a28cbb3d03ee (Syslog -> SyslogAdaptor) and then again in https://github.com/Logflare/logflare/pull/2985/commits/da3ce14830ef9328381d5f7627255240a747ab21

lib/logflare/backends/adaptor/tcp_adaptor.ex
ruslandoga: Renamed to SyslogAdaptor in https://github.com/Logflare/logflare/pull/2985/commits/1093195004ade541d536f80abdd7b049eb984440 (and than also in https://github.com/Logflare/logflare/pull/2985/commits/da3ce14830ef9328381d5f7627255240a747ab21)

lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex
ruslandoga: Moved in https://github.com/Logflare/logflare/pull/2985/commits/0ffd4d7c256fb00d057f039dc630140bec1a1cac

lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex
ruslandoga: It's now something like

```elixir
# at the top of that module
@levels %{
  "emergency" => 0,
  "emer" => 0,
  "alert" => 1,
  "critical" => 2,
  "crit" => 2,
  "error" => 3,
  "err" => 3,
  "warning" => 4,
  "warn" => 4,
  "notice" => 5,
  "informational" => 6,
  "info" => 6,
  "debug" => 7
}

@default_level Map.fetch!(@levels, "info")

# ...

defp severity_code(level) when level in 0..7, do: level
defp severity_code(nil), do: @default_level

defp severity_code(str) when is_binary(str) do
  str = String.downcase(str)
  Map.get(@levels, str, @default_level)
end
```

Commits: https://github.com/Logflare/logflare/pull/2985/commits/0ffd4d7c256fb00d057f039dc630140bec1a1cac and https://github.com/Logflare/logflare/pull/2985/commits/da3ce14830ef9328381d5f7627255240a747ab21

lib/logflare/backends/adaptor/tcp_adaptor/syslog.ex
ruslandoga: The reason I used `"body"` was that there are two "bodies" before we get to the "user-supplied" params.

That is, from this

```elixir
body = %{
  "message" => "hello from opentelemetry",
  "level" => "debug",
  "resource" => %{
    "cluster" => "versioned",
    "name" => "Logflare (from resource.name)",
    "node" => ~s[:"logflare-versioned@10.0.0.123"],
    "service" => %{"name" => "Logflare", "version" => "1.26.25"}
  }
}

log_event = build(:log_event, source: source, body: body)
```

we get this

```elixir
%Logflare.LogEvent{
  id: "fa6b82f8-1a68-48e3-bd6f-43d12bb3cf9c",
  body: %{
    "body" => %{
      "level" => "debug",
      "message" => "hello from opentelemetry",
      "resource" => %{
        "cluster" => "versioned",
        "name" => "Logflare (from resource.name)",
        "node" => ":\"logflare-versioned@10.0.0.123\"",
        "service" => %{"name" => "Logflare", "version" => "1.26.25"}
      }
    },
    "event_message" => "test-msg",
    "id" => "fa6b82f8-1a68-48e3-bd6f-43d12bb3cf9c",
    "timestamp" => 1765463387926222
  },
  valid: true,
  drop: false,
  is_from_stale_query: nil,
  ingested_at: ~N[2025-12-11 14:29:47.939186],
  origin_source_id: :"13763990-6863-471c-a42a-41b37571d436",
  via_rule: nil,
  retries: 0,
  source_id: 2004,
  pipeline_error: nil
}
```

and the original `body["level"]` becomes `log_event.body["body"]["level"]`

lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex
ruslandoga: Added `resource` in https://github.com/Logflare/logflare/pull/2985/commits/22a0b0dd773ef22be19432bd7915e09b5646bdec

lib/logflare/backends/adaptor/tcp_adaptor/syslog.ex
ruslandoga: I'll try something like this, where we prefer `body["body"]` for lookups over `body` (which I called "root" body)

```elixir
def format(log_event, cipher_key \\ nil) do
  %LogEvent{id: id, body: root_body} = log_event

  body = root_body["body"]
  metadata = get_in(body["metadata"]) || root_body["metadata"]
  # resource comes from opentelemetry
  resource = get_in(body["resource"]) || root_body["resource"]
  level = get_in(metadata["level"]) || get_in(body["level"]) || root_body["level"]
  pri = 16 * 8 + severity_code(level)

  timestamp =
    root_body
    |> Map.fetch!("timestamp")
    |> DateTime.from_unix!(:microsecond)
    |> DateTime.truncate(:millisecond)
    |> DateTime.to_iso8601()

  hostname =
    get_in(metadata["host"]) ||
      get_in(resource["node"]) ||
      get_in(body["host"]) ||
      root_body["host"]

  # we default to `"logflare"` since telegraf rejects events without APP-NAME
  app_name =
    get_in(metadata["app_name"]) ||
      get_in(resource["name"]) ||
      get_in(body["app_name"]) ||
      root_body["app_name"] ||
      "logflare"

  procid = get_in(metadata["procid"]) || get_in(body["procid"]) || root_body["procid"]
```

Commit: https://github.com/Logflare/logflare/pull/2985/commits/22a0b0dd773ef22be19432bd7915e09b5646bdec

Ziinc: ah i think there is some misunderstanding in how the factory works:
```
iex> log_event = build(:log_event, source: source, some: "metadata", other: %{"field" =>  123})
iex> log_event
%LogEvent{
  body: %{
    "some" => "metadata",
    "other" => %{"field" => 123}
  },
  ...
}
```
we only take certain keys to build the event, then everything else gets set to the :body key.

ruslandoga: Thank you!

Fixed in https://github.com/Logflare/logflare/pull/2985/commits/31ed494b5ce6ec03663f42641e5dc5732f3376d9

ruslandoga: One other thing I am not sure if I am doing correctly is ingestion in tests. Right now it seems to take ~2 seconds per test: one second comes from BufferProducer interval, and then another second is spent in Broadway batcher. This makes Logflare.Backends.Adaptor.SyslogAdaptorTest add 13 seconds to the test suite.

```
Logflare.Backends.Adaptor.SyslogAdaptorTest [test/logflare/backends/adaptor/syslog_adaptor_test.exs]
  * test extracts level from input (2545.0ms) [L#67]
  * test sends message over mTLS (2154.4ms) [L#99]
  * test can send encrypted message (2187.5ms) [L#139]
  * test handles opentelemetry metadata (2062.9ms) [L#40]
  * test basic fields check (2153.0ms) [L#15]
  * test replaces invalid or empty log level with `info` severity code (2143.8ms) [L#83]

Finished in 13.3 seconds (0.00s async, 13.3s sync)
```

I wonder if there is a standard way to make it faster while still keeping end-to-end "benefits" (still going through the flow of "real" messages: BufferProducer -> syslog broadway pipeline -> syslog pool) but with maybe much smaller wait intervals and timeouts.

---

**UPDATE:**

For example, these changes

```diff
@@ -22,14 +22,16 @@ defmodule Logflare.Backends.Adaptor.SyslogAdaptor.Pipeline do
     Broadway.start_link(__MODULE__,
       name: name,
       producer: [
-        module: {Backends.BufferProducer, backend_id: backend.id, source_id: source.id},
+        module:
+          {Backends.BufferProducer,
+           backend_id: backend.id, source_id: source.id, interval: to_timeout(millisecond: 50)},
         transformer: {__MODULE__, :transform, []}
       ],
       processors: [
         default: [min_demand: 1]
       ],
       batchers: [
-        syslog: [concurrency: 5, batch_size: 50]
+        syslog: [concurrency: 5, batch_size: 50, batch_timeout: to_timeout(millisecond: 50)]
       ],
       context: %{
         source_id: source.id,
```

result in 

```
Logflare.Backends.Adaptor.SyslogAdaptorTest [test/logflare/backends/adaptor/syslog_adaptor_test.exs]
  * test can send encrypted message (538.2ms) [L#139]
  * test sends message over mTLS (322.3ms) [L#99]
  * test replaces invalid or empty log level with `info` severity code (130.8ms) [L#83]
  * test validates PEM configuration (3.5ms) [L#118]
  * test handles opentelemetry metadata (251.5ms) [L#40]
  * test extracts level from input (226.7ms) [L#67]
  * test basic fields check (234.5ms) [L#15]

Finished in 1.8 seconds (0.00s async, 1.8s sync)
```

But backend config probably shouldn't contain `batch_timeout` and Backends.BufferProducer `interval` values since they are test-only.

lib/logflare/backends/adaptor/syslog_adaptor/pipeline.ex
```diff
@@ -0,0 +1,82 @@
...
+    %{pool: pool, cipher_key: cipher_key} = context
+
+    content =
+      for %Broadway.Message{data: log_event} <- messages do
+        Syslog.format(log_event, cipher_key)
+      end
+
+    case Pool.send(pool, content) do
+      :ok -> messages
+      {:error, reason} -> fail_batch(messages, reason)
```
ruslandoga, line 60: I wonder if we need to add `handle_failed` and retry there. We can keep something like `:attempt` in the ack metadata to stop retrying after `attempt > 3` or something like that.

lib/logflare/backends/adaptor/syslog_adaptor/pool.ex
ruslandoga: I am certainly missing context, but why don't backend adapters "just" get restarted after an update? That would accommodate configuration changes. Downtime?

lib/logflare/backends/adaptor/syslog_adaptor/pool.ex
ruslandoga: How would it work exactly? In my current understanding it would be:
- pass just backend_id to the pool init, don't make any args for gen_tcp/ssl
- on checkout
  - if no socket, connect by pulling backend from cache by backend_id, build args for gen_tcp/ssl, connect
  - if socket, pull backend from cache by backend_id, build args for gen_tcp/ssl, compare with args for that socket
    - if equal, give socket
    - if not equal, remove socket (nimble_pool will attempt to checkout another socket)
    
Is that about right? I don't like cache lookup and building args and comparing them on every send, but I guess it's amortized thanks to batching in Broadway pipeline.

---

**UPDATE:** I did it this way in https://github.com/Logflare/logflare/pull/2985/changes/e4eb23a99e542ba84ee2ab0a151a32b9885eafb8 but the new test with config change doesn't pass since the Backend cache is not reset on update (cached backend still contains old config).

---

**UPDATE:** ah, the cache seems to get busted on WAL changes (?), but they probably (?) don't work in Ecto sandbox. I'll reset the cache manually in the test. Commit that made the tests pass: https://github.com/Logflare/logflare/pull/2985/changes/3f4f4c19127d4dd5ddcddc007f72dcbe2d832d2c

---

**UPDATE:** comparing `connect_args` is unnecessary, we can compare backend configs instead (current one from the cache against the one from when the socket was opened): https://github.com/Logflare/logflare/pull/2985/changes/6efe15e2bc0c0bfd27978086575ae8d6632b49df

lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -378,6 +378,36 @@
               Region of your cluster
             </small>
           </div>
+        <% "tcp" -> %>
```
ruslandoga:
```suggestion
        <% "syslog" -> %>
```

lib/logflare/backends/adaptor/syslog_adaptor/pipeline.ex
Ziinc: lets default to 1 retry which is what all backends are standardizing to

ruslandoga: I only see retries in ClickHouse, but it's tied into some consolidation logic. Is there a simpler adapter I could model retries after? I think I'll go with `handle_failed` + single retry (via re-enqueue) for now.

ruslandoga: I've added retires in https://github.com/Logflare/logflare/pull/2985/changes/447d63f8300f996df149e1ee0c394455b170efed

These tests are quite slow, and can be potentially sped up by providing custom checkout timeout to the pipeline, related to https://github.com/Logflare/logflare/pull/2985#discussion_r2611693765

ruslandoga:
@Ziinc I still want to add one integration-like more test for pipeline retries, but other than that I think it's ready for review! Sorry for the delay!

lib/logflare/backends/adaptor/syslog_adaptor/pipeline.ex
```diff
@@ -0,0 +1,145 @@
...
+      name: name,
+      producer: [
+        module: {Backends.BufferProducer, backend_id: backend.id, source_id: source.id},
+        transformer: {__MODULE__, :transform, []}
+      ],
+      processors: [
+        default: [min_demand: 1]
+      ],
+      batchers: [
+        syslog: [concurrency: 5, batch_size: 50]
```
Ziinc, line 33: would we need a custom batcher to keep it under 64KB perg msg?

test/logflare/backends/adaptor/syslog_adaptor_test.exs
```diff
@@ -0,0 +1,510 @@
...
+                 build(:log_event,
+                   message: "dirty headers",
+                   metadata: %{"app_name" => "My App Name", "host" => "My Host Name 🚀"}
+                 )
+               ],
+               source
+             )
+  end
+
+  test "sanitizes dangerous header values" do
```
Ziinc, line 126: can consider property testing for more comprehensive verification

test/logflare/backends/adaptor/syslog_adaptor_test.exs
```diff
@@ -0,0 +1,510 @@
...
+                 "appname" => "My_App_Name",
+                 "hostname" => "My_Host_Name_____"
+               }
+             }
+           ] =
+             ingest_syslog(
+               [
+                 build(:log_event,
+                   message: "dirty headers",
+                   metadata: %{"app_name" => "My App Name", "host" => "My Host Name 🚀"}
```
Ziinc, line 119: can consider property testing for comprehensiveness

lib/logflare/backends/adaptor/syslog_adaptor/pipeline.ex
ruslandoga: I missed that Syslog messages are supposed to be within 64KB. I think we can truncate them in https://github.com/ruslandoga/logflare/blob/790e271841d71cc09aecacf9d9c312af3333c0f9/lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex#L33-L106

Ziinc: we can have it as a int configuration option like `:max_message_bytes` for truncating, as max message length depends on the receiver. can default to no truncation since RFC does not state an upper limit.

PR #2986, opened 2025-12-02
Restrict ClickHouse pipeline to log events that are within a 72 window

amokan:
As the title implies, this change drops events that are older than 72 hours based on the `LogEvent.body["timestamp"]` attribute to combat the issues with ClickHouse partitions when handling events that are substantially older (_as we've recently seen_).

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
@@ -94,6 +94,8 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
       }
     )
 
+    message_count = length(messages)
+
     result =
       OpenTelemetry.Tracer.with_span :clickhouse_pipeline, %{
         attributes: %{
@@ -108,11 +110,20 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
         backend = Backends.Cache.get_backend(backend_id)
         cutoff_us = System.system_time(:microsecond) - @max_event_age_us
 
-        # only include events that are within our cutoff threshold
         events =
           for %{data: le} <- messages,
               le.body["timestamp"] >= cutoff_us,
               do: le
+
+        event_count = length(events)
+
+        if event_count < message_count do
+          Logger.warning(
+            "Dropping #{message_count - event_count} of #{message_count} ClickHouse event(s) older than 72 hours",
+            source_token: source_token,
+            backend_id: backend_id
+          )
+        end
 
         ClickHouseAdaptor.insert_log_events({source, backend}, events)
       end


lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -103,7 +106,13 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
       } do
         source = Sources.Cache.get_by_id(source_id)
         backend = Backends.Cache.get_backend(backend_id)
-        events = for %{data: le} <- messages, do: le
+        cutoff_us = System.system_time(:microsecond) - @max_event_age_us
+
+        # only include events that are within our cutoff threshold
+        events =
+          for %{data: le} <- messages,
+              le.body["timestamp"] >= cutoff_us,
```
chasers, line 115: We need some logging here to know when things are dropped. 

Either just a Logger so we know on the backend (for now), or (eventually) log into our new system logs table so that Logflare users can see the fact that a message was dropped.

chasers: And we probably don't want a log event for each message which could overflow Logger.

We likely want to know if a batch included dropped logs? And how many?

chasers: Or if logs were dropped while making a batch...

amokan: 👍 I'll make that happen

amokan: Done

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -103,7 +108,22 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
         backend = Backends.Cache.get_backend(backend_id)
-        events = for %{data: le} <- messages, do: le
+        cutoff_us = System.system_time(:microsecond) - @max_event_age_us
+
+        events =
+          for %{data: le} <- messages,
+              le.body["timestamp"] >= cutoff_us,
+              do: le
+
+        event_count = length(events)
```
SteffenDE, line 118: Probably not too big of a difference with batches of size 1500, but if this is a hot path and you want to reduce the amount of times you traverse the lists, you can count the number of events and discarded events while traversing:

```elixir
{events, message_count, discarded_events} =
  for %{data: le} <- messages, reduce: {[], 0, 0} do
    {events, n, discarded_events} ->
      if le.body["timestamp"] >= cutoff_us do
        {[le | events], n + 1, discarded_events}
      else
        {events, n + 1, discarded_events + 1}
      end
  end

events = Enum.reverse(events)

if discarded_events > 0 do
  ...
end
```

Instead of going through messages twice (length + for) and then going over the events again (I assume most of the time those are similarly sized), this only goes over the messages once and then does a reverse on the result.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -24,6 +24,9 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
   @batch_size 1_500
   @max_retries 1
 
+  # 72 hour max event age, based on timestamp
+  @max_event_age_us 72 * 3_600 * 1_000_000
```
SteffenDE, line 28: If it is fine to support only Elixir 1.17+, I like `to_timeout` a lot:

```suggestion
  @max_event_age_us to_timeout(hour: 72) * 1000
```

test/logflare/backends/adaptor/clickhouse_adaptor/pipeline_test.exs
```diff
@@ -115,6 +115,45 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.PipelineTest do
...
+      messages = [
+        %Message{data: old_event, acknowledger: {Pipeline, :ack_id, :ack_data}},
+        %Message{data: recent_event, acknowledger: {Pipeline, :ack_id, :ack_data}}
+      ]
+
+      result = Pipeline.handle_batch(:ch, messages, %{size: 2, trigger: :flush}, context)
+
+      assert result == messages
+
+      Process.sleep(200)
```
SteffenDE, line 141: Is this because Clickhouse needs a bit before the result is visible in the select? If you're fine with just asserting on the insert to be called, mocking out the database with Mox or Mimic might be a cleaner solution, assuming you have some other integration tests as well.

I've seen that the other tests use the same pattern, so I assume you're fine with it and you can of course keep this as is - it just always bothers me when I see sleep calls in tests :D

SteffenDE (commented):
You can reduce the amount of list iterations by one, but overall looks good to me! :)

test/logflare/backends/adaptor/clickhouse_adaptor/pipeline_test.exs
amokan: Thanks for the push! Something I've been meaning to tighten up for a bit now and will do that today.

PR #2988, opened 2025-12-02
feat: display select fields in search results

msmithstubbs:
This PR adds support for displaying fields selected as part of a query in the search results list.

The last nested path key is used as the label, unless an alias is specified. Labels are truncated if necessary.

If a select rule is used the modal link (view, context, etc) are moved underneath the select fields.

CleanShot 2025-12-15 at 10 16 41@2x

They remain inline, after the log event message, when select rules are not used.

CleanShot 2025-12-15 at 10 18 59@2x

Also
* convert SearchLV EEx templates have been converted to HEEx components
* convert live_modal_show_link to a function component and update references.


test/logflare_web/live/search_live/log_event_components_test.exs
```diff
@@ -0,0 +1,63 @@
...
+      }
+
+      html =
+        rendered_to_string(~H"""
+        <LogEventComponents.selected_fields log_event={@log_event} select_fields={@select_fields} />
+        """)
+
+      assert html =~ "metadata.user_id:"
+      assert html =~ "user_123"
+      assert html =~ "metadata.store.city:"
```
Ziinc: should also consider long field paths, in case it execeds timestamp col width, would need to truncate with ...

lib/logflare_web/live/search_live/log_event_components.ex
```diff
@@ -5,14 +5,92 @@ defmodule LogflareWeb.SearchLive.LogEventComponents do
...
+                >
+                  <span>context</span>
+                </.modal_link>
+
+                <.link
+                  class="tw-text-[0.65rem] group-hover:tw-visible tw-invisible"
+                  phx-click={
+                    JS.dispatch("logflare:copy-to-clipboard",
+                      detail: %{
+                        text: "#{LogflareWeb.SearchLive.LogEventComponents.formatted_timestamp(log, assigns[:search_timezone])}    #{log.body["event_message"]}"
```
Ziinc: woudl need to include selected fields, each as a new line.
```
my.field: value

long_field:
.....
`
- inline if <=64 char
- newline if longer

Ziinc (changes requested):
mostly styling adjustments

PR #2997, opened 2025-12-03
Introduce rules cache

bblaszkow06:
The PR adds a separate Rules cache. Change aims at removing the Rules preload and preventing duplicated rules storage with each cached source.

It also fixes 2 problems: 217795b and 99ab681


lib/logflare/context_cache/cache_buster.ex
```diff
@@ -68,7 +70,7 @@ defmodule Logflare.ContextCache.CacheBuster do
         :telemetry.execute([:logflare, :cache_buster, :to_bust], %{count: length(records)})
 
         GenServer.cast(
-          {:via, PartitionSupervisor, {__MODULE__.Supervisor, records}},
+          {:via, PartitionSupervisor, {CacheBusterWorker.Supervisor, records}},
```
Ziinc: oh dear, we'll need some regression checks for this. very big find.

lib/logflare/context_cache/cache_buster.ex
```diff
@@ -84,6 +86,14 @@ defmodule Logflare.ContextCache.CacheBuster do
     {Sources, String.to_integer(id)}
   end
 
+  defp handle_record(%UpdatedRecord{
+         relation: {_schema, "rules"},
+         record: record,
+         old_record: old_record
+       }) do
+    {Rules, handle_rule_record(record) ++ handle_rule_record(old_record)}
```
Ziinc, line 92: the busting works on the primary key of the Rules table

lib/logflare/rules/cache.ex
```diff
@@ -0,0 +1,56 @@
...
+          ]
+        ]
+      }
+    }
+  end
+
+  def list_by_source_id(id), do: apply_repo_fun(__ENV__.function, [id])
+  def list_by_backend_id(id), do: apply_repo_fun(__ENV__.function, [id])
+
+  def bust_by(kw) do
```
Ziinc, line 44: i think that this isn't needed, because the cache buster works by filtering on either the values stored in the cache, checking the struct's pkey or any struct's pkey in a list.
this essentially performs busting by the key while original logic busts by the returned value, which is more comprehensive.

lib/logflare/rules/cache_warmer.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  use Cachex.Warmer
+
+  @impl true
+  def execute(_state) do
+    sources =
+      from(s in Source,
+        where: s.log_events_updated_at >= ago(1, "day"),
+        order_by: {:desc, s.log_events_updated_at},
+        limit: 100,
```
Ziinc:
```suggestion
        limit: 250,
```

lib/logflare/sources.ex
```diff
@@ -507,6 +507,7 @@ defmodule Logflare.Sources do
   def preload_for_dashboard(sources) do
     sources
     |> Enum.map(&preload_defaults/1)
+    |> Enum.map(&preload_rules/1)
```
Ziinc: can consider getitng rid of this as we don't really need it for dashboard usage, except in the relevant rules editing pages (source rules/backend rules)

lib/logflare/context_cache/cache_buster.ex
```diff
@@ -184,11 +194,9 @@ defmodule Logflare.ContextCache.CacheBuster do
 
   defp handle_record(%NewRecord{
          relation: {_schema, "rules"},
-         record: %{"id" => _id, "source_id" => source_id}
-       })
-       when is_binary(source_id) do
-    # When new records are created they were previously cached as `nil` so we need to bust the :not_found keys
-    {Sources, String.to_integer(source_id)}
```
Ziinc, line 192: this was due to the ingest routing using preloading to check, should not use source preloading if we are getting rid of this and fetch it in the routing function instead.

lib/logflare/rules/cache.ex
bblaszkow06: > while original logic busts by the returned value, which is more comprehensive.

That's exactly the problem here.
Firstly, it does nothing better, as the entire entry is removed (not updated) when a rule with a matching primary key is found in the list.
Then, there's the performance: buster for Rules would iterate over every single cache entry (i.e., every active source) and then every rule inside the list (only one entry has the rule, so others will be iterated fully).
This results in roughly linear complexity with the total number of rules. Sounds like something we'd like to eliminate when aiming for 100k rules.

lib/logflare/backends/source_sup_worker.ex
```diff
@@ -33,7 +33,7 @@ defmodule Logflare.Backends.SourceSupWorker do
 
   defp do_check(source) do
     backends = Backends.list_backends(source_id: source.id)
-    rules = Rules.list_rules(source)
+    rules = Rules.Cache.list_rules(source)
```
Ziinc: might be good to do a Stream in the future, so that we don't hold too many rules in the cache.

Ziinc: don't think using cache for this is a good idea since this is a janitor worker to ensure correctness in event of stale data/dropped messages/race conditions

test/logflare/rules/rule_cache_test.exs
```diff
@@ -0,0 +1,87 @@
...
+  use Logflare.DataCase
+
+  alias Logflare.Rules
+  alias Logflare.Sources
+
+  @subject Rules.Cache
+
+  setup do
+    Cachex.reset(@subject)
+    on_exit(fn -> Cachex.reset!(@subject) end)
```
Ziinc: lets move cache resetting to DataCase and ConnCase, there is already some logic there

test/logflare/sources_test.exs
```diff
@@ -219,7 +219,6 @@ defmodule Logflare.SourcesTest do
       sources = Sources.preload_for_dashboard(sources)
 
       assert Enum.all?(sources, &Ecto.assoc_loaded?(&1.user))
-      assert Enum.all?(sources, &Ecto.assoc_loaded?(&1.rules))
```
Ziinc, line 222: lets assert that it isn't loaded so that we don't accidentally shoot ourselves in the foot in the future.

lib/logflare/rules/cache_warmer.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  use Cachex.Warmer
+
+  @impl true
+  def execute(_state) do
+    sources =
+      from(s in Source,
+        where: s.log_events_updated_at >= ago(1, "day"),
+        order_by: {:desc, s.log_events_updated_at},
+        limit: 250,
```
Ziinc:
```suggestion
        where: s.log_events_updated_at >= ago(2, "hour"),
        order_by: {:desc, s.log_events_updated_at},
        limit: 500,
```

Ziinc: can also do a repo-stream so that we can eventually do >1k preloads. in the future.
https://hexdocs.pm/ecto/Ecto.Repo.html#c:stream/2

lib/logflare/context_cache.ex
```diff
@@ -82,6 +83,10 @@ defmodule Logflare.ContextCache do
     {:ok, busted}
   end
 
+  defp bust_key({Rules, kw}) do
+    Rules.Cache.bust_by(kw)
+  end
+
```
Ziinc:
```suggestion
  defp bust_key({context, kw}) when is_list(kw) do
    context_cache = cache_name(context)
    context_cache.bust_by(kw)
  end
```
this would be applicable to almost all caches that receive kw args

lib/logflare/rules/cache.ex
Ziinc: ok.
some follow up PRs:
- lets do a behaviour for the context cache for this optional callback, as per the code suggestion for making it more generalized for other caches
- Let's also add in a comparison benchmark script to verify this assumption on the bust_by vs bust_keys for instances with large number of keys with large lists (many sources with many rules for each)

test/logflare/context_cache/cache_buster_test.exs
```diff
@@ -0,0 +1,36 @@
...
+    change = %DeletedRecord{
+      relation: {"public", "sources"},
+      old_record: %{"id" => Integer.to_string(source_id)}
+    }
+
+    test_pid = self()
+
+    Mimic.expect(ContextCache, :bust_keys, fn arg ->
+      send(test_pid, arg)
+    end)
```
Ziinc, line 35: this does not test the full flow of the key busting.
should set some initial cache keys that are to be busted, then do an expectation that syncing functions are called.

Ziinc (changes requested):

lib/logflare/backends/source_sup_worker.ex
bblaszkow06: I'll remove the cache and replace the query, filtering out non-backend rules

lib/logflare/rules/cache_warmer.ex
bblaszkow06: > repo-stream  
This isn't a straight-forward change as stream doesn't support preload, putting it on a list

test/logflare/context_cache/cache_buster_test.exs
bblaszkow06: The actual key busting by ContextCache is tested in `context_cache_test.exs`. I've added a `call_original` here in mock and test the cache size, but I'm not sure if it's the best way.

The other approach would be mocking the `Cachex.del`, but that would depend on the implementation of ContextCache even further, in particular on the shape of the key.
I'm in favour of what's now in the PR, but I'm open to a second opinion 😉

test/logflare/sources_test.exs
bblaszkow06: Turns out the test was invalid, as `insert_list` preloads the value that it inserts, including the provided user and a default empty list of rules

test/logflare/context_cache/cache_buster_test.exs
Ziinc: ah was referring to test coverage of the bust_by (not the bust_keys/1 which isn't really touched here)

https://github.com/Logflare/logflare/pull/2997/files#diff-d66885b4a65333e731543f8107fca9e3ede1fe3d57dc36e8dfd5b87bc03c2e3dR71-R83

its fine for now to leave it in the rules cache test, but if we start extending it to other caches then it will make sense to ensure that it is covered wholly in a more generic test.

PR #2999, opened 2025-12-03
fix: recent events touch refactor, cache buster worker name reference

Ziinc:
two big fixes:
- fixes partition supervisor name - discovered by @bblaszkow06 🙏 
- Removes GenSingleton usage from non-stable processes

High volume of starting/stopping of SourceSup across the clusters result in broadcast storms across the cluster, with syn getting overloaded with messages of procs starting/stopping and resulting in high message queue counts.

as the conflict resolutino is handled on a singular process on a per-scope basis, backing up of the message queue can result in bootloops, even if we always choose the original process with nanosecond resolution. although clock time discrepancies is a possibility, i am doubtful that it is the root cause for the broadcast storm we are seeing on prod.
Previous partitioning of the recent events scope was done to mitigate this, but removing GenSingleton usage altogether will fix the root problem.

Furthermore, the fact that it starts to occur across all clusters in a synchronized  manner leads me to believe strongly that it is related to the scheduled SourceSup shutdown which happens every half hour across all nodes.

Given above hypothesis, refactoring achieves the following:
- moves ui procs to its own :syn scope
- moves general quantum scheduler to its own global process using GenSingleton
- runs the recent events touch every 5 minutely on at most 500 sources.
- performs at most 5 transactions each 5 minutes when updating sources
- increased auto shutdown interval to hourly on all nodes.

To achieve the global job scheduler, i added in an additional config to Quantum to prevent the startup of the inbuilt Task.Supervisor, so now the global run strategy works without duplicating jobs.  
- Upstream PR has been opened [here](https://github.com/quantum-elixir/quantum-core/pull/662) for the new config option 

Expected outcomes of this PR:
- significant reduction in DB transactions per second
- no message queue buildup on :syn_gen_scope for the :core scope
- less frequent memory reclaiming from SourceSup auto-shutdown
- cross cluster cache busting working again


lib/logflare/scheduler.ex
```diff
@@ -2,5 +2,14 @@ defmodule Logflare.Scheduler do
...
   Quantum scheduler for periodic tasks.
   """
-  use Quantum, otp_app: :logflare
+  use Quantum, otp_app: :logflare, restart: :transient
+
+  @doc """
+  Returns the scheduler :via name used for syn registry.
+  """
+  def scheduler_name do
+    ts = DateTime.utc_now() |> DateTime.to_unix(:nanosecond)
```
josevalim: DateTime.utc_now uses `System.os_time(:nanosecond)` to build the DateTime, so you can avoid the building and conversion by calling it directly:

```suggestion
    ts = System.os_time(:nanosecond)
```

lib/logflare/scheduler/run_strategy_all
josevalim, line 1: Missing the `.ex` extension, so this file likely isn't being compiled.

lib/logflare/sources.ex
```diff
@@ -665,4 +665,48 @@ defmodule Logflare.Sources do
...
+      {:cached, %Source{id: id, log_events_updated_at: log_events_updated_at}} ->
+        if log_events_updated_at < threshold do
+          id
+        else
+          nil
+        end
+
+      _val ->
+        nil
+    end)
```
josevalim: So `nil`s don't propagate down:

```suggestion
    |> Stream.flat_map(fn
      {:cached, %Source{id: id, log_events_updated_at: log_events_updated_at}} ->
        if log_events_updated_at < threshold do
          [id]
        else
          []
        end

      _val ->
        []
    end)
```

lib/logflare/scheduler/run_strategy_all
Ziinc: removed

lib/logflare/sources.ex
Ziinc: ah yes much nicer indeed

PR #3022, opened 2025-12-10
feat: add LQL count distinct operator

Ziinc:
- add countd chart aggregation parsing and encoding support
- implement distinct count handling for chart queries across BigQuery, ClickHouse, and Postgres with coverage in LQL tests
Closes ANL-1133

Fixed a schema tooltip issue with raw html being shown:

Closes ANL-1266

CleanShot 2025-12-17 at 19 00 49@2x


Ziinc:
- [ ] docs

PR #3030, opened 2025-12-12
Filters `LogEvents` before adding to buffer If they are older than 72 hours or 1 hour in the future

amokan:
This takes the logic that was originally implemented in the ClickHouse pipeline and makes it universal for all backends.

- Dumps events older than 72 hours
- Dumps events timestamped more than 1 hour in the future
- Logs drops at the batch level
- Cleans up existing tests that had hardcoded timestamps outside these boundaries

### DIFF
--- a/.github/workflows/docker-base-runner-daily.yml
+++ b/.github/workflows/docker-base-runner-daily.yml
@@ -34,16 +34,16 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: docker context create builders
-      - uses: docker/setup-buildx-action@v2
+      - uses: docker/setup-buildx-action@v3
         with:
           endpoint: builders
       - name: Login to DockerHub
-        uses: docker/login-action@v2
+        uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - id: build
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v6
         with:
           file: ./Dockerfile.base
           push: true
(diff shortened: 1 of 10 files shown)


lib/logflare/backends.ex
```diff
@@ -500,34 +504,66 @@ defmodule Logflare.Backends do
...
+              %LogEvent{} = le ->
+                le
+
+              param ->
+                LogEvent.make(param, %{source: source})
+            end
+            |> maybe_mark_le_dropped_by_lql(source)
+            |> LogEvent.apply_custom_event_message(source)
+
+          case le do
```
SteffenDE: Since the function is quite big already, if you want to reduce it a bit here, I think extracting the upper part into a `preprocess_param` / `param_to_log_event` or however you'd call it and then only do


```suggestion
          case param_to_log_event(param) do
```

could be a good refactor.

lib/logflare/backends.ex
```diff
@@ -500,34 +504,66 @@ defmodule Logflare.Backends do
...
+            le ->
+              {[le | events], errors, total + 1, dropped_old, dropped_future}
+          end
       end
-    end)
+
+    if dropped_old > 0 do
+      Logger.warning(
+        "Dropping #{dropped_old} of #{total} event(s) with timestamp older than 72 hours",
+        source_id: source.id,
```
Ziinc:
```suggestion
        source_id: source.token,
```

lib/logflare/backends.ex
```diff
@@ -500,34 +504,66 @@ defmodule Logflare.Backends do
...
+        source_token: source.token
+      )
+    end
+
+    if dropped_future > 0 do
+      Logger.warning(
+        "Dropping #{dropped_future} of #{total} event(s) with timestamp more than 1 hour in the future",
+        source_id: source.id,
```
Ziinc:
```suggestion
        source_id: source.token,
```

PR #3031, opened 2025-12-12
fix: apply_select_rules should be performed after apply_filter_rules

msmithstubbs:
Ensures that the positional bindings are referenced correctly.

Adds test to verify the correct order.

### DIFF
--- a/test/logflare/logs/search_test.exs
+++ b/test/logflare/logs/search_test.exs
@@ -1,0 +1,101 @@
+defmodule Logflare.Logs.SearchTest do
+  use Logflare.DataCase, async: false
+
+  alias Logflare.Backends.Adaptor.BigQueryAdaptor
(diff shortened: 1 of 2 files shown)


test/logflare/logs/search_test.exs
```diff
@@ -0,0 +1,101 @@
...
+    end)
+
+    [source: source]
+  end
+
+  describe "search_events/1" do
+    test "with nested and top-level field filters", %{source: source} do
+      GoogleApi.BigQuery.V2.Api.Tables
+      |> stub(:bigquery_tables_patch, fn _conn, _project_id, _dataset_id, _table_name, _opts ->
+        {:ok, %{}}
```
Ziinc, line 32: shouldn't need this since we aren't spinning up the Schema module

PR #3038, opened 2025-12-12
chore: add phoenix_test_playwright

nelsonmestevao:
Depends on https://github.com/Logflare/logflare/pull/3036 and https://github.com/Logflare/logflare/pull/3037.

Initial work on ANL-1220.


config/runtime.exs
```diff
@@ -1,6 +1,24 @@
 import Config
+
 alias Logflare.Utils
 
+if config_env() == :test do
+  active? = fn key ->
+    System.get_env(key, "false") in ["t", "true", "1", "yes"]
```
Ziinc:
```suggestion
    System.get_env(key, "false") in ["true", "1"]
```

.envrc
```diff
@@ -0,0 +1,15 @@
+# shellcheck shell=bash
```
Ziinc: is this accidentally committed?

flake.nix
```diff
@@ -0,0 +1,58 @@
+{
```
Ziinc: accidentally committed? we have no plans to use nix

Ziinc: can add to gitignore if usage is preferred.

config/test.exs
```diff
@@ -1,16 +1,16 @@
...
   env: :test,
   cache_stats: true,
+  sql_sandbox: true,
   encryption_key_default: "Q+IS7ogkzRxsj+zAIB1u6jNFquxkFzSrBZXItN27K/Q="
 
+config :logflare, LogflareWeb.Endpoint,
+  url: [host: "localhost", scheme: "http", port: 4001],
+  http: [port: 4001],
+  server: true
+
```
Ziinc, line 13: splitting e2e testing to a separate config that extends/overwrites config/test.exs by importing it would be better, as we don't want to allocate a port for each test run since this limits concurrent testing.
having a separate test-e2e.exs and environment would be better.

lib/logflare_web/endpoint.ex
```diff
@@ -1,15 +1,20 @@
...
+
   @session_options [store: :cookie, key: "_logflare_key", signing_salt: "INPMyhPE"]
 
   socket("/socket", LogflareWeb.UserSocket, websocket: [compress: true])
 
   socket("/logs", LogflareWeb.LogSocket, websocket: [compress: true])
 
-  socket("/live", Logflare.LiveView.Socket,
-    websocket: [connect_info: [session: @session_options], compress: true]
+  socket("/live", Phoenix.LiveView.Socket,
```
Ziinc, line 10: removal of Logflare.LiveView.Socket seems separate to the playwright addition, can we move it to a separate PR please

lib/logflare_web/hooks/allow_test_sandbox.ex
```diff
@@ -0,0 +1,15 @@
+defmodule LogflareWeb.Hooks.LiveAcceptance do
+  import Phoenix.LiveView
+  import Phoenix.Component
+
+  def on_mount(:default, _params, _session, socket) do
```
Ziinc, line 39: needs docs

lib/logflare_web/hooks/live_acceptance.ex
```diff
@@ -0,0 +1,15 @@
+defmodule LogflareWeb.Hooks.LiveAcceptance do
```
Ziinc: LiveAcceptance probably isn't a good module name for what this is doing. 
also, hooks are inherently a liveview thing so we don't need to prepend `Live` to the module name.
Maybe `AllowTestSandbox`?

lib/logflare_web/router.ex
```diff
@@ -16,6 +16,10 @@ defmodule LogflareWeb.Router do
   alias Opentelemetry.Proto.Collector.Metrics.V1.ExportMetricsServiceRequest
   alias Opentelemetry.Proto.Collector.Logs.V1.ExportLogsServiceRequest
 
+  @common_on_mount_hooks if Application.compile_env(:logflare, :sql_sandbox),
+                           do: [LogflareWeb.Hooks.LiveAcceptance],
```
Ziinc: since sql_sandbox is only true during test, we should instead check if it is a `:test` or `:test-e2e` env?  adding a global key for this seems unnecessary 🤔

lib/logflare_web/router.ex
```diff
@@ -207,7 +211,7 @@ defmodule LogflareWeb.Router do
   scope "/endpoints", LogflareWeb do
     pipe_through([:browser, :require_auth])
 
-    live_session :endpoints, on_mount: LogflareWeb.AuthLive do
+    live_session :endpoints, on_mount: @common_on_mount_hooks ++ [LogflareWeb.AuthLive] do
```
Ziinc: we can have a `@dashboard_hooks`

lib/logflare_web/router.ex
```diff
@@ -535,7 +539,7 @@ defmodule LogflareWeb.Router do
     end
   end
 
-  if Mix.env() == :dev do
+  if Application.compile_env(:logflare, :dev_routes) do
```
Ziinc, line 543:
```suggestion
  if Application.get_env(:logflare, :env) == :dev do
```

Ziinc (changes requested):

lib/logflare_web/router.ex
nelsonmestevao: Configuring options through keys is the way the phoenix generators are following now, so that is why I am doing the same here, that's how they suggest in the guides https://hexdocs.pm/phoenix_ecto/Phoenix.Ecto.SQL.Sandbox.html#module-example.

I think one advantage is it easier to follow and make any environment configure it at any time, example the dev_routes are useful for development but might also be nice in e2e test to confirm an email is received in the browser. However, checking in the env can also work. Let me know if you strong prefer that approach :pray:

lib/logflare_web/router.ex
nelsonmestevao: I got that from the new phoenix generators that are using that. 

But the idea was initiated in this comment https://github.com/Logflare/logflare/pull/3038#discussion_r2657614256.

lib/logflare_web/hooks/live_acceptance.ex
nelsonmestevao: I used the name from the guide https://hexdocs.pm/phoenix_ecto/Phoenix.Ecto.SQL.Sandbox.html#module-acceptance-tests-with-liveviews. 

I will change the name for your suggestion :pray:

nelsonmestevao: Done in 9d4e101f3e1c76445b96b47d7ce8a81224cc502b :heavy_check_mark:

lib/logflare_web/router.ex
nelsonmestevao: For AuthLive right?

nelsonmestevao: Done in 0d21a7b34d141d867e0cc5903c31baa2d73f5723 :heavy_check_mark:

flake.nix
nelsonmestevao: Done in 2f0656bc3f1ee6f479de10c4a20f13faef119e15 :heavy_check_mark:

.envrc
nelsonmestevao: Done in 2f0656bc3f1ee6f479de10c4a20f13faef119e15 :heavy_check_mark:

lib/logflare_web/endpoint.ex
nelsonmestevao: True, done in https://github.com/Logflare/logflare/pull/3064 :heavy_check_mark:

lib/logflare_web/router.ex
Ziinc: i see, ok lets go with additional keys then. my only gripe with this approach is the way that defaults should be handled, having defaults for a configuration flag all around application code is not desirable.

having all default keys declared config.exs and then overriding it on a per env is a fine approach to avoid the above, but human error might occur.

config/test.exs
nelsonmestevao: I split it [here](https://github.com/Logflare/logflare/pull/3038/changes#diff-e43c5cbf91db7a8062b6cb860cbf118296c1b4c7ee32fdcf702e54234ba38092R9). I tried a custom configuration file, but it could not be done for testing without using MIX_ENV=test.

PR #3041, opened 2025-12-15
Cache benchmark and ContextCache behaviour

bblaszkow06:


test/profiling/context_cache_bust_by_kw_bench.exs
```diff
@@ -0,0 +1,122 @@
...
+
+        {_k, _v} ->
+          true
+      end)
+      |> Enum.reduce(0, fn {k, _v}, acc ->
+        Cachex.del(cache, k)
+        acc + 1
+      end)
+    end,
+    "bust_keys by key with Cachex.execute" => fn [{source, _rule} | _] ->
```
Ziinc: we can slim down the benchmark to just this case, since it is already implemented and the only one that we really care about for busting kw lists. we just need this for the baseline.

Ziinc: Also, instead of benchmkaring the internal logic, lets instead call the `bust_keys/1` directly to simulate this.
Can add in a 2nd opts arg to it to allow flagging of different code blocks for different benchmarks for comparison.
doing internal logic benchmarks like this makes it hard to maintain as things change.

test/profiling/context_cache_bust_keys_bench.exs
```diff
@@ -1,40 +1,88 @@
...
+           {:==, {:map_get, :id, {:element, 2, :value}}, pkey}}
+        }
+
+      query = Cachex.Query.build(where: filter, output: {:key, :value})
+
+      Sources.Cache
+      |> Cachex.stream!(query)
+      |> Stream.filter(fn
+        {_k, {:cached, v}} when is_list(v) ->
+          Enum.any?(v, &(&1.id == pkey))
```
Ziinc: this looks like it is redundant since ets filter already catches this

test/profiling/context_cache_bust_keys_bench.exs
```diff
@@ -1,40 +1,88 @@
...
-          acc + 1
+        {_k, _v} ->
+          true
+      end)
+      |> Enum.reduce(0, fn {k, _v}, acc ->
+        Cachex.del(Sources.Cache, k)
+        acc + 1
+      end)
+    end,
+    "bust_keys with ETS filter under Cachex.execute" => fn [source | _] = sources ->
```
Ziinc: as mentioned above, lets used an `opts` arg to allow flagging for different code block changes to make the benchmarks more streamlined and easy to revert.

Ziinc (changes requested):
benchmark needs tweaking, but Cachex.execute/2 optimization looks good

test/profiling/context_cache_bust_keys_bench.exs
bblaszkow06: ETS filter cannot browse lists (and that's the main reason why it is so slow), so this is needed

test/profiling/context_cache_bust_by_kw_bench.exs
bblaszkow06: I've managed to avoid the opts arg while keeping 2 cases and the benchmark result.

If you think it's better to make it even slimmer, let me know

test/profiling/context_cache_bust_keys_bench.exs
bblaszkow06: I'm not convinced that using the `opt` arg and putting dead code with flags and ifs is an elegant solution.

Here, I simply left only the baseline benchmark. In general IMO it would be best to optionally pass a module with an alternative implementation of some function. The module might then be only defined in benchmark, somewhat like it was here https://github.com/bblaszkow06/logflare/blob/5ba6f106ae682bc8a83b94be5fa2376e040005e5/test/profiling/context_cache_bust_by_kw_bench.exs

PR #3047, opened 2025-12-16
perf: optimise loading saved searches

msmithstubbs:
This PR improves performance of loading saved searches in the dashboard.

* saved searches are loaded once, rather than N+1 queries by source
* Adds SavedSearches.Cache - assuming saved searches aren't updated regularly so should be a good candidate for caching
* Adds composite index to saved_searches table to exclude `saved_by_user: false` rows

Follow on PR will move saved searches from dashboard to modal, which will still benefit from these improvements.

Part of ANL-1178

### DIFF
--- a/lib/logflare/saved_searches/saved_searches.ex
+++ b/lib/logflare/saved_searches/saved_searches.ex
@@ -51,15 +51,6 @@ defmodule Logflare.SavedSearches do
   end
 
   @doc """
-  Completely deletes a saved search.
-  TODO: remove, unused.
-  """
-  @spec delete(SavedSearch.t()) :: {:ok, SavedSearch.t()}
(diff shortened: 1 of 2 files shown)


lib/logflare/saved_searches/cache.ex
```diff
@@ -0,0 +1,53 @@
...
+      id: __MODULE__,
+      start:
+        {Cachex, :start_link,
+         [
+           __MODULE__,
+           [
+             hooks:
+               [
+                 if(stats, do: Utils.cache_stats()),
+                 Utils.cache_limit(100_000)
```
Ziinc, line 20:
```suggestion
                 Utils.cache_limit(10_000)
```

PR #3054, opened 2025-12-19
perf: exclude system source logging from metrics

amokan:
This is a very naive assessment, so tell me if I'm missing something here @Ziinc.

One thing I noticed when digging through the user monitoring/system source stuff is that it appears we end up processing system source metrics through the same filter process in `Logflare.Backends.UserMonitoring.log_interceptor/1`. I get the 60 second aggregation window prevents this from going forever, but it seems like we could prevent this altogether with an extra metadata key indicating that it is a system source log.

### DIFF
--- a/lib/logflare/backends/user_monitoring.ex
+++ b/lib/logflare/backends/user_monitoring.ex
@@ -128,7 +128,8 @@ defmodule Logflare.Backends.UserMonitoring do
   Intercepts Logger messages related to specific users, and send them to the respective
   System Source when the user has activated it
   """
-  def log_interceptor(%{meta: %{system_source: true}} , _), do: :ignore
+  def log_interceptor(%{meta: %{system_source: true}}, _), do: :ignore
+
   def log_interceptor(%{meta: %{user_id: user_id} = meta} = log_event, _)
       when is_integer(user_id) do
     with %{system_monitoring: true} <- Users.Cache.get(user_id),
--- a/lib/logflare/backends/adaptor/bigquery_adaptor.ex
+++ b/lib/logflare/backends/adaptor/bigquery_adaptor.ex
@@ -55,7 +55,12 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
     project_id = backend.config.project_id
     dataset_id = backend.config.dataset_id
     # TODO: remove source_id metadata to reduce confusion
-    Logger.metadata(source_id: source.token, source_token: source.token, user_id: user.id, system_source: source.system_source)
+    Logger.metadata(
+      source_id: source.token,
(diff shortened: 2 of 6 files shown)


lib/logflare/backends/user_monitoring.ex
```diff
@@ -133,6 +135,8 @@ defmodule Logflare.Backends.UserMonitoring do
   Intercepts Logger messages related to specific users, and send them to the respective
   System Source when the user has activated it
   """
+  def log_interceptor(%{meta: %{system_source: true}}, _), do: :ignore
+
   def log_interceptor(%{meta: meta} = log_event, _) do
```
Ziinc:
```suggestion
  def log_interceptor(%{meta: %{user_id: _} = meta} = log_event, _) do
```
this might be a simpler way to do it, we would save on a cache lookup.

lib/logflare/backends/user_monitoring.ex
```diff
@@ -133,6 +135,8 @@ defmodule Logflare.Backends.UserMonitoring do
   Intercepts Logger messages related to specific users, and send them to the respective
   System Source when the user has activated it
   """
+  def log_interceptor(%{meta: %{system_source: true}}, _), do: :ignore
```
Ziinc, line 131: not surfacing errors to users for system sources is fine i think 👍

lib/logflare/backends/adaptor/http_based/pipeline.ex
```diff
@@ -76,7 +77,8 @@ defmodule Logflare.Backends.Adaptor.HttpBased.Pipeline do
         "source_id" => context[:source_id],
         "source_uuid" => context[:source_token],
         "backend_id" => context[:backend_id],
-        "backend_uuid" => context[:backend_token]
+        "backend_uuid" => context[:backend_token],
+        "system_source" => context[:system_source]
```
Ziinc, line 81: will need to also modify the extract_tags/2 to exclude this field from the attributes

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -146,7 +148,11 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
 
   @spec handle_message(any, Broadway.Message.t(), any) :: Broadway.Message.t()
   def handle_message(_processor_name, message, context) do
-    Logger.metadata(source_id: context.source_token, source_token: context.source_token)
+    Logger.metadata(
+      source_id: context.source_token,
```
Ziinc, line 153:
```suggestion
      user_id: context.user_id,
      source_id: context.source_token,
```

PR #3059, opened 2025-12-22
feat: sync alert jobs

ruslandoga:
- ANL-1154 -- Periodically perform a Alerts scheduler sync, every 60 minutes
- ANL-1156 -- Fetch the alert definition just before executing the query, to avoid using stale data -- this is done by storing and using `alert_query_id` instead of `%AlertQuery{}` struct in the Quantum job.

### DIFF
--- a/lib/logflare/alerting.ex
+++ b/lib/logflare/alerting.ex
@@ -241,22 +241,18 @@ defmodule Logflare.Alerting do
 
   defp do_sync_alert_jobs do
     wanted_jobs = init_alert_jobs()
+    wanted_jobs_set = MapSet.new(wanted_jobs, & &1.name)
     current_jobs = AlertsScheduler.jobs()
 
-    wanted_jobs_lookup = Map.new(wanted_jobs, &{&1.name, &1})
-    current_jobs_lookup = Map.new(current_jobs)
-
-    # Delete jobs that are no longer in the DB
-    Enum.each(current_jobs_lookup, fn {name, _job} ->
-      if not Map.has_key?(wanted_jobs_lookup, name) do
+    # Delete jobs that are no longer wanted
+    Enum.each(current_jobs, fn {name, _job} ->
+      if not MapSet.member?(wanted_jobs_set, name) do
         AlertsScheduler.delete_job(name)
       end
     end)
 
     # Upsert all wanted jobs
-    Enum.each(wanted_jobs_lookup, fn {name, job} ->
-      AlertsScheduler.add_job(job)
-    end)
+    Enum.each(wanted_jobs, &AlertsScheduler.add_job/1)
   end
 
   @doc """


lib/logflare/alerting.ex
```diff
@@ -268,8 +268,17 @@ defmodule Logflare.Alerting do
 
   Send notifications if necessary configurations are set. If no results are returned from the query execution, no alert is sent.
   """
-  @spec run_alert(AlertQuery.t(), :scheduled) ::
+  @spec run_alert(AlertQuery.t() | integer(), :scheduled) ::
           :ok | {:error, :not_enabled} | {:error, :below_min_cluster_size}
+  def run_alert(alert_id, :scheduled) when is_integer(alert_id) do
+    # sync the alert job for the next run
+    sync_alert_job(alert_id)
```
Ziinc: would be slightly better to return the alert job (if present) as an :ok tuple from the sync job, then that would allow us to avoid a 2nd query to re-fetch the alert job.

ruslandoga: Which second query do you mean? The `run_alert/2` below seems to operate on `AlertQuery`, not `Quantum.Job`, and the job only knows about `alert_id` now (since we don't really need to put `%AlertQuery{}` in the job anymore if we refetch the alert definition each time).

Ziinc: sync_alert_job would perform 1 db query on the scheduler_node, then perform another db query at get_alert_query_by at line 285, so that would result in 2 db queries being performed.

alternative is to fetch the alert_query first then pass it to sync_alert_job, reversing the order and reducing db queries to 1

test/logflare/alerting_test.exs
```diff
@@ -408,10 +408,62 @@ defmodule Logflare.AlertingTest do
...
+                 task: {Logflare.Alerting, :run_alert, [^alert_id, :scheduled]}
                } = Alerting.get_alert_job(alert_id)
       end)
     end
+
+    test "run_alert/2 uses fresh alert query data", %{user: user} do
+      # ensure config allows execution
+      old_config = Application.get_env(:logflare, Logflare.Alerting)
+      Application.put_env(:logflare, Logflare.Alerting, min_cluster_size: 0, enabled: true)
+      on_exit(fn -> Application.put_env(:logflare, Logflare.Alerting, old_config) end)
```
Ziinc: should put this in a common setup

test/logflare/alerting_test.exs
```diff
@@ -408,10 +408,62 @@ defmodule Logflare.AlertingTest do
...
+      expect(GoogleApi.BigQuery.V2.Api.Jobs, :bigquery_jobs_query, 1, fn _conn, _proj_id, opts ->
+        assert opts[:body].query =~ "unique_fresh_data_check"
+        {:ok, TestUtils.gen_bq_response([%{"testing" => "123"}])}
+      end)
+
+      Alerting.run_alert(alert.id, :scheduled)
+    end
+
+    test "run_alert/2 unschedules job if alert is missing", %{user: user} do
+      {:ok, alert} =
```
Ziinc, line 447: this is missing the configuration setup, if config is disabled then this would pass without verifying the logic.

ruslandoga: Done in https://github.com/Logflare/logflare/pull/3059/changes/55fc12256712f7aba326bc5be5ac5f227bc8e2ab

test/logflare/alerting_test.exs
ruslandoga: Done in https://github.com/Logflare/logflare/pull/3059/changes/55fc12256712f7aba326bc5be5ac5f227bc8e2ab

lib/logflare/alerting.ex
```diff
@@ -397,11 +416,19 @@ defmodule Logflare.Alerting do
...
+      pid when is_pid(pid) ->
+        pid
+        |> node()
+        |> Cluster.Utils.rpc_call(func)
+
+      {_name, node} ->
+        Cluster.Utils.rpc_call(node, func)
+
+      nil ->
+        raise "Alerting scheduler node not found"
```
ruslandoga, line 431: Previously that function was able to silently be a no-op if scheduler node wasn't found, I wonder if this `raise` is a good idea? It would make these cases more noticeable, if they ever happen.

lib/logflare/alerting.ex
ruslandoga: Do we actually need to call `sync_alert_job` which "re-adds" the job from `run_query`, or is it enough to unschedule the job for alerts that no longer exist, i.e., https://github.com/Logflare/logflare/pull/3059/changes/4519e145038437f2881dae91f5cfc49630605b0c. Or maybe it can just be no-op and sync would then be done periodically (every 60 minutes), but that has another problem: https://github.com/Logflare/logflare/pull/3059#discussion_r2668660044

Re-adding the job from `run_query/1` feels off for cron jobs somehow, ~~but in light of https://github.com/Logflare/logflare/pull/3059#discussion_r2668660044 it might be the only way to reliably re-sync active jobs.~~ One possible problem with that approach is when an rare alert gets modified to become less rare, the update won't take place until it runs at least once which might be surprising.

config/config.exs
```diff
@@ -173,6 +173,11 @@ config :logflare, Logflare.Scheduler,
       run_strategy: Quantum.RunStrategy.Local,
       schedule: "*/15 * * * *",
       task: {Logflare.Sources, :recent_events_touch, []}
+    ],
+    alerts_scheduler_sync: [
+      run_strategy: Quantum.RunStrategy.Local,
+      schedule: "0 * * * *",
```
ruslandoga, line 179: I wonder if it's possible for alert schedule be rarer than alerts_scheduler_sync schedule, less than once in 60 minutes? That would probably mean that those jobs would never run, since in the current `do_sync_alert_jobs` they would keep getting re-added (which in Quantum doesn't seem (?) to execute the job right away but kind of reschedules and effectively postpones them indefinitely).

lib/logflare/alerting.ex
```diff
@@ -243,23 +249,18 @@ defmodule Logflare.Alerting do
...
+      :ok = AlertsScheduler.add_job(job)
     else
-      # alert query does not exist, maybe remove from scheduler
-      job = AlertsScheduler.find_job(to_job_name(alert_id))
-
-      if job do
-        AlertsScheduler.delete_job(job.name)
-      end
+      AlertsScheduler.delete_job(to_job_name(alert_id))
+      {:error, :not_found}
```
ruslandoga, line 273: It's not an error though, since sync completed as intended. More like `{:ok, :removed}` or `:ok` or `:removed` maybe?

config/config.exs
Ziinc: Yes there are jobs that run on a minutely basis, or every hour

Ziinc: Hmm that behaviour would not be good. Perhaps separate sync schedule jobs, syncing once a day for hourly schedules and once a minute for jobs less than an hour

lib/logflare/alerting.ex
ruslandoga: I think I'd like to go with no syncing in `run_alert/1` since it already uses the up-to-date alert query thanks to fetching it by id on each run (and being no-op if it doesn't exist), and try and update `sync_alert_jobs` to be a bit smarter (instead of "full" delete followed by "full" insert) to avoid the problem in https://github.com/Logflare/logflare/pull/3059#discussion_r2668660044

---

**UPDATE:** done.

config/config.exs
ruslandoga: Actually, I was wrong and misunderstood how Quantum works. The original "wipe and re-add" approach is safe for all but the jobs coinciding with `alerts_scheduler_sync` schedule ("0 * * * *") so I think it still makes sense to make `do_sync_alert_jobs` a bit smarter. Right now I'm going with something like this

```elixir
  defp do_sync_alert_jobs do
    wanted_jobs = init_alert_jobs()
    wanted_jobs_set = MapSet.new(wanted_jobs, & &1.name)
    current_jobs = AlertsScheduler.jobs()

    # Delete jobs that are no longer wanted
    Enum.each(current_jobs, fn {name, _job} ->
      if not MapSet.member?(wanted_jobs_set, name) do
        AlertsScheduler.delete_job(name)
      end
    end)

    # Upsert all wanted jobs
    Enum.each(wanted_jobs, &AlertsScheduler.add_job/1)
  end
```

lib/logflare/alerting.ex
```diff
@@ -209,9 +209,15 @@ defmodule Logflare.Alerting do
   """
   @spec create_alert_job_struct(AlertQuery.t()) :: Quantum.Job.t()
   def create_alert_job_struct(%AlertQuery{} = alert_query) do
+    %AlertQuery{id: alert_query_id, cron: cron} = alert_query
+
+    if is_nil(alert_query_id) do
+      raise ArgumentError, "AlertQuery is missing id: #{inspect(alert_query)}"
+    end
+
```
Ziinc, line 217:
```suggestion
def create_alert_job_struct(%AlertQuery{id: alert_query_id, cron: cron} = alert_query) when alert_query_id != nil do
```
its fine for the match to fail since it effectively will result in the same error raising. the nice error message is a a nice to have but not really necessary since it is not expected behaviour.

lib/logflare/alerting.ex
Ziinc, line 431:
```suggestion
```

Ziinc: we don't need to explicitly raise since this is unexpected behaviour and the case matcherrorr will be sufficient for us to pinpoint the issue. we should always have a global scheduler present.

Ziinc:
@ruslandoga can you do a follow up PR for the changes 🙏 changes look great, thanks for digging into the scheduler behaviour

PR #3075, opened 2026-01-08
Ingest rules optimization

bblaszkow06:


lib/logflare/context_cache/cache_buster.ex
```diff
@@ -330,7 +330,7 @@ defmodule Logflare.ContextCache.CacheBuster do
   end
 
   defp handle_rule_record(record) do
```
Ziinc: Match directly and direct list construction would be faster

lib/logflare/sources/source_router/rules_tree.ex
```diff
@@ -0,0 +1,145 @@
...
+
+  defp accumulate([], acc), do: acc
+  defp accumulate([h | tail], acc), do: accumulate(tail, accumulate(h, acc))
+
+  defp accumulate({rule_id, _filters_num}, acc) when is_map_key(acc, rule_id),
+    do: %{acc | rule_id => acc[rule_id] - 1}
+
+  defp accumulate({rule_id, filters_num}, acc), do: Map.put(acc, rule_id, filters_num - 1)
+
+  defp stringify(v) when is_integer(v) do
```
Ziinc: Utils has a stringify function, i think

lib/logflare/sources/source_router/rules_tree.ex
```diff
@@ -0,0 +1,145 @@
...
+
+  defp stringify(v) when is_float(v) do
+    Float.to_string(v)
+  end
+
+  defp stringify(v) when is_binary(v), do: v
+  defp stringify(v), do: inspect(v)
+
+  # Groups all the rules associated with source by path in a tree
+  def build(rules) do
```
Ziinc, line 179: this might be quite expensive, should benchmark this as well to see what kind of ips we are looking at.

lib/logflare/sources/source_router/rules_tree.ex
```diff
@@ -0,0 +1,145 @@
...
+    |> match_cond(rule_ids, acc)
+  end
+
+  defp match_rule(le_value, {operator, expected}, {:route, rule_ids}, acc) do
+    check_op(operator, le_value, expected)
+    |> match_cond(rule_ids, acc)
+  end
+
+  defp check_op(operator, le_value, expected) do
+    case operator do
```
Ziinc, line 89: some of these should be separate function heads as guards

lib/logflare/sources/source_router/rules_tree.ex
```diff
@@ -0,0 +1,145 @@
...
+  defp match_cond(true, rule_ids, acc), do: accumulate(rule_ids, acc)
+  defp match_cond(false, _rule_ids, acc), do: acc
+
+  defp accumulate([], acc), do: acc
+  defp accumulate([h | tail], acc), do: accumulate(tail, accumulate(h, acc))
+
+  defp accumulate({rule_id, _filters_num}, acc) when is_map_key(acc, rule_id),
+    do: %{acc | rule_id => acc[rule_id] - 1}
+
+  defp accumulate({rule_id, filters_num}, acc), do: Map.put(acc, rule_id, filters_num - 1)
```
Ziinc: not the easiest to read and understand, this module would need some docs for clarity and posterity.

lib/logflare/sources/source_router/rules_tree.ex
```diff
@@ -0,0 +1,145 @@
...
+
+  @impl true
+  def matching_rules(event, source) do
+    rule_set = Rules.Cache.rules_tree_by_source_id(source.id)
+
+    matching_rule_ids(event, rule_set)
+    |> Enum.flat_map(fn
+      {id, 0} -> [id]
+      {_id, _matches_left} -> []
+    end)
```
Ziinc, line 46: instead of flatmapping to get the keys, matching_rule_ids/2 can return a tuple instead. might not even need to return  the map, maybe just list of IDs would do, then we can avoid the iteration entirely

lib/logflare/sources/source_router/rules_tree.ex
Ziinc: also will need some docs + illustrative code examples on how the rules tree structure.

something to consider is also that Iteraptor has a to_flatmap function, which might be an easier to work with without having much nesting

lib/logflare/sources/source_router/rules_tree.ex
bblaszkow06: Benchmarked that with the implementation below, and there was no difference (44.02 K vs 43.94 K), so only readability matters. Which version do you think is better?
```elixir
  defp check_op(nil, _le_value, _expected), do: false
  defp check_op(:=, le_value, expected) when le_value == expected, do: true
  defp check_op(:<, le_value, expected) when le_value < expected, do: true
  defp check_op(:>, le_value, expected) when le_value > expected, do: true
  defp check_op(:<=, le_value, expected) when le_value <= expected, do: true
  defp check_op(:>=, le_value, expected) when le_value >= expected, do: true

  defp check_op(:range, le_value, {lvalue, rvalue})
       when le_value >= lvalue and le_value <= rvalue, do: true

  defp check_op(:list_includes, le_value, expected), do: expected in le_value

  defp check_op(:list_includes_regexp, le_value, expected),
    do: stringify(le_value) =~ ~r/#{expected}/u

  defp check_op(:"~", le_value, expected), do: stringify(le_value) =~ ~r/#{expected}/u

  defp check_op(:string_contains, le_value, expected),
    do: String.contains?(stringify(le_value), stringify(expected))

  defp check_op(_op, _le_value, _expected), do: false
```

lib/logflare/context_cache/cache_buster.ex
bblaszkow06: Cache buster isn't called during ingestion 😉

lib/logflare/sources/source_router/rules_tree.ex
bblaszkow06: I wanted to ensure it can be inlined as a private function, but it doesn't seem to matter after benchmarking, so removing that

lib/logflare/sources/source_router/rules_tree.ex
bblaszkow06: That's what I had initially. The problem is that to match a rule, all FilterRules must match. I started by returning a list and calculating the number of matches, but it was slower for a large number of matches and roughly the same for a few matches.

A problematic case is the nested objects with a list, as one FilterRule may match multiple times, breaking my approach 😢 I will need to fix that next week, I'm thinking of using bit flags

lib/logflare/sources/source_router/rules_tree.ex
```diff
@@ -0,0 +1,145 @@
...
+      target = {rule.id, filters_num}
+
+      reverse_path = String.split(filter.path, ".") |> Enum.reverse()
+
+      reverse_path
+      |> Enum.reduce(%{to_command(filter) => {:route, target}}, fn k, acc ->
+        %{{:get, k} => acc}
+      end)
+    end
+    |> Enum.reduce(&deep_merge/2)
```
bblaszkow06: note: conversion to lists instead of maps cuts down memory usage and slightly increases performance

lib/logflare/sources/source_router/rules_tree.ex
bblaszkow06: > something to consider is also that Iteraptor has a to_flatmap function, which might be an easier to work with without having much nesting

The nesting is intended as an optimization, grouping all the paths in rules to ensure each key is accessed at most once. I've tried using iteraptor to generate this data structure via `from_flatmap`, but it tries to convert keys to existing atoms without an option to disable that behaviour.

lib/logflare/context_cache/cache_buster.ex
Ziinc: Yea, regardless any small improvement adds up since cache buster is still considered a hot code path

Ziinc: Yea, regardless any small improvement adds up since cache buster is still considered a hot code path

lib/logflare/sources/source_router/rules_tree.ex
Ziinc: Ok let me know if forking the repo to get that change in is needed, we can do that if it is simpler

lib/logflare/sources/source_router/rules_tree.ex
Ziinc: I see we can keep if there is no meaningful difference then

Ziinc: I see we can keep if there is no meaningful difference then

PR #3077, opened 2026-01-08
Consolidated Ingest - Phase 2

amokan:
Adds dual-key support to `IngestEventQueue` for consolidated ingest.

_Note that this branch is currently based off of #3076_

### DIFF
--- a/assets/js/source.js
+++ b/assets/js/source.js
@@ -186,5 +186,6 @@ window.Source = {
   initLogsUiFunctions,
   trackScroll,
   scrollBottom,
-  scrollOverflowBottom
+  scrollOverflowBottom,
+  switchDateFormat
 }


lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -125,10 +170,81 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+          source_backend_pid()
+          | queues_key()
+          | consolidated_queues_key()
+          | consolidated_table_key(),
+          [LogEvent.t()],
+          Keyword.t()
+        ) :: :ok | {:error, :not_initialized}
   def add_to_table(sid_bid_or_sid_bid_pid, batch, opts \\ [])
 
+  def add_to_table({:consolidated, bid} = key, batch, opts) when is_integer(bid) do
```
Ziinc, line 183: for this new func, let's just implement the optimized version (check_queue_size=true, no_get_tid=true, chunk_size=100)

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -125,10 +170,81 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+    chunk_size = Keyword.get(opts, :chunk_size, 100)
+    no_get_tid = Keyword.get(opts, :no_get_tid, true)
+    check_queue_size = Keyword.get(opts, :check_queue_size, true)
+    startup_queue = {:consolidated, bid, nil}
+
+    reducer =
+      if check_queue_size do
+        fn
+          {{{:consolidated, _, nil}, _tid}, _}, acc -> acc
+          {_obj, count}, acc when count >= @max_queue_size -> acc
```
Ziinc, line 193: i think consolidated queues need to have a much larger max count.

In current state, each queue would at most have 15k events (we're gonna revert the 30k), so events in queue for the backend would be 15k*cores , which isn't actually a lot and will lead to a lot of load shedding. 

maybe something large like 150k

amokan: Was intentionally treating these early PRs more like placeholders for some of these items. I can make these adjustments now or handle them with some benchmarking in PR 5 or 6 of the series before ClickHouse is actually enabled as a consolidated ingest backend.

amokan:
@Ziinc I am working on a new branch based off this one for PR 3 that will include the queue safety / OOM prevention items you list above. Hopefully that is cool with you. We'll have another 3-4 PRs after that to iron out additional parameter tweaks and so-on.

Ziinc (approved):

PR #3079, opened 2026-01-09
feat: Display saved searches in modal

msmithstubbs:
* Adds a modal to the search page to list and delete saved searches.
* Saved searches now appear in a modal for sources in the dashboard

https://github.com/user-attachments/assets/ebaa4b39-3d10-4dd2-a4f0-e08e93459f6c

ANL-1178

### DIFF
--- a/lib/logflare_web/live/dashboard_live/dashboard_source_components.ex
+++ b/lib/logflare_web/live/dashboard_live/dashboard_source_components.ex
@@ -4,6 +4,8 @@ defmodule LogflareWeb.DashboardLive.DashboardSourceComponents do
   use Phoenix.Component
 
   alias Logflare.Sources.Source
+  alias LogflareWeb.ModalLiveHelpers
+  alias LogflareWeb.SearchLive.SavedSearchesModalComponent
   alias Phoenix.LiveView.JS
 
   attr :source, Source, required: true
@@ -22,6 +24,9 @@ defmodule LogflareWeb.DashboardLive.DashboardSourceComponents do
       </div>
       <div>
         <div class="float-right">
+          <ModalLiveHelpers.modal_link component={SavedSearchesModalComponent} modal_id={:"saved-searches-#{@source.id}"} title="Saved Searches" class="dashboard-links tw-mr-2" phx-value-source-id={@source.id}>
+            <i class="fas fa-bookmark"></i>
+          </ModalLiveHelpers.modal_link>
           <.team_link href={~p"/sources/#{@source}/edit"} team={@team} class="dashboard-links">
             <i class="fas fa-edit"></i>
           </.team_link>
--- a/lib/logflare_web/live/search_live/logs_search_lv.ex
+++ b/lib/logflare_web/live/search_live/logs_search_lv.ex
@@ -234,7 +234,7 @@ defmodule LogflareWeb.Source.SearchLV do
       <:path>
         ~/logs/<.team_link team={@team} href={~p"/sources/#{@source}"} class="text-primary">{@source.name}</.team_link>/search
       </:path>
-      <SubheadComponents.subhead_actions user={@user} search_timezone={@search_timezone} search_op_error={@search_op_error} search_op_log_events={@search_op_log_events} search_op_log_aggregates={@search_op_log_aggregates} />
+      <SubheadComponents.subhead_actions user={@user} source={@source} search_timezone={@search_timezone} search_op_error={@search_op_error} search_op_log_events={@search_op_log_events} search_op_log_aggregates={@search_op_log_aggregates} />
     </.subheader>
     <div class="container source-logs-search-container console-text">
       <div id="logs-list-container">
--- a/lib/logflare_web/live/search_live/saved_searches_modal_component.ex
+++ b/lib/logflare_web/live/search_live/saved_searches_modal_component.ex
@@ -11,7 +11,12 @@ defmodule LogflareWeb.SearchLive.SavedSearchesModalComponent do
 
   @spec update(map(), Phoenix.LiveView.Socket.t()) :: {:ok, Phoenix.LiveView.Socket.t()}
   def update(assigns, socket) do
-    socket = assign(socket, assigns)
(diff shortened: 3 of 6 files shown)


lib/logflare_web/live/search_live/saved_searches_modal_component.ex
```diff
@@ -0,0 +1,128 @@
...
+  @spec handle_event(String.t(), map(), Phoenix.LiveView.Socket.t()) ::
+          {:noreply, Phoenix.LiveView.Socket.t()}
+  def handle_event("delete_saved_search", %{"id" => saved_search_id}, socket) do
+    user = socket.assigns.user
+
+    socket =
+      with %Logflare.SavedSearch{} = saved_search <- SavedSearches.get(saved_search_id),
+           true <- Sources.get_by_user_access(user, saved_search.source_id) |> is_struct(),
+           {:ok, _response} <- SavedSearches.delete_by_user(saved_search) do
+        _ = SavedSearches.Cache.bust_by(source_id: saved_search.source_id)
```
msmithstubbs: I found busting the cache directly was required. The ContextCache buster wasn't clearing the cached list in time for the UI to fetch it.

Ziinc:
@msmithstubbs instead of a consolidated modal on the dashboard, I think a separate modal icon button on the left of the icon button for each source will be much better, so that users can open up the same search modal and go to a saved search quickly.
Not a fan of the full consolidated list, too messy and noisy.
Let's do the icon modal in the same PR.

lib/logflare_web/live/search_live/saved_searches_modal_component.ex
Ziinc: Would be better to do the busting in the context instead.

Ziinc: Would be better to do the busting in the context instead.

lib/logflare_web/live/search_live/subhead_components.ex
```diff
@@ -27,6 +27,11 @@ defmodule LogflareWeb.SearchLive.SubheadComponents do
             <span class="hide-on-mobile">scroll down</span>
           </a>
         </li>
+        <li>
+          <.modal_link component={LogflareWeb.SearchLive.SavedSearchesModalComponent} modal_id={:saved_searches_link} title="Saved Searches">
+            <i class="fas fa-bookmark"></i><span class="hide-on-mobile"> saved searches </span>
```
Ziinc: Should use bookmark icon for the save button as well then, so that there is some connection between to the two functions.
Maybe "saved" is enough?

lib/logflare_web/live/search_live/saved_searches_modal_component.ex
msmithstubbs: I realised it would be simpler to just fetch the saved searches from the database directly. It's an async request for the user anyway, and we optimistically remove the deleted search from the UI so there's no UI delay.

lib/logflare_web/live/search_live/subhead_components.ex
msmithstubbs: Changed the icon to bookmark and label to **Saved**

CleanShot 2026-01-15 at 12 02 38@2x

msmithstubbs:
> a separate modal icon button on the left of the icon button for each source will be much better, so that users can open up the same search modal and go to a saved search quickly.

Added a button to the dashboard for each source. Not sure if I've understood the location correctly?
CleanShot 2026-01-15 at 12 04 44@2x

lib/logflare/saved_searches/saved_searches.ex
```diff
@@ -111,14 +111,12 @@ defmodule Logflare.SavedSearches do
...
+  @spec list_saved_searches_by_source(number()) :: [SavedSearch.t()]
+  def list_saved_searches_by_source(source_id) do
     SavedSearch
     |> where([s], s.saved_by_user == true)
-    |> join(:inner, [s], src in Source, on: s.source_id == src.id)
-    |> where([s, src], src.user_id == ^user_id)
-    |> order_by([s, src], asc: src.name, asc: s.inserted_at)
-    |> preload([s, src], source: src)
+    |> where([s], s.source_id == ^source_id)
+    |> order_by([s], asc: fragment("lower(?)", s.querystring))
```
Ziinc: Probably can make this as an option in future so that we can in future order by creation date or a future name column

lib/logflare_web/live/dashboard_live/dashboard_source_components.ex
```diff
@@ -22,6 +24,9 @@ defmodule LogflareWeb.DashboardLive.DashboardSourceComponents do
       </div>
       <div>
         <div class="float-right">
+          <ModalLiveHelpers.modal_link component={SavedSearchesModalComponent} modal_id={:"saved-searches-#{@source.id}"} title="Saved Searches" class="dashboard-links tw-mr-2" phx-value-source-id={@source.id}>
```
Ziinc, line 27: not a fan of generating so many atoms due to high cardinality, ok for now but should be using strings for such IDs

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -487,7 +487,7 @@ defmodule LogflareWeb.Source.SearchLV do
 
     %Billing.Plan{limit_saved_search_limit: limit} = Billing.get_plan_by_user(user)
 
-    if Enum.count(source.saved_searches) < limit do
+    if SavedSearches.list_saved_searches_by_source(source.id) |> Enum.count() < limit do
```
Ziinc, line 490: can have a separate context (and cache) function for this

Ziinc (approved):

lib/logflare_web/live/search_live/logs_search_lv.ex
msmithstubbs: Sounds good. My thinking was that since this is only checked during a save_search event caching wasn't as critical here. Will submit a follow up PR.

PR #3081, opened 2026-01-12
Consolidated Ingest - Phase 3 ("consolidated queue safety")

amokan:
This PR addresses a number of comments made in [#3077](https://github.com/Logflare/logflare/pull/3077)

- Adds `IngestEventQueue.pop_pending/2` for atomic event removal from queues
- QueueJanitor: handle consolidated keys with 10x larger max threshold
- Skip recent logs listing for consolidated backends
- Add consolidated key support to `truncate_table` and `PubSubRates.Cache`

_This branch is based off the pending changes in #3077_

PR #3083, opened 2026-01-12
Consolidated Ingest - Phase 4 (Routing Changes)

amokan:
The 4th chunk of the consolidated backend effort focused on routing and `SourceSup` adjustments.

- `Backends.dispatch_to_backends/3` uses `{:consolidated, backend_id}` for consolidated backends
- SourceSup filters out consolidated (_next PR handles the supervision of consolidated backends_)
- `SourceSup.start_backend_child/2` returns `:noop` for consolidated backends (_as well as `start_rule_child/1` if a rule is related to a consolidated backend_)

_This branch is based off #3081 so it will look larger than it really is until that is merged_

### DIFF
--- a/lib/logflare/backends/source_sup.ex
+++ b/lib/logflare/backends/source_sup.ex
@@ -91,7 +91,7 @@ defmodule Logflare.Backends.SourceSup do
 
   This allows for zero-downtime ingestion, as we don't restart the SourceSup supervision tree.
   """
-  @spec start_rule_child(Rule.t()) :: Supervisor.on_start_child() | :skip
+  @spec start_rule_child(Rule.t()) :: Supervisor.on_start_child() | :noop
   def start_rule_child(%Rule{backend_id: backend_id} = rule) do
     backend = Backends.Cache.get_backend(backend_id) |> Map.put(:register_for_ingest, false)
     source = Sources.Cache.get_by_id(rule.source_id)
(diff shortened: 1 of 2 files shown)


lib/logflare/backends.ex
```diff
@@ -602,6 +602,23 @@ defmodule Logflare.Backends do
   end
 
   # send to a specific backend
+  defp dispatch_to_backends(source, %Backend{consolidated_ingest?: true} = backend, log_events) do
+    telemetry_metadata = %{backend_type: backend.type}
+
+    :telemetry.span([:logflare, :backends, :ingest, :dispatch], telemetry_metadata, fn ->
+      log_events = maybe_pre_ingest(source, backend, log_events)
+      IngestEventQueue.add_to_table({:consolidated, backend.id}, log_events)
```
Ziinc: If only this line is different, maybe just an if condition for the queue key is enough. The code duplication seems unnecessary

lib/logflare/backends/source_sup.ex
```diff
@@ -98,8 +101,12 @@ defmodule Logflare.Backends.SourceSup do
   @doc """
   Starts a given backend-souce combination when SourceSup is already running.
   This allows for zero-downtime ingestion, as we don't restart the SourceSup supervision tree.
+
+  Consolidated backends are excluded.
   """
-  @spec start_backend_child(Source.t(), Backend.t()) :: Supervisor.on_start_child()
+  @spec start_backend_child(Source.t(), Backend.t()) :: Supervisor.on_start_child() | :skip
+  def start_backend_child(%Source{}, %Backend{consolidated_ingest?: true}), do: :skip
```
Ziinc:
```suggestion
  def start_backend_child(%Source{}, %Backend{consolidated_ingest?: true}), do: :noop
```
Nit: Follows rest of codebase

Ziinc (approved):

lib/logflare/backends.ex
amokan: Good catch

PR #3089, opened 2026-01-13
Consolidated Ingest - Phase 6 (ClickHouse Backend Changes)

amokan:
This is the final PR in the effort to consolidate ingestion for ClickHouse.

- Converts the ClickHouse backend (and related modules) to act as a consolidated backend
- Changes the ingest table naming schema for ClickHouse to `ingest_<backend_token>` with an added `source_id` attribute.
- Removes mocks used in prior tests in favor of the local ClickHouse instance

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
@@ -95,11 +95,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
   @doc false
   @spec encode_batch([LogEvent.t()]) :: iodata()
   def encode_batch([%LogEvent{} | _] = rows) do
-    rows
-    |> Enum.sort_by(fn %LogEvent{origin_source_uuid: sid, body: body} ->
-      {sid, body["timestamp"]}
-    end)
-    |> Enum.map(&encode_row/1)
+    Enum.map(rows, &encode_row/1)
   end
 
   @doc false


lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -78,9 +81,12 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
 
   @doc false
   @spec encode_row(LogEvent.t()) :: iodata()
-  def encode_row(%LogEvent{body: body}) do
+  def encode_row(%LogEvent{body: body, origin_source_id: origin_source_id}) do
```
Ziinc:
```suggestion
  def encode_row(%LogEvent{body: body, origin_source_uuid: origin_source_uuid}) do
```

amokan: `origin_source_uuid` is not an attribute in the LogEvent struct right now. Are you suggesting we add a new one or rename the existing attribute?

Ziinc: renamine for clarity is fine, since it is easy to get tripped up with the integer ids and uuids

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -89,7 +95,11 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
   @doc false
   @spec encode_batch([LogEvent.t()]) :: iodata()
   def encode_batch([%LogEvent{} | _] = rows) do
-    Enum.map(rows, &encode_row/1)
+    rows
+    |> Enum.sort_by(fn %LogEvent{origin_source_id: sid, body: body} ->
+      {sid, body["timestamp"]}
+    end)
+    |> Enum.map(&encode_row/1)
```
Ziinc: > That said, pre-sorting is an optional optimization—not a requirement. ClickHouse sorts data highly efficiently using parallel processing, and in many cases, server-side sorting is faster or more convenient than pre-sorting client-side.
the double iteration might be quite expensive for such large batches, maybe lets try without the premature optimization first so that we have a baseline to determine if client-side sorting for our situation is worth, since the insert part is not in the millions.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -212,58 +171,21 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
-  """
-  @spec pipeline_alive?(source_backend_tuple()) :: boolean()
-  def pipeline_alive?({%Source{}, %Backend{}} = args) do
-    case pipeline_pid(args) do
-      nil -> false
-      pid -> Process.alive?(pid)
+  @spec clickhouse_ingest_table_name(Backend.t()) :: String.t()
+  def clickhouse_ingest_table_name(%Backend{token: token}) do
+    token_str = String.replace(token, "-", "_")
+    table_name = "#{QueryTemplates.default_table_name_prefix()}_#{token_str}"
```
Ziinc, line 180: this means that creating two backend records with exactly the same configs would result in two different ingestion tables.
is this the expected behaviour?

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -15,20 +16,21 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
-  alias Logflare.Sources
   alias Logflare.Utils
 
   @producer_concurrency 1
-  @processor_concurrency 5
-  @batcher_concurrency 10
-  @batch_size 10_000
-  @batch_timeout 4_000
+  @processor_concurrency 2
+  @batcher_concurrency 1
```
Ziinc: batcher shouldn't be 1, since batchers fill up sequentially and while one batcher is waiting for network io, other batchers will then get filled up.
keeping same 2:1 processor/batcher ratio would be better since with larger batches, more processors will be needed.

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
amokan: Changed `origin_source_id` to `origin_source_uuid` throughout the code.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -19,23 +21,16 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
-  typedstruct do
-    field(:source, Source.t())
-    field(:backend, Backend.t())
-    field(:ingest_connection, tuple())
-  end
 
-  @type source_backend_tuple :: {Source.t(), Backend.t()}
-  @type via_tuple :: {:via, Registry, {module(), {pos_integer(), {module(), pos_integer()}}}}
+  @min_pipelines 1
+  @max_pipelines 2
```
Ziinc: these are too low, was thinking that keeping max at Systems.schedulers_online/0 would be good, to factor in large spikes. Each pipeline is one queue, so if we were to limit this to only 2 queues, then things would implode quite fast. tweaking the resolve logic to up slower and down faster would be something to consider

Ziinc: considering that this is for every single source, it would make sense to have a higher ceiling imo

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
```diff
@@ -73,12 +73,13 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.QueryTemplates do
       """
       CREATE TABLE IF NOT EXISTS #{db_table_string} (
         `id` UUID,
+        `source_id` UUID,
```
Ziinc:
```suggestion
        `source_uuid` UUID,
```

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
```diff
@@ -73,12 +73,13 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.QueryTemplates do
...
         `body` String,
         `timestamp` DateTime64(6)
       )
       ENGINE = #{engine}
       PARTITION BY toYYYYMMDD(timestamp)
-      ORDER BY (timestamp)
+      ORDER BY (source_id, timestamp)
```
Ziinc:
```suggestion
      ORDER BY (source_uuid, timestamp)
```

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
```diff
@@ -73,12 +73,13 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.QueryTemplates do
       """
       CREATE TABLE IF NOT EXISTS #{db_table_string} (
         `id` UUID,
+        `source_id` UUID,
         `body` String,
         `timestamp` DateTime64(6)
       )
       ENGINE = #{engine}
       PARTITION BY toYYYYMMDD(timestamp)
```
Ziinc, line 81: can probably partition by source_uuid as well

Ziinc: unsure of query performance impact though

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
amokan: Yep. I don't really see a _safer_ solution in that scenario. Seems like if you're creating another backend with the same config, you're clearly trying to isolate something, right?

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
amokan: I have increased the processor concurrency to 4 and set the batcher concurrency to 2.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
amokan: Max pipelines has been changed to use `Systems.schedulers_online/0` and the resolver logic has been updated as suggested.

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
amokan: Done

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
amokan: Done

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
amokan: Pre-sorting removed

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
amokan: We discussed this and decided to leave as-is for now. Will handle a migration if it's determined we want to add that partition later and/or adjust the timestamp partition to something other than daily.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
Ziinc: hmm perhaps we can make it as an overridable config for customization when we offer db drains.

PR #3093, opened 2026-01-15
feat: backend dropdown for /query view

msmithstubbs:
Allows users to select which backend to run a query again. 

- [x]  Adds a backends dropdown to the /query view 
- [x] Warns user when the query uses a source with out a configured backend
- [x] Support in /api/query

* Requests to `/api/query` accept an optional `backend_id`. 
* When `backend_id` and `sql` params are used the sql dialect will be inferred from the selected backend. 
  * If no backend is specified the default is BigQuery with BigQuery SQL dialect.
  * If the request includes a`ch_sql`, `pg_sql`, or `bq_sql` param this determines the sql dialect used. 

CleanShot 2026-01-15 at 16 40 42@2x

### Error messages

An backend error from running a query is rendered in the UI under the form. 

CleanShot 2026-06-23 at 15 21 59@2x


lib/logflare_web/live/query_live.ex
```diff
@@ -193,13 +198,12 @@ defmodule LogflareWeb.QueryLive do
...
+      |> assign_form(params)
 
     query_string =
-      if q != nil and socket.assigns.query_string == nil do
-        q
-      else
-        "SELECT id, timestamp, metadata, event_message \nFROM `YourSource` \nWHERE timestamp > '#{DateTime.utc_now() |> DateTime.to_iso8601()}'"
-      end
+      q ||
+        socket.assigns.query_string ||
```
msmithstubbs, line 217: Logic here changed as the default query was overwriting `query_string` when we had an `assigns.query_string` and a `q` param-

lib/logflare_web/live/query_live.ex
```diff
@@ -193,13 +198,12 @@ defmodule LogflareWeb.QueryLive do
...
+        "SELECT id, timestamp, metadata, event_message \nFROM YourSource \nWHERE timestamp > '#{DateTime.utc_now() |> DateTime.to_iso8601()}'"
```
msmithstubbs, line 218: Removed backticks from example query so it is compatible across non-BigQuery backends.--

lib/logflare_web/live/query_live.ex
```diff
@@ -287,26 +311,80 @@ defmodule LogflareWeb.QueryLive do
     {:noreply, LiveMonacoEditor.set_value(socket, formatted, to: "query")}
   end
 
+  def handle_event("set_backend", %{"backend" => params}, socket) do
+    {:noreply, assign_form(socket, params)}
+  end
```
msmithstubbs, line 339: The `backend_id` URL param is not updated here. This maintains consistency with updating the `q` param. Both are updated on `run_query`

test/logflare_web/controllers/api/query_controller_test.exs
```diff
@@ -125,5 +126,117 @@ defmodule LogflareWeb.Api.QueryControllerTest do
...
+    end
+
+    test "?ch_sql= with backend_id", %{conn: conn, user: user} do
+      backend = insert(:backend, user: user, type: :clickhouse)
+
+      expect(ClickHouseAdaptor, :execute_query, fn _backend, query, _opts ->
+        assert {~S|SELECT now() AS "my_time"|, _, _, %{language: :ch_sql}} = query
+
+        {:ok, [%{"my_time" => "123"}]}
+      end)
```
Ziinc: shouldn't stub if can, ideally we run it against the test CH backend with test setup for it.

test/logflare_web/controllers/api/query_controller_test.exs
```diff
@@ -125,5 +126,117 @@ defmodule LogflareWeb.Api.QueryControllerTest do
...
+
+      assert %{"result" => [%{"my_time" => _}]} = response
+    end
+  end
+
+  describe "backend_id parameter" do
+    test "?sql= with backend_id uses backend's language", %{conn: conn, user: user} do
+      backend = insert(:backend, user: user, type: :clickhouse)
+
+      expect(ClickHouseAdaptor, :execute_query, fn _backend, query, _opts ->
```
Ziinc: as below.

lib/logflare/endpoints.ex
```diff
@@ -616,6 +621,51 @@ defmodule Logflare.Endpoints do
     find_backend_by_type_or_default(user_id, backend_type)
   end
 
+  @spec validate_backend_sources(Query.t(), Backend.t(), String.t()) :: :ok | {:error, String.t()}
+  defp validate_backend_sources(%Query{backend_id: nil}, _backend, _query_string), do: :ok
+
+  defp validate_backend_sources(
```
Ziinc: i think instead of validating, since this can become  quite complex, we just let the error propagate to the user as an error message, either as a flash or an alert (which might be more readable).
if a table doesn't exist then it should be quite self-explanatory that something is not configured right.

Ziinc (changes requested):
overall changes look good, just hesitant on the validation logic, looks very heavy for what we need.

test/logflare_web/controllers/api/query_controller_test.exs
msmithstubbs: Done,

test/logflare_web/controllers/api/query_controller_test.exs
msmithstubbs: and done.

lib/logflare/endpoints.ex
msmithstubbs: Removed, and query run error is now rendered under the form.

.dialyzer_ignore.exs
```diff
@@ -66,6 +66,7 @@
   {"lib/logflare_web/controllers/stripe_controller.ex", :pattern_match_cov},
   {"lib/logflare_web/live/billingaccount_live/payment_method_component.ex", :call},
   {"lib/logflare_web/live/monaco_editor_component.ex", :pattern_match},
+  {"lib/logflare_web/live/endpoints/endpoints_live.ex", :unused_fun},
```
msmithstubbs: Added to get test typings to pass.

lib/logflare_web/controllers/api/query_controller.ex
```diff
@@ -99,33 +103,73 @@ defmodule LogflareWeb.Api.QueryController do
...
 
-  def query(%{assigns: %{user: user}} = conn, %{"ch_sql" => sql}) do
-    with {:ok, %{rows: rows}} <- Endpoints.run_query_string(user, {:ch_sql, sql}) do
-      json(conn, %{result: rows})
-    end
+  @spec extract_query(map()) :: {:ok, atom(), String.t()} | {:error, String.t()}
+  defp extract_query(%{"sql" => sql}), do: {:ok, :infer, sql}
+  defp extract_query(%{"bq_sql" => sql}), do: {:ok, :bq_sql, sql}
+  defp extract_query(%{"ch_sql" => sql}), do: {:ok, :ch_sql, sql}
+  defp extract_query(%{"pg_sql" => sql}), do: {:ok, :pg_sql, sql}
```
msmithstubbs, line 132: Do we need to infer language from the param name? Seems like we could just treat `ch_sql`, `bq_sql`, `pg_sql`, and `sql` as interchangeable and infer the language from the selected backend. Is there a reason to send, for example, BigQuery sql to a non-BigQuery backend?

lib/logflare_web/components/query_components.ex
```diff
@@ -3,12 +3,57 @@ defmodule LogflareWeb.QueryComponents do
...
+  attr :show_language, :boolean, default: true
+  attr :label, :string, default: nil
+
+  slot :help
+
+  def backend_select(assigns) do
+    assigns =
+      assign_new(assigns, :options, fn %{backends: backends} ->
+        backend_options = Enum.map(backends, &{"#{&1.name} (#{&1.type})", &1.id})
+        [{"Default (BigQuery)", nil}] ++ backend_options
```
Ziinc, line 28: It depends on the system backend, for self hosters in single tenant mode it could be postgres (or CH in future)

lib/logflare_web/controllers/api/query_controller.ex
Ziinc: We might be able to do so, it makes sense to deprecate the prefix query params since we can infer from the backend directly. 
Need to be careful about doing so for self hosted supabase mode as the dialect translation from BQ to PG might result in unexpected behaviours. 
IIRC it should be fine to remove it

Ziinc (approved):

PR #3098, opened 2026-01-20
refactor: replace scrivener_html with numerator

msmithstubbs:
This PR drops scrivener_html and replaces with a pagination component based on [numerator](https://hex.pm/packages/numerator).

This is in preparation for updating React, which will allow some chart improvements (ANL-1136)
[scrivener_html](https://hex.pm/packages/scrivener_html) does not support phoneix_html beyond 3.0 and that's blocking updating phoenix_react_component to 0.6.

CleanShot 2026-01-20 at 13 14 23@2x


lib/logflare_web/templates/admin/sources.html.heex
```diff
@@ -0,0 +1,46 @@
...
+  </ul>
+  <%= if @sources.total_pages > 1 do %>
+    <PaginationComponents.pagination_links page={@sources} path={~p"/admin/sources"} params={[sort_by: @conn.params["sort_by"]]} />
+  <% end %>
+</div>
+<div id="__phx-assigns__" data-source-tokens={Jason.encode!(Enum.map(@sources, &Atom.to_string(&1.token)))}></div>
+<script>
+  document.addEventListener("DOMContentLoaded", async () => {
+    await Dashboard.main()
+  })
```
Ziinc, line 45: i think this should be safe to remove

lib/logflare_web/templates/admin/accounts.html.heex
```diff
@@ -0,0 +1,56 @@
...
+    <%= if @accounts.total_pages > 1 do %>
+      <PaginationComponents.pagination_links page={@accounts} path={~p"/admin/accounts"} params={[sort_by: @conn.params["sort_by"]]} />
+    <% end %>
+  </div>
+</div>
+
+<script>
+  document.addEventListener("DOMContentLoaded", async () => {
+    await Dashboard.main()
+  })
```
Ziinc, line 55: this should be safe to remove, i think. would need to double check. would want to remove these kind of adhoc scripts from templates.

lib/logflare_web/templates/admin/sources.html.heex
```diff
@@ -0,0 +1,46 @@
...
+      Sort by: {@conn.params["sort_by"]}
+    </button>
+    <div class="dropdown-menu" aria-labelledby="dropdownMenuButton">
+      <%= for option <- @sort_options do %>
+        {link(option, to: Routes.admin_path(@conn, :sources, sort_by: option), class: "dropdown-item")}
+      <% end %>
+    </div>
+  </div>
+  <ul class="list-group">
+    <%= for source <- @sources do %>
```
Ziinc, line 21: on a side note, as of now, the /sources admin page in production is non-functional due to all sources getting loaded. we probably can remove this page.

Ziinc: for support we usually use the accounts "become" functionality instead

PR #3099, opened 2026-01-20
chore: update phoenix_html

msmithstubbs:
Follows on from #3098, part of ANL-1136

* Update phoenix_html to 4.0
* Update React to 18


lib/logflare_web/controllers/admin_controller.ex
```diff
@@ -11,7 +11,7 @@ defmodule LogflareWeb.AdminController do
 
   require Logger
 
-  @page_size 50
+  @page_size 5
```
Ziinc: 5 per page would be to few and not very efficient

Ziinc: is this change needed?

Ziinc (changes requested):

lib/logflare_web/controllers/admin_controller.ex
msmithstubbs: No, sorry, accidental commit. That was hanging around from when I was testing the pagination changes in #3098 

Fixed.

Ziinc (approved):

PR #3109, opened 2026-01-23
Replace via_rule with via_rule_id

bblaszkow06:
* Reduces the size of each LogEvent
* Unlocks further optimizations in SourceRouter


lib/logflare/rules/cache.ex
```diff
@@ -45,23 +45,20 @@ defmodule Logflare.Rules.Cache do
...
-        value
+  def get_lql_rule(id), do: apply_repo_fun(__ENV__.function, [id])
 
-      {:ok, {:cached, value}} ->
-        value
-    end
+  def get_lql_rules(ids) do
+    Cachex.execute!(__MODULE__, fn cache ->
+      for id <- Enum.uniq(ids) do
+        ContextCache.fetch(cache, {:get_lql_rule, [id]}, fn -> Rules.get_lql_rule(id) end)
```
Ziinc: Perhaps filtering out those that need to be fetched and then fetching all in one pass with a Repo.all would be more efficient... it would definitely reduce db load.

lib/logflare/sources/source/channel_topics.ex
```diff
@@ -26,9 +27,15 @@ defmodule Logflare.Sources.Source.ChannelTopics do
...
+    lql_rule =
+      case le do
+        %{via_rule_id: nil} -> nil
+        %{via_rule_id: rule_id} -> Rules.Cache.get_lql_rule(rule_id)
+      end
+
     maybe_broadcast("source:#{source.token}", "source:#{source.token}:new", %{
       body: body,
-      via_rule: le.via_rule && Map.take(le.via_rule, [:regex]),
+      via_lql_rule: lql_rule,
```
Ziinc: we can get rid of the lql rule broadcasting, it was effectively broken (regex rules was removed a long time ago) so there is no benefit in keeping it here.

lib/logflare_web/controllers/source_controller.ex
```diff
@@ -462,21 +463,42 @@ defmodule LogflareWeb.SourceController do
...
+        |> Map.put(:via_lql_rule, lql_rule)
+        |> case do
+          %{body: %{"metadata" => %{"level" => level}}} = le
+          when level in ~W(debug info warning error alert critical notice emergency) ->
+            body = Map.put(le.body, "level", level)
+            %{le | body: body}
+
+          le ->
+            le
+        end
```
Ziinc: never been a fan the body manipulation logic here, lets take the opportunity to rip it out

lib/logflare_web/templates/source/show.html.heex
```diff
@@ -49,8 +49,8 @@
       <div class="collapse metadata" id={"metadata-#{inx}"}>
         <pre class="pre-metadata"><code><%= JSON.encode!(log.body, pretty: true) %></code></pre>
       </div>
-      <%= if log.via_rule do %>
-        <span data-toggle="tooltip" data-placement="top" title={"Matching #{ log.via_rule.lql_string } routing from #{log.origin_source_uuid }"} style="color: ##5eeb8f;">
+      <%= if log.via_lql_rule do %>
+        <span data-toggle="tooltip" data-placement="top" title={"Matching #{ log.via_lql_rule } routing from #{ log.origin_source_uuid }"} style="color: #5eeb8f;">
```
Ziinc: i think we can just show a "Routed" icon if the `log.via_rule_id` is present, save us the trouble of fetching the data.

lib/logflare/rules.ex
```diff
@@ -79,6 +79,12 @@ defmodule Logflare.Rules do
   @spec get_rule(non_neg_integer()) :: Rule.t() | nil
   def get_rule(id), do: Repo.get(Rule, id)
 
+  @doc "Retrieves LQL representation of a rule"
+  @spec get_rule(non_neg_integer()) :: String.t() | nil
+  def get_lql_rule(id) do
+    Repo.one(from(r in Rule, where: r.id == ^id, select: r.lql_string))
+  end
+
```
Ziinc: if we remove the lql string rendering then we don't really need this.

lib/logflare/rules/cache.ex
bblaszkow06: It makes sense for the DB fetching, but not quite for caching, as we wouldn't want to duplicate entries for fetching e.g. `[1, 2, 3]` and `[2, 3]`, and there's no straightforward way to fetch multiple ids from ETS cache. We could split the list, and group fetch the rest (that is missing from cache). Since `get_lql_rules` is removed, I'll consider it in a following PR for using in `get_rules` replacement

test/logflare_web/controllers/source_controller_test.exs
```diff
@@ -141,7 +142,7 @@ defmodule LogflareWeb.SourceControllerTest do
 
     test "show source's recent logs", %{conn: conn, source: source} do
       start_supervised!({SourceSup, source})
-      le = build(:log_event, source: source, metadata: %{"level" => "debug"})
+      le = LogEvent.make(%{"level" => "debug", "event_message" => "Log!"}, %{source: source})
```
Ziinc: should use factory for consistency, seems like an unnecessary change?

Ziinc (commented):
Small nit, looks nice

test/logflare_web/controllers/source_controller_test.exs
bblaszkow06: The change was necessary to move `"level": "debug"` outside of `"metadata"` key; however, it could be done with the factory as well.

PR #3116, opened 2026-01-28
test(e2e): check if expand button is on screen after scroll

laraujo7:


test/e2e/supabase/features/schema_modal.spec.ts
```diff
@@ -26,8 +26,38 @@ test('copies correct log id and status to clipboard', async ({ page, browserName
...
+
+  const expandButton = page.getByRole('button', { name: 'Expand' });
+
+  expect(await isVisibleOnScreen(expandButton)).toBe(false);
+
+  await expandButton.scrollIntoViewIfNeeded();
+
+  await page.waitForTimeout(200);
+
+  expect(await isVisibleOnScreen(expandButton)).toBe(true);
```
Ziinc: not so concerned buttons and other ui elements (since those may change), but more concerned about data flow:
1. on click on log event message,
2. should show the metadata of the events __somewhere__ , whether it is a modal/drawer/etc we don't really care, so long as the event metadata becomes visible

the actual request data carries more importance in this case.

Ziinc (changes requested):
assert on event metadata instead.

test/e2e/supabase/features/schema_modal.spec.ts
laraujo7: Yeap all good, thought the issue was to only test out the scroll but like that makes it better 👍

PR #3120, opened 2026-01-29
feat: quick filter button

msmithstubbs:
Adds a quick filter button the event viewer.

* quick filter appends a `FilterRule` to the LQL query string of the current search.
* the button is only displayed if the value is less than 500 characters
* array values will use the list includes search operator
* `timestamp` field the search will be +/- 1 of the current chart duration (eg. +1 hour and -1 hour).

The quick filter button is implemented as slot passed to the JSON view component. We could easily add more actions to the UI (even user configurable ones). 

CleanShot 2026-01-29 at 16 57 48@2x

## Todo

- [x] let's shrink the icon more
- [x] let's only show the icon on hover, we can always show icon for the event message and timestamp fields.
- [x] we should use the same tailing behaviour all the time even if it is a timestamp field,
- [x] ~~extract the lql manipulation logic to Lql.Rules~~
- [x] use lql_schema from  parent liveview in QueryComponents
- [x] source_schema_flat_map in liveview

## Demo

https://github.com/user-attachments/assets/563a9559-22fa-4308-a6e3-62d77aae19c0


lib/logflare_web/components/query_components.ex
```diff
@@ -69,4 +97,149 @@ defmodule LogflareWeb.QueryComponents do
...
+          String.t() | nil
+  def append_to_query(lql, %{key: key, path: path, value: value}, source, is_tailing) do
+    flat_map = source_schema_flat_map(source)
+
+    case normalize_array_key(key, path, flat_map) do
+      {normalized_key, list_includes?} ->
+        resolved_path = resolve_lql_path(normalized_key, flat_map)
+        updated_lql = append_filter(lql, resolved_path, value, source, list_includes?)
+
+        is_tailing = if key == "timestamp", do: false, else: is_tailing
```
Ziinc: Hmm I'm of the opinion we should use the same tailing behaviour all the time even if it is a timestamp field, since this implies zooming in on historical data

lib/logflare_web/components/query_components.ex
```diff
@@ -69,4 +97,149 @@ defmodule LogflareWeb.QueryComponents do
...
+
+        ~p"/sources/#{source}/search?#{%{querystring: updated_lql, tailing?: is_tailing}}"
+
+      nil ->
+        nil
+    end
+  end
+
+  defp append_filter(lql, path, value, source, list_includes?) do
```
Ziinc: We can extract the lql manipulation logic to Lql.Rules,not the first time we're doing such rule manipulation

lib/logflare_web/components/query_components.ex
```diff
@@ -69,4 +97,149 @@ defmodule LogflareWeb.QueryComponents do
...
+        path: path,
+        operator: if(list_includes?, do: :list_includes, else: :=),
+        value: value,
+        modifiers: if(is_binary(value), do: %{quoted_string: true}, else: %{})
+      )
+
+    rules ++ [filter_rule]
+  end
+
+  defp lql_schema(source) do
```
Ziinc: The parent liveview should have this already

lib/logflare_web/components/query_components.ex
```diff
@@ -69,4 +97,149 @@ defmodule LogflareWeb.QueryComponents do
...
+    end
+  end
+
+  @doc """
+  Checks the path + key exists in the schema.
+  If it doesn't the keypath may be an array index (eg. `tags.0`), so
+  recursively drops the last path segment until it finds an array type.
+  """
+  @spec normalize_array_key(String.t(), [String.t()], map()) :: {String.t(), boolean()} | nil
+  def normalize_array_key(key, path, flat_map) do
```
Ziinc: From my recollection, stored source schema flatmaps do not store the array index, could you double check the behaviour?
It would only store the inner types and not an individual key per index.

lib/logflare_web/components/query_components.ex
```diff
@@ -69,4 +97,149 @@ defmodule LogflareWeb.QueryComponents do
...
+
+  defp resolve_lql_path(key, schema_flat_map) do
+    if Map.has_key?(schema_flat_map, key) do
+      key
+    else
+      "metadata.#{key}"
+    end
+  end
+
+  defp source_schema_flat_map(source) do
```
Ziinc: Can hoist this to parent liveview

lib/logflare_web/components/query_components.ex
msmithstubbs: That's correct, the array index key was coming from the JSON viewer component. I used the array index as a key when rendering arrays.

CleanShot 2026-03-17 at 14 36 19@2x


I've updated the JSON component so it doesn't pass the array index as a key anymore. 
This means a little more duplicate in the JSON component, but less logic in QueryComponents.

lib/logflare_web/components/query_components.ex
msmithstubbs: > we should use the same tailing behaviour all the time even if it is a timestamp field

I've removed the special case so tailing will be preserved from the current liveview state. 
Is that how you wanted it to work?

Ziinc: Yea only want to trigger tailing on first view of the search page, other flows don't really make sense for the associated cost.

lib/logflare_web/components/query_components.ex
Ziinc: I see. There could be edge case bugs where stringified numbers are used for field keys and the walk up will skip that level, could you check 🙏
Walk up logic is fine to keep though

msmithstubbs: > There could be edge case bugs where stringified numbers are used for field keys

It's a good point. I thought I'd handled that but it didn't work if the numeric key was in the middle of the field (`metadata.0.info`). 

I'm sticking with passing the full path through from the view component. This should match the field name used in the schema flat map.

msmithstubbs: Added test coverage, too.

lib/logflare_web/components/query_components.ex
msmithstubbs: I switched this to use `Rules.upsert_filter_rule_by_path` https://github.com/Logflare/logflare/blob/b764c031ec199320b702f5dc100a3cd502ba272a/lib/logflare/lql/rules.ex#L253

This will append the rule if doesn't exist, or replace if it does, which I think is the correct behaviour. I don't think we'd want the quick filter action to create a search with two rules (` m.qty:12 m.qty:11` for example) since that wouldn't return any results.

Ziinc (approved):

PR #3140, opened 2026-02-08
feature: partition consolidated queue by log type

amokan:
Changes:
- Use the `LogEvent.log_type` as the batcher key in the consolidated pipeline, allowing each type to build up a decent batch.
- Adds schema DDL defs to the CH query template module for `otel_logs`, `otel_metrics`, and `otel_traces` (_Not dialed in fully, but enough to support the other changes in the PR_).

Known Gaps:
- Super naive mapping between a `LogEvent` struct and the encoding for the destination table. This is by design as more sophisticated mapping is coming after.
- The `encode_log_row`, `encode_metric_row`, and `encode_trace_row` functions in `Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester` are super simple at the moment. Same reason as above.
- The otel table schemas will likely need some adjustments.
- The SQL aspect around `transform_source_name` for ClickHouse had to be adjusted to just passing through the name for now. Will have to think through how SQL and LQL interact with the otel_* tables given that a source can produce data that could be in 3 different tables.

### DIFF
--- a/test/logflare/sql_test.exs
+++ b/test/logflare/sql_test.exs
@@ -557,16 +557,16 @@ defmodule Logflare.SqlTest do
       table_name = ClickHouseAdaptor.clickhouse_ingest_table_name(backend, :log)
 
       cte_query =
-        "with src as (select body from #{table_name}) select body from src"
+        "with src as (select event_message from #{table_name}) select event_message from src"
 
-      consumer_query = "select body from src"
+      consumer_query = "select event_message from src"
 
       assert {:ok, transformed} = Sql.transform(:ch_sql, {cte_query, consumer_query}, user)
       assert {:ok, results} = ClickHouseAdaptor.execute_query(backend, transformed, [])
       assert length(results) == 2
 
       # cannot access the source table directly
-      consumer_query_accessing_table = "select body from #{table_name}"
+      consumer_query_accessing_table = "select event_message from #{table_name}"
 
       assert {:error, err} =
                Sql.transform(:ch_sql, {cte_query, consumer_query_accessing_table}, user)
@@ -574,7 +574,7 @@ defmodule Logflare.SqlTest do
       assert String.downcase(err) =~ "table not found in cte"
 
       # cannot access another known source that exists but is not in the CTE
-      consumer_query_accessing_other_source = "select body from #{other_source.name}"
+      consumer_query_accessing_other_source = "select event_message from #{other_source.name}"
 
       assert {:error, err} =
                Sql.transform(:ch_sql, {cte_query, consumer_query_accessing_other_source}, user)


lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -80,30 +93,91 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
...
-  @doc false
-  @spec encode_batch([LogEvent.t()]) :: iodata()
-  def encode_batch([%LogEvent{} | _] = rows) do
-    Enum.map(rows, &encode_row/1)
+  @spec encode_metric_row(LogEvent.t()) :: iodata()
+  defp encode_metric_row(%LogEvent{
+         id: id,
+         body: body,
+         origin_source_uuid: origin_source_uuid,
+         origin_source_name: origin_source_name
```
Ziinc, line 146: side note: we don't really need the `origin_` prefix since there is already `:source_id` on the struct

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -80,30 +93,91 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
...
+      RowBinaryEncoder.datetime64(timestamp, 9),
+      RowBinaryEncoder.json(body)
+    ]
+  end
+
+  @spec body_timestamp_to_datetime(integer() | nil) :: DateTime.t()
+  defp body_timestamp_to_datetime(nil), do: DateTime.utc_now()
+
+  defp body_timestamp_to_datetime(timestamp_us) when is_integer(timestamp_us) do
+    DateTime.from_unix!(timestamp_us, :microsecond)
```
Ziinc: this can get quite expensive actually as the conversion is slow, clickhouse can't ingest it as a unix timestamp directly? it's one of the reasons why i ripped out the iso timestamp conversions from Processor modules

Ziinc: would be good if we can offload the conversion compute to clickhouse

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -16,26 +17,38 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
   @pool_timeout 8_000
   @receive_timeout 30_000
 
+  @log_columns ~w(id source_uuid source_name project timestamp body log_attributes)
+  @metric_columns ~w(id source_uuid source_name project time_unix start_time_unix metric_type attributes)
+  @trace_columns ~w(id source_uuid source_name project timestamp span_attributes)
```
Ziinc: should also have start_time_unix and end_time_unix equivalents

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -16,26 +17,38 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
...
+      when is_list(log_events) and is_log_type(log_type) do
     with {:ok, connection_opts} <- build_connection_opts(backend) do
-      insert(connection_opts, table, log_events)
+      insert(connection_opts, table, log_events, log_type)
     end
   end
 
-  def insert(connection_opts, table, [%LogEvent{} | _] = log_events)
-      when is_list(connection_opts) and is_non_empty_binary(table) do
+  def insert(connection_opts, table, [%LogEvent{log_type: log_type} | _] = log_events, log_type)
```
Ziinc, line 47: `LogEvent.log_type` - probably makes more sense to call this `:event_type` instead

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
```diff
@@ -41,18 +38,187 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.QueryTemplates do
...
+    {db_table, engine, ttl_days} = extract_opts(table, opts)
+
+    Enum.join([
+      """
+      CREATE TABLE IF NOT EXISTS #{db_table} (
+        `id` UUID,
+        `source_uuid` LowCardinality(String) CODEC(ZSTD(1)),
+        `source_name` LowCardinality(String) CODEC(ZSTD(1)),
+        `project` LowCardinality(String) CODEC(ZSTD(1)),
+        `time_unix` DateTime64(9) CODEC(Delta(8), ZSTD(1)),
```
Ziinc, line 118: no `timestamp` or `event_message` column, expected? means that less compatabiilty with existing LQL and query building logic. also means LQL querying will break if we were to support clickhouse as a supported self-hosted backend

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
```diff
@@ -41,18 +38,187 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.QueryTemplates do
...
+        `project` LowCardinality(String) CODEC(ZSTD(1)),
+        `timestamp` DateTime64(9) CODEC(Delta(8), ZSTD(1)),
+        `timestamp_time` DateTime DEFAULT toDateTime(timestamp),
+        `trace_id` String CODEC(ZSTD(1)),
+        `span_id` String CODEC(ZSTD(1)),
+        `trace_flags` UInt8,
+        `severity_text` LowCardinality(String) CODEC(ZSTD(1)),
+        `severity_number` UInt8,
+        `service_name` LowCardinality(String) CODEC(ZSTD(1)),
+        `body` String CODEC(ZSTD(1)),
```
Ziinc: as mentioned in other comment, this might break LQL stuff

Ziinc: might need a callback on Adaptor module to determine the event_message field for LQL compatability

Ziinc (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor/query_templates.ex
amokan: Again, not a final schema and CH really wanted `Body` to be the name for the sake of HyperDX but I think that is working now after they made some adjustments.

PR #3141, opened 2026-02-09
feat: backend connection testing

bblaszkow06:
Closes ANL-707


lib/logflare_web/controllers/api/backend_controller.ex
```diff
@@ -101,4 +102,27 @@ defmodule LogflareWeb.Api.BackendController do
...
+    summary: "Test backend connection",
+    parameters: [token: [in: :path, description: "Backend Token", type: :string]],
+    responses: %{
+      200 => BackendConnectionTest.response(),
+      404 => NotFound.response()
+    }
+  )
+
+  def test_connection(%{assigns: %{user: user}} = conn, %{"token" => token}) do
+    with {:ok, backend} <- Backends.fetch_backend_by(token: token, user_id: user.id) do
```
bblaszkow06, line 116: Would it make sense to introduce cache here? This function has not been cached yet.

lib/logflare_web/controllers/api/backend_controller.ex
```diff
@@ -101,4 +102,27 @@ defmodule LogflareWeb.Api.BackendController do
...
+      case Backends.test_connection(backend) do
```
bblaszkow06, line 117: IMO, this should be guarded by some rate-limiting. Is it OK to merge without it? Should it be a priority?

lib/logflare_web/controllers/api/backend_controller.ex
Ziinc: yes that makes sense 👍

lib/logflare_web/controllers/api/backend_controller.ex
Ziinc: downstream service will rate limit, so this is fine for now

lib/logflare_web/controllers/api/backend_controller.ex
bblaszkow06: Stacked that on top of this in https://github.com/bblaszkow06/logflare/pull/21

PR #3142, opened 2026-02-09
feat: KeyValues initial implementation

Ziinc:
This PR implements the initial KeyValues logic, a ingest-time data enrichment transformation step that will allow us to lookup and enrich events as they get ingested and stored.

Limitations is at the billing level, set to 10M for paid plans.
Page is also feature flagged to staging via config cat.

Closes ANL-1346

### DIFF
--- a/docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
+++ b/docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
@@ -158,7 +158,7 @@ Destination field names must match the BigQuery colummn name specification, or t
 ### Key-Value Enrichment
 
 :::info Pro Plan / Self-Hosted
-Key-value enrichment is available on Pro plans and for self-hosted deployments.
+Key-value enrichment is only available on Pro plans and for self-hosted deployments.
 :::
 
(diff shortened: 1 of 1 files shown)


lib/logflare_web/open_api_schemas.ex
```diff
@@ -140,6 +140,16 @@ defmodule LogflareWeb.OpenApiSchemas do
     use LogflareWeb.OpenApi, properties: @properties, required: [:name]
   end
 
+  defmodule KeyValueApiSchema do
+    @properties %{
+      id: %Schema{type: :integer},
+      key: %Schema{type: :string},
+      value: %Schema{type: :string}
```
chasers: Hmm, should we do jsonb for values? There is a need possibly to enrich with other values from a key.

I guess you could do `key:something: value` instead. Would keep it more straight forward when we go to enrich the logs.

Ziinc: done

Ziinc: Allows both setting of the entire json and on a specific field of the json

PR #3143, opened 2026-02-09
fix(e2e): correct PostgRest flaky test

laraujo7:


nelsonmestevao (approved):
This should fix the flakyness of postgrest-logs 🙏

PR #3146, opened 2026-02-10
feat: store last chosen team in session

msmithstubbs:
Default to the last selected team when no `t=` team param is present.

* when a user selects a team from the team switcher dropdown the `last_team_id` is stored in the session
* When visiting a URL without a `t=` param the `last_team_id` is used as the current team

## Demo

This recording demonstrates moving between teams and falling back to the last selected team.

* From the dashboard of my home team  I selected another team (Happy Team `team_id=13`)
* Click through to a source. 
* Remove the `t=` param from the URL.
* click dashboard: the last selected team is preserved

https://github.com/user-attachments/assets/2a3560c9-8ce3-47e7-b3c2-a7645c9b3996

Closes ANL-1278

### DIFF
--- a/lib/logflare_web/live/auth_live.ex
+++ b/lib/logflare_web/live/auth_live.ex
@@ -13,8 +13,8 @@ defmodule LogflareWeb.AuthLive do
   alias Logflare.Teams.TeamContext
   require Logger
 
-  def on_mount(:default, params, %{"current_email" => email}, socket) do
-    team_id = Map.get(params, "t")
+  def on_mount(:default, params, %{"current_email" => email} = session, socket) do
+    team_id = Map.get(params, "t") || session["last_switched_team_id"]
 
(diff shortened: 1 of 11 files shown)


lib/logflare_web/live/auth_live.ex
```diff
@@ -13,8 +13,8 @@ defmodule LogflareWeb.AuthLive do
   alias Logflare.Teams.TeamContext
   require Logger
 
-  def on_mount(:default, params, %{"current_email" => email}, socket) do
-    team_id = Map.get(params, "t")
+  def on_mount(:default, params, %{"current_email" => email} = session, socket) do
+    team_id = Map.get(params, "t") || session["last_team_id"]
```
Ziinc: maybe more explicit like `last_switched_team_id`

PR #3147, opened 2026-02-10
fix: Don't rely on `total_bytes_processed` in `query_live`

ruslandoga:
Right now running [`select 1`](http://localhost:4000/query?q=select+1%3B) fails on a local instance started with `make start.st.pg`:

```
[error] [] GenServer #PID<0.5885.0> terminating
** (CaseClauseError) no case clause matching:

    {:ok, %{rows: [%{"?column?" => 1}]}}

    (logflare 1.30.8) lib/logflare_web/live/query_live.ex:297: LogflareWeb.QueryLive.run_query/3
    (logflare 1.30.8) lib/logflare_web/live/query_live.ex:260: LogflareWeb.QueryLive.handle_event/3
    (phoenix_live_view 1.0.18) lib/phoenix_live_view/channel.ex:509: anonymous fn/3 in Phoenix.LiveView.Channel.view_handle_event/3
    (telemetry 1.3.0) /Users/x/Developer/logflare/deps/telemetry/src/telemetry.erl:324: :telemetry.span/3
    (phoenix_live_view 1.0.18) lib/phoenix_live_view/channel.ex:260: Phoenix.LiveView.Channel.handle_info/2
    (stdlib 6.2.2.2) gen_server.erl:2345: :gen_server.try_handle_info/3
    (stdlib 6.2.2.2) gen_server.erl:2433: :gen_server.handle_msg/6
    (stdlib 6.2.2.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
Last message: %Phoenix.Socket.Message{topic: "lv:phx-GJLkcBOkaadc1IEh", event: "event", payload: %{"event" => "run-query", "meta" => %{}, "type" => "form", "value" => "live_monaco_editor%5Bquery%5D=SELECT+1%3B"}, ref: "10", join_ref: "4"}
```

---

This PR removes `total_bytes_processed` from the match clause since it is not guaranteed to exist in adapters/backends returning only rows. First clause here: https://github.com/Logflare/logflare/blob/a3b441747dd70b74efb4c7e03dab6eaccadd8b6c/lib/logflare/endpoints.ex#L590-L598


Ziinc:
@ruslandoga can we get a LV for this error please for self-hosted postgres mode? 🙏 even if it is a relatively small bug.

test/logflare_web/live_views/query_live_test.exs
```diff
@@ -189,4 +189,45 @@ defmodule LogflareWeb.QueryLiveTest do
...
+
+      backend =
+        insert(:backend,
+          type: :postgres,
+          config: %{url: url},
+          user: user,
+          sources: [source]
+        )
+
+      stub(Logflare.Backends, :get_default_backend, fn _user -> backend end)
```
ruslandoga: @Ziinc I wonder if there is a more "correct" way to make this liveview use postgres backend. Stubbing and modifying app env both seem wrong ...

Ziinc: Can you reference SingleTenant tests? there are macros for setting up the postgres repo and configuring the default backends etc.

ruslandoga: Thank you for the pointer. Done in https://github.com/Logflare/logflare/pull/3147/changes/16b2bef4c59c3891aa9a6b322640babfff0ffc18

Ziinc (approved):

test/logflare_web/live_views/query_live_test.exs
```diff
@@ -189,4 +189,27 @@ defmodule LogflareWeb.QueryLiveTest do
...
+        value: "select 42 as answer, 'and some other expected string'"
+      })
+
+      html =
+        view
+        |> element("form")
+        |> render_submit(%{})
+
+      assert html =~ "Ran query successfully"
+      assert html =~ "0 bytes processed"
```
Ziinc, line 209: We should hide this since it doesn't make sense in postgres self hosted.
Can you remove in a follow up PR? thanks

ruslandoga: Hidden in https://github.com/Logflare/logflare/pull/3217 if `total_bytes_processed` is nil.

PR #3153, opened 2026-02-12
changes to otel_traces schema

amokan:
_I'm kind of guessing on these because I have **no idea** what would go in these fields from our existing sources :shrug:_

Either way, I have changed the field names to PascalCase to align with the request from CH. I also figured I should probably turn `events.Attributes` and `links.Attributes` back into `Array(Map(LowCardinality(String), String))` rather than `Array(JSON(max_dynamic_paths=0, max_dynamic_types=1))`.

Worth noting that I did not change anything on `otel_metrics` because they didn't say to. But that also has nested attributes in the form of `exemplars.filtered_attributes`, `exemplars.time_unix`, `exemplars.span_id`, and `exemplars.trace_id`.

This will require the table to be re-paved, so I'll need to kill CH ingestion and drop the existing tables before rolling this out.


chasers (approved):

amokan:
killing this one

PR #3156, opened 2026-02-12
refactor: switch cachex stats to telemetry polling

ruslandoga:
ANL-1033

Removed log-based Cachex stats collection in favor of telemetry-based polling system.


lib/telemetry.ex
```diff
@@ -6,15 +6,18 @@ defmodule Logflare.Telemetry do
 
   def start_link(arg), do: Supervisor.start_link(__MODULE__, arg, name: __MODULE__)
 
+  context_caches =
+    Logflare.ContextCache.Supervisor.list_caches()
+    |> Enum.map(fn cache ->
+      ["Logflare", context, "Cache"] = Module.split(cache)
+      {cache, context |> Macro.underscore() |> String.to_atom()}
+    end)
+
```
ruslandoga, line 10: This evaluates to

```elixir
[
  {Logflare.TeamUsers.Cache, :team_users},
  {Logflare.Partners.Cache, :partners},
  {Logflare.Users.Cache, :users},
  {Logflare.Backends.Cache, :backends},
  {Logflare.Sources.Cache, :sources},
  {Logflare.Billing.Cache, :billing},
  {Logflare.SourceSchemas.Cache, :source_schemas},
  {Logflare.Auth.Cache, :auth},
  {Logflare.Endpoints.Cache, :endpoints},
  {Logflare.Rules.Cache, :rules},
  {Logflare.SavedSearches.Cache, :saved_searches}
]
```

or should it be hardcoded?

test/logflare/telemetry_test.exs
```diff
@@ -60,4 +60,29 @@ defmodule Logflare.TelemetryTest do
       refute_receive {:telemetry_event, ^event, _, _}
     end
   end
+
+  describe "cachex_metrics/0" do
+    test "retrieves and emits stats for caches with all metrics" do
+      # test for one of the caches
+      event = [:cachex, :sources]
+      ref = :telemetry_test.attach_event_handlers(self(), [event])
```
ruslandoga:
```suggestion
      ref = :telemetry_test.attach_event_handlers(self(), [event])
      on_exit(fn -> :telemetry.detach(ref) end)
```

lib/telemetry.ex
Ziinc: can you inverse the logic so that we can do `list_contexts_to_cache()` and in `list_caches` it will call `list_contexts_to_cache/0` and concat the `.Cache`

Ziinc: the module.split isn't very pretty 😆

Ziinc: besides that, its fine with keeping it dynamic.

ruslandoga: I went into a slightly different direction and added

```elixir
  def list_caches_with_metrics do
    [
      {TeamUsers.Cache, :team_users},
      {Partners.Cache, :partners},
      {Users.Cache, :users},
      {Backends.Cache, :backends},
      {Sources.Cache, :sources},
      {Billing.Cache, :billing},
      {SourceSchemas.Cache, :source_schemas},
      {Auth.Cache, :auth},
      {Endpoints.Cache, :endpoints},
      {Rules.Cache, :rules},
      {KeyValues.Cache, :key_values},
      {SavedSearches.Cache, :saved_searches}
    ]
  end
```

I think it's a bit safer for atoms to exist in the source code.

ruslandoga:
I'll now try and test it with some otel collector.

PR #3157, opened 2026-02-12
refactor: scheduler/observer telemetry

ruslandoga:
closes ANL-1268

Removed log-based scheduler/observer metrics and converted to more poll-based telemetry.


lib/logflare/system_metrics/observer.ex
```diff
@@ -1,29 +1,26 @@
...
-
     :telemetry.execute([:logflare, :system, :observer, :metrics], observer_metrics)
     :telemetry.execute([:logflare, :system, :observer, :memory], mem_metrics)
   end
 
   defp get_memory do
-    :erlang.memory() |> Enum.map(fn {k, v} -> {k, div(v, 1024 * 1024)} end) |> Enum.into(%{})
+    %{memory: persistent_term_memory} = :persistent_term.info()
+    memory = [persistent_term: persistent_term_memory | :erlang.memory()]
+    Map.new(memory, fn {k, v} -> {k, div(v, 1024 * 1024)} end)
```
ruslandoga: I wonder if it'd be more idiomatic keeping native sizes here and transforming in telemetry_metrics instead.

https://hexdocs.pm/telemetry_metrics/Telemetry.Metrics.html#module-converting-units

lib/logflare/system_metrics/observer.ex
ruslandoga, line 33: `version` and `otp_release` are probably not measurements, but metadata.

lib/telemetry.ex
```diff
@@ -103,7 +103,22 @@ defmodule Logflare.Telemetry do
...
+      last_value("logflare.system.observer.metrics.port_count"),
+      last_value("logflare.system.observer.metrics.ets_count"),
+      last_value("logflare.system.observer.metrics.atom_count"),
+      last_value("logflare.system.observer.memory.total", unit: :megabyte),
+      last_value("logflare.system.observer.memory.processes", unit: :megabyte),
+      last_value("logflare.system.observer.memory.binary", unit: :megabyte),
+      last_value("logflare.system.observer.memory.ets", unit: :megabyte),
+      last_value("logflare.system.observer.memory.atom", unit: :megabyte),
+      last_value("logflare.system.observer.memory.code", unit: :megabyte),
+      last_value("logflare.system.scheduler.utilization", tags: [:name, :type])
```
ruslandoga, line 136: OK to add these or should we postpone it?

lib/logflare/system_metrics/schedulers/schedulers.ex
```diff
@@ -1,31 +1,44 @@
...
-            utilization: Kernel.floor(utilization * 100),
-            utilization_percentage: utilization_percentage
-          }
+  def async_dispatch_stats(duration \\ to_timeout(second: 1)) do
+    Logflare.Utils.Tasks.start_child(fn ->
+      :erlang.system_flag(:scheduler_wall_time, true)
+      prev_sample = :scheduler.get_sample_all()
+      Process.sleep(duration)
+      next_sample = :scheduler.get_sample_all()
+      :erlang.system_flag(:scheduler_wall_time, false)
```
ruslandoga: This implementation is different from `poller.ex`, instead of keeping prev sample around it samples over one second each time. I think it might produce a more accurate result.

lib/logflare/system_metrics/observer.ex
Ziinc: yes that would be more appropriate

lib/logflare/system_metrics/observer.ex
Ziinc: can remove

lib/telemetry.ex
Ziinc: yes these are fine

lib/logflare/system_metrics/schedulers/schedulers.ex
```diff
@@ -1,31 +1,46 @@
...
-  def scheduler_utilization(sample_a, sample_b) do
-    formatter(:scheduler.utilization(sample_a, sample_b))
+
+  def async_dispatch_stats(duration \\ to_timeout(second: 1)) do
+    Logflare.Utils.Tasks.start_child(fn -> collect_dispatch_stats(duration) end)
   end
 
-  defp formatter(scheduler_utilization) do
-    Enum.map(scheduler_utilization, fn x ->
+  defp collect_dispatch_stats(duration \\ to_timeout(second: 1)) do
```
ruslandoga:
```suggestion
  def collect_dispatch_stats(duration \\ to_timeout(second: 1)) do
```

PR #3162, opened 2026-02-13
perf: on every pull request, run playwright tests only on chromium

nelsonmestevao:
The all browsers playwright tests are taking longer than 15 minutes, for that reason it is cancelling the pipeline on every push in main branch. Even with the cache optimization done previously. So this PR reactivates only chromium on pull request and stops this workflow to run on every push (I think it could be beneficial long term but I think it is not worth it atm).

See the example [below](https://github.com/Logflare/logflare/actions/runs/21991706284/job/63540342501).
image

### DIFF
--- a/.github/workflows/integration-supabase.yml
+++ b/.github/workflows/integration-supabase.yml
@@ -1,24 +1,10 @@
 name: Integration Tests with Supabase
 
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - 'config/**'
-      - 'assets/**'
-      - 'lib/**'
-      - 'priv/**'
-      - 'mix.exs'
-      - 'mix.lock'
-      - 'Dockerfile'
-      - 'run.sh'
-      - 'test/e2e/supabase/**'
-      - '.github/workflows/integration-supabase.yml'
   pull_request:
     branches:
       - main
-    paths:
+    paths: &watched_paths
       - 'config/**'
       - 'assets/**'
       - 'lib/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-supabase.yml
```diff
@@ -91,12 +82,21 @@ jobs:
...
-    - name: Run Playwright tests (all browsers)
-      if: ${{ github.event_name != 'pull_request' }}
-      run: npx playwright test
+    # TODO: Find a way to make it take less than 15 minutes to reactive it on main
+    # - name: Install Playwright Browsers
+    #   if: ${{ github.event_name != 'pull_request' }}
+    #   run: npx playwright install chromium --with-deps
+    # - name: Run Playwright tests (all browsers)
+    #   if: ${{ github.event_name != 'pull_request' }}
+    #   run: npx playwright test
```
Ziinc, line 91: can do a matrix to spawn multiple workflow jobs

Ziinc (approved):

.github/workflows/integration-supabase.yml
nelsonmestevao: One per browser? That makes sense and might be below the 15 minutes but we will not reuse the setup cache for each (it will not be shared).

I will do that then :+1:

nelsonmestevao: Done in https://github.com/Logflare/logflare/pull/3190. We get ~13 minutes (which is fine for our case but I wish we could make it run faster under 5 min), I don't see a way to improve these. The main issue is building the logflare docker image which takes ~9 min. The caching is not very effective because once we do COPY we loose all following layers. I might rearrange the order.

PR #3145, opened 2026-02-10
feat: suggest extended search range when older events appear in chart

msmithstubbs:
Adds a button to extend search time range when there are no results but the chart returns events 

CleanShot 2026-02-10 at 13 15 13@2x

Closes ANL-1085

PR #3163, opened 2026-02-16
refactor: clean up SearchLVTest

msmithstubbs:
Cleans up a few things in SearchLVTest, including feedback from #3145 

* Tidy up assertions, remove unnecessary `retry_assert/1`
* Replace `sleep/1` with `wait_for_render/1` in SearchLVTest. Eliminates around 10 seconds

### DIFF
--- a/test/logflare_web/live/search_live/logs_search_lv_test.exs
+++ b/test/logflare_web/live/search_live/logs_search_lv_test.exs
@@ -1,6 +1,6 @@
 defmodule LogflareWeb.Source.SearchLVTest do
   @moduledoc false
-  use LogflareWeb.ConnCase, async: true
+  use LogflareWeb.ConnCase, async: false
 
   import Phoenix.LiveViewTest
 


test/logflare/backends/consolidated_sup_test.exs
```diff
@@ -130,7 +130,11 @@ defmodule Logflare.Backends.ConsolidatedSupTest do
       IngestEventQueue.upsert_tid({:consolidated, backend.id, nil})
 
       on_exit(fn ->
-        ConsolidatedSup.stop_pipeline(backend.id)
+        try do
+          ConsolidatedSup.stop_pipeline(backend.id)
+        catch
+          value -> value
+        end
```
msmithstubbs, line 137: This cleanup kept throwing an exception and causing CI to fail.

lib/logflare/logs/search_query_executor.ex
```diff
@@ -17,17 +17,20 @@ defmodule Logflare.Logs.SearchQueryExecutor do
...
-    GenServer.start_link(__MODULE__, Keyword.put(args, :caller, self()), hibernate_after: 5_000)
+    callers = [self() | Process.get(:"$callers", [])]
+    GenServer.start_link(__MODULE__, Keyword.put(args, :callers, callers), hibernate_after: 5_000)
   end
 
   @impl true
   def init(args) do
     source = Keyword.get(args, :source)
+    callers = Keyword.get(args, :callers)
+    Process.put(:"$callers", callers)
```
Ziinc: not a fan of manipulating the process dictionary directly for this since it is subject to otp internal changes.
don't think we need to update it either since it isn't used in any further logic

msmithstubbs: Removed this and restored the Sandbox.allow/4 calls

Ziinc (approved):

PR #3167, opened 2026-02-17
feat: oban scheduler for alerts, enable/disable, execution history

Ziinc:
This PR adds in the following:
- Oban for global scheduling for cross-cluster execution and execution history 
- adds in enable/disable of alerts
- adds in a simple testing interface in edit page
- changes json viewer of results to table viewer
- allows enabling/disabling of alerts execution on a per-instance basis via env var

Prod setup remains the same, alerts is still maintained and functioning in only one cluster.

Part of O11Y-321
Closes O11Y-1461
Closes O11Y-195


lib/logflare_web/live/alerts/components/run_query_result.html.heex
```diff
@@ -9,8 +9,28 @@
...
+    <table class="table table-sm tw-text-white tw-font-mono tw-text-xs">
+      <thead>
+        <tr>
+          <th :for={col <- Map.keys(hd(@query_result_rows))}>{col}</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr :for={row <- @query_result_rows}>
+          <td :for={col <- Map.keys(hd(@query_result_rows))}>
+            <% val = Map.get(row, col) %>
```
josevalim, line 23: Do you still need to do the Map.keys dance or could you traverse on the values directly?

```suggestion
          <td :for={{col, val} <- row}>
```

lib/logflare_web/live/alerts/alerts_live.ex
```diff
@@ -340,14 +456,94 @@ defmodule LogflareWeb.AlertsLive do
...
+
+  defp truncate_reason(reason, max_length \\ 80)
+
+  defp truncate_reason(reason, max_length)
+       when is_binary(reason) and byte_size(reason) > max_length do
+    String.slice(reason, 0, max_length) <> "..."
+  end
+
+  defp truncate_reason(reason, _max_length) when is_binary(reason), do: reason
+  defp truncate_reason(reason, max_length), do: truncate_reason(inspect(reason), max_length)
```
josevalim, line 515: It may be worth extracting some of these helpers (truncate_reason, format_bytes, time_ago) into a separate module for unit testing in a future commit.

josevalim (commented):
It looks great to me, just two minor comments.

I have to say: it is really nice that we can just emit regular queries when we need to manipulate existing Oban jobs!

PR #3168, opened 2026-02-17
fix: include recommended query fields for 'view' event modal

msmithstubbs:
This is an attempt to fix a reported bug that would result in the event modal showing "not found" while the permalink works correctly.  I haven't been able to reproduce the bug so this is a speculative fix. 

Both code paths call `LogEvents.get_event_with_fallback` so I don't think the issue is related to the cache. 
One difference was the 'view' link wasn't including recommended fields. It's possible the user was working with a BigQuery table that required a field to be selected that wasn't.

I haven't included tests for this change, but can if requested. It would simply be a regex on the event anchor element.

ANL-1368


Ziinc:
under observation.

PR #3182, opened 2026-02-20
native clickhouse protocol implementation for inserts

amokan:
_This is a big one, but nearly all logic is contained within the `Logflare.Backends.Adaptor.ClickHouseAdaptor.NativeIngester` namespace._

The thinking here is that we're already seeing some weird behavior around HTTP-based connection churn so it made sense to give this a go as the native wire protocol is _much_ more efficient and allows us greater flexibility at the scale we are pushing data.

If this proves to be the better approach in production as I suspect it will - we can remove the HTTP-based inserts entirely and move this to its own library with little effort, if we wanted to go that route 🤷‍♂️ 

----

### Overview

Implements ClickHouse native TCP protocol client for INSERT operations which can be enabled at runtime using the Backend config through the Logflare UI.

When enabled, this eliminates HTTP overhead (connection setup, JSON/RowBinary serialization, content negotiation) and some of the issues around HTTP-based connections. Also provides LZ4-compression.

- Wire protocol implementation (`NativeIngester.Protocol`, `NativeIngester.Connection`)
- Full client-side implementation of the ClickHouse native TCP handshake, query packets, data block framing, and server response parsing. Supports protocol revision 54483 (_ClickHouse 25.x_).
- Column-oriented block encoding (`NativeIngester.BlockEncoder`)
- Encodes typed column data into ClickHouse's native block format. Supports all Logflare column types: UInt/Int/Float variants, String, Bool, DateTime64, UUID, Nullable(T), Array(T), Enum8, JSON (_normalized to String_), and LowCardinality (_stripped to inner type_). Batches larger than 10K rows are automatically split into sub-blocks to reduce peak memory usage.
- LZ4 compression with CityHash checksums (`NativeIngester.ChCompression`, `NativeIngester.Compression`) via Rust NIF providing CityHash v1.0.2 checksums and LZ4 raw block compression, matching ClickHouse's on-wire compressed block envelope format.
- Connection pooling (`NativeIngester.Pool`, `NativeIngester.PoolSup`) with NimblePool-based connection pool with per-backend pool isolation, lazy connection creation, and idle connection health checks via handle_ping (_30s idle timeout_).
- Retry logic (`NativeIngester`) — Retries transient errors (_socket errors, checkout timeouts, ClickHouse exception codes 159/164/202/241/252_) once after 500ms with a fresh connection. Non-transient errors (_column mismatch, unknown table, etc._) fail immediately.

### Current Tradeoffs

_Things I chose to skip for now as the ClickHouse server can work these out just fine in the current form. If we keep this approach or move it to a dedicated library, these two items can be resolved with a little effort._

- LowCardinality types are stripped to their inner type - We send `low_cardinality_allow_in_native_format=0` so the server reports LowCardinality(String) as plain String. The server rebuilds the dictionary on insert. This trades some wire bandwidth (_no dictionary compression for low-cardinality columns like `severity_text`_) for avoiding the most complex serialization format in the native protocol.
- JSON columns are serialized as String - JSON values are pre-encoded to strings via Jason.encode!/1 and sent as String columns with `input_format_native_allow_types_conversion=1`, which lets ClickHouse parse them back server-side. The native protocol's Object serialization format would require ~500+ lines of schema inference and dynamic type encoding. Produces identical stored data with negligible server-side overhead.


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -130,21 +135,35 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
       :pool_size,
       :async_insert,
-      :read_only_url
+      :read_only_url,
+      :native_insert_protocol,
+      :native_port,
+      :native_pool_size
     ])
     |> Logflare.Utils.default_field_value(:async_insert, false)
+    |> Logflare.Utils.default_field_value(:native_insert_protocol, false)
```
Ziinc: Hmm shouldn't we be using db default for default values?

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -120,7 +122,10 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
        port: :integer,
        pool_size: :integer,
        async_insert: :boolean,
-       read_only_url: :string
+       read_only_url: :string,
+       native_insert_protocol: :boolean,
```
Ziinc: Why not do an enum like `protocol` type like `http` or `native`? Keeps it more flexible and we can support grpc/arrow in the future if needed

amokan: Converted to an enum named `insert_protocol`.

test/logflare/backends/adaptor/clickhouse_adaptor_test.exs
```diff
@@ -314,6 +345,44 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptorTest do
...
+
+      ddl = QueryTemplates.create_table_statement(table_name, :log, ttl_days: 0)
+      {:ok, _} = ClickHouseAdaptor.execute_ch_query(query_backend, ddl)
+
+      log_event = build_mapped_log_event(source: source, message: "native route test")
+
+      assert :ok = ClickHouseAdaptor.insert_log_events(native_backend, [log_event], :log)
+
+      pool_pid = GenServer.whereis(NativeIngester.Pool.via(native_backend))
+      assert is_pid(pool_pid)
```
Ziinc, line 368: Should also have a refute to assert that the http logic is not called at all

lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -217,19 +217,37 @@
             </small>
           </div>
           <div class="form-group">
-            {label(f_config, :port, "Port")}
+            {label(f_config, :insert_protocol, "Insert Protocol")}
+            {select(f_config, :insert_protocol, [{"HTTP", "http"}, {"Native (TCP)", "native"}],
```
Ziinc, line 221: Should also add in a configcat flag in case we need to remotely disable this feature and can't access the node due to catastrophic failure

PR #3191, opened 2026-02-23
test: improve coverage on fetch resource plug

nelsonmestevao:


lib/logflare_web/controllers/plugs/fetch_resource.ex
```diff
@@ -101,14 +104,13 @@ defmodule LogflareWeb.Plugs.FetchResource do
     end
   end
 
-  defp uuid?(_), do: false
+  defp get_source_from_headers(conn) do
```
nelsonmestevao, line 107: This was a bug where x-collection would never be called.

Ziinc: Let's remove support for x-collection and collection-related terminology, it is unused and we won't be changing the terminology any time soon.
could you do a separate PR for this? 🙏

nelsonmestevao: @Ziinc done in PR https://github.com/Logflare/logflare/pull/3197 🙏

PR #3201, opened 2026-02-24
test: add ConnectionManager lifecycle resolution tests

laraujo7:


test/logflare/backends/adaptor/clickhouse_adaptor/connection_manager_test.exs
```diff
@@ -102,6 +102,81 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.ConnectionManagerTest do
...
+
+      Process.sleep(:timer.seconds(2))
+
+      state_after = :sys.get_state(manager_pid)
+      assert is_integer(state_after.last_activity)
+      assert Process.alive?(pool_pid)
+    end
+
+    test "stops pool after exceeding inactivity timeout", %{backend: backend} do
+      {:ok, manager_pid} = ConnectionManager.start_link(backend)
```
nelsonmestevao: We could move the start link to the setup 🙏

test/logflare/backends/adaptor/clickhouse_adaptor/connection_manager_test.exs
```diff
@@ -102,6 +102,81 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.ConnectionManagerTest do
...
+      state_after = :sys.get_state(manager_pid)
+      assert state_after.last_activity > activity_before
+      assert ConnectionManager.pool_active?(backend)
+    end
+
+    test "initializes activity timestamp on resolve when not previously set", %{backend: backend} do
+      {:ok, manager_pid} = ConnectionManager.start_link(backend)
+
+      assert :ok == ConnectionManager.ensure_pool_started(backend)
+      state_with_pool = :sys.get_state(manager_pid)
```
nelsonmestevao: I think it would be better in terms of  maintainability, to extract this `:sys.get_state/1` call to a function like `get_connection_manager_state/1` (aux function and reuse them). Ideally we should have one like `get_pool_pid/1` but it is not easy to reuse it on the previous tests.

test/logflare/backends/adaptor/clickhouse_adaptor/connection_manager_test.exs
```diff
@@ -102,6 +102,81 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.ConnectionManagerTest do
...
+      {:ok, manager_pid} = ConnectionManager.start_link(backend)
+
+      state_before = :sys.get_state(manager_pid)
+
+      Process.sleep(:timer.seconds(2))
+
+      state_after = :sys.get_state(manager_pid)
+
+      assert state_before.backend_id == state_after.backend_id
+      assert is_reference(state_after.resolve_timer_ref)
```
nelsonmestevao: Could we check that resolve_timer_ref was different in state_before?

I am not sure what we are making sure in this one test though.

test/logflare/backends/adaptor/clickhouse_adaptor/connection_manager_test.exs
```diff
@@ -102,6 +102,81 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.ConnectionManagerTest do
...
+      pool_pid = state_with_pool.pool_pid
+      assert Process.alive?(pool_pid)
+
+      :sys.replace_state(manager_pid, fn state -> %{state | last_activity: nil} end)
+
+      Process.sleep(:timer.seconds(2))
+
+      state_after = :sys.get_state(manager_pid)
+      assert is_integer(state_after.last_activity)
+      assert Process.alive?(pool_pid)
```
nelsonmestevao: Here, instead of checking pool_pid we can use the public API, the ` ConnectionManager.pool_active?/1` call, that way we avoid knowing so much about the internals 🙏

test/logflare/backends/adaptor/clickhouse_adaptor/connection_manager_test.exs
nelsonmestevao: Check if these are really needed and if we can instead rely on this https://github.com/Logflare/logflare/pull/3201#discussion_r2848897570.

Ziinc (changes requested):
should avoid directly calling `:sys.get_state/1` as we would then be testing implementation details. consider exposing public functions instead on the module.

PR #3202, opened 2026-02-24
fix latency variance in clickhouse native protocol

amokan:
Was seeing some high variance on native inserts when comparing to HTTP and found some improvements to get those times consistent:

- avoid un-necessary pings on recently used connections
- when performing an insert, cache the schema details returned in the native wire handshake to avoid the overhead of rebuilding column details on each subsequent data blocks
- minimize sub-block processing only for cases where the cache schema details have changed or the first time they are returned. now pre-encodes full blocks of data when the schema cache is present.

Worth noting that if the schema information changes mid-insert, the logic will handle this situation by rebuilding the block and updating the cache.

Includes a version bump to 1.33.1


test/logflare/backends/adaptor/clickhouse_adaptor/native_ingester/pool_test.exs
```diff
@@ -292,6 +293,101 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.NativeIngester.PoolTest do
...
+                 {"id", "UInt64", [2]},
+                 {"name", "String", ["second"]}
+               ]) do
+            {:ok, updated_conn} -> {:ok, updated_conn}
+            {:error, _} = error -> {error, :remove}
+          end
+        end)
+
+      assert result == :ok
+
```
Ziinc, line 358: can add an assert on the SchemaCache just to be sure that cache is populated.

lib/logflare/backends/adaptor/clickhouse_adaptor/native_ingester/schema_cache.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  @spec start_link(term()) :: GenServer.on_start()
+  def start_link(_arg) do
+    GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
+  end
+
+  @spec get(pos_integer(), String.t()) :: Connection.column_info() | nil
+  def get(backend_id, cache_key)
+      when is_integer(backend_id) and is_non_empty_binary(cache_key) do
+    case :ets.lookup(@table, {backend_id, cache_key}) do
```
Ziinc, line 29: do we need to consider expiry? what happens if stuff goes out of sync (say support engineers mess with the db schema)?

Ziinc: would batches get rejected if stale schema is used?

amokan: it gets resolved during the comparison with the metadata that is returned in the insert handshake - basically it adapts and rebuilds the cache when needed.

amokan: The only 'loss' in that scenario is the time it took to pre-encode the block, which gets rebuilt.

PR #3207, opened 2026-02-26
add (temporary) option to write clickhouse data to a set of simpler table schemas

amokan:
This is to test some theories around the JSON data type.

- adds a `use_simple_schemas` boolean option to the ClickHouse backend
- provisions 6 tables when starting (rather than 3), with a set of the new `simple_otel_*` variants that use `Map(String, String)` rather than JSON types (_only when option is enabled_)
- adds a new `flat_map` option to the mapper NIF that handles the coercion to string values on the Rust side
- fixed an issue with the map encoding that was already present

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/native_ingester/connection.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/native_ingester/connection.ex
@@ -179,6 +179,12 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.NativeIngester.Connection 
     transport.close(socket)
   end
 
+  @spec transfer_ownership(t(), pid()) :: :ok | {:error, term()}
+  def transfer_ownership(%__MODULE__{socket: socket, transport: transport}, new_owner)
+      when is_pid(new_owner) do
+    transport.controlling_process(socket, new_owner)
+  end
+
   @spec ping(t()) :: {:ok, t()} | {:error, term()}
   def ping(%__MODULE__{} = conn) do
     packet = Protocol.encode_varuint(Protocol.client_ping())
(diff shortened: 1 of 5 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -245,6 +291,146 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
...
+      RowBinaryEncoder.string(body["status_code"] || ""),
+      RowBinaryEncoder.string(body["status_message"] || ""),
+      RowBinaryEncoder.string(body["scope_name"] || ""),
+      RowBinaryEncoder.string(body["scope_version"] || ""),
+      RowBinaryEncoder.map_string_string(body["resource_attributes"] || %{}),
+      RowBinaryEncoder.map_string_string(body["span_attributes"] || %{}),
+      RowBinaryEncoder.array(body["events.timestamp"] || [], &RowBinaryEncoder.int64/1),
+      RowBinaryEncoder.array_string(body["events.name"] || []),
+      RowBinaryEncoder.array(
+        body["events.attributes"] || [],
```
Ziinc, line 419: this is an array? 👀 
weird that it differes from resource_attributes and span_attributes. is there some special handling for this in the row binary encoding?

test/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults_test.exs
```diff
@@ -402,4 +405,202 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaultsTest do
...
+      }
+
+      result = Mapper.map(payload, compiled)
+
+      assert result["metric_name"] == "http_requests_total"
+      assert result["value"] == 42.5
+
+      # attributes should be flat
+      attrs = result["attributes"]
+      assert is_map(attrs)
```
Ziinc, line 507: hmm seems to differ from the ingester.ex

Ziinc:
would need a version bump

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
amokan: Yep and yep - https://clickhouse.com/docs/use-cases/observability/clickstack/ingesting-data/schemas#traces

PR #3217, opened 2026-02-27
fix: hide bytes processed when `nil` on `/query` page

ruslandoga:
Context: https://github.com/Logflare/logflare/pull/3147#discussion_r2856130459

If `total_bytes_processed` is missing from the query result, the `#{total_bytes_processed} bytes processed` is no longer rendered on the page.

### DIFF
--- a/.github/workflows/integration-supabase.yml
+++ b/.github/workflows/integration-supabase.yml
@@ -39,10 +39,11 @@ jobs:
         browsers:
           - name: Chrome
             project: chromium
-          - name: Firefox
-            project: firefox
-          - name: Safari
-            project: webkit
+          # TODO: reactivate when Firefox/Webkit tests are fixed
+          # - name: Firefox
+          #   project: firefox
+          # - name: Safari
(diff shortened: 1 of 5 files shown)


lib/logflare_web/live/query_live.ex
```diff
@@ -93,7 +93,7 @@ defmodule LogflareWeb.QueryLive do
     <section :if={@query_result_rows} class="container mx-auto">
       <div class="tw-flex tw-justify-between tw-items-end">
         <h3>Query result</h3>
-        <div class="tw-mb-1">
+        <div class="tw-mb-1" :if={@total_bytes_processed}>
```
ruslandoga: On second thought, it should probably avoid hiding the `div` since it has bottom margin on it. I'll move `:if` to `QueryComponents` instead.

ruslandoga: Ah, no. It doesn't seem to affect the spacing between the "Query result" title and the results table below it.

PR #3218, opened 2026-02-27
feat: add partial broadcast for cached values

ruslandoga:
ANL-1352


lib/logflare/context_cache.ex
```diff
@@ -162,4 +163,18 @@ defmodule Logflare.ContextCache do
...
   end
+
+  defp maybe_broadcast_cached(cache, cache_key, value) do
+    Logflare.Utils.Tasks.start_child(fn ->
+      configs = Application.get_env(:logflare, :cache_broadcast)
+
+      if config = Map.get(config, cache) do
+        %{ratio: ratio, max_nodes: max_nodes} = config
+        peers = Logflare.Utils.Cluster.peer_list_partial(ratio, max_nodes)
+        :erpc.multicast(peers, Cachex, :put_new, [cache, value])
```
ruslandoga: There are two race conditions between CacheBuster and this `put_new` call (negative lookups and updates/deletes). I'll try adding something like tombstones ~~or cuckoo/bloom filter/custom short ttl for broadcasted entries~~ to handle it.

lib/logflare/context_cache.ex
```diff
@@ -162,4 +163,18 @@ defmodule Logflare.ContextCache do
       end)
     end)
   end
+
+  defp maybe_broadcast_cached(cache, cache_key, value) do
+    Logflare.Utils.Tasks.start_child(fn ->
```
Ziinc: thinking that instead of creating a task for each broadcast, we can have a dedicated genserver for this to avoid the overhead of ephemeral task procs

ruslandoga: ~~I can add a benchmark to compare the two but I'd rather avoid a single process handling cross-node broadcasts. AFAIK `:erpc.multicast` still awaits on `spawn_request` reply, so if one of the nodes connection is slow, it could potentially block all of the cache broadcast from sending and grow the process's message queue.~~

I looked into it a bit more and it seems like Tasks are unnecessary since `multicast` uses [`spawn_request` with `reply=no`](https://github.com/erlang/otp/blob/c8fca7e34e4db5d977ccf292cc2490ddb05b5879/lib/kernel/src/erpc.erl#L1202-L1203) which means it should be returning immediately.

```elixir
# since node=nil doesn't exist, it probably means we are not waiting for `send/2` to finish
iex> :erlang.spawn_request(_bad_node_name = nil, :erpc, :execute_cast, [Function, :identity, [1]], reply: :no)
#==> #Reference<0.2294785123.1844445189.249014>

# and it's pretty much instant
iex> :timer.tc fn -> :erlang.spawn_request(_bad_node_name = nil, :erpc, :execute_cast, [Function, :identity, [1]], reply: :no) end
#==> {12, #Reference<0.2294785123.1844445189.249084>}
```

Side-note: maybe we can add global cache broadcast rate limiting similar to https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/logger_handler/rate_limiter.ex to avoid sending out too many messages, just as a pre-caution.

config/runtime.exs
```diff
@@ -474,3 +474,49 @@ config :logflare, Oban,
...
+default_cache_broadcast_enabled = if config_env() == :test, do: "true", else: "false"
+
+cache_broadcast_enabled? =
+  System.get_env("LOGFLARE_CACHE_BROADCAST_ENABLED", default_cache_broadcast_enabled) == "true"
+
+cache_broadcast_ratio =
+  "LOGFLARE_CACHE_BROADCAST_RATIO" |> System.get_env("0.1") |> String.to_float()
+
+cache_broadcast_max_nodes =
+  "LOGFLARE_CACHE_BROADCAST_MAX_NODES" |> System.get_env("5") |> String.to_integer()
```
ruslandoga: I'm not sure how it should be configured. For now I went with "global" env vars that configure all context caches with the same values.

lib/logflare/cluster/utils.ex
```diff
@@ -6,6 +6,19 @@ defmodule Logflare.Cluster.Utils do
     [Node.self() | Node.list()]
   end
 
+  @spec peer_list_partial(float, pos_integer) :: [Node.t()]
+  def peer_list_partial(ratio, max_nodes) do
```
ruslandoga, line 14: Not sure if this function belongs here or in ContextCache.

lib/logflare/context_cache/supervisor.ex
```diff
@@ -76,6 +76,33 @@ defmodule Logflare.ContextCache.Supervisor do
     Enum.map(list_caches_with_metrics(), fn {cache, _} -> cache end)
   end
 
+  defp wal_tombstone_specs do
```
ruslandoga: I didn't want to create another module for it since it's not like a normal ContextCache. So I wrote the spec here. This "cache" is used to remember recently deleted records so that they are dropped from "cache broadcasts" to avoid race conditions (record fetch miss, lookup -> cache broadcast -> record deleted -> wal broadcast -> wal broadcast arrives faster -> cache broadcast arrives later).

test/logflare/context_cache_test.exs
```diff
@@ -78,4 +78,92 @@ defmodule Logflare.ContextCacheTest do
       assert_received %Cainophile.Changes.Transaction{}
     end
   end
+
+  describe "broadcasts" do
```
ruslandoga: I can also add a more realistic test similar to Phoenix PubSub [distributed tests](https://github.com/phoenixframework/phoenix_pubsub/blob/main/test/phoenix/distributed_pubsub_test.exs) where extra nodes are started and connected and "cache broadcast" side-effects are tested on them. Or I can use mocks.

config/runtime.exs
```diff
@@ -474,3 +474,49 @@ config :logflare, Oban,
...
 
 config :logflare, Logflare.Alerting, enabled: enable_alerting?
+
+default_cache_broadcast_enabled = if config_env() == :test, do: "true", else: "false"
+
+cache_broadcast_enabled? =
+  System.get_env("LOGFLARE_CACHE_BROADCAST_ENABLED", default_cache_broadcast_enabled) == "true"
+
+cache_broadcast_ratio =
+  "LOGFLARE_CACHE_BROADCAST_RATIO" |> System.get_env("0.1") |> String.to_float()
```
Ziinc: lets hardcode this value in the logic and let the user set the max broadcast nodes

Ziinc: default of 0.05 would be better (5% of cluster)
so in a 100 node cluster, at most 5 would receive an update (assuming user sets a high max)

config/runtime.exs
Ziinc: lets default to 3 nodes.
`LOGFLARE_CACHE_GOSSIP_MAX_NODES`

config/runtime.exs
Ziinc: `LOGFLARE_CACHE_GOSSIP_RATIO`

config/runtime.exs
```diff
@@ -474,3 +474,49 @@ config :logflare, Oban,
...
+  Logflare.ContextCache.Supervisor.list_caches_with_metrics()
+  |> Enum.map(fn {name, short_name} -> {Atom.to_string(short_name), name} end)
+  |> Map.new()
+
+# allows excluding heavy caches via: LOGFLARE_CACHE_BROADCAST_EXCLUDE="auth,saved_searches"
+excluded_caches =
+  System.get_env("LOGFLARE_CACHE_BROADCAST_EXCLUDE", "")
+  |> String.split(",", trim: true)
+  |> Enum.map(&String.trim/1)
+  |> MapSet.new()
```
Ziinc: i don't think having an explicit exclusion is a good approach, would be too much configuration for this which should be a transparent optimization.

config/runtime.exs
```diff
@@ -474,3 +474,49 @@ config :logflare, Oban,
...
+    config = %{
+      ratio: cache_broadcast_ratio,
+      max_nodes: cache_broadcast_max_nodes,
+      enabled: cache_broadcast_enabled? and not MapSet.member?(excluded_caches, short_name)
+    }
+
+    {name, config}
+  end)
+
+config :logflare, :cache_broadcasts, cache_broadcasts
```
Ziinc: storing the ratio and the max_nodes on app env would be enough, this seems unnecessary. all caches should be broadcasted.

lib/logflare/cluster/utils.ex
Ziinc: here is good.
maybe `node_list_random/2` might be a better name (to be similar to `node_list_all/0`

lib/logflare/context_cache.ex
```diff
@@ -159,4 +166,111 @@ defmodule Logflare.ContextCache do
...
+  # Negative lookups (`nil` or `[]`) are not broadcasted. If Node A caches `nil`,
+  # and the record is immediately created, a delayed `nil` broadcast to Node B
+  # would cause phantom "not found" lookups while the record actually exists in the database.
+  defp maybe_broadcast(_cache, _key, nil), do: :ok
+  defp maybe_broadcast(_cache, _key, []), do: :ok
+
+  # Explicitly ignore high-volume/ephemeral caches
+  defp maybe_broadcast(Logflare.Logs.LogEvents.Cache, _key, _value), do: :ok
+  defp maybe_broadcast(Logflare.Logs.RejectedLogEvents, _key, _value), do: :ok
+  defp maybe_broadcast(Logflare.PubSubRates.Cache, _key, _value), do: :ok
```
Ziinc: these caches do not go through the context cache, we can remove

lib/logflare/context_cache.ex
```diff
@@ -159,4 +166,111 @@ defmodule Logflare.ContextCache do
...
+  @doc false
+  def receive_broadcast(cache, key, value) do
+    meta = %{cache: cache}
+
+    :telemetry.span([:logflare, :context_cache, :receive_broadcast], meta, fn ->
+      action =
+        cond do
+          # do nothing if the node already has this cache key
+          Cachex.exists?(cache, key) == {:ok, true} ->
+            :dropped_exists
```
Ziinc: should check if values are identical, then bump the ttl

lib/logflare/context_cache.ex
```diff
@@ -159,4 +166,111 @@ defmodule Logflare.ContextCache do
...
+            Cachex.put(cache, key, {:cached, value})
+            :cached
+        end
+
+      {:ok, Map.put(meta, :action, action)}
+    end)
+  end
+
+  @doc false
+  def record_tombstones(context_pkeys) when is_list(context_pkeys) do
```
Ziinc: moving tombstoning/gossip broadcasting logic to a separate module like ContextCache.Gossip is better since wal tombstone is only used in that context.

lib/logflare/context_cache/supervisor.ex
Ziinc: a separate ContextCache.WalTombstoneCache would probably be better, we have a few ephemeral caches like these as you have already identified so it is fine.
They're differentiated by the `.Cache` suffix.

lib/telemetry.ex
```diff
@@ -7,11 +7,13 @@ defmodule Logflare.Telemetry do
   def start_link(arg), do: Supervisor.start_link(__MODULE__, arg, name: __MODULE__)
 
   context_caches_with_metrics = Logflare.ContextCache.Supervisor.list_caches_with_metrics()
+  wal_tombstones = Logflare.ContextCache.wal_tombstones_cache_name()
```
Ziinc: separate module name would make this easier.

lib/telemetry.ex
```diff
@@ -225,6 +227,26 @@ defmodule Logflare.Telemetry do
...
+      ),
+      counter("logflare.context_cache.receive_broadcast.count",
+        event_name: "logflare.context_cache.receive_broadcast.stop",
+        tags: [:cache, :action],
+        description: "Total cache broadcasts received and their outcome (cached or dropped)"
+      ),
+      distribution("logflare.context_cache.receive_broadcast.stop.duration",
+        tags: [:cache, :action],
+        unit: {:native, :millisecond},
+        description: "Latency of processing an incoming cache broadcast"
```
Ziinc: should make clear in event_name that it is for the local `gossip` mechanism of the ContextCache on fetching nodes, and not necessarily from the node with TransactionBroadcaster that does broadcasting of wal updates.
clarification is needed due to upcoming changes in [here](https://github.com/Logflare/logflare/pull/3184)  where the TransactionBroadcaster will be broadcasting updated values with the wal.

test/logflare/context_cache_test.exs
Ziinc: a distributed test harness would be good for this and other broadcasting tests.
i had tried to use [LocalCluster](https://hexdocs.pm/local_cluster/LocalCluster.html) before for this but couldn't get a nice testing flow.

Ziinc: would prefer avoiding mocks where possible

Ziinc (changes requested):

config/runtime.exs
ruslandoga: I wonder if we really want to name it gossip. In the current implementation it's not really a gossip, it's just a one-hop partial broadcast. And from the discussion on Slack we don't actually want real gossiping since it would result in more network load due to redundant messaging (at least in my current understanding).

lib/logflare/cluster/utils.ex
ruslandoga: It's different in `node_list_all` in that it's not listing itself, so I tried to highlight that with `peer` in the name. But I guess it's not that important.

lib/logflare/context_cache.ex
ruslandoga: Somehow `maybe_broadcast/3` was called for at least `Logflare.Logs.LogEvents.Cache` in tests. I'll double check.

config/runtime.exs
Ziinc: for initial implementation we're looking at 1 hop, but if network messaging does not go through the roof then multi-hop is ideal due to benefits of the cluster size being large.
having the hop count, hop probability, and the max message throughput as env var configs would be good so that we can experiment with this.

lib/logflare/cluster/utils.ex
Ziinc: ah i see, would add `@doc` on use case if that is so.

for multi-hop scenario, perhaps we could partition the node list and send that with the message, then the receiver will know which partition to not select and broadcast to, to avoid duplication of messages. wonder if it would be worthwhile 🤔 
just a thought.

lib/telemetry.ex
ruslandoga: The metrics now have `context_cache_gossip` in their name

```elixir
"logflare.context_cache_gossip.multicast.count"
"logflare.context_cache_gossip.multicast.stop.duration"
"logflare.context_cache_gossip.receive.count"
"logflare.context_cache_gossip.receive.stop.duration"
```

lib/telemetry.ex
ruslandoga: I've added `Logflare.ContextCache.Tombstones.Cache`

lib/logflare/context_cache.ex
ruslandoga: I removed the clauses for `Logflare.Logs.RejectedLogEvents` and `Logflare.PubSubRates.Cache` but kept the `Logflare.Logs.LogEvents.Cache`

config/runtime.exs
ruslandoga: Now it's

```elixir
config :logflare, :context_cache_gossip, %{
  enabled: cache_gossip_enabled?,
  ratio: cache_gossip_ratio,
  max_nodes: cache_gossip_max_nodes
}
```

config/runtime.exs
ruslandoga: Removed.

config/runtime.exs
ruslandoga: I've updated the defaults and renamed the env vars.

test/logflare/context_cache_test.exs
ruslandoga: I've added `Logflare.ContextCache.GossipClusterTest`, it's a bit clunky but seems to work for this use case.

test/logflare/context_cache/gossip_test.exs
```diff
@@ -0,0 +1,128 @@
...
+  use Logflare.DataCase, async: false
+  import Logflare.Factory
+
+  alias Logflare.ContextCache.Tombstones
+  alias Logflare.Sources
+
+  @moduletag :cluster
+
+  setup_all do
+    System.cmd("epmd", ["-daemon"])
```
ruslandoga: I am not sure about this and open to suggestions :)

lib/logflare/context_cache.ex
ruslandoga: I've added a call to `Cachex.refresh` here. Not sure if that or `Cachex.touch` should be used or both, but judging just by the docs, refresh is the one that bumps the ttl.

lib/logflare/context_cache.ex
ruslandoga: I've extracted the new logic into `ContextCache.Gossip` in https://github.com/Logflare/logflare/pull/3218/commits/e62958e751e67354f792c9fa14731fce39c1e2e9.

test/logflare/context_cache/gossip_test.exs
ruslandoga: I added `epmd -daemon` to CI before tests, and added a big error banner when net_kernel fails to start in the gossip/cluster tests.

lib/logflare/context_cache/gossip.ex
```diff
@@ -0,0 +1,120 @@
...
+
+  # Negative lookups (`nil` or `[]`) are not gossiped. If Node A caches `nil`,
+  # and the record is immediately created, a delayed `nil` cast to Node B
+  # would cause phantom "not found" lookups while the record actually exists in the database.
+  def maybe_gossip(_cache, _key, nil), do: :ok
+  def maybe_gossip(_cache, _key, []), do: :ok
+
+  # Explicitly ignore high-volume/ephemeral caches
+  def maybe_gossip(Logflare.Logs.LogEvents.Cache, _key, _value), do: :ok
+  def maybe_gossip(Logflare.Auth.Cache, _key, _value), do: :ok
```
ruslandoga: I've added auth cache here as well, but not sure about it. For one thing it has unusual return types like `{:ok, value, value}`, and it's scary to mis-cache those.

lib/logflare/context_cache/gossip.ex
```diff
@@ -0,0 +1,120 @@
...
+
+  defp extract_pkeys(values) when is_list(values) do
+    Enum.flat_map(values, &extract_pkeys/1)
+  end
+
+  defp extract_pkeys({:ok, value}), do: extract_pkeys(value)
+  defp extract_pkeys(%{id: id}), do: [id]
+
+  defp extract_pkeys(v) do
+    Logger.warning("Unable to extract primary key from gossip value: #{inspect(v)}")
```
ruslandoga: Some values I've noticed that we can't extract primary keys for (and therefore can't tombstone/prevent race conditions):

- ` [{"event_message", [{{:string_contains, "testing"}, {:route, {1312, 0}}}]}]` from Rules cache at key `{:rules_tree_by_source_id, [12104]}` -- what's the rule id?
- `{:error, :token_not_found}` for Auth cache at key `{:verify_access_token, ["vIby58vR7h", []]}` -- looks like a memoized call, should those be gossiped at all?
- `{:ok, %Logflare.OauthAccessTokens.OauthAccessToken{}, %Logflare.User{}}` for Auth cache at key `{:verify_access_token, ["940e5d63954fa0e99f2dc2ae0fa1d032f5d07cda655459b4b2f2fc4df6d255d4", ["private"]]}` -- is tombstone for user or token?

I'll prevent Auth and Rules caches from gossiping for now since it's unclear how best to extract pkeys for tombstoning.

lib/logflare/context_cache/gossip.ex
```diff
@@ -0,0 +1,143 @@
...
+  require Logger
+  require Cachex.Spec
+
+  alias Logflare.Cluster.Utils, as: ClusterUtils
+  alias Logflare.ContextCache
+  alias Logflare.ContextCache.Tombstones
+
+  @telemetry_handler_id "context-cache-gossip-logger"
+
+  def attach_logger do
```
ruslandoga, line 14: This can aid troubleshooting or just checking in with the node. The logs I'm adding there are probably too noisy to be on full-time. These functions can be removed later.

lib/logflare/context_cache/gossip.ex
```diff
@@ -0,0 +1,188 @@
...
+          :cached ->
+            Logger.notice("Cached gossip for #{cache} #{inspect(key)} in #{duration}ms")
+
+          :refreshed ->
+            Logger.notice("Refreshed gossip for #{cache} #{inspect(key)} in #{duration}ms")
+        end
+    end
+  end
+
+  def multicast(Cachex.Spec.cache(name: cache), key, value) do
```
Ziinc, line 81: needs @doc

lib/logflare/context_cache/gossip.ex
```diff
@@ -0,0 +1,188 @@
...
+  end
+
+  defp pkeys_from_cached_value({:ok, value}), do: pkeys_from_cached_value(value)
+  defp pkeys_from_cached_value(%{id: id}), do: [id]
+  defp pkeys_from_cached_value(_value), do: []
+
+  @doc false
+  def record_tombstones(context_pkeys) when is_list(context_pkeys) do
+    # Writes a short-lived marker for a primary key indicating it was recently updated or deleted.
+    # Incoming cache multicasts check this tombstone cache to determine if their payload could be stale.
```
Ziinc, line 166: inline docs should be `@doc`

test/logflare/context_cache/gossip/distributed_test.exs
```diff
@@ -0,0 +1,132 @@
+defmodule Logflare.ContextCache.Gossip.DistributedTest do
+  use Logflare.DataCase, async: false
+  import Logflare.Factory
+
+  alias Ecto.Adapters.SQL.Sandbox, as: EctoSandbox
+  alias Logflare.{ContextCache, Users}
```
Ziinc, line 6:
```suggestion
  alias Logflare.ContextCache
  alias Logflare.Users
```

test/logflare/context_cache/gossip/distributed_test.exs
```diff
@@ -0,0 +1,132 @@
...
+          Failed to start distributed Erlang for tests.
+
+          Please make sure `epmd` is running:
+
+              epmd -daemon
+
+          =============================================
+          Underlying error: #{inspect(reason)}
+          """
+      end
```
Ziinc, line 28: i think we should teardown `:net_kernel` to avoid funky behaviour in other tests. it is fine for now, but would not be the case if we were to introduce more similar distributed testing

Ziinc: also, tearing it down will _slightly_ improve ci test speed

Ziinc (approved):

lib/logflare/context_cache/gossip.ex
ruslandoga: Added in https://github.com/Logflare/logflare/pull/3375

lib/logflare/context_cache/gossip.ex
ruslandoga: Moved in https://github.com/Logflare/logflare/pull/3375

test/logflare/context_cache/gossip/distributed_test.exs
ruslandoga: Adding teardown in https://github.com/Logflare/logflare/pull/3376

PR #3219, opened 2026-02-27
Adjustment to timestamp handling logic in `LogEvent.make`

amokan:
Within `LogEvent.make` we currently set the timestamp attribute to the current time if it was not provided or was unable to be parsed. There are cases in upstream processing where having a signal of this occurrence is important for mapping decisions (_otel traces being a big one_).

This PR adds a new boolean attribute to the LogEvent struct named `timestamp_inferred` that allows us to understand if the timestamp was created in Logflare or if it came with the event body. So in a case where the timestamp is being set on the Logflare side - this attribute will be `true`. If the timestamp was parsed from the incoming payload - the attribute is `false`

I also did some tidying within the LogEvent module - adding typespecs and reordering the require, import, and alias statements at the top.

### DIFF
--- a/test/logflare/log_event_test.exs
+++ b/test/logflare/log_event_test.exs
@@ -342,6 +342,55 @@ defmodule Logflare.LogEventTest do
     assert le.body["message"] == nil
   end
 
+  describe "timestamp_inferred flag" do
+    test "is true when no timestamp is provided", %{source: source} do
+      le = LogEvent.make(%{"event_message" => "test"}, %{source: source})
+      assert le.timestamp_inferred == true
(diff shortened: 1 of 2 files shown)


test/logflare/log_event_test.exs
```diff
@@ -342,6 +342,55 @@ defmodule Logflare.LogEventTest do
     assert le.body["message"] == nil
   end
 
+  describe "timestamp_inferred flag" do
```
Ziinc, line 345: probably can consolidate this into one or two tests instead of so many 👀

PR #3222, opened 2026-03-02
perf: initial keyvalues cache replication

Ziinc:
- replicates the entire KeyValues table into memory via Cachex
- cache performs periodic warming to sync recently inserted values

### DIFF
--- a/lib/logflare/key_values/cache_warmer.ex
+++ b/lib/logflare/key_values/cache_warmer.ex
@@ -44,7 +44,7 @@ defmodule Logflare.KeyValues.CacheWarmer do
   def warm_recent do
     entries =
       KeyValue
-      |> where([kv], kv.inserted_at >= ago(1, "hour"))
+      |> where([kv], kv.updated_at >= ago(1, "hour"))
       |> Repo.all()
       |> Enum.map(&to_cache_entry/1)
 


lib/logflare/key_values.ex
```diff
@@ -116,7 +116,7 @@ defmodule Logflare.KeyValues do
       end)
 
     Repo.insert_all(KeyValue, rows,
-      on_conflict: {:replace, [:value]},
+      on_conflict: {:replace, [:value, :updated_at]},
```
amokan, line 123: I _think_ this will set `updated_at` to NULL on an upsert as `Repo.insert_all` did not auto-generate timestamps on conflict last I recall.

May need to adjust the logic above to do something like:

```
now = DateTime.utc_now()
rows =
  Enum.map(entries, fn entry ->
    %{
      user_id: user_id,
      key: entry[:key] || entry["key"],
      value: entry[:value] || entry["value"],
      inserted_at: now,
      updated_at: now
    }
  end)
```

lib/logflare/key_values/cache_warmer.ex
```diff
@@ -0,0 +1,55 @@
...
+        entries = Enum.map(chunk, &to_cache_entry/1)
+        Cachex.put_many(Cache, entries)
+      end)
+    end)
+  end
+
+  defp warm_recent do
+    entries =
+      KeyValue
+      |> where([kv], kv.inserted_at >= ago(1, "hour"))
```
amokan: Should this be looking at `updated_at` rather than `inserted_at` ?

lib/logflare/key_values/cache_warmer.ex
```diff
@@ -0,0 +1,55 @@
...
+  import Ecto.Query
+
+  @pt_key {__MODULE__, :initialized}
+
+  @impl true
+  def execute(_state) do
+    if initialized?() do
+      warm_recent()
+    else
+      warm_full()
```
amokan: Worth noting that if `warm_full` fails - we still set the persistent term and then only do `warm_recent` after. May be worth checking the result of `warm_full` ?

PR #3229, opened 2026-03-03
fix: redaction bug structs

Ziinc:
Properly fixes redaction of Tesla structs, adds Backend struct redaction.


lib/logflare/utils.ex
```diff
@@ -336,7 +337,8 @@ defmodule Logflare.Utils do
   """
   @spec redact_sensitive_headers(map()) :: list(tuple())
```
amokan, line 338: Looks like this typespec needs updated, but let's see if dialyzer catches it before doing anything. I can adjust it, if so.

PR #3230, opened 2026-03-04
feat: quick search fields inline with query search field

msmithstubbs:
Moves quick search filters inline with query field. Labels are truncated with the full field name available as a tool tip.

CleanShot 2026-03-04 at 11 55 24@2x

CleanShot 2026-03-04 at 12 08 35@2x

## mobile layout

CleanShot 2026-03-04 at 14 37 18@2x

fixes O11Y-1536


Ziinc:
Looks good. think we can get it in today?

msmithstubbs:
> There is a minor issue with mobile layout but happy to fix in a follow up PR so we can merge this.

Fixed.

PR #3239, opened 2026-03-05
feat: Monaco editor with LQL support for search field

msmithstubbs:
Use the Monaco editor component for the search field

* Replaces the search field with a `LiveMonacoEditor/1` component
* Syntax highlight for LQL
* Autocomplete for LQL operators, aggregates, source fields, and saved searches 
* Add vitest for JS testing framework (currently just covers `lql_language.js`)

I chose to replace the search `<form>` element and have the monaco editor instance communicate using liveview hooks and events. This simplified keeping the editor in sync with a hidden text input.

Stacked on #3306 and #3307 

Fixes O11Y-1309

**LQL syntax highlighting**
 
CleanShot 2026-03-05 at 15 54 54@2x

**Supports aggregate completions**

CleanShot 2026-03-10 at 11 49 12@2x

**Saved searches are included in autocomplete**

CleanShot 2026-03-10 at 11 47 36@2x

**Source query and recommended fields adopt similar styling**

Same background and text colours, but not Monaco editor.

CleanShot 2026-03-12 at 14 15 03@2x
CleanShot 2026-03-12 at 14 15 30@2x

https://github.com/user-attachments/assets/bd99ddfa-0bd0-4095-852d-623f22e552f5


Ziinc:
> What about suggested searches? We use have a <datalist> for those currently, but they could be pulled into monaco editor's autocomplete, too?

Yes the demo looks great, perhaps for the first character, we can limit it to saved searches only

CleanShot 2026-03-12 at 13 15 54@2x

i think we can consolidate `metadata:` into `m:` and maybe add a helper description so something like `m:   (alias metadata:)` and for  `timestamp:` into `t:` too, since shorthand is more commonly used.

Probably can also consolidate the `c:group_by` together as well.

should be `s:` instead of `s:*`, don't want to encourage users to select the world, would be too many fields and bad for ux.

assets/js/lql_language.js
```diff
@@ -0,0 +1,503 @@
...
+  monaco.languages.setLanguageConfiguration("lql", {
+    wordPattern: /[a-zA-Z_]\w*/,
+  });
+}
+
+const LQL_KEYWORDS = [
+  { label: "t:", detail: "timestamp filter", insertText: "t:" },
+  { label: "timestamp:", detail: "timestamp filter", insertText: "timestamp:" },
+  { label: "m.", detail: "metadata field", insertText: "m." },
+  { label: "metadata.", detail: "metadata field", insertText: "metadata." },
```
Ziinc: as mentioned, these can be consolidated

assets/js/lql_language.js
```diff
@@ -0,0 +1,503 @@
...
+  { label: "s:*", detail: "select all fields", insertText: "s:*" },
```
Ziinc:
```suggestion
  { label: "s:", detail: "select all fields", insertText: "s:" },
```

assets/js/lql_language.js
```diff
@@ -0,0 +1,503 @@
...
+  { label: "c:count(*)", detail: "chart count", insertText: "c:count(*)" },
```
Ziinc:
```suggestion
  { label: "c:count()", detail: "chart count", insertText: "c:count($0)" },
```

assets/js/lql_language.js
Ziinc, line 1: can we get this file's logic covered with tests? using vitest would be great, but whatever works.

Ziinc: too much chance for breakage for such an important component.

assets/js/lql_editor_wrapper_hook.js
```diff
@@ -0,0 +1,174 @@
...
+        return;
+      }
+
+      this.disposeEditorBindings();
+      this._editor = standaloneEditor;
+      this.applyPendingEditorValue();
+
+      const monaco = window.monaco;
+
+      registerLqlLanguage(monaco);
```
Ziinc: should wrap this in a try/do to prevent mount failure.

Ziinc (changes requested):
considering this is a very important flow, lets add an integration test using PhoenixTest that executes a search using this.

would likely need to add postgres lql support for it to run on CI. can be separate PR with this stacked.

assets/js/source_lv_hooks.js
```diff
@@ -161,43 +164,68 @@ hooks.SourceLogsSearch = {
...
-    $daterangepicker.on("cancel.daterangepicker", (e, picker) => {
+    $daterangepicker.on("cancel.daterangepicker.sourceLogsSearch", () => {
       hook.pushEvent("soft_play", {})
     })
 
+    $daterangepicker.on("show.daterangepicker.sourceLogsSearch", () => {
+      hook.pushEvent("soft_pause", {})
+    })
+  },
+  pushChartControlsUpdate() {
```
Ziinc, line 175: should cover this hook with an integration test as well.

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -1192,4 +1183,27 @@ defmodule LogflareWeb.Source.SearchLV do
...
+    case SourceSchemas.Cache.get_source_schema_by(source_id: source.id) do
+      %{schema_flat_map: flat_map} when is_map(flat_map) ->
+        Enum.map(flat_map, fn {name, type} ->
+          %{name: name, type: format_schema_type(type)}
+        end)
+
+      _ ->
+        []
+    end
+  end
```
Ziinc: can't we pass the flatmap directly? or this needs to be done elixir side?

Ziinc (changes requested):

assets/js/lql_editor_wrapper_hook.js
msmithstubbs: Wrapped `registerLqlLanguage` to catch failure. Was that where you think we'd get an exception? Or are you concerned monaco might not be loaded at all?

lib/logflare_web/live/search_live/logs_search_lv.ex
msmithstubbs: We can pass the flatmap as is if we only want the field names. If we'd like the field types, too, we need to encode any tuples (such as `{:list, :string}`) because they can't currently be JSON encoded.

I've simplified the schema fields helper, and moved it to FormComponents which is a little bit simpler because it reduces passing another assign and attribute. I don't love that a helper in FormComponents is fetching a record which might hit the db, but it's cached so shouldn't be a problem in practice.

Ziinc: i think would be better to move the fetch out of the template and into the liveview. feels like a code smell, plus we can consolidate the calls (liveview also stores source schema in state)

msmithstubbs: Agree with that. I couldn't see anywhere SearchLV stores source schema currently but there would be some refactoring if we stack this on top of #3120

assets/js/source_lv_hooks.js
msmithstubbs: Will include in the follow up integration test PR. 

That should be an integration test with Playwright in `test/e2e/features/` ?

Ziinc: Yes should cover it with an integration test with playwright

assets/package.json
```diff
@@ -3,6 +3,7 @@
   "license": "MIT",
   "scripts": {
     "deploy": "node ./build.mjs",
+    "test": "vitest run",
```
Ziinc, line 6: Need to add this to ci

Ziinc (approved):

assets/package.json
msmithstubbs: Already added:

https://github.com/Logflare/logflare/blob/d99ab0b7ccc6a752f60e98d214cb418b67812833/.github/workflows/elixir-ci.yml#L184-L185

Would it be it be better in a separate GitHub workflow?

assets/js/source_lv_hooks.js
msmithstubbs: Sure, already has some coverage but will see if I can go further.

PR #3248, opened 2026-03-05
chore: start caching cargo crates and improve dialyzer caching

nelsonmestevao:
Closes O11Y-1541.

### DIFF
--- a/.github/workflows/elixir-ci.yml
+++ b/.github/workflows/elixir-ci.yml
@@ -22,13 +22,27 @@ on:
 permissions:
   contents: read
 
-env:
-  MIX_ENV: test
-
 jobs:
+  versions:
+    name: Get versions
+    runs-on: ubuntu-latest
+    outputs:
+      elixir:  ${{ steps.v.outputs.elixir }}
+      erlang:  ${{ steps.v.outputs.erlang }}
+      rust:    ${{ steps.v.outputs.rust }}
+    steps:
+      - uses: actions/checkout@v6
+      - name: Read versions from .tool-versions
+        id: v
+        run: |
+          echo "elixir=$(grep '^elixir '  .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
+          echo "erlang=$(grep '^erlang '  .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
+          echo "rust=$(grep   '^rust '    .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
+
   check:
     name: Checks
     runs-on: ubuntu-latest
+    needs: versions
     strategy:
       matrix:
         commands:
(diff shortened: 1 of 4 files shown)


.github/workflows/elixir-ci.yml
```diff
@@ -84,35 +104,56 @@ jobs:
...
+        with:
+          path: dialyzer
+          key: plt-${{ runner.os }}-${{ needs.versions.outputs.erlang }}-${{ needs.versions.outputs.elixir }}-${{ hashFiles('mix.lock') }}
+          restore-keys: |
+            plt-${{ runner.os }}-${{ needs.versions.outputs.erlang }}-${{ needs.versions.outputs.elixir }}-
+            plt-${{ runner.os }}-${{ needs.versions.outputs.erlang }}-
+
       - name: Install dependencies
-        run: mix do deps.get, deps.compile
+        run: mix do deps.get + deps.compile
```
amokan, line 149: I believe the original comma syntax will short-circuit so that if fetching fails - the compilation will not be attempted.

Please tell me if I'm wrong on that though 😄

nelsonmestevao: I don't think they changed behaviour. I believe it was just to improve for windows users 🙏

nelsonmestevao: Image

https://github.com/elixir-lang/elixir/blob/v1.19.5/CHANGELOG.md#mix-10

https://hexdocs.pm/mix/1.19.5/Mix.Tasks.Do.html

.github/workflows/elixir-ci.yml
```diff
@@ -24,8 +24,28 @@ permissions:
...
+      - name: Read versions from .tool-versions
+        id: v
+        run: |
+          echo "elixir=$(grep '^elixir '  .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
+          echo "erlang=$(grep '^erlang '  .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
+          echo "rust=$(grep   '^rust '    .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
+
   check:
     name: Checks
     runs-on: ubuntu-latest
```
nelsonmestevao, line 44: I am actually not sure. The Github Actions API stopped working 

https://www.githubstatus.com/

I will read the docs :pray:

.github/workflows/elixir-ci.yml
```diff
@@ -24,11 +24,29 @@ permissions:
...
+    needs: versions
```
nelsonmestevao, line 45: Yes, this is needed according to the docs 🙏 

https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#needs-context

nelsonmestevao:
> CleanShot 2026-03-10 at 18 09 25@2x
> 
> CleanShot 2026-03-10 at 18 09 11@2x
> 
> @nelsonmestevao looks like caching isn't working still

The parallel jobs are overriding each others? I will need to be more specific in the keys, I guess. Or maybe does not matter if at least one job does save the cache. I will investigate and do more tests 🙏

nelsonmestevao:
@Ziinc I think this is normal, right?

https://github.com/Logflare/logflare/actions/runs/22893933813/job/66423244054?pr=3248

Only one of the checks in the matrix will be able to store the caching of the mix dependencies, so I believe the first one probably :point_down: 

image

PR #3250, opened 2026-03-05
fix: handle case of source not being in cache

nelsonmestevao:


lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -119,31 +119,35 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
         %{}
       end
 
-    source = Sources.Cache.get_by_id(sid)
```
nelsonmestevao, line 122: image


If we don't handle the case of this being nil, then it will have a bad map error in `Utils.stringify(source.token)` [down below](https://github.com/Logflare/logflare/pull/3250/changes#diff-59842ad33ede05564d481491226a223e0f07d403f29e2356c4c46d88f68fa4deR142) 👇

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -119,31 +119,35 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
-
-    for %{data: le} <- successful do
-      # delete immediately if not default backend or if avg rate is above 100
-      if metrics.avg > 100 or bid != nil do
-        IngestEventQueue.delete(queue, le)
-      end
+    case Sources.Cache.get_by_id(sid) do
+      nil ->
```
amokan, line 123: I think we'll want to delete the items in the queue if nil otherwise they'll not be cleaned up.

So maybe something like

```
Logger.warning("Source not found for ack!", source_id: sid)
for %{data: le} <- successful do
  IngestEventQueue.delete(queue, le)
end
```

lib/logflare/sources/source/bigquery/pipeline.ex
amokan, line 197: We do have the `if source` here - but worth noting that this could also be nil if we needed to handle any logic in here.

amokan: This could be an issue if source is `nil`

lib/logflare/sources/source/bigquery/pipeline.ex
nelsonmestevao: Okay, let me do that then 🙏

lib/logflare/sources/source/bigquery/pipeline.ex
nelsonmestevao: I added a new condition in last push 🙏

lib/logflare/sources/source/bigquery/pipeline.ex
nelsonmestevao: Done in last push 🙏

amokan (approved):

PR #3272, opened 2026-03-11
BQ storage write api - Mint pool

bblaszkow06:
Initial implementation of the Mint-adapter-based Storage Write API.

Retries and error handling will be covered by following PRs to keep this manageable.

Closes O11Y-1513


test/logflare_grpc/logs/server_test.exs
```diff
@@ -8,6 +8,7 @@ defmodule LogflareGrpc.Logs.ServerTest do
   setup do
     insert(:plan)
     start_supervised!(AllLogsLogged)
+    start_supervised!(GRPC.Client.Supervisor)
```
Ziinc, line 11: Seems like unnecessary change as there is no client interaction in this test

test/logflare_grpc/metrics/server_test.exs
```diff
@@ -8,6 +8,7 @@ defmodule LogflareGrpc.Metrics.ServerTest do
   setup do
     insert(:plan)
     start_supervised!(AllLogsLogged)
+    start_supervised!(GRPC.Client.Supervisor)
```
Ziinc, line 11: as above

test/logflare_grpc/trace/server_test.exs
```diff
@@ -8,6 +8,7 @@ defmodule LogflareGrpc.Trace.ServerTest do
   setup do
     insert(:plan)
     start_supervised!(AllLogsLogged)
+    start_supervised!(GRPC.Client.Supervisor)
```
Ziinc, line 12: As above

config/runtime.exs
```diff
@@ -96,7 +96,8 @@ config :logflare,
          encryption_key_retired: System.get_env("LOGFLARE_DB_ENCRYPTION_KEY_RETIRED"),
          metadata: logflare_metadata,
          health: logflare_health,
-         http_connection_pools: http_connection_pools
+         http_connection_pools: http_connection_pools,
+         bq_write_api_pool_size: System.get_env("LOGFLARE_BQ_WRITE_API_POOL_SIZE")
```
Ziinc, line 100:
```suggestion
         bq_write_api_pool_size: System.get_env("LOGFLARE_BIGQUERY_WRITE_API_POOL_SIZE")
```

In my opinion we should automatically manage connection pool. But we can give this a try first. Needs a sensible default too.

Ziinc (changes requested):

lib/logflare/networking.ex
```diff
@@ -148,4 +134,18 @@ defmodule Logflare.Networking do
...
+  defp grpc_pools do
+    if Application.get_env(:logflare, :env) == :test do
+      []
+    else
+      [
+        GRPC.Client.Supervisor,
+        {Logflare.Networking.GrpcPool,
+         name: GoogleApiClient.connetion_pool_name(),
+         url: "https://bigquerystorage.googleapis.com",
+         size: Application.get_env(:logflare, :bq_write_api_pool_size, 25)}
```
Ziinc, line 147: Default should be set at runtime.exs instead

Ziinc (approved):
Will merge in first for testing. Please address comments in follow up PRs 🙏

test/logflare_grpc/logs/server_test.exs
bblaszkow06: Actually, there is - in line 35, there's `Client.connect` that sends a request via localhost.

The change was needed due to `:grpc` version bump. The supervisor used to be started with `:grpc` app, but in the current version, it has to be started manually and isn't started in tests

PR #3308, opened 2026-03-24
test(stripe_controller): add webhook endpoint tests

laraujo7:
Also fix get_payment_method_by/1 typespec to include nil return and minor formatting cleanup

### DIFF
--- a/test/logflare_web/controllers/stripe_controller_test.exs
+++ b/test/logflare_web/controllers/stripe_controller_test.exs
@@ -240,10 +240,8 @@ defmodule LogflareWeb.StripeControllerTest do
 
       assert json_response(conn, 202) == %{"message" => "event type not implimented"}
 
-      assert Billing.get_billing_account_by(stripe_customer: ba.stripe_customer).id == ba.id
-
-      assert Billing.get_billing_account_by(stripe_customer: ba.stripe_customer).lifetime_plan ==
-               ba.lifetime_plan
+      assert Billing.get_billing_account_by(stripe_customer: ba.stripe_customer) ==
+               TestUtils.reset_associations(ba)
     end
 
     test "returns not implemented for unknown event in previous_attributes clause", %{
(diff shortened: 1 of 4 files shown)


test/logflare_web/controllers/stripe_controller_test.exs
```diff
@@ -0,0 +1,352 @@
...
+
+      conn =
+        conn
+        |> put_req_header("content-type", "application/json")
+        |> post(~p"/webhooks/stripe", payload)
+
+      assert json_response(conn, 202) == %{"message" => "conflict"}
+    end
+
+    test "returns not implemented for unknown event type in detach clause", %{
```
Ziinc: The unknown event and catch all tests can be combined and final clause should assert that no changes in fixtures.

test/logflare_web/controllers/stripe_controller_test.exs
```diff
@@ -0,0 +1,352 @@
...
+      conn =
+        conn
+        |> put_req_header("content-type", "application/json")
+        |> post(~p"/webhooks/stripe", payload)
+
+      assert json_response(conn, 202) == %{"message" => "conflict"}
+    end
+  end
+
+  describe "event/2 - unhandled event with customer field" do
```
Ziinc: Can combine with unhandled event tests below.

Ziinc (changes requested):

test/logflare_web/controllers/stripe_controller_test.exs
```diff
@@ -0,0 +1,371 @@
...
+        conn
+        |> put_req_header("content-type", "application/json")
+        |> post(~p"/webhooks/stripe", payload)
+
+      assert json_response(conn, 202) == %{"message" => "event type not implimented"}
+
+      assert Billing.get_billing_account_by(stripe_customer: ba.stripe_customer).id == ba.id
+
+      assert Billing.get_billing_account_by(stripe_customer: ba.stripe_customer).lifetime_plan ==
+               ba.lifetime_plan
```
Ziinc: Why compare on this field specifically? Why not do a before/after struct comparison?

Ziinc (changes requested):

test/logflare_web/controllers/stripe_controller_test.exs
Ziinc: Doing so with a struct comparison would be more comprehensive as well

nelsonmestevao: ~~Wouldn't the id be enough though? I can do a struct comparison if you see it as better in this case 🙏~~

nelsonmestevao: I did full struct comparison and extracted the util `TestUtils.reset_associations/1` 🙏

nelsonmestevao (approved):

Ziinc (approved):

PR #3340, opened 2026-04-03
test: improve SearchOperationsTest postgres coverage

msmithstubbs:
Implement feedback on test improvements from #3307 

* use `TestUtils.setup_single_tenant' macro
* remove unnecessary mocks
* stub exception for postgres error

### DIFF
--- a/lib/logflare/logs/search_operations.ex
+++ b/lib/logflare/logs/search_operations.ex
@@ -73,6 +73,8 @@ defmodule Logflare.Logs.SearchOperations do
     backend = postgres_backend(so)
 
     PostgresAdaptor.execute_query(backend, so.query, query_type: :search)
+  rescue
+    error -> {:error, error}
   end
 
   defp execute_backend_query(%SO{} = so) do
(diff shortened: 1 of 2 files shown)


amokan (commented):
Hey @msmithstubbs :wave:

The test changes look good. My only concern is the addition of the `rescue` block in search operations - which does change production behavior a bit.

If there is a need to include the rescue block, could we at least tighten that up to only rescue `Postgrex.Error`?

msmithstubbs:
> The test changes look good. My only concern is the addition of the `rescue` block in search operations - which does change production behavior a bit.
> 
> If there is a need to include the rescue block, could we at least tighten that up to only rescue `Postgrex.Error`?

Yeah, you're right: we don't need it. I've removed it and updated the test to assert any exceptions are passed through. 

Might be worth considering rescuing *some* exceptions to align the Postgres code path closer to BigQuery but that should be a separate PR.

Ziinc (approved):

PR #3364, opened 2026-04-13
fix: correctly build base image logflare

nelsonmestevao:
`cargo fetch --locked` fails during the Docker base image build because Cargo requires that every workspace member listed in `Cargo.toml` has a valid crate structure — at minimum a `src/lib.rs` (or `src/main.rs`). See https://github.com/rust-lang/cargo/issues/2644#issuecomment-3819570767.

The only alternative other than this, was to start using [cargo-chef](https://github.com/LukeMathWalker/cargo-chef). See how it would work [here](https://lpalmieri.com/posts/fast-rust-docker-builds/).

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -44,4 +44,7 @@ RUN cargo version
 # pre-fetch cargo dependencies for layer caching
 COPY Cargo.toml Cargo.lock ./
 COPY --parents native/*/Cargo.toml ./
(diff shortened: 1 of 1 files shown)


Ziinc:
@nelsonmestevao can you add to the description on why this is needed? 🙏

nelsonmestevao:
Hi @Ziinc, left a comment in Dockerfile and improved PR description :pray:

Ziinc (approved):

PR #3365, opened 2026-04-13
chore: add additional histogram buckets for batch size telemetry

amokan:
The existing histogram buckets for `logflare.backends.pipeline.handle_batch.batch_size` were capped out at 2k. The more recent work around consolidated pipelines for the sake of ClickHouse will use batch sizes as large as 50k - so this change adds some extra bucket sizes to reflect that.


lib/telemetry.ex
```diff
@@ -418,6 +418,26 @@ defmodule Logflare.Telemetry do
...
+        1_000,
+        2_000,
+        5_000,
+        10_000,
+        20_000,
+        25_000,
+        35_000,
+        50_000
+      ]
+    ]
```
chasers: Should we stick to the same number of buckets? I am not sure how many metrics this currently throws off.

chasers (changes requested):

chasers (approved):

PR #3367, opened 2026-04-14
feat: copy log events copy clipboard

msmithstubbs:
Add a copy log events to clipboard button. 

* Copies timestamp and event message in plain text
* Copies directly from DOM to avoid extra page weight or network round trip.

Closes O11Y-1560

CleanShot 2026-04-14 at 13 24 27@2x

## Tests

No tests but after #3239 is merged I could add a E2E feature test with PlaywrightEx

## Demo

https://github.com/user-attachments/assets/178cf08e-bfde-42ed-ade8-65af25c371ee

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -102,6 +102,24 @@ document.addEventListener("DOMContentLoaded", (e) => {
   initLiveReact();
 });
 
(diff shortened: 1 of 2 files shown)


Ziinc (approved):
An integration test would be good, thanks.

PR #3368, opened 2026-04-14
feat: render standard event fields first

msmithstubbs:
JSON viewer will render promoted fields (`id`, `timestamp`, `event_message`) first.

Closes O11Y-1596

CleanShot 2026-04-14 at 16 57 49@2x

### DIFF
--- a/lib/logflare_web/components/json_viewer_component.ex
+++ b/lib/logflare_web/components/json_viewer_component.ex
@@ -15,27 +15,63 @@ defmodule LogflareWeb.JSONViewerComponent do
   attr :data, :any, required: true, doc: "List or map to render as a tree"
   attr :class, :string, default: "", doc: "Additional CSS classes"
   attr :id, :string, default: nil, doc: "Optional prefix for DOM ids"
+
+  attr :promoted_fields, :list,
+    default: ~w(id timestamp event_message),
+    doc: "Top-level fields to render first, in the given order"
+
   attr :rest, :global, doc: "Global attributes"
 
   slot :action
 
   def json_viewer(assigns) do
-    data =
+    {promoted_entries, remaining_entries} =
       case assigns.data do
         list when is_list(list) ->
-          Enum.with_index(list, fn v, index -> {to_string(index), v} end)
+          {[], Enum.with_index(list, fn v, index -> {to_string(index), v} end)}
 
         map when is_map(map) ->
-          map
+          partition_promoted_fields(map, assigns.promoted_fields)
       end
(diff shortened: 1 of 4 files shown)


test/logflare/backends/user_monitoring_test.exs
```diff
@@ -266,7 +266,7 @@ defmodule Logflare.Backends.UserMonitoringTest do
 
       :timer.sleep(2500)
 
-      assert_receive {:insert_all, [%{json: %{"attributes" => _}} | _] = rows}, 5_000
+      assert_receive {:insert_all, [%{json: %{"attributes" => _}} | _] = rows}, 15_000
```
msmithstubbs, line 269: Not part of this feature but CI kept failing.

Ziinc (approved):

test/logflare/backends/user_monitoring_test.exs
Ziinc: @nelsonmestevao can you investigate this? We should also convert the sleep into a retry

lib/logflare_web/components/json_viewer_component.ex
```diff
@@ -15,29 +15,65 @@ defmodule LogflareWeb.JSONViewerComponent do
   attr :data, :any, required: true, doc: "List or map to render as a tree"
   attr :class, :string, default: "", doc: "Additional CSS classes"
   attr :id, :string, default: nil, doc: "Optional prefix for DOM ids"
+
+  attr :promoted_fields, :list,
```
Ziinc: This feels like we're over engineering, a sort_by function would work just as well and would be a lot simpler

Ziinc (changes requested):

Ziinc:
Seems really overly complex for sorting of keys

lib/logflare_web/components/json_viewer_component.ex
msmithstubbs: Will do.

Ziinc (approved):

test/logflare/backends/user_monitoring_test.exs
nelsonmestevao: I am not sure how to change this. I could not make this test pass reliable with a smaller assert receive time. However, I could at least remove the sleep in this [PR](https://github.com/Logflare/logflare/pull/3411).

I will investigate how `ingest_logs` could be more reliable for the tests. I believe we had some similar issue that I worked with @laraujo7.

PR #3370, opened 2026-04-14
fix: trigger RecentInsertsCacher in fetch_latest_timestamp test to avoid flaky timeout

laraujo7:
image

### DIFF
--- a/lib/logflare/backends/recent_inserts_cacher.ex
+++ b/lib/logflare/backends/recent_inserts_cacher.ex
@@ -12,7 +12,7 @@ defmodule Logflare.Backends.RecentInsertsCacher do
 
   require Logger
 
-  @cache_every if Application.compile_env(:logflare, :env) == :test, do: 100, else: 5_000
+  @cache_every 5_000
 
   ## Server
   def start_link(args) do
@@ -114,6 +114,11 @@ defmodule Logflare.Backends.RecentInsertsCacher do
   end
 
   defp schedule_cache do
-    Process.send_after(self(), :do_cache, @cache_every + :rand.uniform(1000))
+    jitter =
+      if Application.get_env(:logflare, :env) == :test,
+        do: 0,
+        else: @cache_every + :rand.uniform(1000)
+
+    Process.send_after(self(), :do_cache, jitter)
   end
 end


Ziinc:
Fixing a flaky test like this isn't addressing the root cause. We should adjust internal intervals and wait times during test env such that async stuff processes faster.

Ziinc (changes requested):
Needs fixing of root cause

laraujo7:
> Needs fixing of root cause


This test is flaky because it depended on a timing-based race condition.

After ingesting a log event, the test asserts that `fetch_latest_timestamp` returns a non-zero value. But this timestamp isn't updated during ingestion — it's updated asynchronously by `RecentInsertsCacher`, a GenServer that polls on a timer (`@cache_every + :rand.uniform(1000)`, which is 101–1100ms in tests).

The test was using `retry_assert` (50ms sleep, 5000ms duration) to wait for the cacher to fire. While 5000ms should theoretically be enough for a 1100ms timer, under heavy CI load the BEAM scheduler can delay `Process.send_after` timers beyond their specified time, causing the assertion to fail.

I came up with another solution, because this one was definitly a bad-aid, and instead of waiting for the cacher's timer to fire, we look up the cacher's PID via `Backends.lookup(RecentInsertsCacher, source)` and call `RecentInsertsCacher.do_cache(cacher_pid)` directly inside the retry loop. This forces the cacher to run its update logic immediately, removing the timing dependency.

djwhitt:
@laraujo7 your analysis of the timer race is correct — `Process.send_after` plus `:rand.uniform(1000)` puts worst case at ~1100ms in test, and under CI load BEAM can push timers past the retry window.

A few issues with the current fix:

1. Adds a permanent public `do_cache/1` to `RecentInsertsCacher` just for the test. The cacher is named-registered via `Backends.via_source/2`, so the test could `send(Backends.via_source(source, RecentInsertsCacher), :do_cache)` without touching the GenServer.

2. `send/2` is async — the test enqueues `:do_cache` and immediately asserts. If the assert fails, `retry_assert` queues another, etc. A synchronous `GenServer.call(:flush)` (or similar) would actually wait for the update to complete instead of racing the mailbox it just poked.

3. The test-env jitter looks unintentional. Before commit b6da32ecb (Oct 2025), the predecessor `broadcast` formula was `max(@broadcast_every, round(:rand.uniform(cluster_size * 200)))` — single-node tests effectively had zero jitter. The refactor flattened that to hard-coded `:rand.uniform(1000)`, picking up 0–1000ms of jitter in test as a side effect. Gating the jitter on `:env` (or dropping it from the test arm of `@cache_every`) is a one-line root-cause fix that handles this test and any other latent flakes leaning on the same timer.

I'd lean toward (3) on its own — no test changes needed.

djwhitt:
Thanks @laraujo7 for the original analysis and fix here — the diagnosis of the timer race was spot-on. I've pulled your commit into #3465 (preserved as d17d67386 with your authorship intact) and stacked an adjustment on top:

- **Root cause** turned out to be upstream: b6da32ecb (Oct 2025) replaced the predecessor `RecentInsertsBroadcaster`'s `max(@cadence, :rand.uniform(cluster_size * 200))` with an unconditional `:rand.uniform(1_000)`, regressing the test envelope from `[100, 200] ms` to `[101, 1100] ms`. The adjustment gates the jitter ceiling on `:env` (100 ms in test, 1_000 ms in prod) instead of dropping the reschedule delay to zero, which restores the pre-refactor envelope without introducing a near-hot-reschedule loop in the test suite.
- **Test-side**: drops `retry_assert` in `caches latest timestamp correctly` in favor of an explicit `send(cacher, :do_cache) + :sys.get_state/1` fence — defense in depth so future tuning of the interval doesn't silently re-tripwire this test.

Closing in favor of #3465.

PR #3374, opened 2026-04-15
feat: Error handling in BigQuery Storage Write API

bblaszkow06:
Closes O11Y-1679

The PR introduces:
* retries on errors with exponential backoff
* tests for GoogleApiClient
* Fix for missing value errors (default missing value interpretation setting)

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -89,7 +89,8 @@ config :logflare,
          metadata: logflare_metadata,
          health: logflare_health,
          http_connection_pools: http_connection_pools,
-         bq_write_api_pool_size: System.get_env("LOGFLARE_BQ_WRITE_API_POOL_SIZE")
+         bq_write_api_pool_size:
+           System.get_env("LOGFLARE_BQ_WRITE_API_POOL_SIZE", "10") |> String.to_integer()
        ]
        |> filter_nil_kv_pairs.()
 
--- a/lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
+++ b/lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
@@ -95,22 +95,20 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor.GoogleApiClient do
   end
 
   defp send_requests(requests, channel) do
-    try do
-      stream = BigQueryWrite.Stub.append_rows(channel)
+    stream = BigQueryWrite.Stub.append_rows(channel)
 
-      stream =
-        Enum.reduce(
-          requests,
-          stream,
-          fn request, stream ->
-            GRPC.Stub.send_request(stream, request)
-          end
-        )
+    stream =
+      Enum.reduce(
+        requests,
+        stream,
+        fn request, stream ->
+          GRPC.Stub.send_request(stream, request)
+        end
+      )
 
-      {:ok, GRPC.Stub.end_stream(stream)}
-    catch
-      :exit, reason -> {:error, Exception.format_exit(reason)}
-    end
+    {:ok, GRPC.Stub.end_stream(stream)}
+  catch
+    :exit, reason -> {:error, Exception.format_exit(reason)}
   end
 
   defp retry_call(requests, retry_attempt) do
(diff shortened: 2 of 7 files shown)


config/runtime.exs
```diff
@@ -89,7 +89,8 @@ config :logflare,
          metadata: logflare_metadata,
          health: logflare_health,
          http_connection_pools: http_connection_pools,
-         bq_write_api_pool_size: System.get_env("LOGFLARE_BQ_WRITE_API_POOL_SIZE")
+         bq_write_api_pool_size:
+           System.get_env("LOGFLARE_BQ_WRITE_API_POOL_SIZE", "10") |> String.to_integer()
```
Ziinc, line 93: Feels odd that we need to dictate the pool size, as now we manage it dynamically. Should also scale it dynamically, otherwise this would mean that we have a hard limit per box. 

```suggestion
           System.get_env("LOGFLARE_BIGQUERY_WRITE_API_POOL_SIZE", "10") |> String.to_integer()
```
Perhaps poolboy's bursting behaviour might be ideal here for storage write api in the event that we saturated the grpc connections and need more

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
```diff
@@ -51,62 +51,102 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor.GoogleApiClient do
...
+      )
+
+    {:ok, GRPC.Stub.end_stream(stream)}
+  catch
+    :exit, reason -> {:error, Exception.format_exit(reason)}
+  end
+
+  defp retry_call(requests, retry_attempt) do
+    # 200, 400, 800, 1600, 2000, ...
+    retry_wait = min(2000, 200 * 2 ** retry_attempt)
```
Ziinc, line 116:
```suggestion
    retry_wait = min(1000, 100 * 2 ** retry_attempt)
```
2s blocking the batcher would not be good for queueing and memory

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
```diff
@@ -51,62 +51,102 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor.GoogleApiClient do
...
+        err ->
+          OpenTelemetry.Tracer.set_status(:error, inspect(err))
           err
       end
     end
+    |> case do
+      :ok ->
+        :ok
+
+      failed_reqs when is_list(failed_reqs) and retry_attempt < 5 ->
```
Ziinc, line 86:
```suggestion
      failed_reqs when is_list(failed_reqs) and retry_attempt < 3 ->
```

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
```diff
@@ -51,62 +51,102 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor.GoogleApiClient do
...
+      {request, {:ok, %AppendRowsResponse{response: {:error, %{message: msg}}}}}, acc ->
+        Logger.warning("Storage Write API AppendRows response with error msg - #{inspect(msg)}")
+
+        [request | acc]
+
+      {_request, {:ok, %AppendRowsResponse{row_errors: []}}}, acc ->
+        acc
+
+      {_request, {:ok, %AppendRowsResponse{row_errors: errors}}}, acc when is_list(errors) ->
+        Logger.warning("Storage Write API AppendRows row errors - #{inspect(errors)}")
```
Ziinc, line 139: Should use :error_string field

lib/logflare/backends/adaptor/bigquery_adaptor/google_api_client.ex
```diff
@@ -51,62 +51,102 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor.GoogleApiClient do
...
+    requests
+    |> Stream.zip(responses)
+    |> Enum.reduce([], fn
+      {request, {:error, response}}, acc ->
+        Logger.error("Storage Write API AppendRows response error - #{inspect(response)}")
+
+        [request | acc]
+
+      {request, {:ok, %AppendRowsResponse{response: {:error, %{message: msg}}}}}, acc ->
+        Logger.warning("Storage Write API AppendRows response with error msg - #{inspect(msg)}")
```
Ziinc, line 131: Should use error_string field as this is potentially long

Ziinc (approved):

Ziinc:
Needs some tweaks but look good. 
We can omit the env var from docs until this is stable

Ziinc:
Please handle.comments in follow-up

PR #3378, opened 2026-04-15
chore: scheduler name utilization fix

Ziinc:
fixes scheduler utilization naming for related telmeetry


ruslandoga:
Sorry!

I wonder if we could use something like https://hexdocs.pm/telemetry_metrics/1.1.0/Telemetry.Metrics.ConsoleReporter.html to make catching these mistakes locally easier? Maybe it could be adapted to tests as well, where we would attach it in a test and await for metric to arrive.

Related: https://github.com/beam-telemetry/telemetry_metrics/issues/129

Ziinc:
@ruslandoga possible, though i don't think that the output of ConsoleReporter would get captured by `ExUnit.CaptureLogs`

PR #3379, opened 2026-04-15
feat: Add message truncation in Syslog

ruslandoga:
closes O11Y-1158

Context: https://github.com/Logflare/logflare/pull/2985#discussion_r2944367247


lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex
```diff
@@ -95,14 +92,47 @@ defmodule Logflare.Backends.Adaptor.SyslogAdaptor.Syslog do
...
+
+    {message, message_length} =
+      if cipher_key do
+        message = encrypt(message, cipher_key)
+        {message, IO.iodata_length(message)}
+      else
+        {message, message_length}
+      end
+
+    [Integer.to_string(headers_length + message_length), ?\s, headers | message]
```
ruslandoga, line 114: The octet-counting prefix (`MSG-LEN` and the space) is not counted in the message size limit.

RFC 6587 defines that prefix as "transport framing". Receivers would probably strip it before applying their size limits to the actual `<PRI>...` payload.

ruslandoga:
Oh, forgot to update the docs.

---

**UPDATE:** added `max_message_bytes` description to the docs.

docs/docs.logflare.com/docs/backends/syslog.mdx
```diff
@@ -24,6 +24,7 @@ The following values are required when creating a Syslog backend:
 The following values are optional:
 
 - `tls`: (`boolean`) If true, connects via SSL/TLS. Defaults to false (TCP).
+- `max_message_bytes`: (`integer`) Maximum size of the syslog message frame in bytes. If a log exceeds this limit, the JSON body is truncated to fit. If the headers alone exceed this limit, the body is dropped entirely.
```
Ziinc: should included default value here.

Ziinc (changes requested):

docs/docs.logflare.com/docs/backends/syslog.mdx
ruslandoga:
```suggestion
- `max_message_bytes`: (`integer`) Maximum size of the syslog message frame in bytes. If a log exceeds this limit, the JSON body is truncated to fit. If the headers alone exceed this limit, the body is dropped entirely. Defaults to no limit.
```

docs/docs.logflare.com/docs/backends/syslog.mdx
```diff
@@ -24,6 +24,7 @@ The following values are required when creating a Syslog backend:
 The following values are optional:
 
 - `tls`: (`boolean`) If true, connects via SSL/TLS. Defaults to false (TCP).
+- `max_message_bytes`: (`integer`) Maximum size of the syslog message frame in bytes. If a log exceeds this limit, the JSON body is truncated to fit. If the headers alone exceed this limit, the body is dropped entirely. Defaults to no limit.
```
ruslandoga:
```suggestion
- `max_message_bytes`: (`integer`) Maximum size of the syslog message frame in bytes. If a log exceeds this limit, the JSON body is truncated to fit. If the headers alone exceed this limit, the body is dropped entirely. If not set, the message size is unlimited.
```

docs/docs.logflare.com/docs/backends/syslog.mdx
ruslandoga: I've added description of the default behavior in https://github.com/Logflare/logflare/pull/3379#discussion_r3123788792

docs/docs.logflare.com/docs/backends/syslog.mdx
```diff
@@ -24,6 +24,7 @@ The following values are required when creating a Syslog backend:
 The following values are optional:
 
 - `tls`: (`boolean`) If true, connects via SSL/TLS. Defaults to false (TCP).
+- `max_message_bytes`: (`integer`) Maximum size of the syslog message frame in bytes. If a log exceeds this limit, the JSON body is truncated to fit. If the headers alone exceed this limit, the body is dropped entirely. If not set, the message size is unlimited.
```
Ziinc: We should set a sensible default for users ( like 75% of conventional max or something like that ) so that this does not need to be explicitly configured

Ziinc (changes requested):

Ziinc:
Value can be hardcoded in the pipeline. We should also adjust the Broadway pipeline batcher to batch based on potential batch sizes. Will need to calculate it with a batcher function, can refer to bigquery pipeline as reference

docs/docs.logflare.com/docs/backends/syslog.mdx
ruslandoga: I've set it to 50000 in https://github.com/Logflare/logflare/pull/3379/changes/2c58c21a5249a7d896d5ecce06a44e2837fe6073 (50000/65536 = 0.76).

Also here's what syslog-ng does:

`syslog-ng` defaults `log-msg-size()` to `65536` bytes / 64 KiB, and oversized IETF syslog messages are dropped by default because `trim-large-messages()` defaults to `no`. If trimming is enabled, `syslog-ng` trims to `log-msg-size()` and tags the message.

lib/logflare_web/live/display_helpers.ex
ruslandoga, line 22: We could also allow "non secret" values like booleans (e.g., for syslog's `tls` on/off flag) and small integers (like ports, max_message_bytes, etc.).

Ziinc (approved):

PR #3383, opened 2026-04-15
feat: Read from replicas in caches

ruslandoga:
Closes O11Y-1354


lib/telemetry.ex
```diff
@@ -93,13 +93,29 @@ defmodule Logflare.Telemetry do
     ]
 
     database_metrics = [
-      distribution("logflare.repo.query.total_time", unit: {:native, :millisecond}),
+      distribution("logflare.repo.query.total_time",
+        unit: {:native, :millisecond},
+        tags: [:role],
+        tag_values: &add_repo_role/1
```
ruslandoga: Not sure if this is needed at this stage.

config/runtime.exs
```diff
@@ -506,3 +506,15 @@ config :logflare, :context_cache_gossip, %{
   ratio: cache_gossip_ratio,
   max_nodes: cache_gossip_max_nodes
 }
+
+# LOGFLARE_READ_REPLICAS: Comma-separated list of PostgreSQL read replica URLs to distribute
+# context cache queries across. If unset or empty, all queries go to the primary database.
+# Example: "postgresql://user:pass@replica1:5432/db,postgresql://user:pass@replica2:5432/db"
+read_replicas =
+  "LOGFLARE_READ_REPLICAS"
```
Ziinc, line 514: Let's do comma separated hostnames and assume all other configs are equal

Ziinc: We can support uris later on if needed

lib/logflare/repo.ex
```diff
@@ -29,4 +30,38 @@ defmodule Logflare.Repo do
         0
     end
   end
+
+  defp fetch_read_replicas! do
+    Application.fetch_env!(:logflare, :read_replicas)
+  end
+
+  defp random_read_replica do
+    case fetch_read_replicas!() do
```
Ziinc: We want to load balance across all including the primary for first pass at this.

lib/telemetry.ex
Ziinc: Let's exclude this for now.

lib/telemetry.ex
```diff
@@ -93,13 +93,29 @@ defmodule Logflare.Telemetry do
...
+      distribution("logflare.repo.query.queue_time",
+        unit: {:native, :millisecond},
+        tags: [:role],
+        tag_values: &add_repo_role/1
+      ),
+      distribution("logflare.repo.query.idle_time",
+        unit: {:native, :millisecond},
+        tags: [:role],
+        tag_values: &add_repo_role/1
+      )
```
Ziinc: As above, let's exclude

Ziinc (changes requested):

Ziinc:
Also needs documentation updates for self hosting

lib/telemetry.ex
ruslandoga: Removed this `role` tag from all metrics in https://github.com/Logflare/logflare/pull/3383/changes/03ac56f72e3f668c23e0807bfc53910b53aa305b

config/runtime.exs
ruslandoga: Changed to hostnames in https://github.com/Logflare/logflare/pull/3383/changes/03ac56f72e3f668c23e0807bfc53910b53aa305b, but I'll need to update tests to make sure we are indeed starting pools for these hostnames since `:url` takes precedence, and if `Logflare.Repo` is configured with `:url` then the `URI.parse(url).host` from there would take precedence and all replica repos would connect to primary.

---

Updated the tests in https://github.com/Logflare/logflare/pull/3383/changes/9ce2a73c73af3ea5509646ffc4d72ddc0aaf5c03

lib/logflare/repo.ex
ruslandoga: Added primary to the set of choices in https://github.com/Logflare/logflare/pull/3383/changes/03ac56f72e3f668c23e0807bfc53910b53aa305b

lib/telemetry.ex
ruslandoga: Removed in https://github.com/Logflare/logflare/pull/3383/changes/03ac56f72e3f668c23e0807bfc53910b53aa305b

ruslandoga:
> Also needs documentation updates for self hosting

I've added some documentation in https://github.com/Logflare/logflare/pull/3383/changes/674ee36af72203b7d808cc7e618ec6272574e85d.

Off-topic: there is no description on that page for most of the other database configuration options like `DB_HOSTNAME`, `DB_PORT`, `DB_USERNAME`, `DB_PASSWORD`, etc. They only appear in the Docker Compose example. I think maybe those can be added as well for completeness.

test/logflare/repo_test.exs
```diff
@@ -0,0 +1,67 @@
...
+      start_read_replicas(_no_replicas = [])
+
+      assert Repo.get_dynamic_repo() == Repo
+      assert Repo.apply_with_random_repo(Repo, :get_dynamic_repo, []) == Repo
+    end
+
+    test "uses replica repo during execution and reverts afterward" do
+      start_read_replicas(["127.0.0.1", "::1"])
+
+      # since random repo choice includes body primary and replicas,
```
ruslandoga:
```suggestion
      # since random repo choice includes both primary and replicas,
```

test/logflare/repo_test.exs
```diff
@@ -0,0 +1,67 @@
...
+
+      assert Repo.get_dynamic_repo() == Repo
+      assert Repo.apply_with_random_repo(Repo, :get_dynamic_repo, []) == Repo
+    end
+
+    test "uses replica repo during execution and reverts afterward" do
+      start_read_replicas(["127.0.0.1", "::1"])
+
+      # since random repo choice includes both primary and replicas,
+      # we may need to retry our check until we hit a replica
```
ruslandoga:
```suggestion
      # we may need to perform several calls to hit a replica
```

Ziinc (approved):

Ziinc:
@ruslandoga can you add the remaining missing db env vars in a follow up PR? 🙏

PR #3386, opened 2026-04-16
fix: inspect button in log event modal

msmithstubbs:
Fixes the inspect button in the log event modal.

Added test coverage. There's an opportunity to refactor some of those modal tests to reduce repeated boiler plate if we want.

Fixes O11Y-1680

https://github.com/user-attachments/assets/04aa8485-20be-41ae-952c-d2ad2a04fa77


Ziinc (approved):
👍 please follow up with a refactor of the tests, thanks

PR #3392, opened 2026-04-17
test: add tests for SetPlan plugs

nelsonmestevao:


test/logflare_web/controllers/plugs/set_plan_from_cache_test.exs
```diff
@@ -0,0 +1,109 @@
...
+      user = insert(:user)
+      source = insert(:source, user_id: user.id)
+
+      insert(:backend, sources: [source], type: :webhook, config: %{url: "some url"})
+
+      # warm caches so the pipeline resolves user/source from cache
+      Sources.Cache.get_by_and_preload_rules(token: Atom.to_string(source.token))
+      Sources.Cache.get_source_by_token(source.token)
+      Users.Cache.get(user.id)
+      Users.Cache.get_by(api_key: user.api_key)
```
Ziinc, line 84: Should assert that the underlying context function is not called in the test with a mimic reject

Ziinc (changes requested):

test/logflare_web/controllers/plugs/set_plan_from_cache_test.exs
nelsonmestevao: Done in last push 🙏

Ziinc (approved):

PR #3395, opened 2026-04-17
feat: coerce ClickHouse map field values for numeric and boolean LQL filters

amokan:
LQL numeric and boolean filters against ClickHouse `Map(String, String)` columns now coerce the stored string value during SQL generation so comparisons evaluate correctly instead of erroring on type mismatch.

### What's Coerced

- Numeric comparisons and ranges
  - `log_attributes.response_time:>200` -> `toFloat64OrNull(log_attributes['response_time']) > 200`
- Operators: `:>`, `:<`, `:>=`, `:<=`, `:=`, and `:range` with numeric bounds
- Boolean equality
  - `log_attributes.is_error:true` -> `accurateCastOrNull(log_attributes['is_error'], 'Bool') = true`
  - Handles `"true"/"false"/"1"/"0"/"yes"/"on"` variants (_case-insensitive_)

### Not Coerced By Design

- Top-level typed columns
- Quoted string values (_e.g. `log_attributes.foo:"true"` stays as string compare_)
- `:NULL` equality
- String-oriented operators (`:~`, `:string_contains`, `:list_includes`, `:list_includes_regexp`) - even if the value parses as numeric

### Existing Bugs Fixed

- Range filters on map columns previously raised due to pinning a dynamic inside a raw `where/3` macro. Now wrapped in `dynamic/2`.
- LQL parser now casts numeric-string range bounds to int/float so they render as typed SQL literals instead of quoted strings.

### DIFF
--- a/lib/logflare/lql/backend_transformer/clickhouse.ex
+++ b/lib/logflare/lql/backend_transformer/clickhouse.ex
@@ -535,24 +535,17 @@ defmodule Logflare.Lql.BackendTransformer.ClickHouse do
   @spec negated?(map()) :: boolean()
   defp negated?(modifiers), do: Map.get(modifiers, :negate)
 
-  @spec field_expr(String.t()) :: Ecto.Query.dynamic_expr()
-  defp field_expr(field_path) do
-    case split_map_path(field_path) do
-      {:map_access, column, key} ->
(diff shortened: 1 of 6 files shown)


Ziinc (approved):
Missing arrays and nested objects in arrays. Not too clear on the handling of those yet... Those should be stringified to json right?

PR #3396, opened 2026-04-20
feat: add Endpoint history with paper_trail

msmithstubbs:
Store snapshot versions of `Endpoint` when updated.

* adds paper_trail and `versions` table
* add LogflareWeb.EndpointsVersionsLive for viewing versions
* Endpoint snapshot displayed in modal 
* Refactored EndpointLive#show to reuse components from the modal

The version list shows changes between each version, including a lightweight formatted diff of the query.

Clicking a version opens up the modal to view the snapshot of the endpoint with the full, formatted SQL query.

### PaperTrail limitations

Using the paper_trail package has some limitation that required working around:

1. paper_trail only stores changes. In future we want to be able to route an API request to a nominated version of an endpoint query, so we need to snapshot the query at the time the version was created. To support this the query is stored as a map of attributes which can be restored as a Endpoint later.
2. To specify the endpoint query at a point in time we store a version number in `meta.version_number`. This increments each time a new version is created. 
3. paper_trail supports an originator field to identify the author of the change, but only supports mapping that to one table out of the box (typically a users table). We need to support users, team users, and API keys. To work around this we store the `origin` string reference (email address for a user or team_user, or the API key description). This also means we can maintain the origin description should the user or API key record be removed.

### TODO

- [x] Paginate by showing 25 rows with a load more button
- [x] modal to open version to view full snapshot
- [x] better diff display of query
- [x] tests
- [x] tidy up mobile layout

Fixes O11Y-1025

## Version history

CleanShot 2026-04-22 at 16 40 15@2x

### Responsive layout

CleanShot 2026-04-22 at 16 40 48@2x

### Endpoint show

Reuses components from the snapshot modal

CleanShot 2026-04-22 at 16 41 39@2x

### Demo

https://github.com/user-attachments/assets/bb68ea82-9114-4ee4-a3cb-cf3a6937e42a


lib/logflare/endpoints.ex
```diff
@@ -151,42 +208,144 @@ defmodule Logflare.Endpoints do
...
+  defp to_query_words(nil), do: []
+
+  @spec version_origin(origin()) :: String.t() | nil
+  defp version_origin(%User{email: email}), do: email
+  defp version_origin(%TeamUser{email: email}), do: email
+
+  defp version_origin(%OauthAccessToken{description: description})
+       when is_non_empty_binary(description), do: "API: #{description}"
+
+  defp version_origin(%OauthAccessToken{id: id}), do: "API: id #{id}"
```
msmithstubbs, line 247: Fallback for a API key with no description. Could alternatively use a obfuscated key (eg: `abcd-****12`)

test/logflare_web/live_views/endpoints_versions_live_test.exs
```diff
@@ -0,0 +1,318 @@
...
+               3
+               Changes
+               Description:
+               latest description
+               Query:select current_date() updated_value from endpoint_versions
+               Author
+               #{user.email}
+               Updated
+               #{Calendar.strftime(version_3.inserted_at, "%Y-%m-%d %H:%M:%S UTC")}
+               """)
```
msmithstubbs, line 93: These are intended as snapshot tests to compare the content in the page without having to worry about the DOM structure. The snapshot could be a little more readable with some additional work.

I find this more readable and concise than a series of DOM assertions but interested in others feedback here.

lib/logflare_web/controllers/api/endpoint_controller.ex
```diff
@@ -57,7 +57,8 @@ defmodule LogflareWeb.Api.EndpointController do
 
   def create(%{assigns: %{user: user}} = conn, params) do
     with :ok <- authorize_backend_id(user, params),
-         {:ok, query} <- Endpoints.create_query(user, params) do
+         origin <- conn.assigns.access_token,
```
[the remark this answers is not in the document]
msmithstubbs: Fixed.

lib/logflare/endpoints.ex
```diff
@@ -151,42 +189,165 @@ defmodule Logflare.Endpoints do
...
+  @spec to_query_words(String.t() | nil) :: [String.t()]
+  defp to_query_words(query) when is_binary(query), do: String.split(query)
+  defp to_query_words(nil), do: []
+
+  @spec version_origin(origin()) :: String.t() | nil
+  defp version_origin(%User{email: email}), do: email
+  defp version_origin(%TeamUser{email: email}), do: email
+
+  defp version_origin(%OauthAccessToken{description: description})
+       when is_non_empty_binary(description), do: "API: #{description}"
```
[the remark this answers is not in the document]
msmithstubbs: 👎

lib/logflare/endpoints.ex
```diff
@@ -151,42 +189,165 @@ defmodule Logflare.Endpoints do
...
+    from(version in Version,
+      where: version.item_type == "Query" and version.item_id == ^endpoint_id,
+      select: fragment("COALESCE(MAX((?->>'version_number')::integer), 0)", version.meta)
+    )
+    |> Repo.one()
+    |> Kernel.+(1)
+  end
+
+  @spec should_kill_caches?(map()) :: boolean()
+  defp should_kill_caches?(changes) when is_map(changes) do
```
Ziinc, line 270: Can use is_map_key/2 for this

lib/logflare/endpoints.ex
```diff
@@ -151,42 +189,165 @@ defmodule Logflare.Endpoints do
...
+  defp version_meta(changeset, version_number) do
+    %{
+      version_number: version_number,
+      endpoint_snapshot: endpoint_version_snapshot(changeset)
+    }
+    |> maybe_put_query_diff(changeset)
+  end
+
+  @spec version_snapshot_fields() :: [atom()]
+  def version_snapshot_fields do
```
Ziinc: Should place this in the EndpointQuery module so that schema changes can be collocstes

lib/logflare/endpoints.ex
```diff
@@ -126,6 +138,32 @@ defmodule Logflare.Endpoints do
...
+  def update_query(%Query{} = query, params, origin) when is_map(params) do
+    Repo.transact(fn ->
+      query = lock_endpoint_query(query)
+      version_number = next_endpoint_version_number(query.id)
+      changeset = Query.update_by_user_changeset(query, params)
+      opts = paper_trail_opts(changeset, origin, version_number)
+
+      case PaperTrail.update(changeset, opts) do
+        {:ok, %{model: query}} ->
+          {:ok, {query, should_kill_caches?(changeset.changes)}}
```
Ziinc: We don't need to compute this here, we can compute it in the match branch below

Ziinc: Also we probably can merge the two functions together

lib/logflare/endpoints.ex
```diff
@@ -151,42 +189,165 @@ defmodule Logflare.Endpoints do
...
+      lock: "FOR UPDATE"
+    )
+    |> Repo.one!()
+    |> Repo.preload(:user)
+  end
+
+  @spec next_endpoint_version_number(integer()) :: integer()
+  defp next_endpoint_version_number(endpoint_id) do
+    from(version in Version,
+      where: version.item_type == "Query" and version.item_id == ^endpoint_id,
```
Ziinc, line 262: Should use a more specific item type name like EndpointQuery as we will eventually need to do this for alert query as well

lib/logflare/endpoints.ex
```diff
@@ -151,42 +189,165 @@ defmodule Logflare.Endpoints do
...
+    for pid <- Resolver.list_caches(endpoint) do
+      Utils.Tasks.async(fn ->
+        ResultsCache.invalidate(pid)
+      end)
     end
+    |> Task.await_many(30_000)
+  end
+
+  @spec get_endpoint_query_version_by_version_number(integer(), integer()) :: Version.t() | nil
+  def get_endpoint_query_version_by_version_number(endpoint_id, version_number)
```
Ziinc:
```suggestion
  def get_endpoint_query_version(endpoint_id, version_number)
```
I thinkk it's self explanatory enough

Ziinc (changes requested):

lib/logflare/endpoints.ex
Ziinc: Access tokens will have a description tagged to it, we can snapshot that for identification, as once it gets deleted we won't have any trace of it

Ziinc (commented):
More comments.

lib/logflare/endpoints.ex
```diff
@@ -126,6 +138,32 @@ defmodule Logflare.Endpoints do
...
 
+  @spec update_query(Query.t(), map(), origin()) :: {:ok, Query.t()} | {:error, any()}
+  def update_query(%Query{} = query, params, origin) when is_map(params) do
+    Repo.transact(fn ->
+      query = lock_endpoint_query(query)
+      version_number = next_endpoint_version_number(query.id)
+      changeset = Query.update_by_user_changeset(query, params)
+      opts = paper_trail_opts(changeset, origin, version_number)
+
+      case PaperTrail.update(changeset, opts) do
```
Baishan: The two cases could be condensed like

```suggestion
      with  {:ok, %{model: query}} <- PaperTrail.update(changeset, opts) do
          maybe_kill_endpoint_caches(query, should_kill_caches?(changeset.changes))
          {:ok, query}
      end
```

lib/logflare_web/live/endpoints/components.ex
```diff
@@ -0,0 +1,151 @@
...
+      <div class="tw-text-sm tw-text-zinc-300">{@value}</div>
+    </div>
+    """
+  end
+
+  attr :change, :map, required: true
+
+  def change(%{change: %{query_diff: _}} = assigns) do
+    ~H"""
+    <div class="tw-grid tw-w-full tw-grid-cols-[12rem_minmax(0,1fr)] tw-items-start tw-gap-2 tw-rounded-sm tw-px-2 tw-py-1 tw-text-sm tw-text-zinc-300">
```
Baishan, line 97: It might be nice to eventually move to a component library like daisyUI or flesh out an scss file so the browser can cache it and reduce the size of liveview diffs

lib/logflare_web/live/endpoints/endpoints_versions_live.ex
```diff
@@ -0,0 +1,302 @@
...
+      end
+
+    {:noreply, socket}
+  end
+
+  def handle_params(_params, _uri, socket), do: {:noreply, socket}
+
+  @impl true
+  def handle_event("load-more", _params, socket) do
+    case socket.assigns.next_cursor_id do
```
Baishan: You can pattern match the assigns in the header. Then you can have a catch all handle_event. I've found over time people will eventually change something that causes a key error or bad match crashing the liveview.

lib/logflare_web/live/endpoints/endpoints_versions_live.ex
```diff
@@ -0,0 +1,302 @@
...
+      nil ->
+        {:noreply, socket}
+
+      after_version_id ->
+        {versions, next_cursor_id} = fetch_page(socket.assigns.endpoint.id, after_version_id)
```
Baishan: I'd put this in an async_assign so it doesn't block the liveview while it fetches from the database.
These tend to add up and can make the page feel laggy

Baishan: Something like this

```
def handle_event("load-more", _params, %{assigns: %{next_cursor_id: nil}} = socket) do
  {:noreply, socket}
end

def handle_event("load-more", _params, socket) do
  %{endpoint: %{id: endpoint_id}, next_cursor_id: after_version_id} = socket.assigns

  {:noreply,
   socket
   |> assign(:loading, true)
   |> start_async(:fetch_versions, fn ->
     fetch_page(endpoint_id, after_version_id)
   end)}
end

def handle_async(:fetch_versions, {:ok, {versions, next_cursor_id}}, socket) do
  {:noreply,
   socket
   |> assign(:loading, false)
   |> assign(:next_cursor_id, next_cursor_id)
   |> stream(:versions, versions)}
end

def handle_async(:fetch_versions, {:exit, reason}, socket) do
  {:noreply,
   socket
   |> assign(:loading, false)
   |> put_flash(:error, "Failed to load more: #{inspect(reason)}")}
end
```

lib/logflare_web/live/endpoints/endpoints_versions_live.ex
```diff
@@ -0,0 +1,302 @@
...
+          </div>
+        </div>
+      </:col>
+    </.versions>
+
+    <.live_component
+      :if={@selected_version}
+      module={LogflareWeb.ModalComponent}
+      id="endpoint-version-snapshot-modal"
+      title={snapshot_to_endpoint(@selected_version).description}
```
Baishan: Could store this in your assign whenever the version is changed to prevent it calling the function whenever the page re-renders

Baishan (changes requested):
This looks great. The only change I'd push for not calling the database in handle_event  and instead use assign_async or start_async.

lib/logflare/endpoints.ex
msmithstubbs: Agree, but this is determined by Papertrail. Ok to rename `Logflare.Endpoints.Query` to `Logflare.Endpoints.EndpointQuery`?

lib/logflare/endpoints.ex
msmithstubbs: Didn't want to call kill caches inside the transaction, so I recomputed the changes and passed them to the merged function. There’s a small chance the query struct passed to update and the locked query aren’t identical, but I think the impact is minimal?

lib/logflare/endpoints.ex
msmithstubbs: The intention was to avoid calling kill caches inside the transaction. Consolidated `should_kill_caches?` and `maybe_kill_endpoint_caches` and moved both out of the transaction.

Requires recomputing changeset, but I think the risk is  minimal.

Should we just call expire caches in the transaction? Prefer to keep transactions small as possible

lib/logflare_web/live/endpoints/components.ex
msmithstubbs: For sure, there's a few different approaches to UI in the codebase. Moving to a consistent approach would be good.

lib/logflare_web/live/endpoints/endpoints_versions_live.ex
msmithstubbs: Moved it to an assign in handle_params

lib/logflare_web/live/endpoints/endpoints_versions_live.ex
msmithstubbs: Good idea, added async fetching

lib/logflare_web/live/endpoints/endpoints_versions_live.ex
msmithstubbs: Generally prefer to not have a catch all that swallows things we shouldn't be doing, but can do if required.

Baishan (approved):

lib/logflare_web/live/endpoints/endpoints_versions_live.ex
Baishan: Not required, this liveview is small enough. I've seen huge ones that crash during database timeouts or other unexpected events and opted to catchall and log/flash so that only a specific component would break rather than have the page reload.

lib/logflare/endpoints.ex
Ziinc: Yea we should rename this

test/logflare_web/live_views/endpoints_versions_live_test.exs
Ziinc: we can omit the assertion on the exact updated time, more verbose than necessary.
For ui testing we should shift down more comprehensive testing to the context level instead of doing it at the UI level. keeping UI tests for core interaction tests is more valuable than asserting on data display, which makes tests less coupled with things like styling tweaks.

Ziinc (approved):

Ziinc:
@msmithstubbs Please do a follow up for the module renaming 🙏

PR #3401, opened 2026-04-23
fix: don't adjust shorthand timestamps timezone

msmithstubbs:
This PR corrects how local timestamps are adjusted to the search timezone.

The LQL parser sets a `timestamp_origin` modifier to `absolute` for timestamps that should not be adjusted from the search timezone to UTC. Timestamps that have a timezone suffix (eg Z or +08:00), are a unix timestamp, or a timestamp shorthand, are all treated as `:absolute` and not adjusted.

Part of O11Y-1717

## Background
In a prior reverted PR #3399 the LQL parser identified timestamps with an explicit timezone suffix (Z or a +/- offset).

However, it did not handle shorthand timestamps (`last@1day` or `today`) correctly. The parser maps these to a UTC date time but because they lacked a Z suffix in the LQL string they were then subsequently adjusted based on the search timezone. This caused incorrect search results.

This PR replaces the `explicit_timezone` modifier flag with `timestamp_origin`.

### DIFF
--- a/lib/logflare/lql/parser/helpers.ex
+++ b/lib/logflare/lql/parser/helpers.ex
@@ -541,6 +541,10 @@ defmodule Logflare.Lql.Parser.Helpers do
     {lvalue, lmodifiers} = unwrap_value_modifiers(lvalue)
     {rvalue, rmodifiers} = unwrap_value_modifiers(rvalue)
 
+    if lmodifiers[:timestamp_origin] != rmodifiers[:timestamp_origin] do
+      throw("Timestamp range cannot mix timestamps with and without timezone offsets")
+    end
+
     {{:range_operator, [lvalue, rvalue]}, Map.merge(lmodifiers, rmodifiers)}
   end
 
--- a/lib/logflare/lql/parser/helpers.ex
+++ b/lib/logflare/lql/parser/helpers.ex
@@ -559,12 +559,9 @@ defmodule Logflare.Lql.Parser.Helpers do
     Enum.map(values, &set_timestamp_origin(&1, origin))
   end
 
-  defp set_timestamp_origin(value, origin) do
-    modifiers =
-      if origin,
-        do: %{timestamp_origin: :absolute},
-        else: %{timestamp_origin: :local}
+  defp set_timestamp_origin(value, true),
+    do: {:with_modifiers, value, %{timestamp_origin: :absolute}}
 
-    {:with_modifiers, value, modifiers}
-  end
+  defp set_timestamp_origin(value, false),
+    do: {:with_modifiers, value, %{timestamp_origin: :local}}
 end
--- a/lib/logflare/lql/parser/helpers.ex
+++ b/lib/logflare/lql/parser/helpers.ex
@@ -173,9 +173,12 @@ defmodule Logflare.Lql.Parser.Helpers do
   # ============================================================================
 
   @spec parse_date_or_datetime_with_range(list()) ::
-          [Date.t() | NaiveDateTime.t() | {:with_modifiers, NaiveDateTime.t(), map()}]
+          [
+            {:with_modifiers, Date.t() | NaiveDateTime.t(),
+             %{timestamp_origin: :absolute | :local}}
+          ]
   def parse_date_or_datetime_with_range(result) when is_list(result) do
-    explicit_timezone =
+    absolute_timestamp =
       Enum.any?(result, fn
         {:timezone, _timezone} -> true
         _ -> false
@@ -197,7 +200,7 @@ defmodule Logflare.Lql.Parser.Helpers do
           [Map.put(lacc, k, lv), Map.put(racc, k, rv)]
       end)
       |> Enum.map(&build_date_or_datetime_from_parts/1)
-      |> maybe_mark_explicit_timezone(explicit_timezone)
+      |> set_timestamp_origin(absolute_timestamp)
 
     if lv == rv do
       [lv]
@@ -210,15 +213,10 @@ defmodule Logflare.Lql.Parser.Helpers do
   def parse_date_or_datetime([{tag, result}]), do: parse_iso8601_value!(tag, result)
 
   @spec parse_timestamp_datetime([{:datetime | :datetime_tz, String.t()}]) ::
-          NaiveDateTime.t() | {:with_modifiers, NaiveDateTime.t(), map()}
+          {:with_modifiers, NaiveDateTime.t(), map()}
   def parse_timestamp_datetime([{tag, result}]) when tag in [:datetime, :datetime_tz] do
-    value = parse_iso8601_value!(:datetime, result)
-
-    if tag == :datetime_tz do
-      {:with_modifiers, value, %{explicit_timezone: true}}
-    else
-      value
-    end
+    parse_iso8601_value!(:datetime, result)
+    |> set_timestamp_origin(tag == :datetime_tz)
   end
 
   @spec parse_datetime_literal(term()) :: {:ok, Date.t() | NaiveDateTime.t()} | {:error, term()}
@@ -254,7 +252,7 @@ defmodule Logflare.Lql.Parser.Helpers do
   def parse_unix_timestamp_literal([value]) do
     case parse_unix_timestamp(value) do
       {:ok, parsed} ->
-        {:with_modifiers, parsed, %{explicit_timezone: true}}
+        set_timestamp_origin(parsed, true)
 
       {:error, :invalid_format} ->
         throw(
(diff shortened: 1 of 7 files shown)


lib/logflare/lql/parser/helpers.ex
```diff
@@ -549,9 +555,16 @@ defmodule Logflare.Lql.Parser.Helpers do
...
   end
 
-  defp maybe_mark_explicit_timezone(values, true) when is_list(values) do
-    Enum.map(values, &{:with_modifiers, &1, %{explicit_timezone: true}})
+  defp set_timestamp_origin(values, origin) when is_list(values) do
+    Enum.map(values, &set_timestamp_origin(&1, origin))
   end
 
-  defp maybe_mark_explicit_timezone(values, false), do: values
+  defp set_timestamp_origin(value, origin) do
```
amokan: Minor nit - but we could use pattern matching here:

```
defp set_timestamp_origin(value, true), 
    do: {:with_modifiers, value, %{timestamp_origin: :absolute}}

defp set_timestamp_origin(value, false),
    do: {:with_modifiers, value, %{timestamp_origin: :local}}
```

lib/logflare/lql/parser/helpers.ex
```diff
@@ -271,19 +269,24 @@ defmodule Logflare.Lql.Parser.Helpers do
...
+    value = %{Timex.now() | microsecond: {0, 0}} |> set_timestamp_origin(true)
+    %{value: value, shorthand: "now"}
   end
 
   def timestamp_shorthand_to_value(["today"]) do
     dt = Timex.today() |> Timex.to_datetime()
-    value = {:range_operator, [dt, Timex.shift(dt, days: 1, seconds: -1)]}
+
+    value =
+      {:range_operator, [dt, Timex.shift(dt, days: 1, seconds: -1)] |> set_timestamp_origin(true)}
```
amokan: The pipe inside the tuple could be changed to something like this which may be a little more clear later on.

```
range = set_timestamp_origin([dt, Timex.shift(dt, days: 1, seconds: -1)], true)
value = {:range_operator, range}
```

lib/logflare/lql/parser/helpers.ex
amokan, line 548: I could be missing something, but `Map.merge` is _right side wins_ so let's imagine we have some mixed origin range come in like `t:2026-04-23T10:00:00Z..2026-04-23T11:00:00`.

After parsing, that would produce a pair of tuples like:
- `{:with_modifiers, ~N[2026-04-23 10:00:00], %{timestamp_origin: :absolute}}`
- `{:with_modifiers, ~N[2026-04-23 11:00:00], %{timestamp_origin: :local}}`

Then we go into the merge with `Map.merge(%{timestamp_origin: :absolute}, %{timestamp_origin: :local})` and it will return `%{timestamp_origin: :local}`.

When we apply the timestamp corrections (_assuming the user's search timezone is Australia/Brisbane_) we'll end up with "BETWEEN '2026-04-23 00:00:00Z' AND '2026-04-23 01:00:00Z'", which has the left side off by 10 hours as the 10:00Z bound was already UTC but got treated as a Brisbane wall-clock time and shifted.

Seems like we should consider switching the logic here in `unwrap_value_modifiers` to OR the origins and if either side is `:absolute` - the whole range should be handled as absolute.

Worth calling out that this mixed-range scenario appears to have just flipped sides with the changes. Pre-PR, the merge landed on `explicit_timezone: true` and skipped the shift (_right side off by 10 hours_), now it lands on `:local` and shifts both (_left side off by 10 hours_). Bug didn't really go away here, it just moved.

If we are not concerned about handling the mixed range situation, maybe we could pop a test in to at least help us understand if we ever changed this behavior later on accident 🤷‍♂️

amokan (commented):
Some minor nits but just want to make sure we're thinking about a scenario where a mixed range may cause confusion.

lib/logflare/lql/parser/helpers.ex
msmithstubbs: Yeah this is a good catch. 

The handling of mixed timezones in ranges here is not clear so should be fixed. I've thought a bit more about this and actually think we should reject a range with inconsistent timezones. 

A query with `t:2026-04-23T10:00:00Z..2026-04-23T11:00:00` makes sense logically but the ergonomics are poor. It's easy for a user to miss the Z embedded in the middle of the range and not realise they are mixing timezones. 

We can absolutely support mixed timezones in a range, and if there is a use case that needs it then I would say let's do that. But if not it’s likely to add confusion and increase the risk of mistakes.

I don't know enough about real world use, though. Thoughts?

lib/logflare/lql/parser/helpers.ex
msmithstubbs: Sure - I've moved the tuple inside the map. I think that's still readable?

```elixir
range = set_timestamp_origin([dt, Timex.shift(dt, days: 1, seconds: -1)], true)
    
%{value: {:range_operator, range}, shorthand: "yesterday"}
```

lib/logflare/lql/parser/helpers.ex
amokan: I think a clear rejection of mixed timezones is the way to go.

msmithstubbs: Added parser error when using mixed local and absolute timezones.
CleanShot 2026-04-28 at 10 04 55@2x

lib/logflare/lql/parser/helpers.ex
```diff
@@ -549,9 +558,13 @@ defmodule Logflare.Lql.Parser.Helpers do
...
   end
 
-  defp maybe_mark_explicit_timezone(values, true) when is_list(values) do
-    Enum.map(values, &{:with_modifiers, &1, %{explicit_timezone: true}})
+  defp set_timestamp_origin(values, origin) when is_list(values) do
+    Enum.map(values, &set_timestamp_origin(&1, origin))
   end
 
-  defp maybe_mark_explicit_timezone(values, false), do: values
+  defp set_timestamp_origin(value, true),
```
Ziinc, line 565:
```suggestion
  defp set_absolute_timestamp_origin(value, true),
```

test/logflare/lql/parser/helpers_test.exs
```diff
@@ -344,38 +344,70 @@ defmodule Logflare.Lql.Parser.HelpersTest do
...
-               {:with_modifiers, ~N[2023-01-15 08:30:00], %{explicit_timezone: true}}
+               {:with_modifiers, ~N[2023-01-15 08:30:00], %{timestamp_origin: :absolute}}
              ]
     end
 
     test "timestamp_shorthand_to_value handles 'now'" do
       result = Helpers.timestamp_shorthand_to_value(["now"])
       assert result.shorthand == "now"
-      assert %DateTime{} = result.value
+      assert {:with_modifiers, %DateTime{}, %{timestamp_origin: :absolute}} = result.value
```
Ziinc, line 354: Shouldn't this be local?

test/logflare/lql/parser/helpers_test.exs
```diff
@@ -344,38 +344,70 @@ defmodule Logflare.Lql.Parser.HelpersTest do
...
 
     test "timestamp_shorthand_to_value handles 'today'" do
       result = Helpers.timestamp_shorthand_to_value(["today"])
       assert result.shorthand == "today"
-      assert match?({:range_operator, [%DateTime{}, %DateTime{}]}, result.value)
+
+      assert match?(
+               {:range_operator,
+                [
+                  {:with_modifiers, %DateTime{}, %{timestamp_origin: :absolute}},
```
Ziinc, line 364: Should be local, no?

test/logflare/logs/search_operations_test.exs
```diff
@@ -336,21 +335,73 @@ defmodule Logflare.Logs.SearchOperationsTest do
...
+          lql_ts_filters: [timestamp_filter],
+          search_timezone: "Australia/Brisbane"
+        }
+
+      corrected = SearchOperations.apply_local_timestamp_correction(so)
+
+      assert %FilterRule{
+               operator: :range,
+               shorthand: "last@1hour",
+               modifiers: %{timestamp_origin: :absolute}
```
Ziinc, line 402: If search tz is Brisbane and shorthand is used, then shouldn't the origin be based on user local tz?

Ziinc (changes requested):
The situation where shorthand usage results in a DateTime and not NaiveDateTime is confusing, as it then results in `:absolute` which smells off

Ziinc:
Can we also add in a test against postgres single tenant to verify the behaviour is as expected?

msmithstubbs:
> The situation where shorthand usage results in a DateTime and not NaiveDateTime is confusing, as it then results in :absolute which smells off

We can have shorthand timestamps resolve to NaiveDateTime but will need to pass the search timezone into the parser. That OK or will that blur the component boundaries?

amokan (approved):
Assuming @Ziinc is good with the changes he requested, this looks good to me @msmithstubbs

Ziinc (approved):

PR #3402, opened 2026-04-23
chore: Fix DB SSL client cert filename check in runtime config

ruslandoga:
## Summary
- make the DB SSL runtime guard check for `"db-client-cert.pem"` since that's what used in the config later on

## Testing
- not run (no focused runtime config test file for this path)


config/runtime.exs
ruslandoga, line 350: It uses `"db-client-cert.pem"` here but the `if` was checking for `"db-client-ca.pem"`. Codex caught that, not me ...

Ziinc (approved):

PR #3404, opened 2026-04-24
fix: left outer join for metadata

msmithstubbs:
Fixes an issue where a search on a  source with a schema  that includes `metadata.level` would only return events that include a metadata field.

Part of O11Y-1717


Ziinc:
@msmithstubbs we will need more extensive testing to prevent regressions.

Ziinc (approved):

msmithstubbs:
Add more tests in #3407. Any particular approaches to testing you'd like me to add?

PR #3406, opened 2026-04-24
fix: slack signed state

Ziinc:
Closes PRODSEC-24


lib/logflare_web/controllers/auth/oauth_controller.ex
```diff
@@ -48,43 +48,12 @@ defmodule LogflareWeb.Auth.OauthController do
...
-          conn
-          |> put_flash(:error, "You do not have access to that alert.")
-          |> redirect(to: ~p"/dashboard")
-        else
-          alert_callback(conn, alert_query, url, id)
-        end
+    with {:ok, payload} <- SlackOauthState.verify(current_user, state) do
+      handle_slack_callback(conn, current_user, auth, payload)
+    else
+      {:error, _reason} -> auth_error_redirect(conn)
```
amokan: Seems like we may want to consider logging the error here in the else?

test/logflare_web/controllers/oauth_controller_test.exs
```diff
@@ -88,6 +89,105 @@ defmodule LogflareWeb.OAuthControllerTest do
...
+        "provider" => "slack",
+        "state" => state_signed_by_attacker
+      })
+
+    assert html_response(conn, 302)
+    assert conn.assigns.flash["error"] =~ "Authentication error"
+    refute Sources.get(victim_source.id).slack_hook_url == slack_hook_url
+  end
+
   test "attacker cannot overwrite another user's source slack hook by tampering state",
```
amokan: Prob should remove the "_attacker cannot overwrite another user's source slack hook by tampering state_" test as it is functionally the same as the new "_slack oauth2 rejects unsigned plain-JSON state_" test above.

amokan (approved):
Some minor items to address but other than that it looks ok 👍

PR #3413, opened 2026-04-29
feat: IncidentioAdaptor connection testing

bblaszkow06:
Probes the alert events endpoint with a `status: "resolved"` payload and a stable deduplication key (`logflare-connection-test-<backend.id>`), so probes never escalate and dedupe across runs to a single resolved test alert per backend.


Ziinc (approved):
needs ci cleanup, looks good

PR #3416, opened 2026-04-30
feat: Add BigqueryStorageWriteApi ConfigCat flag to BigQuery pipeline

Ziinc:
This PR adds a config cat flag to the storage write api ingestion pipeline. if enabled at the source level, then all events entering that source will be ingested via storage write API. 

Uses `source.id` as the identifier for when we enable percentage-based rollout via ConfigCat

Part of O11Y-1340

https://claude.ai/code/session_01YEZ4Vibn7frjAPuKHcipJn

https://claude.ai/code/session_01PsqMdm4TvhM8ePYCCdtpuX


djwhitt:
@Ziinc The change itself looks okay, but the test failure appears to be related.

djwhitt (dismissed):
@Ziinc The change itself looks okay, but the test failure appears to be related.

test/logflare/utils_test.exs
```diff
@@ -70,17 +73,104 @@ defmodule Logflare.Utils.FlagTest do
...
+        "other" => "true"
+      })
+
+      assert Utils.flag("BigqueryStorageWriteApi", "some-id") == false
+      assert Utils.flag("other") == true
+      assert Utils.flag("unset_feature") == false
+    end
+
+    test "BigqueryStorageWriteApi can be enabled via feature_flag_override" do
+      prev = Application.get_env(:logflare, :feature_flag_override)
```
amokan:
```suggestion
```

`prev` is unused.

test/support/test_utils.ex
```diff
@@ -9,6 +9,46 @@ defmodule Logflare.TestUtils do
...
+    merged = Map.merge(prev || %{}, %{flag => value})
+    Application.put_env(:logflare, :feature_flag_override, merged)
+    :ok
+  end
+
+  @doc """
+  restores the previous feature flag override. Should be used in a test or setup block.
+  To use outside of a test or setup block, use `:setup` option to wrap in a setup block.
+  """
+  defmacro reset_feature_flag_overrides(opts \\ [setup: false]) do
```
amokan:
```suggestion
  defmacro restore_feature_flag_overrides_on_exit(opts \\ [setup: false]) do
```

Nit. Since this doesn't actually reset the current value it may make sense to give a clearer name.

lib/logflare/utils.ex
```diff
@@ -13,26 +13,33 @@ defmodule Logflare.Utils do
...
   Checks if a feature flag is enabled.
   If SDK key is not set, will always return false.
-  In test mode, will always return true.
 
   ### Example
     iex> flag("my-feature")
     true
   """
-  def flag(feature, identifer \\ nil) when is_binary(feature) do
+  def flag(feature, identifier \\ nil) when is_binary(feature) do
```
amokan:
```suggestion
  def flag(feature, identifier \\ nil) when is_non_empty_binary(feature) do
```

Nit. Could use the `is_non_empty_binary` guard here.

lib/logflare/utils.ex
```diff
@@ -13,26 +13,33 @@ defmodule Logflare.Utils do
...
     config_cat_key = Application.get_env(:logflare, :config_cat_sdk_key)
-    env = Application.get_env(:logflare, :env)
-    overrides = Application.get_env(:logflare, :feature_flag_override, %{})
+
+    overrides =
+      Application.get_env(:logflare, :feature_flag_override) ||
+        %{}
+
+    test_env? = Application.get_env(:logflare, :env) in [:test]
```
amokan:
```suggestion
    test_env? = Application.get_env(:logflare, :env) == :test
```

Nit.

lib/logflare/utils.ex
```diff
@@ -13,26 +13,33 @@ defmodule Logflare.Utils do
...
 
     cond do
-      env == :test ->
+      test_env? and feature in @test_disabled_features ->
+        false
```
amokan:
```suggestion
      test_env? and feature in @test_disabled_features ->
        Map.get(overrides, feature, "false") in ["true", true]
```

This was short-circuiting before consulting overrides. The reason the tests are passing is still because we have `bq_storage_write_api: true` set on those sources.

amokan (dismissed):
This is better than the last iteration, but a couple minor things. If we fix that cond logic in `utils.ex` - we can get a test that exercises pipeline.ex taking the storage write branch via the flag alone 🙏

test/logflare/utils_test.exs
Ziinc: switched to a mock-based approach

test/support/test_utils.ex
Ziinc: removed in favour of mocks

lib/logflare/utils.ex
Ziinc: removed

lib/logflare/utils.ex
Ziinc: removed in favour of mocks

PR #3417, opened 2026-04-30
Prevent XSS in Monaco editor completions by escaping HTML

Ziinc:
This PR fixes a potential XSS vulnerability in the Monaco editor component by properly escaping HTML characters in completion names before rendering them in JavaScript.

Completion names (from sources and endpoints) were being JSON-encoded but not HTML-escaped before being embedded directly into a JavaScript string literal. An attacker could inject a completion name like `</script><script>alert(document.domain)</script>` which would break out of the string and execute arbitrary JavaScript.

Closes PRODSEC-42

https://claude.ai/code/session_013Fn6xnMTBQCf8P9ddUNLQE


amokan (approved):

amokan (approved):
Looks like we may need to fix the test - but aside from that 👍

PR #3419, opened 2026-04-30
fix: prevent mass assignment of token scopes via user-controlled input

Ziinc:
## Reviewer focus

- Is the scope allow-list (`public`, `ingest`, `query`, `private`, plus `ingest:(source|collection):<int>` and `query:endpoint:<int>`) complete for all user-creatable tokens? Anything the UI doesn't currently expose but a legitimate caller depends on?
- Is `create_access_token_with_token/2` sufficiently isolated? Today only `Logflare.SingleTenant` calls it for env-seeded tokens; the docstring warns against forwarding user-controlled params. Worth confirming no other code path is tempted to reach for it.
- The API now returns 422 with a changeset-error body for any invalid scope, where previously it returned 401 specifically for `partner`. External clients that match on 401 for this case will need to update.

## Summary

Closes a mass-assignment issue in the access-token create path. User-controlled params could assign privileged or invalid scopes, and in deployments where `User.id` and `Partner.id` can collide (self-hosted Supabase, dev environments), a forged token carrying the `partner` scope authenticated as a `Partner` and granted partner-tier API access.

The pre-existing controller check (`"partner" not in scopes_input` → 401) blocked the obvious HTTP body case, but two routes around it remained:

- **LiveView form** — crafted websocket payloads with `scopes_main: ["partner"]` flowed through `Auth.create_access_token/2` with no equivalent check, since the Auth changeset itself cast `:scopes` unconditionally.
- **`:token` mass-assignment** — the controller forwarded the full request body into `Auth.create_access_token/2`, which cast `:token` for both branches. An attacker could propose the literal token value through `POST /api/access-tokens`.

## Approach

Scope policy moves into a `validate_scopes/1` changeset validation on the user/team path; partner tokens have their scope force-set to exactly `"partner"` via `put_change` and do not run the validation. The controller-level `partner` check is removed as redundant. The public `create_access_token/2` drops `:token` from both cast lists; trusted callers that legitimately need a server-controlled token value (single-tenant env seeding) call the new `create_access_token_with_token/2` helper. The OAuth insert + changeset update is wrapped in `Repo.transaction` so validation failure rolls back the underlying `ExOauth2Provider.AccessTokens.create_token/3` row.

## Compatibility impact

- **Invalid user-supplied `scopes` now returns 422 with `errors.scopes`.** Previously, `scopes: "partner"` returned 401 (controller-level) and unrecognized scopes (e.g. `"admin"`, `"ingest:source:not-a-number"`) were silently cast through. API clients matching specifically on 401 for the partner case need updating.
- **Partner tokens always carry exactly `"partner"`.** Previously `cast(:scopes) |> update_change(fn s -> s <> " partner" end)` could yield compound scopes; no current caller passes scopes to the Partner branch, so no observable impact today.
- **Public `create_access_token/2` no longer honors `:token` in attrs.** Both user and partner cast lists drop the field. Trusted in-tree callers (single-tenant env seeding) have been migrated to `create_access_token_with_token/2`. Out-of-tree callers passing `:token` via attrs will see it silently dropped and get a server-generated token.

## Behavioral notes

- LiveView changeset errors surface to the form as a flash via `LogflareWeb.Utils.stringify_changeset_errors/2`. `coerce_scope_list/1` normalizes non-list payloads (string, map) before forwarding to Auth, but the security check happens at the Auth changeset — the LiveView guards are shape, not policy.
- Resource-id ownership for `ingest:source:<id>` / `query:endpoint:<id>` is still enforced downstream by `VerifyResourceAccess` at request time; the changeset only validates shape.
- OAuth token rows now roll back on validation failure; previously a failed `Repo.update` left the ExOauth2Provider-created row in the table with default scopes.

## Changes

- `lib/logflare/auth.ex` — adds `validate_scopes/1` and the allow-list module attributes, splits the user and partner branches through `do_create_user_access_token/3` + `run_create_access_token/3` (transaction wrapper), drops `:token` from both branches' cast lists, adds the privileged `create_access_token_with_token/2`.
- `lib/logflare_web/controllers/api/access_token_controller.ex` — drops the controller-level `partner` check; forwards only `description` + `scopes` from the request body to Auth; adds a 422 response in the OpenAPI spec.
- `lib/logflare_web/live/access_tokens_live.ex` — adds `coerce_scope_list/1` / `coerce_scope_fields/2` shape-guards; surfaces Auth changeset errors as a form flash instead of letting `{:ok, token} = ...` raise.
- `lib/logflare/single_tenant.ex` — switches the two env-seeded creates to `create_access_token_with_token/2`.

## Test plan

- [x] `mix test test/logflare/auth_test.exs` — `:scopes` and `:token` mass-assignment (atom and string keys, user and partner branches), unrecognized-scope rejection, partner-token scope invariance, team-clause attrs forwarding, `create_access_token_with_token/2` honoring its `:token` arg
- [x] `mix test test/logflare_web/controllers/api/access_tokens_controller_test.exs` — `scopes: "partner"` and `scopes: "ingest partner"` both return 422 with changeset errors; valid `scopes: "ingest"` round-trips; attacker-supplied `token` in the body is ignored and the response carries a server-generated token
- [x] `mix test test/logflare_web/live/access_tokens_live_test.exs` — crafted submit/change payloads with smuggled `partner`, unknown sub-scopes, and non-list scope params (string, map) all reject without creating a token
- [x] `mix test test/logflare/auth/cache_test.exs test/logflare_web/controllers/plugs/verify_api_access_test.exs test/logflare_web/controllers/plugs/verify_resource_access_test.exs test/logflare/single_tenant_test.exs` — adjacent suites covering the Auth signature surface; no files under these paths were modified on this branch
- [x] Manually: `POST /api/access-tokens` with `{"scopes": "partner"}` returns 422 with `errors.scopes` (verified against single-tenant dev server)
- [x] Manually: `POST /api/access-tokens` with no `scopes` field returns a token whose scopes contain `"public"` (verified against single-tenant dev server)
- [x] Manually: legitimate `ingest` / `query` / `private` token creation via UI still works

### DIFF
--- a/lib/logflare/auth.ex
+++ b/lib/logflare/auth.ex
@@ -127,36 +127,41 @@ defmodule Logflare.Auth do
   end
 
   @doc "Creates an Oauth access token with no expiry, linked to the given user or team's user."
-  @typep create_attrs :: %{description: String.t()} | map()
-  @spec create_access_token(Team.t() | User.t() | Partner.t(), create_attrs()) ::
+  @typep create_attrs :: %{optional(:description) => String.t()} | map()
+  @typep create_opts :: [scopes: String.t()]
+  @spec create_access_token(Team.t() | User.t() | Partner.t(), create_attrs(), create_opts()) ::
           {:ok, OauthAccessToken.t()} | {:error, term()}
-  def create_access_token(user_or_team_or_partner, attrs \\ %{})
+  def create_access_token(user_or_team_or_partner, attrs \\ %{}, opts \\ [])
 
-  def create_access_token(%Team{user_id: user_id}, attrs) do
+  def create_access_token(%Team{user_id: user_id}, attrs, opts) do
     user_id
     |> Logflare.Users.get()
-    |> create_access_token(attrs)
+    |> create_access_token(attrs, opts)
   end
 
-  def create_access_token(%User{} = user, attrs) do
+  def create_access_token(%User{} = user, attrs, opts) do
     with {:ok, token} <- ExOauth2Provider.AccessTokens.create_token(user, %{}, env_oauth_config()) do
       token
-      |> Changeset.cast(attrs, [:token, :scopes, :description])
+      |> Changeset.cast(attrs, [:token, :description])
+      |> maybe_put_scopes(Keyword.get(opts, :scopes))
       |> Changeset.unique_constraint(:token)
       |> Repo.update()
     end
   end
 
-  def create_access_token(%Partner{} = partner, attrs) do
+  def create_access_token(%Partner{} = partner, attrs, _opts) do
     with {:ok, token} <-
            ExOauth2Provider.AccessTokens.create_token(partner, %{}, env_partner_oauth_config()) do
       token
-      |> Changeset.cast(attrs, [:token, :scopes, :description])
-      |> Changeset.update_change(:scopes, fn scopes -> scopes <> " partner" end)
+      |> Changeset.cast(attrs, [:token, :description])
+      |> Changeset.put_change(:scopes, "partner")
       |> Changeset.unique_constraint(:token)
       |> Repo.update()
     end
   end
+
+  defp maybe_put_scopes(changeset, nil), do: changeset
+  defp maybe_put_scopes(changeset, scopes), do: Changeset.put_change(changeset, :scopes, scopes)
 
   @doc """
   Verifies a `%OauthAccessToken{}` or string access token.
--- a/lib/logflare/auth.ex
+++ b/lib/logflare/auth.ex
@@ -143,19 +143,37 @@ defmodule Logflare.Auth do
   end
 
   def create_access_token(%User{} = user, attrs) do
-    run_create_access_token(user, env_oauth_config(), fn token ->
-      token
-      |> Changeset.cast(attrs, [:token, :scopes, :description])
-      |> validate_scopes()
-      |> Changeset.unique_constraint(:token)
-    end)
+    do_create_user_access_token(user, attrs, [:scopes, :description])
   end
 
   def create_access_token(%Partner{} = partner, attrs) do
     run_create_access_token(partner, env_partner_oauth_config(), fn token ->
       token
-      |> Changeset.cast(attrs, [:token, :description])
+      |> Changeset.cast(attrs, [:description])
       |> Changeset.put_change(:scopes, "partner")
+      |> Changeset.unique_constraint(:token)
+    end)
+  end
+
+  @doc """
+  Creates an access token with a caller-supplied `:token` value.
+
+  Privileged variant of `create_access_token/2` that opts into casting `:token`.
+  Only call from trusted code paths where the token value is server-controlled
+  (e.g. single-tenant environment seeding). Never forward user-controlled
+  request params through this function.
+  """
+  @spec create_access_token_with_token(User.t(), create_attrs()) ::
+          {:ok, OauthAccessToken.t()} | {:error, Changeset.t()}
+  def create_access_token_with_token(%User{} = user, attrs) do
+    do_create_user_access_token(user, attrs, [:token, :scopes, :description])
+  end
+
+  defp do_create_user_access_token(user, attrs, cast_fields) do
+    run_create_access_token(user, env_oauth_config(), fn token ->
+      token
+      |> Changeset.cast(attrs, cast_fields)
+      |> validate_scopes()
       |> Changeset.unique_constraint(:token)
     end)
   end
--- a/lib/logflare/auth.ex
+++ b/lib/logflare/auth.ex
@@ -192,7 +192,7 @@ defmodule Logflare.Auth do
 
   defp validate_scopes(changeset) do
     Changeset.validate_change(changeset, :scopes, fn field, scopes ->
-      tokens = String.split(scopes)
+      tokens = String.split(scopes) |> String.trim()
 
       cond do
         "partner" in tokens ->
--- a/lib/logflare/auth.ex
+++ b/lib/logflare/auth.ex
@@ -140,10 +140,12 @@ defmodule Logflare.Auth do
   end
 
   def create_access_token(%User{} = user, attrs, opts) do
-    with {:ok, token} <- ExOauth2Provider.AccessTokens.create_token(user, %{}, env_oauth_config()) do
+    with {:ok, scopes} <- sanitize_user_scopes(Keyword.get(opts, :scopes)),
+         {:ok, token} <-
+           ExOauth2Provider.AccessTokens.create_token(user, %{}, env_oauth_config()) do
       token
       |> Changeset.cast(attrs, [:token, :description])
-      |> maybe_put_scopes(Keyword.get(opts, :scopes))
+      |> maybe_put_scopes(scopes)
       |> Changeset.unique_constraint(:token)
       |> Repo.update()
     end
@@ -162,6 +164,14 @@ defmodule Logflare.Auth do
 
   defp maybe_put_scopes(changeset, nil), do: changeset
   defp maybe_put_scopes(changeset, scopes), do: Changeset.put_change(changeset, :scopes, scopes)
+
+  # User-owned tokens must never carry the partner scope. Guard at the auth
+  # boundary so every caller is protected, not just the HTTP controller.
+  defp sanitize_user_scopes(nil), do: {:ok, nil}
+
+  defp sanitize_user_scopes(scopes) when is_binary(scopes) do
+    if "partner" in String.split(scopes), do: {:error, :unauthorized}, else: {:ok, scopes}
+  end
 
   @doc """
   Verifies a `%OauthAccessToken{}` or string access token.
--- a/lib/logflare/auth.ex
+++ b/lib/logflare/auth.ex
@@ -127,7 +127,7 @@ defmodule Logflare.Auth do
   end
 
   @doc "Creates an Oauth access token with no expiry, linked to the given user or team's user."
-  @typep create_attrs :: %{optional(:description) => String.t()}
+  @typep create_attrs :: %{optional(:token) => String.t(), optional(:description) => String.t()}
   @typep create_opts :: [scopes: String.t()]
   @spec create_access_token(Team.t() | User.t() | Partner.t(), create_attrs(), create_opts()) ::
           {:ok, OauthAccessToken.t()} | {:error, term()}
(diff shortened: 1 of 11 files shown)


test/logflare_web/live/access_tokens_live_test.exs
```diff
@@ -137,8 +174,8 @@ defmodule LogflareWeb.AccessTokensLiveTest do
            |> render_submit(%{
              description: "some description",
              scopes_main: if(scopes =~ ":", do: [], else: [scopes]),
-             scopes_ingest: if(scopes =~ "ingest:", do: [], else: [scopes]),
-             scopes_query: if(scopes =~ "query:", do: [], else: [scopes])
+             scopes_ingest: if(scopes =~ "ingest:", do: [scopes], else: []),
+             scopes_query: if(scopes =~ "query:", do: [scopes], else: [])
```
Baishan, line 222: The does the condition reverse here?

lib/logflare/auth.ex
```diff
@@ -127,37 +127,50 @@ defmodule Logflare.Auth do
...
     user_id
     |> Logflare.Users.get()
-    |> create_access_token(attrs)
+    |> create_access_token(attrs, opts)
   end
 
-  def create_access_token(%User{} = user, attrs) do
-    with {:ok, token} <- ExOauth2Provider.AccessTokens.create_token(user, %{}, env_oauth_config()) do
+  def create_access_token(%User{} = user, attrs, opts) do
+    with {:ok, scopes} <- reject_partner_scope(Keyword.get(opts, :scopes)),
```
Baishan: You could probably do the same by keeping scopes in attrs? The only issue I see doing it like this is that it silently discards scopes if anything calls it the old way.

Baishan (approved):
Looks good. Its a little confusing to know what should go in attrs vs opts.

lib/logflare/auth.ex
```diff
@@ -127,37 +127,50 @@ defmodule Logflare.Auth do
...
+         {:ok, token} <-
+           ExOauth2Provider.AccessTokens.create_token(user, %{}, env_oauth_config()) do
       token
-      |> Changeset.cast(attrs, [:token, :scopes, :description])
```
Ziinc, line 144: Keeping scope in the cast is fine, just that the partner scope should never be allowed (so add in a changeset validation after the cast that blocks the partner scope).
If it is provided then we should mark a changeset error and block the creation.

lib/logflare/auth.ex
Ziinc: Should not silently discard, should do a changeset error

lib/logflare/auth.ex
```diff
@@ -127,37 +127,50 @@ defmodule Logflare.Auth do
   end
 
   @doc "Creates an Oauth access token with no expiry, linked to the given user or team's user."
-  @typep create_attrs :: %{description: String.t()} | map()
-  @spec create_access_token(Team.t() | User.t() | Partner.t(), create_attrs()) ::
+  @typep create_attrs :: %{optional(:token) => String.t(), optional(:description) => String.t()}
+  @typep create_opts :: [scopes: String.t()]
+  @spec create_access_token(Team.t() | User.t() | Partner.t(), create_attrs(), create_opts()) ::
```
Ziinc: Pulling scopes out as a 3rd arity seems unnecessary as the changeset cast already handles the parsing of the string, we just need to harden the validations

lib/logflare_web/live/access_tokens_live.ex
```diff
@@ -151,6 +151,21 @@ defmodule LogflareWeb.AccessTokensLive do
     "scopes_query" => [],
     "scopes_main" => ["ingest"]
   }
+
+  @scopes_main_allowed ~w(ingest query private)
+  @scopes_ingest_pattern ~r/^ingest:source:\d+$/
+  @scopes_query_pattern ~r/^query:endpoint:\d+$/
+
+  defp filter_param(values, allowed) when is_list(values) and is_list(allowed) do
```
Ziinc: The changeset validation will be enough, no need to do liveview level validation. If changeset fails, error will bubble up and get displayed in the form

lib/logflare_web/controllers/api/access_token_controller.ex
```diff
@@ -36,16 +36,17 @@ defmodule LogflareWeb.Api.AccessTokenController do
...
+    opts =
+      case Map.fetch(params, "scopes") do
+        {:ok, scopes} when is_binary(scopes) -> [scopes: scopes]
+        _ -> []
+      end
 
-    with {:scopes, true} <- {:scopes, "partner" not in String.split(scopes_input)},
-         {:ok, access_token} <- Auth.create_access_token(user, params) do
+    with {:ok, access_token} <-
+           Auth.create_access_token(user, Map.take(params, ["description"]), opts) do
```
Ziinc: Agree that we should omit the token field here

lib/logflare_web/controllers/api/access_token_controller.ex
```diff
@@ -36,16 +36,17 @@ defmodule LogflareWeb.Api.AccessTokenController do
   )
 
   def create(%{assigns: %{user: user}} = conn, params) do
-    scopes_input = Map.get(params, "scopes", "")
+    opts =
+      case Map.fetch(params, "scopes") do
+        {:ok, scopes} when is_binary(scopes) -> [scopes: scopes]
+        _ -> []
+      end
```
Ziinc: Not necessary.

Also, side note , function design wise, since the scope to always be provided, it shouldn't be an option.

test/logflare_web/live/access_tokens_live_test.exs
djwhitt: Yes — the original conditions were inverted. For `scopes = "ingest:source:1"` the old test put `[scopes]` into `scopes_query` and left `scopes_ingest` empty (and vice versa). The pre-fix LiveView concatenated everything into one space-separated string so the test still passed, but with the shape-aware changeset validation in place the test has to put `ingest:source:1` through `scopes_ingest` and `query:endpoint:1` through `scopes_query` — which is also what the UI actually does.

lib/logflare/auth.ex
djwhitt: Fixed in a4f8d3f6e. Scopes live in attrs again, and `validate_scopes/1` returns a changeset error for `partner` or any unrecognized scope token — no silent discard.

lib/logflare/auth.ex
djwhitt: Fixed in a4f8d3f6e — `validate_scopes/1` after `cast` blocks `partner` and unrecognized scope tokens with a `changeset.errors[:scopes]` entry, and the OAuth insert is wrapped in `Repo.transaction` so the underlying row rolls back rather than being left orphaned.

lib/logflare/auth.ex
djwhitt: Fixed in a4f8d3f6e — changeset error rather than silent discard.

lib/logflare/auth.ex
djwhitt: Fixed in a4f8d3f6e — collapsed back to `/2`, scopes hardened via `validate_scopes/1`.

lib/logflare_web/live/access_tokens_live.ex
djwhitt: Fixed in a4f8d3f6e — the LiveView allow-list and regex filters are gone. Changeset errors now surface to the form via `LogflareWeb.Utils.stringify_changeset_errors/2` as a flash. The only LiveView-level logic left is `coerce_scope_list/1` which `List.wrap`s + filters non-empty binaries to keep the form handler from crashing on a malformed (non-list) payload — not a security check.

lib/logflare_web/controllers/api/access_token_controller.ex
djwhitt: Fixed in b05451d74. `:token` is dropped from both branches' cast lists in the public `Auth.create_access_token/2`. The controller now forwards only `description` + `scopes`. Trusted callers that legitimately need a server-controlled token value (single-tenant env seeding) use the new `Auth.create_access_token_with_token/2` helper, which is the only path that casts `:token`.

lib/logflare_web/controllers/api/access_token_controller.ex
djwhitt: Fixed in a4f8d3f6e. The opts extraction is gone; the controller just does `Map.take(params, ["description", "scopes"])` and forwards to `Auth.create_access_token/2`. Scopes ride in attrs alongside `description`.

lib/logflare/auth.ex
```diff
@@ -139,23 +143,74 @@ defmodule Logflare.Auth do
...
+        token
+      else
+        {:error, %Changeset{} = changeset} -> Repo.rollback(changeset)
+      end
+    end)
+  end
+
+  defp validate_scopes(changeset) do
+    Changeset.validate_change(changeset, :scopes, fn field, scopes ->
+      tokens = String.split(scopes)
```
Ziinc: should also trim
```suggestion
      tokens = String.split(scopes) |> String.trim()
```

test/logflare/auth_test.exs
```diff
@@ -30,9 +30,60 @@ defmodule Logflare.AuthTest do
...
+        Auth.create_access_token_with_token(user, %{token: explicit, scopes: "public"})
+
+      assert token.token == explicit
+      assert token.scopes == "public"
+    end
+
+    test "user token attrs rejects unrecognized scopes via changeset error", %{user: user} do
+      assert_scope_error(Auth.create_access_token(user, %{scopes: "ingest admin"}))
+      assert_scope_error(Auth.create_access_token(user, %{scopes: "ingest:source:not-a-number"}))
+    end
```
Ziinc, line 74:
```suggestion
    test "user token attrs rejects unrecognized scopes via changeset error", %{user: user} do
      assert_scope_error(Auth.create_access_token(user, %{scopes: "ingest admin"}))
      assert_scope_error(Auth.create_access_token(user, %{scopes: "ingest:source:not-a-number"}))
      assert_scope_error(Auth.create_access_token(user, %{scopes: "ingest  partner"}))
    end
```

lib/logflare/auth.ex
djwhitt: Reverted in 611cfab91 — the applied suggestion piped a list into `String.trim/1` and raised `FunctionClauseError` on every scope validation (existing tests like `"ingest admin"` started crashing, not just the new whitespace case). `String.split/1` with no args already trims leading/trailing whitespace and collapses runs, so no extra trim is needed:

```elixir
iex> String.split("  ingest  partner  ")
["ingest", "partner"]
```

The `"ingest  partner"` regression case passes against plain `String.split(scopes)`.

PR #3421, opened 2026-04-30
Fix BigQuery SQL injection vulnerabilities in query construction pipeline

Ziinc:
This PR introdudces tighter validation around  `users.bigquery_dataset_id` and `backends.config.dataset_id` with additional sanitization to ensure that sql injection is not possible.

Closes PRODSEC-33

https://claude.ai/code/session_0128QMzSJtg1wEPPSy78hLSH


lib/logflare/user.ex
```diff
@@ -272,6 +273,15 @@ defmodule Logflare.User do
     end)
   end
 
+  @bq_identifier_pattern ~r/\A[a-zA-Z0-9_]+\z/
```
amokan: We have an identical module attr in `bigquery_adaptor.ex:248`. May be worth using one as a source of truth or moving to a new module to avoid drift.

lib/logflare_web/utils.ex
```diff
@@ -127,7 +127,7 @@ defmodule LogflareWeb.Utils do
 
       replacement =
```
amokan, line 128: I get this function is for display purposes, but we could tighten a little more by covering backslash escapes.

Honestly not a major issue _unless_ someone used this function for actual execution paths. May be good enough to just add a `@doc` to the function to indicate the use case clearly 🤷‍♂️ 

```suggestion
      replacement =
        case type do
          "STRING" ->
            escaped =
              value
              |> String.replace("\\", "\\\\")
              |> String.replace("'", "''")
        
            "'#{escaped}'"
        
          num when num in ["INTEGER", "FLOAT"] ->
            inspect(value)
        
          _ ->
            inspect(value)
        end
```

amokan (approved):
Looks pretty good. Some minor notes.

PR #3422, opened 2026-04-30
Fix authentication bypass in node shutdown endpoint

Ziinc:
This PR tightens security around the LOGFLARE_NODE_SHUTDOWN_CODE, using a header verification instead of query parameter, and also preventing shutdown via the API if code is not set.
Timing attack prevention has also been added. 

Closes PRODSEC-44

https://claude.ai/code/session_01FTArwTBUZpCbanBxaViw8i

### DIFF
--- a/lib/logflare_web/controllers/admin_controller.ex
+++ b/lib/logflare_web/controllers/admin_controller.ex
@@ -2,6 +2,7 @@ defmodule LogflareWeb.AdminController do
   use LogflareWeb, :controller
 
   import Ecto.Query, only: [from: 2]
+  import Logflare.Utils.Guards, only: [is_non_empty_binary: 1]
 
   alias Logflare.Admin
   alias Logflare.Repo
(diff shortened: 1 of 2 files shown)


test/logflare_web/controllers/admin_controller_test.exs
```diff
@@ -2,6 +2,65 @@ defmodule LogflareWeb.AdminControllerTest do
...
+        |> put(~p"/admin/shutdown")
+
+      assert json_response(conn, 401)
+    end
+
+    test "nil code in query param no longer bypasses auth when env var is unset", %{conn: conn} do
+      conn = put(conn, ~p"/admin/shutdown")
+
+      assert json_response(conn, 401)
+    end
```
amokan, line 64: Maybe worth including a 'happy path' test here?

```elixir
test "shuts down node when secret matches header", %{conn: conn} do
  Application.put_env(:logflare, :node_shutdown_code, "correct-secret")

  expect(Logflare.Admin, :shutdown, fn -> :ok end)

  conn =
    conn
    |> put_req_header("lf-shutdown-code", "correct-secret")
    |> put(~p"/admin/shutdown")

  assert json_response(conn, 200)
end
```

lib/logflare_web/controllers/admin_controller.ex
```diff
@@ -18,6 +18,17 @@ defmodule LogflareWeb.AdminController do
   ]
   defp env_node_shutdown_code, do: Application.get_env(:logflare, :node_shutdown_code)
 
+  defp valid_shutdown_code?(provided) do
```
amokan: Nit.

If we add an `import Logflare.Utils.Guards, only: [is_non_empty_binary: 1]`, this could read a bit clearer.

```elixir
defp valid_shutdown_code?(provided) when is_non_empty_binary(provided) do
  case env_node_shutdown_code() do
    configured when is_non_empty_binary(configured) ->
      Plug.Crypto.secure_compare(configured, provided)

    _ ->
      false
  end
end

defp valid_shutdown_code?(_provided), do: false
```

test/logflare_web/controllers/admin_controller_test.exs
```diff
@@ -2,6 +2,65 @@ defmodule LogflareWeb.AdminControllerTest do
...
+
+      conn =
+        conn
+        |> put_req_header("lf-shutdown-code", "wrong-secret")
+        |> put(~p"/admin/shutdown")
+
+      assert json_response(conn, 401)
+    end
+
+    test "nil code in query param no longer bypasses auth when env var is unset", %{conn: conn} do
```
amokan: Nit.


```suggestion
    test "returns 401 when env is unset and no header sent", %{conn: conn} do
```

amokan (approved):
A couple minor nitpicks - but generally LGTM.

PR #3423, opened 2026-04-30
fix: verify Stripe webhook signatures to prevent forged billing events

Ziinc:
## Summary

- Update to stripity_stripe 3.2
- Replace the unverified `StripeController` with `Stripe.WebhookPlug` placed in the endpoint before `Plug.Parsers`, enabling HMAC-SHA256 signature verification on every incoming webhook request
- Move billing event logic into a new `LogflareWeb.StripeWebhookHandler` implementing the `Stripe.WebhookHandler` behaviour
- Reject all requests with a 400 when `STRIPE_WEBHOOK_SECRET` is not configured (random per-request secret ensures no valid signature can be produced)

## Test plan

- [x] Verify `STRIPE_WEBHOOK_SECRET` is set in the production environment
- [x] Confirm requests with a missing or invalid `Stripe-Signature` header return 400
- [x] Confirm requests with a valid signature continue to process billing events correctly
- [x] Confirm that unsetting `STRIPE_WEBHOOK_SECRET` causes all webhook requests to be rejected

Closes PRODSEC-35

https://claude.ai/code/session_01US7DAoh3UXBjSUn2fDV5Gx

---
_Generated by [Claude Code](https://claude.ai/code/session_01US7DAoh3UXBjSUn2fDV5Gx)_


test/logflare_web/controllers/stripe_controller_test.exs
```diff
@@ -16,183 +18,224 @@ defmodule LogflareWeb.StripeControllerTest do
     {:ok, user: user, billing_account: billing_account}
   end
 
-  describe "event/2 - invoice events" do
-    test "syncs invoices when billing account exists", %{conn: conn, billing_account: ba} do
-      expect(Stripe.Invoice, :list, fn %{customer: "cus_test123"} ->
-        {:ok, %Stripe.List{data: []}}
+  describe "signature verification - no webhook secret configured" do
+    setup do
+      Application.put_env(:logflare, :stripe_webhook_secret, nil)
```
amokan: I think this could cause some weirdness given this is likely async given the global context of `Application.put_env`.

Maybe worth considering setting these to `async: false` or using Mimic to stub `LogflareWeb.Endpoint.stripe_webhook_secret/0`?

lib/logflare_web/stripe_webhook_handler.ex
```diff
@@ -0,0 +1,157 @@
...
+  @spec extract_customer(map() | nil, map() | nil) :: String.t() | nil
+  defp extract_customer(object, prev_attrs) do
+    customer_from_object =
+      object && (Map.get(object, :customer) || Map.get(object, "customer"))
+
+    customer_from_prev =
+      prev_attrs && (get_in(prev_attrs, [:customer]) || get_in(prev_attrs, ["customer"]))
+
+    cond do
+      is_binary(customer_from_object) -> customer_from_object
```
amokan: Nit. Could use the `is_non_empty_binary` guard in `Logflare.Utils.Guards` if there is any opportunity for an empty string value to end up in here 🤷‍♂️ 

Def not required though.

lib/logflare_web/stripe_webhook_handler.ex
```diff
@@ -0,0 +1,157 @@
...
+      Logger.info("Stripe webhook: #{type}")
+      :ok
+    else
+      nil -> log_customer_not_found()
+      err -> log_and_error(err)
+    end
+  end
+
+  defp dispatch("payment_method.attached", object, customer) do
+    stripe_id = Map.get(object, :id) || Map.get(object, "id")
```
amokan: Looks like in the stripity stripe code these are all going to be coming in as atom-based keys (_see `deps/stripity_stripe/lib/stripe/converter.ex:32-36`_).

I'm def not super fluent with the library though - so maybe there is another reason we are ORing all of these? If my atom keys hunch is correct, a number of these could extracted from the head with pattern matching.

amokan (approved):
Looks pretty good. Really the only real concern I have is around that `Application.put_env` in the test setup.

lib/logflare_web/stripe_webhook_handler.ex
Baishan: Just claude trying to make it compatible with v2. I've updated it

lib/logflare_web/stripe_webhook_handler.ex
Baishan: Nice catch, fixed

test/logflare_web/controllers/stripe_controller_test.exs
Baishan: I think conncase is async false by default? Some of the tests write to the db for payments. I've stubbed this out anyway.

config/runtime.exs
```diff
@@ -280,6 +280,10 @@ config :stripity_stripe,
          publishable_key: System.get_env("STRIPE_PUBLISHABLE_KEY")
        )
 
+if config_env() != :test do
+  config :logflare, :stripe_webhook_secret, System.get_env("STRIPE_WEBHOOK_SECRET")
```
Ziinc, line 284: Are we missing the secret for staging? Don't see a change file for it

lib/logflare/billing.ex
```diff
@@ -140,6 +139,13 @@ defmodule Logflare.Billing do
...
   @doc "retrieves the stripe plan stored on the BillingAccount"
   @spec get_billing_account_stripe_plan(BillingAccount.t()) :: nil | map()
+  def get_billing_account_stripe_plan(%BillingAccount{
+        stripe_subscriptions: %{
+          "data" => [%{"items" => %{"data" => [%{"plan" => plan} | _]}} | _]
+        }
+      }),
+      do: plan
+
   def get_billing_account_stripe_plan(%BillingAccount{
```
Ziinc, line 149: So this never gets called?

lib/logflare_web/plugs/stripe_webhook.ex
```diff
@@ -0,0 +1,33 @@
...
+  def init(opts), do: Stripe.WebhookPlug.init(opts)
+
+  def call(conn, opts) do
+    conn
+    |> Stripe.WebhookPlug.call(opts)
+    |> log_rejections()
+  end
+
+  @spec log_rejections(Plug.Conn.t()) :: Plug.Conn.t()
+  defp log_rejections(%{status: 400} = conn) do
```
Ziinc, line 23: I don't think we need to log rejections as we have request traces for this. I think we can add attributes to the trace instead

Ziinc (commented):
Lgtm, please merge when ready

config/runtime.exs
Baishan: We don't have a webhook configured in stripe for it.

lib/logflare/billing.ex
Baishan: Probably not. Accounts stored with v2 will have the top level key but v3 won't. They both should have items but without checking the prod db I was reluctant to remove it in case an older version stored plans without the items key.

lib/logflare_web/plugs/stripe_webhook.ex
Baishan: I don't expect these to be hit often (they shouldn't be hit at all ) but we would see a spike in them during deploy if the secret is misconfigured.

PR #3425, opened 2026-04-30
fix: enforce auth before syslog body parsing and remove log injection

Ziinc:
## Summary

- Move `VerifyApiAccess` ahead of `Plug.Parsers` on ingest pipelines so unauthenticated callers can no longer reach the parsers. This closes the log-injection path through `SyslogParser` and prevents future parser-side bugs from leaking attacker-supplied content on the unauthenticated path.
- Introduce dedicated `:ingest_api` and `:ingest_otlp_api` pipelines for `/logs`, `/api/logs`, `/api/events`, the Cloudflare logpush scope, and `/v1/*` OTLP routes. The shared `:api` pipeline is left alone so `/health`, webhooks, OAuth, OpenAPI etc. keep working without API auth.
- `FetchResource` and `VerifyResourceAccess` still run **after** parsing because some clients (notably the BERT logger at `/logs/elixir/logger`) identify the source via a body field. Auth running first is enough to keep unauthenticated bodies away from the parsers.
- Add a `require_token: true` option to `VerifyApiAccess`. Without it the plug's existing `{:error, :no_token}` clause passes unauthenticated requests through whenever a `resource_type` assign is set — a behavior the public-endpoint query path (`enable_auth: false`) relies on, but which would otherwise silently defeat the new ordering. The ingest pipelines now pass this option; the endpoint and mgmt pipelines do not.
- `VerifyApiAccess` calls `fetch_query_params/1` at the top of `call/2` so legacy `?api_key=` clients still authenticate even when the plug runs before `Plug.Parsers` (which previously fetched them).
- Redact raw bodies from parser error logs. `SyslogParser` no longer logs the raw message string or `?source=` param. `NdjsonParser` was logging `inspect/1` of the `Jason.DecodeError`, which carries the raw failing input in `:data` and `:token` — it now logs only the decode position. Defence-in-depth on the authenticated path; auth ordering already protects the unauthenticated path.
- Add controller-level tests for malformed payloads on `/logs/logplex`, `/logs/cloudflare`, and `/v1/logs` (unauthenticated, authenticated, and invalid-key) asserting (a) parser-error log lines never appear for unauthenticated or invalid-key requests, (b) an attacker-controlled marker never appears in the captured logs even when the parser does run.
- Lock in body-sourced ingest paths against the resource-after-parsing reorder: BERT body `source_name` to `/logs/elixir/logger`, JSON body `source` token to `/logs`, JSON body `source_name` to `/logs`.
- Add plug-level coverage of the new `require_token` option in `verify_api_access_test.exs`.

## Test plan

- [x] `mix test test/logflare_web/controllers/log_controller_test.exs` (29/29)
- [x] `mix test test/logflare_web/controllers/plugs/verify_api_access_test.exs` (14/14)
- [x] `mix test test/logflare_web/controllers/endpoints_controller_test.exs test/logflare_web/controllers/api/endpoint_controller_test.exs` (38/38, 2 skipped)
- [x] `mix lint.diff` clean
- [x] Manual: unauth POST `/logs/logplex`, `/logs/cloudflare`, and `/v1/logs` with malformed bodies all return 401 with no `Syslog message parsing error` / `NDJSON parser error` line in the dev server log (parsers never run)
- [x] Manual: authenticated POST `/logs/logplex` and `/logs/cloudflare` with malformed bodies do log a redacted parser-error line and never contain the request body
- [x] Manual: legacy `?api_key=` query auth on `POST /logs?source=<token>&api_key=...` returns 200 (verifies `fetch_query_params` defensive fetch)
- [x] Manual: BERT body with `source_name` to `POST /logs/elixir/logger` returns 200 (verifies the body-sourced source-name path still works post-reorder)
- [x] Manual: `/health`, OpenAPI spec, mgmt API list/create/delete with bearer auth, mgmt API 401 without auth — all behave as before
- [x] Manual: public endpoint (`enable_auth: false`) reachable unauthenticated by UUID token (verifies the `require_token` opt-in correctly preserves the public-endpoint pathway)

Closes PRODSEC-39

https://claude.ai/code/session_0188UD1xQBYhqoM9Lt1WsPcx

---
_Generated by [Claude Code](https://claude.ai/code/session_0188UD1xQBYhqoM9Lt1WsPcx)_

### DIFF
--- a/test/logflare_web/controllers/log_controller_test.exs
+++ b/test/logflare_web/controllers/log_controller_test.exs
@@ -253,6 +253,37 @@ defmodule LogflareWeb.LogControllerTest do
       refute log =~ "Syslog message parsing error"
     end
 
+    test "invalid api key on /logs/cloudflare 401s before parsing the body", %{conn: conn} do
+      log =
+        capture_log(fn ->
+          conn =
+            conn
+            |> put_req_header("content-type", "application/x-ndjson")
+            |> put_req_header("x-api-key", "this-key-is-not-valid")
+            |> post(~p"/logs/cloudflare", @malformed_ndjson)
+
+          assert json_response(conn, 401)
+        end)
+
+      refute log =~ @attacker_marker
+      refute log =~ "NDJSON parser error"
+    end
+
+    test "invalid api key on /v1/logs 401s before parsing the body", %{conn: conn} do
+      log =
+        capture_log(fn ->
+          conn =
+            conn
+            |> put_req_header("content-type", "application/x-protobuf")
+            |> put_req_header("x-api-key", "this-key-is-not-valid")
+            |> post(~p"/v1/logs", "garbage protobuf #{@attacker_marker}")
+
+          assert json_response(conn, 401)
+        end)
+
+      refute log =~ @attacker_marker
+    end
+
     test "authenticated /logs/logplex with malformed body does not log raw content", %{conn: conn} do
       user = insert(:user)
       source = insert(:source, user: user)
--- a/test/logflare_web/controllers/log_controller_test.exs
+++ b/test/logflare_web/controllers/log_controller_test.exs
@@ -319,6 +319,23 @@ defmodule LogflareWeb.LogControllerTest do
       assert log =~ "NDJSON parser error"
       refute log =~ @attacker_marker
     end
+
+    test "authenticated /logs/logplex error log does not include ?source= query param",
+         %{conn: conn} do
+      user = insert(:user)
+      insert(:plan, name: "Free")
+
+      log =
+        capture_log(fn ->
+          conn
+          |> put_req_header("x-api-key", user.api_key)
+          |> put_req_header("content-type", "application/logplex-1")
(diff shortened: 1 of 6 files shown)


test/logflare_web/controllers/log_controller_test.exs
```diff
@@ -182,6 +183,135 @@ defmodule LogflareWeb.LogControllerTest do
...
+            |> put_req_header("content-type", "application/x-protobuf")
+            |> post(~p"/v1/logs", "garbage protobuf #{@attacker_marker}")
+
+          assert json_response(conn, 401)
+        end)
+
+      refute log =~ @attacker_marker
+    end
+
+    test "invalid api key on /logs/logplex 401s before parsing the body", %{conn: conn} do
```
amokan, line 240: Not a major thing - but should we consider doing a similar assertion for the `/logs/cloudflare` and `/v1/logs` routes?

amokan (approved):
Looks good. Only question I have is if we should consider my point about covering those other routes with a similar test to prevent future regressions.

PR #3426, opened 2026-04-30
test: exhaustive coverage for Lql.BackendTransformer.Postgres

laraujo7:
Also removes dead code, on line 65 of lib/logflare/lql/backend_transformer/postgres.ex

### DIFF
--- a/lib/logflare/lql/backend_transformer/postgres.ex
+++ b/lib/logflare/lql/backend_transformer/postgres.ex
@@ -55,16 +55,10 @@ defmodule Logflare.Lql.BackendTransformer.Postgres do
   def apply_select_rules_to_query(query, [], _opts), do: query
 
   def apply_select_rules_to_query(query, select_rules, _opts) do
-    normalized_rules =
-      case Enum.find(select_rules, & &1.wildcard) do
-        nil -> select_rules
-        _wildcard -> [%{wildcard: true, path: "*"}]
-      end
-
-    case normalized_rules do
-      [] -> query
-      [%{wildcard: true}] -> query
-      rules -> build_combined_select(query, rules)
+    if Enum.any?(select_rules, & &1.wildcard) do
+      query
+    else
+      build_combined_select(query, select_rules)
     end
   end
 
--- a/test/logflare/lql/backend_transformer/postgres_test.exs
+++ b/test/logflare/lql/backend_transformer/postgres_test.exs
@@ -1,11 +1,11 @@
 defmodule Logflare.Lql.BackendTransformer.PostgresTest do
   use Logflare.DataCase, async: true
+  use ExUnitProperties
 
   import Ecto.Query
 
   alias Ecto.Query
   alias Ecto.Query.BooleanExpr
-  alias Ecto.Query.DynamicExpr
   alias Logflare.Lql.BackendTransformer.Postgres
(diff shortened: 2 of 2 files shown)


lib/logflare/lql/backend_transformer/postgres.ex
```diff
@@ -62,7 +62,6 @@ defmodule Logflare.Lql.BackendTransformer.Postgres do
       end
 
     case normalized_rules do
-      [] -> query
```
nelsonmestevao, line 65: Thanks to pattern matching on Line 55, this cannot be empty from the `Enum.find/2` call.

nelsonmestevao (approved):

lib/logflare/lql/backend_transformer/postgres.ex
```diff
@@ -62,7 +62,6 @@ defmodule Logflare.Lql.BackendTransformer.Postgres do
...
       [%{wildcard: true}] -> query
       rules -> build_combined_select(query, rules)
```
Ziinc: looks like we can combine both cases

test/logflare/lql/backend_transformer/postgres_test.exs
```diff
@@ -458,6 +340,16 @@ defmodule Logflare.Lql.BackendTransformer.PostgresTest do
...
     end
+
+    test "applies a nested JSONB field without an alias" do
+      query = from(l in "logs")
+      select_rule = %SelectRule{path: "m.status"}
+
+      result = Postgres.apply_select_rules_to_query(query, [select_rule], [])
+
+      assert %Ecto.Query{select: %{expr: expr}} = result
+      assert expr |> Macro.to_string() =~ "m.status"
```
Ziinc: should assert on the operator as well.

test/logflare/lql/backend_transformer/postgres_test.exs
```diff
@@ -515,6 +407,160 @@ defmodule Logflare.Lql.BackendTransformer.PostgresTest do
     end
   end
 
+  describe "transform_filter_rule/2 — additional top-level operators" do
```
Ziinc: a property test with StreamData would likely work better here, since we're more concerned about the validation of the DynamicExpr generation.

Ziinc (changes requested):

test/logflare/lql/backend_transformer/postgres_test.exs
```diff
@@ -515,6 +407,160 @@ defmodule Logflare.Lql.BackendTransformer.PostgresTest do
...
+          operator: :=,
+          value: "a@b.c",
+          modifiers: %{}
+        )
+
+      assert %DynamicExpr{} = Postgres.transform_filter_rule(filter_rule, %{})
+    end
+  end
+
+  describe "transform_select_rule/2 — error case" do
```
Ziinc: sahould combine with the main block at :384

test/logflare/lql/backend_transformer/postgres_test.exs
Ziinc: also, the tests should assert on the values via `inspect/1`, otherwise their utility will remain low.
https://hexdocs.pm/ecto/dynamic-queries.html#building-dynamic-queries

Ziinc (approved):

PR #3427, opened 2026-04-30
Update stale inspect_fun docs and extend redaction property tests

0xAX:
This PR provides the three minor improvements for the `Logflare.Utils.inspect_fun/3` and private API it uses:

1. The stale docstring is updated with the mentioning all redacted structures.

2. Extended the property test struct generator with `Tesla.Client` and `Tesla.Env` with a nested `__client__`, covering the missing redaction paths in property tests.

3. The @sensitive_header_names attribute was re-used everywhere where possible.

### DIFF
--- a/lib/logflare/utils.ex
+++ b/lib/logflare/utils.ex
@@ -11,6 +11,8 @@ defmodule Logflare.Utils do
   import Cachex.Spec
   import Logflare.Utils.Guards, only: [is_atom_value: 1]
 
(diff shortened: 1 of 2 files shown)


Ziinc (approved):

0xAX:
@Ziinc hello, after fast glance seems failed test is flaky by itself and not really related to PR or I should take a closer look?

PR #3431, opened 2026-05-04
fix: adds the create home team flow to DashboardLive

Ziinc:
This PR adds back the create home team flow back to the DashboardLive.

It also adds a `Users.create_user(%TeamUser{})`, so that we can consolidate some of the logic around user creation inside of the Users context and deprecate/remove old logic in auth controller over time.

CleanShot 2026-05-04 at 19 34 27@2x


lib/logflare/users/users.ex
```diff
@@ -246,6 +248,31 @@ defmodule Logflare.Users do
...
+        :provider_uid,
+        :name,
+        :image,
+        :phone,
+        :email_preferred,
+        :valid_google_account
+      ])
+
+    Repo.transaction(fn ->
+      with {:ok, user} <- insert_user(auth_params),
```
amokan, line 270: Are we cool with the new user not getting BQ IAM access and not getting the welcome email? I know this is a change that will be in transition for a bit, but just wanted to be sure.

lib/logflare/users/users.ex
```diff
@@ -10,6 +10,8 @@ defmodule Logflare.Users do
   alias Logflare.Sources
   alias Logflare.Sources.Source
   alias Logflare.Backends
+  alias Logflare.Generators
```
amokan: Nit: would be nice to order these aliases alphabetically

lib/logflare_web/live/dashboard_live.ex
```diff
@@ -15,6 +16,12 @@ defmodule LogflareWeb.DashboardLive do
   def mount(_, _session, socket) do
     %{user: user} = socket.assigns
 
+    home_team_exists? =
```
amokan: Nit.

What about inverting the logic here and naming this `can_create_home_team?` ?
Could make the call sites a bit more natural to read 🤷‍♂️

lib/logflare/users/users.ex
Ziinc: nice catch, yea we should preserve those. i've moved it down to the context level to make sure its not missed out in the future.

PR #3433, opened 2026-05-05
feat: default LQL rules for source

msmithstubbs:
Adds a default LQL query setting for sources. This LQL query is appended to any new source search.

* default LQL query can be set per source.
* any valid LQL expression can be set. 
* when starting a search the default LQL query will be appended
* editing the query for an existing search will not append the default LQL, allowing the user to remove it.

### Questions

1. Currently all LQL rules (select, filter, etc) are supported. Is this correct?
2. The LQL parser only removes duplicate select rules. Do we need to handle other depublication?
3. Should the source default LQL field be available to mange through the source API? 
 

Follows on from #3432 

Closes O11Y-740

CleanShot 2026-05-06 at 09 09 53@2x

## Demo

https://github.com/user-attachments/assets/b0f463eb-ef07-44b7-9beb-46ded56ebdf2

### DIFF
--- a/lib/logflare_web/controllers/source_controller.ex
+++ b/lib/logflare_web/controllers/source_controller.ex
@@ -323,12 +323,13 @@ defmodule LogflareWeb.SourceController do
 
   def update(
         %{assigns: %{source: source, user: _user, plan: _plan}} = conn,
-        %{"source" => %{"default_search_lql" => lqlstring} = params}
+        %{"source" => %{"default_search_lql" => lqlstring} = _params}
       ) do
     schema = get_bigquery_schema(source)
 
     with {:ok, _lql_rules} <- Lql.Parser.parse(lqlstring, schema),
-         {:ok, _changeset} <- Sources.update_source_by_user(source, %{"default_search_lql" => lqlstring}) do
+         {:ok, _changeset} <-
+           Sources.update_source_by_user(source, %{"default_search_lql" => lqlstring}) do
       conn
       |> put_flash(:info, "Source updated!")
       |> redirect(to: Routes.source_path(conn, :edit, source.id))
--- a/lib/logflare_web/controllers/source_controller.ex
+++ b/lib/logflare_web/controllers/source_controller.ex
@@ -328,7 +328,7 @@ defmodule LogflareWeb.SourceController do
     schema = get_bigquery_schema(source)
 
     with {:ok, _lql_rules} <- Lql.Parser.parse(lqlstring, schema),
-         {:ok, _changeset} <- Sources.update_source_by_user(source, params) do
+         {:ok, _changeset} <- Sources.update_source_by_user(source, %{"default_search_lql" => lqlstring}) do
       conn
       |> put_flash(:info, "Source updated!")
       |> redirect(to: Routes.source_path(conn, :edit, source.id))
--- a/lib/logflare_web/controllers/source_controller.ex
+++ b/lib/logflare_web/controllers/source_controller.ex
@@ -322,6 +322,48 @@ defmodule LogflareWeb.SourceController do
   end
 
   def update(
+        %{assigns: %{source: source, user: _user, plan: _plan}} = conn,
+        %{"source" => %{"default_search_lql" => lqlstring} = params}
+      ) do
+    schema = get_bigquery_schema(source)
+
+    with {:ok, _lql_rules} <- Lql.Parser.parse(lqlstring, schema),
+         {:ok, _changeset} <- Sources.update_source_by_user(source, params) do
+      conn
+      |> put_flash(:info, "Source updated!")
+      |> redirect(to: Routes.source_path(conn, :edit, source.id))
+    else
+      {:error, %Ecto.Changeset{} = changeset} ->
+        conn
+        |> put_status(406)
+        |> put_flash(:error, "Something went wrong!")
+        |> render(
+          "edit.html",
+          changeset: changeset,
+          source: source,
+          notifications_opts: notifications_options()
+        )
+
+      {:error, :field_not_found, _suggested_querystring, error} ->
+        conn
+        |> put_flash(:error, error)
+        |> redirect(to: Routes.source_path(conn, :edit, source.id))
+
+      {:error, error} when is_binary(error) ->
+        conn
+        |> put_flash(:error, "Default search LQL: #{error}")
+        |> redirect(to: Routes.source_path(conn, :edit, source.id))
+
+      e ->
+        Logger.error("Error updating default search LQL.", error_string: inspect(e))
+
+        conn
+        |> put_flash(:error, "Something else went wrong. Contact support if this continues.")
+        |> redirect(to: Routes.source_path(conn, :edit, source.id))
+    end
+  end
+
+  def update(
         %{assigns: %{source: source, user: _user, plan: plan}} = conn,
         %{"source" => %{"notifications_every" => _freq} = params}
       ) do
--- a/lib/logflare_web/live/search_live/logs_search_lv.ex
+++ b/lib/logflare_web/live/search_live/logs_search_lv.ex
@@ -102,7 +102,7 @@ defmodule LogflareWeb.Source.SearchLV do
       uri: nil,
       lql_rules: [],
       saved_searches: saved_searches(source),
-      querystring: Map.get(params, "querystring", @default_qs),
+      querystring: querystring_param_or_default(params, source, flat_map),
       force_query: Map.get(params, "force", "false") == "true"
     )
     |> maybe_assign_user_timezone(team_user, user)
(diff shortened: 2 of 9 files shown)


lib/logflare_web/controllers/source_controller.ex
```diff
@@ -321,6 +321,48 @@ defmodule LogflareWeb.SourceController do
...
   end
 
+  def update(
+        %{assigns: %{source: source, user: _user, plan: _plan}} = conn,
+        %{"source" => %{"default_search_lql" => lqlstring} = params}
+      ) do
+    schema = get_bigquery_schema(source)
+
+    with {:ok, _lql_rules} <- Lql.Parser.parse(lqlstring, schema),
+         {:ok, _changeset} <- Sources.update_source_by_user(source, params) do
```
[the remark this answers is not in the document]
msmithstubbs: This is a reasonable suggestion but isn't consistent with how the other `update/2` branches currently behave.

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -1060,6 +1075,23 @@ defmodule LogflareWeb.Source.SearchLV do
     {:noreply, socket}
   end
 
+  defp append_default_search_lql(source, qs) do
+    "#{qs} #{source.default_search_lql}"
+    |> String.trim()
+  end
```
Ziinc: instead of appending all the time (even if a query param is provided), a better ux would be to only add in the default search lql if there are no queries. otherwise the filter will get appended to the search when sharing links with others.

Ziinc (changes requested):

test/logflare_web/live/search_live/logs_search_lv_test.exs
```diff
@@ -498,6 +503,61 @@ defmodule LogflareWeb.Source.SearchLVTest do
...
+      allow_sandbox(search_executor_pid)
+
+      querystring = find_querystring(html)
+
+      assert querystring =~ "s:m.level"
+      assert querystring =~ "c:count(*) c:group_by(t::minute)"
+    end
+
+    @tag source_attrs: [default_search_lql: "s:m.level"]
+    test "appends source default LQL to an initial querystring param", %{
```
Ziinc: we don't want this to happen, otherwise what is shown won't be reflective of what is shared to users.

Ziinc: esp in situations where a query is built/templated

test/logflare_web/live/search_live/logs_search_lv_test.exs
```diff
@@ -498,6 +503,61 @@ defmodule LogflareWeb.Source.SearchLVTest do
...
+      {:ok, view, html} = live(conn, Routes.live_path(conn, SearchLV, source.id))
+
+      %{executor_pid: search_executor_pid} = get_view_assigns(view)
+      allow_sandbox(search_executor_pid)
+
+      querystring = find_querystring(html)
+
+      assert querystring =~ "s:m.level"
+      assert querystring =~ "c:count(*) c:group_by(t::minute)"
+    end
```
Ziinc, line 520: should add an additional verification that on subsequent searches (say the select rule is removed and more filters added), that the default search lql is not re-added

test/logflare_web/controllers/source_controller_test.exs
```diff
@@ -1040,6 +1040,56 @@ defmodule LogflareWeb.SourceControllerTest do
...
+      conn =
+        patch(conn, ~p"/sources/#{source}", %{
+          "source" => %{"default_search_lql" => ""}
+        })
+
+      assert redirected_to(conn, 302) =~ ~p"/sources/#{source}/edit"
+      assert Phoenix.Flash.get(conn.assigns.flash, :info) == "Source updated!"
+      assert Repo.reload(source).default_search_lql == nil
+    end
+
```
Ziinc: nit: this as a source_test.exs test would be better

Ziinc (changes requested):

test/logflare_web/live/search_live/logs_search_lv_test.exs
msmithstubbs: So the default search LQL should only be added if the user is coming directly from the recent logs (`/sources/:id`) page. We add it even if they have entered a search query.

If they hit the Search view directly (`/sources/:id/search`) we don't add the default LQL unless the query string is empty.

Ziinc: yes, two entry points:
1. from recent logs with a fresh user-provided query
2. directly hitting /sources/:id/search with no query string

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -1060,6 +1075,23 @@ defmodule LogflareWeb.Source.SearchLV do
     {:noreply, socket}
   end
 
+  defp append_default_search_lql(source, qs) do
```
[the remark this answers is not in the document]
msmithstubbs: Outdated. Function no longer exists.

test/logflare_web/controllers/source_controller_test.exs
msmithstubbs: Done. Had to enable that test case first: #3440

test/logflare_web/live/search_live/logs_search_lv_test.exs
msmithstubbs: Tried a slightly different approach. The default search LQL is now set as the default value for the source search field. This provides a hint to the user that the default search will be included. The cursor is positioned after the default query ready for the user to add additional rules. The user can modify the field to remove the default search if they wish. 

This deviates from the requirements in O11Y-740 slightly (“alway add, even if user enters their own filters“) by allowing the user to remove the default search but it is simpler and more predictable.

When the request hits the search live view default LQL search is only added if the query string param is missing, or is empty (`""`) .

`/sources/123/search` -> default LQL search is added 
`/sources/123/search?querystring=` -> default LQL search (querystring is blank)
`/sources/123/search?querystring=foo` -> default lql search not added 


https://github.com/user-attachments/assets/8503bf57-e958-4e32-813e-7aac6446f3e7

lib/logflare_web/live/search_live/logs_search_lv.ex
msmithstubbs: No longer appended to existing querystring unless blank.

test/logflare_web/live/search_live/logs_search_lv_test.exs
msmithstubbs: Added this test, though the new model doesn't append to an existing query anyway.

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -35,7 +35,6 @@ defmodule LogflareWeb.Source.SearchLV do
 
   @tail_search_interval 1000
   @user_idle_interval :timer.minutes(2)
-  @default_qs "c:count(*) c:group_by(t::minute)"
```
msmithstubbs, line 38: Removed as it was being set by rules anyway, eg: 

https://github.com/Logflare/logflare/blob/67f69d327306ae1332ba87a7062c1d782f09d547/lib/logflare_web/live/search_live/logs_search_lv.ex#L181

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -226,17 +227,6 @@ defmodule LogflareWeb.Source.SearchLV do
     {:noreply, socket}
   end
 
-  def handle_params(%{"source_id" => source}, uri, socket) do
-    params = %{"source_id" => source, "querystring" => "", "tailing?" => "true"}
-    handle_params(params, uri, socket)
-  end
-
-  def handle_params(_params, _uri, socket) do
```
msmithstubbs, line 234: Redundant: no route to this view without a source_id

https://github.com/Logflare/logflare/blob/67f69d327306ae1332ba87a7062c1d782f09d547/lib/logflare_web/router.ex#L303-L304

test/logflare_web/live/search_live/logs_search_lv_test.exs
Ziinc: yes this ux looks great 👍

lib/logflare_web/templates/source/show.html.heex
```diff
@@ -110,6 +112,12 @@
 
 <script>
   document.addEventListener("DOMContentLoaded", async () => {
+      const searchInput = document.getElementById("source-search-querystring")
+      if (searchInput && searchInput.value !== "") {
+        searchInput.focus()
+        searchInput.setSelectionRange(searchInput.value.length, searchInput.value.length)
+      }
+
       await Source.main({scrollTracker: true}, {avgEventsPerSecond: <%= @source.metrics.avg %>})
```
Ziinc, line 121: this seems quite unnecessary, we can let the changeset handle this

Ziinc (approved):
nice, very elegant alternative method 💯

lib/logflare_web/templates/source/show.html.heex
msmithstubbs: This JS code moves the cursor to the end of the input so the user can start typing straight away.

CleanShot 2026-05-11 at 09 13 45@2x

I'm not sure how to handle it with a changeset, but maybe the intention of the code isn't clear.

PR #3434, opened 2026-05-05
Revert "chore: Fix DB SSL client cert filename check in runtime config"

Ziinc:
Reverts Logflare/logflare#3402

temp revert


Ziinc:
- [ ]

PR #3435, opened 2026-05-05
fix: ssl_opts should not be used, add :server_name_indication option

Ziinc:
Updates ssl options to use `:ssl` instead of `:ssl_opts`, and add in disabling of server name (which connecting seems to reject GCP's hostnames as mismatching...)


Ziinc:
Merging in for testing.

PR #3436, opened 2026-05-06
wip: New Monaco editor integration

msmithstubbs:
> [!NOTE]
> Early WIP don't review yet.

The goal is to replace the existing LiveMonacoEditor component with a simpler, more testable model. This approach uses a hidden <textarea> element to store the edited code. The Monaco editor instance writes updates to that element, which in turn triggers the standard form change event handled by LiveView. Updates from LiveView still use a hook, but the Monaco editor itself no longer sends events directly to LiveView. This should be much closer to the familiar form-backed LiveView model.

The intention is to enable JavaScript-based test coverage for editor and DOM interactions without requiring a running LiveView instance. I also expect to add a small number of end-to-end Playwright tests to sanity-check the integration. 

Another advantage of this approach is that it allows us to vendor the Monaco JavaScript dependencies directly into our repository.

Follows #3444 

Fixes O11Y-1684

 ## TODO 

- [x] Replace QueryLive editor
- [x] Replace Search LQL editor
- [x] Replace Endpoints editor
- [ ] JS unit tests for monaco_hook
- [ ] E2E integration test.


Ziinc:
@msmithstubbs can you give this lib a try to see if we can have a more lightweight testing strategy for the js areas, especially for Monaco editor testing 

https://github.com/Ziinc/phoenix_test_jsdom

msmithstubbs:
I've got the JSDOM library working locally but the biggest missing piece is cookie support so it can hit authenticated routes. Could submit a PR for this? Or is there another approach to take? If we just want to test the monaco editor in isolation we could just mount it in a harness designed for testing, and leave Playwright to handle full integration testing

msmithstubbs:
Closing, now #3525

PR #3439, opened 2026-05-06
test: stabilize and instrument Supabase E2E logs.spec.ts

djwhitt:
## Summary

Stabilize the intermittent failures in `test/e2e/supabase/features/logs.spec.ts`. Post-fix flake rate measured at **0/30** across 30 sequential reruns (was regularly failing across browsers).

## Root cause

Logflare's startup in supabase mode seeds the predefined sources (`storage.logs.prod.2`, `gotrue.logs.prod`, `postgres.logs`, `cloudflare.logs.prod`, etc. — see `lib/logflare/single_tenant.ex`) inside `startup_tasks`. Bandit binds the HTTP port *before* that seeding completes. POSTs that land in the seeding window get HTTP 401, because the source isn't installed yet.

Vector's HTTP sink classifies 401 as non-retriable and **drops the events permanently** — no retry, no error visible to the test, no entry in the Logflare table. Two adjacent issues stacked on top to produce the observed empty-table flake (most often on `storage_logs`, occasionally elsewhere):

1. The upstream `analytics` Docker healthcheck used `curl http://localhost:4000/health` without `--fail`. `/health` returns 503 while `startup_tasks` is still seeding, but `curl` exits 0 on a non-2xx response unless `--fail` is set — so the container reported "healthy" while seeding was in progress.
2. With analytics reporting healthy too early, Vector started in parallel with seeding. Its first POST to a not-yet-seeded source got 401 and dropped the event.
3. The original `beforeAll` 30s blanket sleep masked this on fast runners most of the time (the seeding window is short relative to 30s) but couldn't recover events that had already been dropped at the Vector → Logflare boundary.

The original PR description on this branch hypothesized that "the 30s sleep isn't enough for the ingestion path to complete on slow CI runners." That hypothesis was wrong. The captured CI failure showed a fully-loaded page rendering an empty-state table, no console errors, and the correct uniqueId in the search URL — i.e. ingestion had completed before the search ran, but the relevant events never arrived. The path wasn't slow; events had been dropped on the floor before Logflare ever saw them. The diagnostic harness shipped earlier in this branch is what made it possible to falsify that hypothesis and find the real one.

The flake didn't reproduce locally even under torture (50/50 with `--repeat-each 50`, 20/20 with shell loop + 4-CPU pressure). Likely because the local Logflare image and seeded state were already warm, so the seeding window was effectively zero by the time Vector started posting. CI cold-starts hit the race window.

The two upstream compose-stack bugs (curl healthcheck without `--fail`, and `vector` not gating on analytics health) are filed at supabase/supabase#45738. Once that lands the overlay below becomes redundant and can be removed.

## Fix

In decreasing order of load-bearingness:

- **`test/e2e/supabase/docker-compose.e2e.yml` overlay** — override the `analytics` healthcheck to use `curl --fail`, so 503-during-seeding actually fails the check. Add `depends_on: analytics: service_healthy` to `vector`. Together these stop Vector from posting into the seeding window. This is the actual race fix. Temporary workaround pending the upstream compose fix tracked in supabase/supabase#45738 — when that ships, this overlay can be deleted.
- **`setup-supabase-services.sh` source-readiness probe** — after `docker compose up`, POST a probe event to each Supabase source until 2xx, with a 60s deadline. Belt-and-suspenders for the test phase. If a source never seeds, the probe times out and dumps the analytics log tail rather than failing silently 30s later as "No results found".
- **`logs.spec.ts` `beforeAll`** — replace the 30s blanket sleep with parallel per-pipeline polling against the same `/api/platform/projects/default/analytics/endpoints/logs.all` endpoint Studio's UI uses. Hygiene rather than a root-cause fix: when the seeding race is open, polling cannot recover dropped events. With the seeding race closed, polling provides faster feedback (typically <1s vs. blanket 30s) and a clean timeout when a pipeline genuinely doesn't deliver.

## Diagnostic harness (separately useful)

The diagnostic infrastructure shipped earlier on this branch is what made the root-cause investigation tractable, and stays in for future flakes:

- `beforeEach`/`afterEach` in `logs.spec.ts` attach the following to the Playwright report on test failure (no-op on success):
  - `table.html` — rendered HTML of the search-results table at the failure moment.
  - `table.txt` — text content of the same table, easier to compare against `expect.toContainText` assertions.
  - `browser-console.json` — captured `console` and `pageerror` events for the test, with timestamps.
  - `network.json` — request/response pairs filtered to relevant URL patterns (`/api/platform/`, `/logs`, `/functions/v1/`, `/auth/v1/`, `/storage/v1/`, `/realtime/v1/`, `/rest/v1/`, `/pg-meta/`). Method, status, content-type, duration, and the first 4 KB of response body. Asset traffic filtered out.
  - `page-url.txt` — the current URL.
- The workflow now also captures `docker logs` for every running container into per-browser `container-logs-*` artifacts on failure.
- `playwright.config.ts` adds the `github` reporter (inline PR annotations) and the `json` reporter (consumed by `ci-failure-summary.sh` to populate `$GITHUB_STEP_SUMMARY`). `expect` timeout bumped 5s → 15s to ride out brief UI/ingestion latency.
- `ci-failure-summary.sh` reads the JSON results and writes a Markdown summary to the GitHub Actions UI on failure, so the failure surface is visible without downloading artifacts.

## Workflow infra

- The pre-existing `actions/upload-artifact@v4` step had two bugs preventing retrieval: the `path` was relative to the repo root rather than `test/e2e/supabase/`, and all three matrix browsers uploaded under the same artifact name (v4 enforces unique names). Both fixed; each browser now produces a separately-named artifact.
- New `scripts/debug-flaky-action.sh` — general-purpose CLI for quantifying flake rate or repro-ing transient failures on any workflow in this repo. `repro` mode reruns until any matching job fails (downloads artifacts, notifies). `sample` mode runs N attempts regardless of outcome and reports the pass/fail rate. Run id is auto-detected from the current branch when not passed; job-name pattern filtering scopes which failures count. See `scripts/debug-flaky-action.sh --help`.

## Verification

- [x] Local: all 7 tests in `logs.spec.ts` pass on chromium in 20.5s with the polling fix in place (vs ~50–70s with the previous 30s blanket sleep). Polling typically completes in <1s on a healthy stack.
- [x] Workflow fix verified: a CI run on this branch produced three uploaded artifacts (`playwright-report-chromium`, `-firefox`, `-webkit`) at ~200 KB each, vs. the previous "No files were found" warning.
- [x] Diagnostic verified end-to-end: the captured Chromium failure that drove the root-cause investigation surfaced `table.html`, `table.txt`, `browser-console.json`, `network.json`, and `page-url.txt` as expected.
- [x] Post-fix flake rate measured via `scripts/debug-flaky-action.sh --mode sample --cap 30` against the `Integration Tests with Supabase` workflow on commit `531de768`: **30/30 success** ([run 25600358994](https://github.com/Logflare/logflare/actions/runs/25600358994), attempts 1–30 spanning 11:46–16:30 UTC on 2026-05-09, ~9–11 min each).

## Out of scope

- The `searchLogs` helper rewrite (`test/e2e/supabase/lib/utils.ts`). Cosmetic now that the seeding race is closed; defensible as a separate "test infra hygiene" change later.
- Splitting the shared `beforeAll` into per-test setup so one slow pipeline doesn't cascade.
- Browser-specific retries / timeouts in `playwright.config.ts`.
- Investigating whether `startup_tasks` ever fails to finish seeding entirely. There is a comment in `setup-supabase-services.sh` flagging evidence of persistent 401 across full test runs in some prior state; the mechanism is unproven. With the probe in place this would now surface as a clean timeout with the analytics log tail dumped, rather than a silent flake.

### DIFF
--- a/test/e2e/supabase/features/logs.spec.ts
+++ b/test/e2e/supabase/features/logs.spec.ts
@@ -1,6 +1,40 @@
-import { test, expect, type Request } from '@playwright/test';
+import { test, expect, type APIRequestContext, type Request } from '@playwright/test';
 import { searchLogs } from '../lib/utils';
 import supabase from '../lib/supabase';
+
+// Poll the logs.all endpoint until at least one row matches `pattern` in `table`.
+// Replaces a fixed sleep that was racy when CI ingestion was slow. The endpoint
+// runs the supplied SQL through Logflare via Studio's BFF; `regexp_contains` is
+// used because `LIKE` returns 500 against this backend.
+async function waitForLogs(
+  request: APIRequestContext,
+  table: string,
+  pattern: string,
+  { timeoutMs = 90_000, intervalMs = 1_000 }: { timeoutMs?: number; intervalMs?: number } = {}
+): Promise<void> {
+  const sql = `select count(*) as c from ${table} where regexp_contains(event_message, '${pattern}')`;
+  const deadline = Date.now() + timeoutMs;
+  let lastError: string | undefined;
+
+  while (Date.now() < deadline) {
+    const start = new Date(Date.now() - 3_600_000).toISOString();
+    const end = new Date(Date.now() + 3_600_000).toISOString();
+    const resp = await request.get('/api/platform/projects/default/analytics/endpoints/logs.all', {
+      params: { iso_timestamp_start: start, iso_timestamp_end: end, sql },
+    });
+    if (resp.ok()) {
+      const data = await resp.json().catch(() => null);
+      const count = data?.result?.[0]?.c ?? 0;
+      if (count > 0) return;
+      lastError = undefined;
+    } else {
+      lastError = `${resp.status()} ${await resp.text().catch(() => '')}`;
+    }
+    await new Promise(r => setTimeout(r, intervalMs));
+  }
+  const suffix = lastError ? ` last response: ${lastError}` : '';
+  throw new Error(`waitForLogs(${table}, "${pattern}") timed out after ${timeoutMs}ms.${suffix}`);
+}
 
 let uniqueId = '';
 
@@ -136,7 +170,9 @@ test.afterEach(async ({ page }, testInfo) => {
 });
 
 test.beforeAll(async ({ request, browserName }) => {
-  test.setTimeout(60_000);
+  // The polling below can take up to ~90s in worst case; bump the hook timeout
+  // to give it headroom on top of the synchronous setup work above.
+  test.setTimeout(180_000);
 
   uniqueId = `${Date.now()}_${browserName}`;
 
@@ -171,7 +207,22 @@ test.beforeAll(async ({ request, browserName }) => {
     query: `select cron.schedule('test_cron_${uniqueId}', '5 seconds', $$SELECT auth.email()$$);`
   }});
 
-  await new Promise(r => setTimeout(r, 30_000))
+  // Wait for ingestion of THIS test's setup events, in parallel. These are
+  // uniqueId-tagged so we can confirm our specific data made it through the
+  // (Vector → Logflare → backend → searchable) path before the test bodies
+  // run their searches. Replaces a fixed 30s sleep that was racy on slow CI
+  // runners.
+  //
+  // The other pipelines (postgrest "Config reloaded", realtime "Billing
+  // metrics", edge functions /home/deno/functions/hello, cron job)
+  // emit periodically or on infra events independent of our setup; the test
+  // bodies' date-windowed searches reliably hit historical entries, so we
+  // don't need to poll those in beforeAll.
+  await Promise.all([
+    waitForLogs(request, 'edge_logs', `function_${uniqueId}`),
+    waitForLogs(request, 'auth_logs', `example_${uniqueId}`),
+    waitForLogs(request, 'storage_logs', `avatars_${uniqueId}`),
+  ]);
 });
 
 test.afterAll(async ({ request }) => {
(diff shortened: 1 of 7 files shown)


test/e2e/supabase/features/logs.spec.ts
```diff
@@ -1,21 +1,202 @@
...
+    const end = new Date(Date.now() + 3_600_000).toISOString();
+    const sampleSql = `select event_message from ${table} order by timestamp desc limit 5`;
+    const resp = await request.get('/api/platform/projects/default/analytics/endpoints/logs.all', {
+      params: { iso_timestamp_start: start, iso_timestamp_end: end, sql: sampleSql },
+    });
+    if (resp.ok()) {
+      const data = await resp.json().catch(() => null);
+      sample = (data?.result ?? []).map((r: { event_message?: string }) => (r.event_message ?? '').slice(0, 300));
+    }
+  } catch (_) { /* ignore — diagnostic only */ }
```
Ziinc: if the logs.all endpoint is timing out then this would also timeout, rendering it redundant.
better to query the database directly for debugging via a pg client.

test/e2e/supabase/setup-supabase-services.sh
```diff
@@ -82,5 +82,79 @@ if ! compose up -d --wait --wait-timeout 180; then
...
+fi
+
+SOURCES=(
+  "cloudflare.logs.prod"
+  "postgres.logs"
+  "deno-relay-logs"
+  "deno-subhosting-events"
+  "gotrue.logs.prod"
+  "realtime.logs.prod"
+  "storage.logs.prod.2"
```
Ziinc, line 121: i think storage has named their source on prod to v3, just a note.

Ziinc (approved):

test/e2e/supabase/features/logs.spec.ts
djwhitt: Good point. Will adjust this.

djwhitt: Worth noting it does provide some value as-is too since the thing it's waiting for is a specific message. This will collect other non-matching messages and make them available for debugging purposes.

PR #3446, opened 2026-05-08
perf: pop events on ingest

Baishan:
Keeping events in ets during the pipeline is wasteful. Current behavour does not seem to retry events marked as :ingested and they are cleaned up by the queue janitor anyway.  

- [ x ] Pop events during ingestion instead of marking them as ingested and waiting for the janitor
- [ x ] Added ability to requeue failed bigquery events
- [ x ] Only pop events when metrics for the source are above @max_avg_before_pop set to 100

closes O11Y-1777

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -162,6 +162,9 @@ config :opentelemetry,
 
 config :logflare, Logflare.Vault, json_library: Jason
 
+config :logflare, Logflare.FeatureFlags,
+  mark_ingested: false #marks events as ingested instead of popping when true
+
 config :broadway, config_storage: :ets
 
 config :mime, :types, %{
--- a/config/config.exs
+++ b/config/config.exs
@@ -27,6 +27,8 @@ config :logflare, :postgres_backend_adapter, pool_size: 3
 
 config :logflare, :bigquery_backend_adaptor, managed_service_account_pool_size: 0
 
+config :logflare, :bigquery_pipeline, max_retries: 0
+
 config :logflare, :clickhouse_backend_adaptor,
   engine: "MergeTree",
   pool_size: 3,
@@ -161,10 +163,6 @@ config :opentelemetry,
   traces_exporter: :none
 
 config :logflare, Logflare.Vault, json_library: Jason
-
-config :logflare, Logflare.FeatureFlags,
-  # marks events as ingested instead of popping when true
-  mark_ingested: false
 
 config :broadway, config_storage: :ets
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -163,7 +163,8 @@ config :opentelemetry,
 config :logflare, Logflare.Vault, json_library: Jason
 
 config :logflare, Logflare.FeatureFlags,
-  mark_ingested: false #marks events as ingested instead of popping when true
+  # marks events as ingested instead of popping when true
+  mark_ingested: false
 
 config :broadway, config_storage: :ets
 
--- a/lib/logflare/backends/buffer_producer.ex
+++ b/lib/logflare/backends/buffer_producer.ex
@@ -26,7 +26,7 @@ defmodule Logflare.Backends.BufferProducer do
           mark_ingested: boolean()
         }
 
-  @type table_key :: {pos_integer(), pos_integer() | nil, pid() | nil}
+  @type table_key :: {pos_integer() | atom(), pos_integer() | nil, pid() | nil}
 
   @default_interval 1_000
 
@@ -86,7 +86,8 @@ defmodule Logflare.Backends.BufferProducer do
       source_token: nil,
       backend_id: backend_id,
       last_discard_log_dt: nil,
-      interval: interval
+      interval: interval,
+      mark_ingested: false
     }
 
     table_key = {:consolidated, backend_id, self()}
--- a/lib/logflare/backends/buffer_producer.ex
+++ b/lib/logflare/backends/buffer_producer.ex
@@ -22,7 +22,8 @@ defmodule Logflare.Backends.BufferProducer do
           source_token: atom() | nil,
           backend_id: pos_integer(),
           last_discard_log_dt: DateTime.t() | nil,
-          interval: pos_integer()
+          interval: pos_integer(),
+          mark_ingested: boolean()
         }
 
   @type table_key :: {pos_integer(), pos_integer() | nil, pid() | nil}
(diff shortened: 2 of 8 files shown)


lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -446,6 +472,43 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
     end
   end
 
+  # Requeue failed events if the number of previous retries is less than @max_retries
+  defp maybe_requeue_failed(_, [], _), do: :ok
+  defp maybe_requeue_failed(_, _, %{max_retries: 0}), do: :ok
+
+  defp maybe_requeue_failed({_sid, _bid} = sid_bid, failed, %{max_retries: max_retries}) do
```
amokan, line 441: I think we could avoid some potential effort by adjusting `maybe_requeue_failed` a little bit in cases where `events` is an empty list.


```elixir
defp maybe_requeue_failed({_sid, _bid} = sid_bid, failed, %{max_retries: max_retries}) do
  events =
    Enum.filter(failed, fn %{data: %LE{retries: retries}} -> retries < max_retries end)
    |> Enum.map(fn %{data: %LE{} = event} ->
      %LE{event | retries: (event.retries || 0) + 1}
    end)

  requeue(sid_bid, events)
end

defp requeue(_sid_bid, []), do: :ok

defp requeue(sid_bid, events) do
  Logger.info("Requeuing #{length(events)} BigQuery events for retry")
  IngestEventQueue.delete_batch(sid_bid, events)
  IngestEventQueue.add_to_table(sid_bid, events)
end
```

test/logflare/bigquery/pipeline_test.exs
```diff
@@ -54,6 +54,37 @@ defmodule Logflare.BigQuery.PipelineTest do
...
+      assert IngestEventQueue.get_table_size(sid_bid_pid) == 0
+
+      mod.ack(ref, [], [message])
+      assert IngestEventQueue.get_table_size(sid_bid_pid) == 1
+
+      {:ok, [m]} = IngestEventQueue.pop_pending(sid_bid_pid, 1)
+
+      assert m.retries == 1
+    end
+
```
amokan, line 44: Prob worth adding a tests to cover the exhausted side of this.

Worth noting that the example below is manually mutating the factory-built LogEvent. If we suspect we'll have other tests that involve retries - may be worth updating the factory to handle them.

```suggestion
test "ack will not requeue failed events that have exhausted retries", %{source: source} do
  sid_bid_pid = {source.id, nil, self()}
  IngestEventQueue.upsert_tid(sid_bid_pid)
  le = build(:log_event) |> Map.put(:retries, 1)
  ref = {sid_bid_pid, source.token, %{mark_ingested: false, max_retries: 1}}
  message = Pipeline.transform(le, ref: ref)
  {mod, ref, _data} = message.acknowledger
  assert IngestEventQueue.get_table_size(sid_bid_pid) == 0

  mod.ack(ref, [], [message])

  assert IngestEventQueue.get_table_size(sid_bid_pid) == 0
end
```

amokan (commented):
Looks good @Baishan 

A couple minor notes. Feel free to push back if I'm missing something!

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: Added below

amokan (approved):
Looks good @Baishan ! Let's have @Ziinc put some eyes on this next week and go from there.

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -139,22 +154,33 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
+
+    case Sources.Cache.get_by_id(sid) do
+      nil ->
+        Logger.warning("Source not found for ack!", source_id: sid)
+
+      source ->
+        for %{data: le} <- successful do
+          # emit telemetry on event
+          emit_event_telemetry(queue, source, le, backend_metadata)
         end
```
Ziinc, line 146: a future perf refactor would be to sum the bytes of all events and emit the counts+total bytes in one telemetry call (vs many telemerty calls)

config/config.exs
```diff
@@ -162,6 +162,10 @@ config :opentelemetry,
 
 config :logflare, Logflare.Vault, json_library: Jason
 
+config :logflare, Logflare.FeatureFlags,
+  # marks events as ingested instead of popping when true
+  mark_ingested: false
+
```
Ziinc:
```suggestion
```
i think change is small enough to not require feature flagging

lib/logflare/backends/buffer_producer.ex
```diff
@@ -60,7 +64,8 @@ defmodule Logflare.Backends.BufferProducer do
       source_token: source.token,
       backend_id: opts[:backend_id],
       last_discard_log_dt: nil,
-      interval: interval
+      interval: interval,
+      mark_ingested: opts[:mark_ingested] || Keyword.get(feature_flags, :mark_ingested, false)
```
Ziinc: whether to pop or not should be determined based on source ingest rate. we can add a virtual field on the log event to mark it as popped or not.

lib/logflare/backends/buffer_producer.ex
```diff
@@ -229,27 +235,43 @@ defmodule Logflare.Backends.BufferProducer do
   end
 
   @spec do_fetch(state :: state(), count :: non_neg_integer()) :: [LogEvent.t()]
-  defp do_fetch(%{consolidated: true, backend_id: bid}, n) do
+  defp do_fetch(%{consolidated: true, backend_id: bid} = state, n) do
```
Ziinc: we should always pop for consolidated queues. these have no ui impact (yet) and we don't need to factor that in for now.

Ziinc (changes requested):

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: I can look at optimizing this in another pr.

config/config.exs
Baishan: Removed, I wanted it there because I wasn't sure the intention of :ingested but you explained its use for the recent event cache.

lib/logflare/backends/buffer_producer.ex
Baishan: Its determined in do_fetch now per metrics.

lib/logflare/backends/buffer_producer.ex
Baishan: These should always be popped, I've made this more explicit.

Ziinc (approved):

PR #3449, opened 2026-05-08
feat: add drop_fields ingest transform

djwhitt:
## Summary

- New ingest transform `drop_fields` that removes specified paths (top-level or nested via dot syntax, with `m.` shorthand for `metadata.`) from the event body. Mirrors the shape of the existing `copy_fields` and `kv_enrich` transforms, including pre-parsing config into a virtual `transform_drop_fields_parsed` field via `Sources.Cache.get_by_and_preload_rules/1` to avoid reparsing per event.
- Pipeline position: runs last, after `bigquery_spec → copy_fields → kv_enrich`, so a value can be copied or used for enrichment first and then dropped from its original location.
- Adds a nullable `text` column `transform_drop_fields` on `sources`, an edit-form textarea, and a docs section (with a ClickHouse caveat about the catch-all `*_attributes` columns).

## Performance changes on this branch

- **Pre-parse `drop_fields` config into a virtual `transform_drop_fields_parsed` field** (commit `8a004dd33`), populated by `Sources.Cache.get_by_and_preload_rules/1`. Parses the config string once at cache load instead of on every ingest event. Natural part of the new feature.
- **Pre-parse `copy_fields` config the same way** (commit `aa22037d9`). Before this branch, `copy_fields` re-parsed its config string on every event; this commit caches the parsed list on the source struct. Slightly out of scope for "add `drop_fields`" — see open question 4 below.
- **Replace `Kernel.put_in`/`Access.key` with a direct `put_at_path/3` recursion** (commit `3511d2122`) in the shared path-merge helper used by `copy_fields` and `kv_enrich`. Removes the `Access.key` indirection; the same recursion shape is reused by `drop_fields` as `drop_field_at/2`. Tightens per-event allocation cost in the path-rewrite stage.

The bench results below quantify the impact.

## Open questions

**1. Should `drop_fields` run last in the pipeline?**

- **Running last (current):** enables the "copy or enrich, then drop the original" pattern (covered by tests).
- **Running earlier:** more downstream savings. `flatten_body` (and the BigQuery key-sanitizer walk, if drop ran before that) scale with leaf count, so dropping fields earlier reduces work in every later stage. The bench on this branch shows a `drop_fields`-only event runs ~41% faster than the no-transform baseline (45.80K ips vs 32.58K ips) because the smaller body cuts `flatten_body` cost. Moving drop earlier would compound that effect further.

**2. User-supplied paths match post-`bigquery_spec` field names — keeping the existing pattern.** The BigQuery key sanitizer (dashes → underscores, leading-underscore prefix) runs before `copy_fields`/`drop_fields`/`kv_enrich`, so config like `food-key:dest` silently misses — the user has to write `_food_key:dest`. `drop_fields` follows the convention `copy_fields` and `kv_enrich` already use, so this PR doesn't change the rule; it just inherits it. The alternative — rewriting user-supplied paths through the same sanitizer at parse time so `food-key` and `_food_key` both work — would be a cross-transform behavior change worth doing in its own PR if we ever want it, not a `drop_fields`-specific decision. The chosen behavior is locked in by the `paths match post-bigquery_spec field names` test on this branch. Documenting the rule in the form helper and ingestion docs is also a follow-up.

**3. Should `id` / `timestamp` / `event_message` be reserved as protected paths across all three transforms?** None of the parsers validate destination/drop paths today, so `copy_fields: my_field:id`, `kv_enrich: some_key:timestamp`, or `drop_fields: id` will silently overwrite or remove fields that downstream code expects. The `LogEvent` struct's `:id` field is captured before transforms run so the struct is unaffected, but the body sent to backends would be corrupted. This is pre-existing behavior for `copy_fields`/`kv_enrich`; `drop_fields` adds a delete-shaped variant. Cleanest fix is at parse time in `Source.parse_*_config/1` — strip or reject paths whose head matches a reserved name.

**4. Should the `copy_fields` pre-parsing change land here or be split into its own PR?** Commit `aa22037d9` adds the parsed-virtual cache to `copy_fields`, which is technically out of scope for "add `drop_fields` transform." It's bundled here because the natural cache-load wiring point (`Sources.Cache.get_by_and_preload_rules/1`) is the same place `drop_fields_parsed` needed to be wired, and the `LogEvent` transform code reads more uniformly when both follow the same parsed-virtual + unparsed-fallback shape. The bench on this branch is the only place the `copy_fields` parsed-vs-unparsed gap is measured, so splitting the commit out wouldn't lose test coverage but would lose the benchmark context. Happy to split if reviewers prefer narrower diffs.

## Test plan

- [x] `mix test test/logflare/log_event_test.exs` (55 tests, 0 failures)
- [x] `mix test test/logflare/sources/sources_cache_test.exs`
- [x] `MIX_ENV=test mix run test/profiling/transform_bench.exs` (transform pipeline benchmark, 10s/scenario, results below)
- [x] Manual end-to-end smoke against a running dev server — see below
- [x] Verify migration applies cleanly on a populated DB

## Benchmark results

`MIX_ENV=test mix run test/profiling/transform_bench.exs` on Apple M5 / Erlang 27 / Elixir 1.19 (10 s/scenario, 2 s warmup, 1 s memory_time). All scenarios hit the real `Logflare.KeyValues.Cache` — no stubs.

```
Name                                      ips        average  deviation         median         99th %
drop_fields (parsed)                  46.67 K       21.43 μs    ±22.65%       20.71 μs       34.63 μs
drop_fields (unparsed fallback)       43.36 K       23.06 μs    ±89.35%       22.29 μs       31.88 μs
baseline (no transforms)              32.43 K       30.84 μs    ±16.91%       30.17 μs       46.88 μs
kv_enrich (parsed)                    27.23 K       36.73 μs    ±31.07%       35.38 μs       65.46 μs
kv_enrich (unparsed fallback)         25.46 K       39.28 μs   ±126.99%       37.42 μs       69.51 μs
copy + kv + drop (parsed)             23.34 K       42.85 μs    ±19.01%       40.50 μs       72.63 μs
copy_fields (parsed)                  20.73 K       48.24 μs    ±20.78%       45.08 μs       85.63 μs
copy_fields (unparsed fallback)       19.66 K       50.87 μs   ±100.87%       47.96 μs       92.08 μs
```

Memory per call:

```
Name                                    average
drop_fields (parsed)                   63.83 KB
drop_fields (unparsed fallback)        66.13 KB
baseline (no transforms)              119.07 KB
kv_enrich (parsed)                    131.98 KB
kv_enrich (unparsed fallback)         134.79 KB
copy + kv + drop (parsed)             159.85 KB
copy_fields (parsed)                  190.87 KB
copy_fields (unparsed fallback)       195.23 KB
```

Highlights:

- `drop_fields` runs **faster than the no-transform baseline** (~44% on ips, ~46% on memory) — dropping fields before `flatten_body` more than pays for the drop work itself.
- `kv_enrich` (~37 μs) sits between baseline and `copy_fields` (~48 μs), not co-equal with `copy_fields` as an earlier Mimic-stubbed version of this bench was suggesting. The kv destination paths are top-level, so the post-lookup write is shallower and cheaper than `copy_fields`' nested writes.
- The parsed-virtual cache wins by a real but modest margin per event (drop 1.08×, copy 1.13×, kv 1.07×). Most ingest cost is in the per-event body walks (`mapper`, `bigquery_spec`, `flatten_body`), not the config parsing — which is also why the parsed-virtual is amortizing parse work across an event batch within a single HTTP request rather than across requests (`Sources.Cache` only stores raw `%Source{}`; the `parse_*_config/1` calls run in the `FetchResource` plug per request).
- The `unparsed fallback` scenarios have higher deviation (±90–127%) than the parsed scenarios (±16–31%) because they re-run the parser allocation on every event, exposing more GC tail noise. Read the medians (close to the parsed medians) for the per-event delta.

## Realistic-payload comparison (`log_event_make_bench.exs`)

The first commit on this branch (`bench: cover copy + kv configurations in log_event_make_bench`) is positioned so it runs cleanly at every point in branch history, including before any feature or perf commits. Sources are set up via `Sources.Cache.get_by_and_preload_rules/1` — the same path the `FetchResource` plug uses on the ingest hot path — so the bench exercises whichever parsed-virtual fields exist at the current commit (kv only pre-PR, all three post-PR), matching production fidelity at each end of the comparison.

Reduction counts (the deterministic signal — same number every run unless code paths actually change):

| Scenario | Before perf (`28d3910ac`) | After perf (`1b33758f0`) | Δ |
|---|---|---|---|
| otel trace | 31.16 K | 31.16 K | 0.00% |
| otel trace + copy | 35.02 K | 34.87 K | −0.43% |
| otel trace + kv | 31.91 K | 31.90 K | −0.03% |
| otel trace + copy + kv | 35.71 K | 35.44 K | −0.76% |
| edge log | 137.49 K | 137.49 K | 0.00% |
| edge log + copy | 142.79 K | 142.28 K | −0.36% |
| edge log + kv | 137.87 K | 137.88 K | +0.01% |
| edge log + copy + kv | 143.19 K | 142.84 K | −0.24% |

All 8 comparable scenarios within ±1%. The PR's perf wins are real but small in absolute terms at the realistic-payload level — the body walks (`mapper`, `bigquery_spec`, `flatten_body`, `MetadataCleaner`) dominate per-event cost, so swapping `copy_fields` from inline-parse + `Kernel.put_in/Access.key` to parsed-virtual + `put_at_path` shaves microseconds that get lost in the hundreds of microseconds spent walking the body. `transform_bench.exs` above measures the same swap on a synthetic 10-leaf payload where transform cost is a larger fraction of total, and shows ~1.10× / 1.13× / 1.07× wins — the realistic bench confirms there's no regression at the realistic level and quantifies how much the body walks dilute the per-transform optimization.

`drop_fields` scenarios are after-only since the feature is new. Numbers from the after run alone:

| Scenario | Reductions | Memory | vs baseline |
|---|---|---|---|
| edge log | 137.49 K | 1321.55 KB | — |
| edge log + drop | 100.76 K | 975.10 KB | **−26.7% / −26%** |
| edge log + all (copy + kv + drop) | 105.91 K | 1029.52 KB | −22.9% / −22% |

Edge log + drop reproduces the headline `drop_fields` perf claim on realistic, deeply-nested Supabase-shaped payloads, not just the synthetic 10-leaf body in `transform_bench.exs`. The full pipeline is still lighter than baseline because dropping four nested paths shrinks the body before `flatten_body` and saves more downstream work than `copy_fields` + `kv_enrich` add — and unlike copy/kv, this saving compounds with body size, so it's the only optimization in the PR with a sizable absolute impact at the realistic-payload level.

## Manual smoke test

Ran end-to-end against a local `make start` dev server. A source was created in the UI with the following transforms configured:

```
# transform_drop_fields
service
namespace
m.routing.region
```

Sent an event via the ingest API:

```bash
curl -X POST "http://localhost:4000/api/logs?source=$SOURCE_TOKEN" \
  -H "Content-Type: application/json" \
  -H "X-API-KEY: $INGEST_KEY" \
  -d '{
    "event_message": "drop fields e2e",
    "service": "router",
    "namespace": "default",
    "keep": "yes",
    "metadata": {"routing": {"region": "us-east", "zone": "a"}, "user": {"id": 42}}
  }'
# => 200 {"message":"Logged!"}
```

Fetched the post-transform body from the recent-events buffer:

```bash
curl "http://localhost:4000/api/sources/$SOURCE_TOKEN/recent" \
  -H "Authorization: Bearer $ACCESS_TOKEN"
```

Persisted body (after dash-to-underscore, drop_fields, and flatten):

```json
{
  "event_message": "drop fields e2e",
  "id": "...",
  "timestamp": ...,
  "keep": "yes",
  "metadata": {
    "routing": {"zone": "a"},
    "user": {"id": 42}
  }
}
```

Verifications:

- `service`, `namespace`, `metadata.routing.region` — dropped
- `keep`, `metadata.user.id`, `metadata.routing.zone`, `event_message` — retained
- `metadata.user` and `metadata.routing` parents preserved with their remaining children
- Empty `transform_drop_fields` (cleared textarea) reverts to a no-op — all fields retained


lib/logflare/sources/source.ex
```diff
@@ -411,6 +417,57 @@ defmodule Logflare.Sources.Source do
...
+  def parse_drop_fields_config(%__MODULE__{transform_drop_fields: blank} = source)
+      when blank in [nil, ""] do
+    %{source | transform_drop_fields_parsed: nil}
+  end
+
+  def parse_drop_fields_config(%__MODULE__{transform_drop_fields: config} = source) do
+    parsed =
+      config
+      |> String.split("\n", trim: true)
+      |> Enum.flat_map(fn raw ->
```
amokan: I do think we should probably guard against the `id`, `event_message`, and `timestamp` fields in here to avoid confusion. Tracing the upstream logic, it looks like the S3 and PG adaptors both read `body["id"]`. Event msg and timestamp for sure are being accessed through body in various places - so keeping them from being dropped is smart.

Bigger picture - would probably be nice to do the same with the copy fields and KV enrichment paths as well, but that can be another PR.

A few ways to handle that check - here is one thought.

```elixir
|> Enum.flat_map(fn raw ->
  case String.trim(raw) do
    "" ->
      []

    path ->
      path = String.replace_prefix(path, "m.", "metadata.")

      case String.split(path, ".") do
        [head | _] when head in ~w(id event_message timestamp) -> []
        segments -> [segments]
      end
  end
end)
```

lib/logflare/sources/source.ex
```diff
@@ -411,6 +417,57 @@ defmodule Logflare.Sources.Source do
...
+    %{source | transform_copy_fields_parsed: parsed}
+  end
+
+  @spec parse_drop_fields_config(%__MODULE__{}) :: %__MODULE__{}
+  def parse_drop_fields_config(%__MODULE__{transform_drop_fields: blank} = source)
+      when blank in [nil, ""] do
+    %{source | transform_drop_fields_parsed: nil}
+  end
+
+  def parse_drop_fields_config(%__MODULE__{transform_drop_fields: config} = source) do
```
amokan: Nit.

You could add `import Logflare.Utils.Guards, only: [is_non_empty_binary: 1]` to the top and leverage the `is_non_empty_binary` guard to invert this logic.

```elixir
def parse_drop_fields_config(%__MODULE__{transform_drop_fields: config} = source)
  when is_non_empty_binary(config) do
```

And then you could have the non-match clause below...

```elixir
def parse_drop_fields_config(%__MODULE__{} = source) do
  %{source | transform_drop_fields_parsed: nil}
end
```

_The same could be said for `parse_copy_fields_config/1` and `parse_key_values_config/1` - but prob could handle those as a cleanup chore outside of this PR_

amokan (changes requested):
Looks great @djwhitt 

I think I'd feel more comfortable with some guard rails around the reserved path situation (`id`, `event_message`, and `timestamp`).

The other comment is not at all a requirement though - just want to be sure you're aware of some of the custom guards in the project. I totally get that the other logic related to copy and enrichment do pretty much the same check currently.

lib/logflare/sources/source.ex
djwhitt: Done in 8b87e483:

- Added a `validate_drop_fields/1` step to `default_validations/2` that rejects `id`, `event_message`, and `timestamp` at the top level with `"cannot drop reserved fields: <names>"` — so users see the error in the UI/API immediately instead of saving a no-op rule.
- Also kept a head-segment filter inside `parse_drop_fields_config/1` so any source path that bypasses the changeset (cached/seeded structs, etc.) can't break ingestion either.

Nested paths like `metadata.id` / `m.timestamp` are still droppable. Tests cover both layers (`test/logflare/sources/source_test.exs` for the changeset, `test/logflare/log_event_test.exs` for parse-time).

lib/logflare/sources/source.ex
djwhitt: Applied to `parse_drop_fields_config/1` and `parse_copy_fields_config/1` in 8b87e483, using the targeted `import Logflare.Utils.Guards, only: [is_non_empty_binary: 1]` form.

Held off on `parse_key_values_config/1` for now — `kv_enrich/2`'s first clause matches `transform_key_values: nil` specifically, so flipping the parser to emit `nil` for `""` would skip all three head clauses and infinite-recurse via the fallback at `log_event.ex:240`. Doable, but needs a companion change to widen that match.

amokan (approved):

docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
```diff
@@ -221,6 +223,47 @@ Key-value pairs can be managed via the [Management API](https://logflare.app/swa
 Key-value enrichment runs after field copying, so copied fields can be used as lookup sources.
 :::
 
+### Drop Fields
+
+A source can be configured to drop specific fields from the event body before it is persisted. This is useful for keys that are needed upstream — for example, by a log shipper for routing — but should not be stored.
+
+One field path per line. Dot syntax addresses nested paths, and `m.` is shorthand for `metadata.`. Missing paths and non-map intermediates are silently ignored.
+
+For example, given a Fluent Bit pipeline that uses `service` and `namespace` for routing but does not need them stored:
```
Ziinc, line 232:
```suggestion
For example, given a FluentBit pipeline that uses `service` and `namespace` for routing but does not need them stored:
```

docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
```diff
@@ -221,6 +223,47 @@ Key-value pairs can be managed via the [Management API](https://logflare.app/swa
...
+{ "service": "router", "namespace": "default", "event_message": "hello" }
+```
+
+becomes:
+
+```json
+{ "event_message": "hello" }
+```
+
+Drop fields runs **after** field copying and key-value enrichment, so a value can be copied or used for enrichment first and then dropped from its original location:
```
Ziinc:
```suggestion
Drop fields runs at the end of the transformation pipeline, so a value can be copied or used for enrichment first and then dropped from its original location:
```

docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
```diff
@@ -221,6 +223,47 @@ Key-value pairs can be managed via the [Management API](https://logflare.app/swa
...
+service:m.routing.service
+
+# transform_drop_fields
+service
+```
+
+The original `service` key is removed from the top level, but `metadata.routing.service` retains the value.
+
+:::note ClickHouse backends
+The ClickHouse backend persists events into fixed OTEL tables via a built-in mapping that already discards unmapped top-level keys. Drop fields is still useful for ClickHouse to keep specific keys out of the catch-all `*_attributes` map columns. Be careful not to drop fields that the mapping reads (e.g. `project`, `trace_id`, `timestamp`) — those will fall back to defaults.
```
Ziinc, line 264: i think our clickhouse backend docs are lacking a little information on what the schema tables are for logs/metrics/traces

test/logflare/sources/source_test.exs
```diff
@@ -23,6 +23,57 @@ defmodule Logflare.Sources.SourceTest do
...
+
+        changeset =
+          Source.update_by_user_changeset(source, %{
+            "transform_drop_fields" => unquote(reserved)
+          })
+
+        refute changeset.valid?
+
+        assert {"cannot drop reserved fields: " <> _, _} =
+                 changeset.errors[:transform_drop_fields]
```
Ziinc, line 40: nit: can be condensed into a single match

test/logflare/sources/source_test.exs
```diff
@@ -23,6 +23,57 @@ defmodule Logflare.Sources.SourceTest do
...
+      end
+    end
+
+    test "rejects when reserved field is mixed with allowed entries" do
```
Ziinc: can be combined with the above test as part of the for comprehension

Ziinc (approved):

docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
djwhitt: Fixed in 3f5867714

test/logflare/sources/source_test.exs
djwhitt: Fixed in 98c018495

test/logflare/sources/source_test.exs
djwhitt: Fixed in 98c018495

docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
djwhitt: Held off on this one — the suggestion drops the space (`Fluent Bit` → `FluentBit`), but the canonical name is two words per fluentbit.io and the CNCF project listing. Also aligned two stale `FluentBit` references elsewhere in the codebase (marketing index alt text + pricing copy) in 0cd63356e.

docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
djwhitt: Confirmed — `backends/clickhouse.mdx` still describes a single ingest table and pre-dates the consolidated OTEL pipeline. The `otel_logs_*` / `otel_metrics_*` / `otel_traces_*` tables, `event_type` partitioning, and catch-all `*_attributes` columns aren't documented there. Out of scope for this PR; will pick up as a follow-up.

PR #3454, opened 2026-05-12
feat: AuthLive hook ensures correct team id param

msmithstubbs:
Adds a LiveView hook the ensure the `t=` param is included in the require URI and matches the resource being requested:

* if the route has a resource id, such as `http://localhost:4000/sources/1/search`, the team for that source will be set, overriding any current team param
  * `http://localhost:4000/sources/1/search` &rarr;  `http://localhost:4000/sources/1/search?t=1`
  * `http://localhost:4000/sources/1/search?t=999` &rarr;  `http://localhost:4000/sources/1/search?t=1`
* if there is no resource id the most recently selected team (`session["last_switched_team_id"]`) will be used
  * `http://localhost:4000/dashboard` &rarr; `http://localhost:4000/dashboard?t=1`
 * Adds `live_with_redirect/3` macro to ConnCase. This is a wrapper around live and avoids the need to include `t=` in the test. 

This PR only affects the dashboard and search live views. Subsequent PRs move other routes into this live session to include the default team param. This PR should be merged first.

* #3466 
* #3468 
* #3469 
* #3470
* #3471 

Part of O11Y-1754

## Demo
Video demonstrates: 
1. visiting /dashboard without a team param: `t` is added
2. Switching to another team, then visiting source search: `t` param is added for the team the source  belongs to.
3. Visiting dashboard again sets the `t` param to last selected team.

https://github.com/user-attachments/assets/f37b45d5-49de-4e66-b80d-c6e10ad483f8


lib/logflare_web/live/auth_live.ex
```diff
@@ -72,4 +79,24 @@ defmodule LogflareWeb.AuthLive do
...
+
+  defp ensure_team_param(_params, uri, %{assigns: %{team: team}} = socket)
+       when is_struct(team, Team) do
+    uri = uri |> local_path_with_query() |> Utils.with_team_param(team)
+    {:halt, push_patch(socket, to: uri, replace: true)}
+  end
+
+  defp ensure_team_param(_params, _uri, %{assigns: %{team: nil}} = socket), do: {:cont, socket}
+
+  defp local_path_with_query(uri) do
```
msmithstubbs: push_patch needs relative path.

Ziinc:
@msmithstubbs this is quite a large far reaching PR, can we split it up into sections and focus on the higer traffic resources first?

lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -31,12 +32,23 @@ defmodule LogflareWeb.Source.SearchLV do
...
+  require Ecto.Query
   require Logger
 
   @tail_search_interval 1000
   @user_idle_interval :timer.minutes(2)
 
   on_mount LogflareWeb.AuthLive
+  on_mount {LogflareWeb.AuthLive, :ensure_team_param}
+
+  def resource_team_id_query(%{"source_id" => source_id}, _uri, user) do
```
msmithstubbs: Might be cleaner to have these query builders in Teams context and match on view. The hook can call directly:

```elixir
Logflare.Teams.resource_team_id_query(LogsSearchLV, source_id, user)
```

Less DB queries in the liveview. Thoughts?

lib/logflare_web/live/auth_live.ex
```diff
@@ -71,5 +92,47 @@ defmodule LogflareWeb.AuthLive do
...
+  defp team_param_matches?(%{"t" => team_id}, %Team{id: id}), do: team_id == to_string(id)
+  defp team_param_matches?(_params, _team), do: false
+
+  defp maybe_assign_team_from_resource(params, uri, %{view: view} = socket) do
+    if function_exported?(view, :resource_team_id_query, 3) do
+      effective_user = socket.assigns[:team_user] || socket.assigns.user
+
+      case view.resource_team_id_query(params, uri, effective_user) do
+        nil -> socket
+        query -> assign_context_by_team_id(socket, Repo.one(query), effective_user.email)
```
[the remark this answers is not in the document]
msmithstubbs: Added raise.

lib/logflare_web/live/search_live/logs_search_lv.ex
Ziinc: yes should not be having these in the liveview modules

lib/logflare_web/live/auth_live.ex
```diff
@@ -56,20 +70,74 @@ defmodule LogflareWeb.AuthLive do
...
+    socket = maybe_assign_team_from_resource(params, uri, socket)
+    team = socket.assigns[:team]
+
+    cond do
+      team_param_matches?(params, team) ->
+        {:cont, socket}
+
+      is_struct(team, Team) ->
+        next_uri = Utils.with_team_param(current_uri, team)
+        {:halt, push_patch(socket, to: next_uri, replace: true)}
```
Ziinc: i think a push_navigate to reset LV assigns would be a cleaner way and would mean we don't need relative path from `local_path_with_query/1`

Ziinc (changes requested):

Ziinc:
@msmithstubbs mind adding a demo so that other team members can understand what the bugfix is about? thanks

lib/logflare_web/live/auth_live.ex
msmithstubbs: Agree push_navigate is a better choice here. Unfortunately it still need s local path, but I did tidy up the local path helper.

lib/logflare_web/live/search_live/logs_search_lv.ex
msmithstubbs: Moved it to TeamContext. Once this PR is approved I'll update the others, too.

msmithstubbs:
Added demo video. 
Switching to `push_navigate` caused some tests to fail since the view process is destroyed and new one invoked. Fixed by ensuring we append t param to link uri where necessary.

Ziinc (approved):

msmithstubbs:
I've updated the follow on branches so they can be merged when ready.
Once they are all merged I'll have a follow up PR to reduce repetition in `TeamContext.resource_team_id_query`

PR #3456, opened 2026-05-12
perf(cleaner): flatten metadata without per-level intermediate lists

djwhitt:
## Summary
- Refactor `MetadataCleaner.flatten/1` to accumulate flattened `{key, value}` pairs in a single list (prepend + `:lists.reverse` + `:maps.from_list`) instead of building and concatenating a fresh list at every recursion level via `Enum.flat_map`.
- Replace `Enum.with_index` on list elements with an inline index counter threaded through the recursion.
- Add a single-element fast path to `build_key/1` so root-level keys skip the `Enum.reverse |> Enum.join(".")` round-trip.
- Add tests covering empty-container-at-leaf passthrough, nested integer-index list paths, mixed-type list elements, and literal-dot-key vs nested-derived-key collision resolution.
- Add a Benchee profiling script (`test/profiling/metadata_cleaner_flatten_bench.exs`) with realistic OTEL trace, edge log, and list-heavy payload shapes.

## Behavior
Equivalent to the previous implementation across all observable cases (DFS source order preserved; duplicate-key resolution still "last wins"; empty containers at leaves still pass through as values; list indices still 0-based stringified).

## Benchmarks
All runs on Apple M5, Elixir 1.19.5 / Erlang 27.3.4.6. Lower is better in every column.

### `flatten/1` in isolation (`test/profiling/metadata_cleaner_flatten_bench.exs`)

| Payload      | Metric     | main      | this branch | Δ       |
| ------------ | ---------- | --------- | ----------- | ------- |
| otel trace   | time       | 1.39 μs   | 1.11 μs     | −20.1%  |
| otel trace   | memory     | 8.77 KB   | 6.86 KB     | −21.8%  |
| otel trace   | reductions | 800       | 440         | −45.0%  |
| edge log     | time       | 8.24 μs   | 7.33 μs     | −11.0%  |
| edge log     | memory     | 38.41 KB  | 31.32 KB    | −18.5%  |
| edge log     | reductions | 2,390     | 1,570       | −34.3%  |
| list heavy   | time       | 50.05 μs  | 44.56 μs    | −11.0%  |
| list heavy   | memory     | 202.59 KB | 156.49 KB   | −22.8%  |
| list heavy   | reductions | 13,930    | 9,430       | −32.3%  |

### End-to-end `LogEvent.make/2` (`test/profiling/log_event_make_bench.exs`)

Shows how the `flatten/1` win translates through the full event-construction path (parse, classify, clean, flatten, build struct). `flatten/1` is a small slice of this work, so the deltas are correspondingly smaller — but they are still in the right direction across all three metrics.

| Payload      | Metric     | main       | this branch | Δ      |
| ------------ | ---------- | ---------- | ----------- | ------ |
| otel trace   | time       | 85.05 μs   | 81.61 μs    | −4.0%  |
| otel trace   | memory     | 0.29 MB    | 0.29 MB     | ~flat  |
| otel trace   | reductions | 31,250     | 30,910      | −1.1%  |
| edge log     | time       | 318.01 μs  | 312.73 μs   | −1.7%  |
| edge log     | memory     | 1.29 MB    | 1.28 MB     | −0.8%  |
| edge log     | reductions | 137,660    | 136,600     | −0.8%  |

## Test plan
- [x] `mix test test/logflare/logs/ingest/metadata_cleaner_test.exs` — 13 tests, 0 failures
- [x] `mix run test/profiling/metadata_cleaner_flatten_bench.exs` against `main` and this branch
- [x] `mix run test/profiling/log_event_make_bench.exs` against `main` and this branch

### DIFF
--- a/lib/logflare/logs/cleaner.ex
+++ b/lib/logflare/logs/cleaner.ex
@@ -4,6 +4,8 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
   Also provides flattening of nested maps into dot-delimited key paths.
   """
 
+  import Logflare.Utils.Guards, only: [is_non_empty_map: 1]
+
   @type flat_key :: String.t()
   @typep pair_acc :: [{flat_key(), term()}]
 
@@ -65,7 +67,7 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
   end
 
   @spec do_flatten_pairs([{term(), term()}], [String.t()], pair_acc()) :: pair_acc()
-  defp do_flatten_pairs([{k, v} | rest], prefix, acc) when is_map(v) and v != %{} do
+  defp do_flatten_pairs([{k, v} | rest], prefix, acc) when is_non_empty_map(v) do
     acc = do_flatten_map(v, [k | prefix], acc)
     do_flatten_pairs(rest, prefix, acc)
   end
@@ -82,7 +84,7 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
   defp do_flatten_pairs([], _prefix, acc), do: acc
 
   @spec do_flatten_list(list(), non_neg_integer(), [String.t()], pair_acc()) :: pair_acc()
-  defp do_flatten_list([v | rest], idx, prefix, acc) when is_map(v) and v != %{} do
+  defp do_flatten_list([v | rest], idx, prefix, acc) when is_non_empty_map(v) do
     acc = do_flatten_map(v, [Integer.to_string(idx) | prefix], acc)
     do_flatten_list(rest, idx + 1, prefix, acc)
   end
--- a/lib/logflare/logs/cleaner.ex
+++ b/lib/logflare/logs/cleaner.ex
@@ -49,43 +49,62 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
   """
   @spec flatten(map()) :: %{flat_key() => term()}
   def flatten(map) when is_map(map) do
-    Map.new(do_flatten_map(map, []))
+    map
+    |> do_flatten_map([], [])
+    |> :lists.reverse()
+    |> :maps.from_list()
   end
 
   def nil_or_empty?(x) when nil_or_empty(x), do: true
   def nil_or_empty?(_), do: false
 
-  @spec do_flatten_map(map(), [String.t()]) :: [{flat_key(), term()}]
-  defp do_flatten_map(map, prefix) do
-    Enum.flat_map(map, fn
-      {k, v} when is_map(v) and v != %{} ->
-        do_flatten_map(v, [k | prefix])
-
-      {k, v} when is_list(v) and v != [] ->
-        do_flatten_list(v, [k | prefix])
-
-      {k, v} ->
-        [{build_key([k | prefix]), v}]
-    end)
+  @spec do_flatten_map(map(), [String.t()], [{flat_key(), term()}]) :: [{flat_key(), term()}]
+  defp do_flatten_map(map, prefix, acc) do
+    do_flatten_pairs(:maps.to_list(map), prefix, acc)
   end
 
-  @spec do_flatten_list(list(), [String.t()]) :: [{flat_key(), term()}]
-  defp do_flatten_list(list, prefix) do
-    list
-    |> Enum.with_index()
-    |> Enum.flat_map(fn
-      {v, idx} when is_map(v) and v != %{} ->
-        do_flatten_map(v, [Integer.to_string(idx) | prefix])
-
-      {v, idx} when is_list(v) and v != [] ->
-        do_flatten_list(v, [Integer.to_string(idx) | prefix])
-
-      {v, idx} ->
-        [{build_key([Integer.to_string(idx) | prefix]), v}]
-    end)
+  @spec do_flatten_pairs([{term(), term()}], [String.t()], [{flat_key(), term()}]) ::
+          [{flat_key(), term()}]
+  defp do_flatten_pairs([{k, v} | rest], prefix, acc) when is_map(v) and v != %{} do
+    acc = do_flatten_map(v, [k | prefix], acc)
+    do_flatten_pairs(rest, prefix, acc)
   end
 
-  @spec build_key([String.t()]) :: flat_key()
+  defp do_flatten_pairs([{k, v} | rest], prefix, acc) when is_list(v) and v != [] do
+    acc = do_flatten_list(v, 0, [k | prefix], acc)
+    do_flatten_pairs(rest, prefix, acc)
+  end
+
+  defp do_flatten_pairs([{k, v} | rest], prefix, acc) do
+    do_flatten_pairs(rest, prefix, [{build_key([k | prefix]), v} | acc])
+  end
+
+  defp do_flatten_pairs([], _prefix, acc), do: acc
+
+  @spec do_flatten_list(list(), non_neg_integer(), [String.t()], [{flat_key(), term()}]) ::
+          [{flat_key(), term()}]
+  defp do_flatten_list([v | rest], idx, prefix, acc) when is_map(v) and v != %{} do
+    acc = do_flatten_map(v, [Integer.to_string(idx) | prefix], acc)
+    do_flatten_list(rest, idx + 1, prefix, acc)
+  end
+
+  defp do_flatten_list([v | rest], idx, prefix, acc) when is_list(v) and v != [] do
+    acc = do_flatten_list(v, 0, [Integer.to_string(idx) | prefix], acc)
+    do_flatten_list(rest, idx + 1, prefix, acc)
+  end
+
+  defp do_flatten_list([v | rest], idx, prefix, acc) do
+    do_flatten_list(rest, idx + 1, prefix, [
+      {build_key([Integer.to_string(idx) | prefix]), v} | acc
+    ])
+  end
+
+  defp do_flatten_list([], _idx, _prefix, acc), do: acc
+
+  @spec build_key([term()]) :: flat_key()
+  defp build_key([single]) when is_binary(single), do: single
+  defp build_key([single]), do: to_string(single)
+
   defp build_key(reversed_parts) do
     reversed_parts
     |> Enum.reverse()
--- a/lib/logflare/logs/cleaner.ex
+++ b/lib/logflare/logs/cleaner.ex
@@ -5,6 +5,7 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
   """
 
   @type flat_key :: String.t()
+  @typep pair_acc :: [{flat_key(), term()}]
 
   defguardp nil_or_empty(x) when x in [%{}, [], "", {}, nil]
 
@@ -50,7 +51,7 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
   @spec flatten(map()) :: %{flat_key() => term()}
   def flatten(map) when is_map(map) do
     map
-    |> do_flatten_map([], [])
+    |> do_flatten_map()
     |> :lists.reverse()
     |> :maps.from_list()
   end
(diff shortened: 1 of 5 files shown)


lib/logflare/logs/cleaner.ex
```diff
@@ -49,43 +49,62 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
...
+          [{flat_key(), term()}]
+  defp do_flatten_pairs([{k, v} | rest], prefix, acc) when is_map(v) and v != %{} do
+    acc = do_flatten_map(v, [k | prefix], acc)
+    do_flatten_pairs(rest, prefix, acc)
+  end
 
-      {k, v} ->
-        [{build_key([k | prefix]), v}]
-    end)
+  defp do_flatten_pairs([{k, v} | rest], prefix, acc) when is_list(v) and v != [] do
```
amokan:
```suggestion
  defp do_flatten_pairs([{k, [_ | _] = v} | rest], prefix, acc) do
```

This could clean up that function head a bit by patten matching for a list containing values.

lib/logflare/logs/cleaner.ex
```diff
@@ -49,43 +49,62 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
...
-      {v, idx} when is_list(v) and v != [] ->
-        do_flatten_list(v, [Integer.to_string(idx) | prefix])
+  defp do_flatten_pairs([], _prefix, acc), do: acc
 
-      {v, idx} ->
-        [{build_key([Integer.to_string(idx) | prefix]), v}]
-    end)
+  @spec do_flatten_list(list(), non_neg_integer(), [String.t()], [{flat_key(), term()}]) ::
+          [{flat_key(), term()}]
+  defp do_flatten_list([v | rest], idx, prefix, acc) when is_map(v) and v != %{} do
```
amokan: Not much of a gain on this one given the nature of maps - but just showing an alternative approach 🤷‍♂️ 

```suggestion
  defp do_flatten_list([v | rest], idx, prefix, acc) when is_map(v) and map_size(v) > 0 do
```

lib/logflare/logs/cleaner.ex
```diff
@@ -49,43 +49,62 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
...
+    acc = do_flatten_map(v, [Integer.to_string(idx) | prefix], acc)
+    do_flatten_list(rest, idx + 1, prefix, acc)
   end
 
-  @spec build_key([String.t()]) :: flat_key()
+  defp do_flatten_list([v | rest], idx, prefix, acc) when is_list(v) and v != [] do
```
amokan:
```suggestion
  defp do_flatten_list([[_ | _] = v | rest], idx, prefix, acc) do
```

lib/logflare/logs/cleaner.ex
```diff
@@ -49,43 +49,62 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
...
+    do_flatten_list(rest, idx + 1, prefix, acc)
+  end
+
+  defp do_flatten_list([v | rest], idx, prefix, acc) do
+    do_flatten_list(rest, idx + 1, prefix, [
+      {build_key([Integer.to_string(idx) | prefix]), v} | acc
+    ])
+  end
+
+  defp do_flatten_list([], _idx, _prefix, acc), do: acc
```
amokan: Nit. Termination conditions will often be placed at the top of the clauses. Not at all saying you need to - but food for thought.

amokan (approved):
A couple opportunities to tighten up some function heads with pattern matching.

Other than that it LGTM 🥳

lib/logflare/logs/cleaner.ex
amokan: Another thought is that this pattern is probably common enough (_check if the arg is a map and non-empty_) that we could probably add a new guard to `Logflare.Utils.Guards`.

Something like `defguard is_non_empty_map(value) when is_map(value) and map_size(value) > 0` 🤷‍♂️

lib/logflare/logs/cleaner.ex
```diff
@@ -49,43 +49,62 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
...
 
   def nil_or_empty?(x) when nil_or_empty(x), do: true
   def nil_or_empty?(_), do: false
 
-  @spec do_flatten_map(map(), [String.t()]) :: [{flat_key(), term()}]
-  defp do_flatten_map(map, prefix) do
-    Enum.flat_map(map, fn
-      {k, v} when is_map(v) and v != %{} ->
-        do_flatten_map(v, [k | prefix])
+  @spec do_flatten_map(map(), [String.t()], [{flat_key(), term()}]) :: [{flat_key(), term()}]
```
Ziinc: can do a `@typep` for the kv pair so that its less repetitive

test/logflare/logs/ingest/metadata_cleaner_test.exs
```diff
@@ -134,5 +134,28 @@ defmodule Logflare.Logs.Ingest.MetadataCleanerTest do
...
+               %{"a.b" => %{}, "a.c" => []}
+    end
+
+    test "nested lists use compound integer index paths" do
+      assert Cleaner.flatten(%{"a" => [[1, 2], [3, 4]]}) ==
+               %{"a.0.0" => 1, "a.0.1" => 2, "a.1.0" => 3, "a.1.1" => 4}
+    end
+
+    test "lists with mixed element types" do
+      assert Cleaner.flatten(%{"xs" => [1, %{"k" => "v"}, [2, 3]]}) ==
```
Ziinc, line 154: note that this map object is considered illegal for logflare ingestion and would be rejected in the pipeline.

should consider adding in a property test to test that the output of `Cleaner.flatten/1` is identical to Iteraptor

lib/logflare/logs/cleaner.ex
```diff
@@ -49,43 +49,62 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
   """
   @spec flatten(map()) :: %{flat_key() => term()}
   def flatten(map) when is_map(map) do
-    Map.new(do_flatten_map(map, []))
+    map
+    |> do_flatten_map([], [])
+    |> :lists.reverse()
```
Ziinc, line 57: if we're creating a map subsequently then we don't need to reverse it
```suggestion
```

lib/logflare/logs/cleaner.ex
```diff
@@ -49,43 +49,62 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
...
   def nil_or_empty?(x) when nil_or_empty(x), do: true
   def nil_or_empty?(_), do: false
 
-  @spec do_flatten_map(map(), [String.t()]) :: [{flat_key(), term()}]
-  defp do_flatten_map(map, prefix) do
-    Enum.flat_map(map, fn
-      {k, v} when is_map(v) and v != %{} ->
-        do_flatten_map(v, [k | prefix])
+  @spec do_flatten_map(map(), [String.t()], [{flat_key(), term()}]) :: [{flat_key(), term()}]
+  defp do_flatten_map(map, prefix, acc) do
```
Ziinc:
```suggestion
  defp do_flatten_map(map, prefix \\ [], acc \\ []) do
```
can turn this arity 1 so that caller at :53 doesn't need to pass in the empty lists.

lib/logflare/logs/cleaner.ex
```diff
@@ -49,43 +49,62 @@ defmodule Logflare.Logs.Ingest.MetadataCleaner do
...
-        do_flatten_list(v, [Integer.to_string(idx) | prefix])
+  defp do_flatten_pairs([], _prefix, acc), do: acc
 
-      {v, idx} ->
-        [{build_key([Integer.to_string(idx) | prefix]), v}]
-    end)
+  @spec do_flatten_list(list(), non_neg_integer(), [String.t()], [{flat_key(), term()}]) ::
+          [{flat_key(), term()}]
+  defp do_flatten_list([v | rest], idx, prefix, acc) when is_map(v) and v != %{} do
+    acc = do_flatten_map(v, [Integer.to_string(idx) | prefix], acc)
```
Ziinc, line 90: if this is subsequently going to be turned into a string then we don't need to do this here in the iteration.
Enum.join can accept mixed types
https://hexdocs.pm/elixir/Enum.html#join/2

```suggestion
    acc = do_flatten_map(v, [idx | prefix], acc)
```

Ziinc (approved):

lib/logflare/logs/cleaner.ex
djwhitt: Ran the numbers and pulled out a concrete example. The trade-off is real but small in both directions.

**Behavior** — collision example:
```elixir
Cleaner.flatten(%{"a.b" => :explicit, "a" => %{"b" => :nested}})
```
|  | result |
|---|---|
| with reverse (current) | `%{"a.b" => :explicit}` — literal flat key wins |
| without reverse | `%{"a.b" => :nested}` — derived path wins |

The test at `metadata_cleaner_test.exs:156` pins the current outcome. Worth noting: the rule "literal wins" only holds for this case because Erlang's small-map representation iterates `"a"` before `"a.b"` in term order — iteration order isn't part of the API, so for maps that fall back to HAMT (>32 entries) the resolution wouldn't be predictable either way. Both behaviors are defensible; "literal wins" feels slightly more intentional, but it's leaning on incidental ordering.

**Perf** (same bench, with vs without reverse):

| Payload | Metric | with reverse | without | Δ |
|---|---|---|---|---|
| otel trace | reductions | 440 | 440 | flat |
| otel trace | memory | 6.86 KB | 6.47 KB | −5.7% |
| edge log | reductions | 1,580 | 1,570 | −0.6% |
| edge log | memory | 31.32 KB | 30.26 KB | −3.4% |
| list heavy | reductions | 9,440 | 9,170 | **−2.9%** |
| list heavy | memory | 156.49 KB | 145.51 KB | **−7.0%** |
| list heavy | time | 47.18 μs | 44.75 μs | **−5.2%** |

(Time deltas on otel/edge are within bench noise — deviation there exceeds the delta. Reductions and memory are the clean signal.)

So removing reverse buys ~3% reductions / ~7% memory / ~5% time on the worst-case payload, in exchange for flipping the collision rule. Personally lean toward keeping it for the explicit-wins property, but happy either way — your call.

lib/logflare/logs/cleaner.ex
djwhitt: Fixed in c443c7818f70f73614804e7aad5942be5d81bcfb

lib/logflare/logs/cleaner.ex
djwhitt: Fixed in c443c7818f70f73614804e7aad5942be5d81bcfb

lib/logflare/logs/cleaner.ex
djwhitt: Fixed in c443c7818f70f73614804e7aad5942be5d81bcfb

lib/logflare/logs/cleaner.ex
djwhitt: Fixed in c443c7818f70f73614804e7aad5942be5d81bcfb

lib/logflare/logs/cleaner.ex
djwhitt: Addressed in d3c74cb504d825b6edbbade339c316f8141ca6c9

I chose not to convert any other occurrences and bloat this PR, but there are a few we can tidy up in the future.

lib/logflare/logs/cleaner.ex
djwhitt: Benchmarked this — the deferred form regresses on the list-heavy payload, which is the case that benefits most from this PR. Numbers:

| Metric | current (eager) | deferred | Δ |
|---|---|---|---|
| time | 45.39 μs | 52.88 μs | **+16.5%** |
| memory | 156.49 KB | 167.52 KB | +7.1% |
| reductions | 9,440 | 13,190 | **+39.7%** |

(otel_trace and edge_log shapes have no list-of-maps traversal, so they show flat memory/reductions and only bench-noise time deltas — deviation there is ±525% and ±93%.)

The reason: eager stringification converts the index **once per list element**, regardless of leaf count beneath it. Deferring pushes the conversion into `Enum.join`'s `String.Chars.to_string/1` call, which fires **once per leaf path** through that element. For lists of maps with N keys each, that's N conversions instead of 1.

lib/logflare/logs/cleaner.ex
djwhitt: Fixed in 36e2c76b6

test/logflare/logs/ingest/metadata_cleaner_test.exs
djwhitt: Added 4 property tests in 6b9c74247: output flatness (no non-empty maps/lists as values), leaf-count preservation on collision-free inputs, path reconstruction via dot-split walk, and idempotence on already-flat maps. 4 properties × ~100 generated inputs each, all passing.

Skipped the Iteraptor oracle. `Cleaner.flatten/1` isn't a port — it was written from scratch in #3305, and the contracts diverge on empty containers at leaves: `flatten(%{"a" => %{"b" => %{}}}) == %{"a.b" => %{}}`, where `Iteraptor.to_flatmap/1` drops them. A strict equivalence property would flag intentional divergences as failures.

Left the mixed-type list test as a function-level contract assertion — cleaner is a general utility and upstream filtering is logically separate.

amokan:
Looks good!

PR #3457, opened 2026-05-12
Exploring `Schema` optimizations

ruslandoga:
Please ignore this PR for now :)

---

Adds a Benchee harness for BigQuery schema helper hot paths under `bench/`.

Adds a pre-cast mailbox guard to `Logflare.Sources.Source.BigQuery.Schema.update/3` so sampled schema updates are dropped when the Schema process already has a backlog.

This targets observed queue spikes where in-process schema rate limiting still allowed ingest pipelines to keep filling the Schema inbox.

Validation: syntax-parsed changed files with `elixir`; `mix test test/logflare/source/bigquery/schema_test.exs` and `mix format` are blocked in this workspace by dependency lock mismatches that require `mix deps.get`.

### DIFF
--- a/bench/schema_update_bench.exs
+++ b/bench/schema_update_bench.exs
@@ -1,0 +1,151 @@
+# Benchmark the local Schema update work without starting the application or
+# touching BigQuery/Postgres. This approximates the work inside the Schema
+# GenServer before a remote patch would be attempted.
+#
+# Run with:
+#   mix run --no-start bench/schema_update_bench.exs
+#
+# Optional env vars:
+#   SCHEMA_UPDATE_BENCH_SCENARIO=all|wide_top_level|cloudflare_nested|otel_trace
+#   BENCH_TIME=5
+#   BENCH_WARMUP=2
+#   BENCH_MEMORY_TIME=0
+#   BENCH_REDUCTION_TIME=0
+#   PROFILE_AFTER=false|eprof|tprof|cprof|fprof
+
+Code.require_file("support/schema_helpers_bench.ex", __DIR__)
+
+defmodule Logflare.Profiling.SchemaUpdateBench do
+  @moduledoc false
+
+  alias Logflare.Google.BigQuery.SchemaUtils
+  alias Logflare.Profiling.SchemaHelpersBench
+  alias Logflare.Sources.Source.BigQuery.SchemaBuilder
+
+  @type scenario :: SchemaHelpersBench.scenario()
+  @type update_result :: :same | {:changed, map(), non_neg_integer()}
+
+  @spec run() :: :ok
+  def run do
+    scenarios = selected_scenarios()
+
+    Benchee.run(
+      %{
+        "existing payload: build + diff" => fn scenario ->
+          existing_payload_update(scenario)
+        end,
+        "one new field: build + diff" => fn scenario ->
+          new_field_update(scenario)
+        end,
+        "one new field: build + diff + flatmap" => fn scenario ->
+          new_field_update_with_flatmap(scenario)
+        end
+      },
+      benchmark_opts(inputs: benchee_inputs(scenarios))
+    )
+
+    :ok
+  end
+
+  @spec existing_payload_update(scenario()) :: update_result()
+  def existing_payload_update(scenario) do
+    scenario
+    |> SchemaHelpersBench.payload_for()
+    |> local_update(SchemaHelpersBench.schema_for(scenario), false)
+  end
+
+  @spec new_field_update(scenario()) :: update_result()
+  def new_field_update(scenario) do
+    scenario
+    |> SchemaHelpersBench.extended_payload_for()
+    |> local_update(SchemaHelpersBench.schema_for(scenario), false)
+  end
+
+  @spec new_field_update_with_flatmap(scenario()) :: update_result()
+  def new_field_update_with_flatmap(scenario) do
+    scenario
+    |> SchemaHelpersBench.extended_payload_for()
+    |> local_update(SchemaHelpersBench.schema_for(scenario), true)
+  end
+
+  defp local_update(payload, old_schema, build_flatmap?) do
+    new_schema = SchemaBuilder.build_table_schema(payload, old_schema)
+
+    if same_schemas?(old_schema, new_schema) do
+      :same
+    else
+      flatmap =
+        if build_flatmap? do
+          SchemaUtils.bq_schema_to_flat_typemap(new_schema)
+        else
+          %{}
+        end
+
+      {:changed, new_schema, map_size(flatmap)}
+    end
+  end
+
+  defp same_schemas?(old_schema, new_schema) do
+    old_schema == new_schema
+  end
+
+  defp benchmark_opts(overrides) do
+    [
+      time: benchmark_seconds("BENCH_TIME", 5.0),
+      warmup: benchmark_seconds("BENCH_WARMUP", 2.0),
+      memory_time: benchmark_seconds("BENCH_MEMORY_TIME", 0.0),
+      reduction_time: benchmark_seconds("BENCH_REDUCTION_TIME", 0.0),
+      profile_after: profile_after(),
+      print: [configuration: false],
+      formatters: [{Benchee.Formatters.Console, extended_statistics: true}]
+    ]
+    |> Keyword.merge(overrides)
+  end
+
+  defp profile_after do
+    case System.get_env("PROFILE_AFTER", "false") do
+      value when value in ["", "false", "0"] -> false
+      value when value in ["true", "1"] -> true
+      "eprof" -> :eprof
+      "tprof" -> :tprof
+      "cprof" -> :cprof
+      "fprof" -> :fprof
+      other -> raise ArgumentError, "Unknown PROFILE_AFTER=#{inspect(other)}"
+    end
+  end
+
+  defp benchmark_seconds(env_var, default) do
+    case System.get_env(env_var) do
+      nil ->
+        default
+
+      value ->
+        case Float.parse(value) do
+          {seconds, ""} when seconds >= 0 -> seconds
+          _ -> raise ArgumentError, "Expected #{env_var} to be a non-negative number"
+        end
+    end
+  end
+
+  defp selected_scenarios do
+    case System.get_env("SCHEMA_UPDATE_BENCH_SCENARIO", "all") do
+      "all" -> SchemaHelpersBench.scenario_names()
+      "wide_top_level" -> [:wide_top_level]
+      "cloudflare_nested" -> [:cloudflare_nested]
+      "otel_trace" -> [:otel_trace]
+      other -> raise ArgumentError, "Unknown SCHEMA_UPDATE_BENCH_SCENARIO=#{inspect(other)}"
+    end
+  end
+
+  defp benchee_inputs(scenarios) do
+    for scenario <- scenarios, into: %{} do
+      {scenario_label(scenario), scenario}
+    end
+  end
+
+  defp scenario_label(:wide_top_level), do: "wide top-level log"
+  defp scenario_label(:cloudflare_nested), do: "cloudflare nested log"
+  defp scenario_label(:otel_trace), do: "otel trace"
+end
+
+Logflare.Profiling.SchemaUpdateBench.run()
--- a/bench/schema_update_bench.exs
+++ b/bench/schema_update_bench.exs
@@ -69,9 +69,10 @@ defmodule Logflare.Profiling.SchemaUpdateBench do
   end
 
   defp local_update(payload, old_schema, build_flatmap?) do
-    new_schema = SchemaBuilder.build_table_schema(payload, old_schema)
+    {new_schema, schema_changed?} =
+      SchemaBuilder.build_table_schema_with_change(payload, old_schema)
 
-    if same_schemas?(old_schema, new_schema) do
+    if not schema_changed? do
       :same
     else
       flatmap =
@@ -83,10 +84,6 @@ defmodule Logflare.Profiling.SchemaUpdateBench do
 
       {:changed, new_schema, map_size(flatmap)}
     end
-  end
-
-  defp same_schemas?(old_schema, new_schema) do
-    old_schema == new_schema
   end
 
   defp benchmark_opts(overrides) do
--- a/bench/support/schema_helpers_bench.ex
+++ b/bench/support/schema_helpers_bench.ex
@@ -142,7 +142,7 @@ defmodule Logflare.Profiling.SchemaHelpersBench do
     |> SchemaUtils.bq_schema_to_flat_typemap()
   end
 
-  @spec repeat(pos_integer(), (() -> term())) :: :ok
+  @spec repeat(pos_integer(), (-> term())) :: :ok
   def repeat(count, fun) when is_integer(count) and count > 0 and is_function(fun, 0) do
     do_repeat(count, fun)
   end
@@ -199,16 +199,29 @@ defmodule Logflare.Profiling.SchemaHelpersBench do
     end
   end
 
-  defp benchmark_opts(overrides \\ []) do
+  defp benchmark_opts(overrides) do
     [
       time: benchmark_seconds("BENCH_TIME", 5.0),
       warmup: benchmark_seconds("BENCH_WARMUP", 2.0),
       memory_time: benchmark_seconds("BENCH_MEMORY_TIME", 3.0),
       reduction_time: benchmark_seconds("BENCH_REDUCTION_TIME", 3.0),
+      profile_after: profile_after(),
       print: [configuration: false],
       formatters: [{Benchee.Formatters.Console, extended_statistics: true}]
     ]
     |> Keyword.merge(overrides)
+  end
+
+  defp profile_after do
+    case System.get_env("PROFILE_AFTER", "false") do
+      value when value in ["", "false", "0"] -> false
+      value when value in ["true", "1"] -> true
+      "eprof" -> :eprof
+      "tprof" -> :tprof
+      "cprof" -> :cprof
+      "fprof" -> :fprof
+      other -> raise ArgumentError, "Unknown PROFILE_AFTER=#{inspect(other)}"
+    end
   end
 
   defp benchmark_seconds(env_var, default) do
--- a/bench/support/schema_helpers_bench.ex
+++ b/bench/support/schema_helpers_bench.ex
@@ -1,0 +1,399 @@
+defmodule Logflare.Profiling.SchemaHelpersBench do
+  @moduledoc false
+
+  alias Logflare.Google.BigQuery.SchemaUtils
+  alias Logflare.Sources.Source.BigQuery.SchemaBuilder
+  alias LogflareWeb.Logs.PayloadTestUtils
+
+  @data_key {__MODULE__, :data}
+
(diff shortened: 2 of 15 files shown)


djwhitt:
Hey @ruslandoga — heads up that #3458 (currently in draft) rewrites `SchemaUtils.flatten_typemap/1` to replace the `Iteraptor.to_flatmap` call with a direct walk, getting ~2-4× faster and ~5-13× less memory on realistic edge_log / OTel fixtures. Looks like we're targeting overlapping hot paths with different harnesses (Benchee here vs. `:tprof` + a snapshot/history file under `test/profiling/` in mine).

Wanted to coordinate before either lands. A few options I can see:

- Land yours first; I rebase #3458 on top. Your baseline numbers measure the slow Iteraptor impl, and you re-baseline after mine.
- Land mine first; your benchmark measures the new direct-walk impl from the start. Probably the more useful ordering for ongoing trend tracking.
- Consolidate harnesses — Benchee for ips/comparison + allocation/reduction snapshots in one place, since two parallel benchmark setups for the same code path is a smell.

Happy to defer to your preference — let me know what works.

ruslandoga:
👋 @djwhitt 

Please feel free to ignore this PR, it was just a one shot from Codex based on Ziinc's task description. I think we should proceed with landing yours first. And if it's clear that a direct walk is an improvement, there is probably little reason to include the current implementation in history :)

ruslandoga:
Added a small standalone benchmark for `Process.info(pid, :message_queue_len)` in `bench/process_info_message_queue_len_bench.exs`.

On my local run with `ITERATIONS=1000000 SENDERS=1`:

```text
sleeping process, empty mailbox:       ~29 ns/call
sleeping process, 100k queued messages: ~29 ns/call
draining process under sender pressure: ~3.1 us/call
```

So the current `Process.info/2` guard is cheap for idle/backlogged targets, but it gets meaningfully more expensive when the target process is actively receiving/draining messages. That still looks acceptable for sampled schema updates, but Ziinc's atomics/counters suggestion is probably the better next shape: keep a per-Schema-process queued/in-flight counter, increment before cast, decrement after handling, and skip casts when the counter is over a small threshold. That would avoid VM mailbox introspection on the hot caller side and track only Schema update work.

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -96,7 +96,12 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
          source: source,
          bigquery_project_id: project_id,
          bigquery_dataset_id: dataset_id,
-         name: Backends.via_source(source, Schema, backend.id)
+         name:
+           Backends.via_source_with_value(
```
ruslandoga, line 100: The Schema process is now registered with a Registry value containing its `:atomics` counter. That keeps the counter tied to the process registration lifetime and lets callers check outstanding schema work before enqueueing another cast.

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -195,6 +208,51 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
...
+
+  defp checkout_registry_update(key) do
+    case Registry.lookup(Backends.SourceRegistry, key) do
+      [{_pid, counter}] -> checkout_counter(counter)
+      _ -> :drop
+    end
+  end
+
+  defp checkout_counter(counter) do
+    queued = :atomics.add_get(counter, 1, 1)
```
ruslandoga, line 238: This counter tracks queued plus in-flight schema update work, not just mailbox length. The increment happens before `GenServer.cast/2`; if the source is already over the configured limit, the increment is rolled back and the cast is skipped.

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -73,20 +85,21 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
...
-  def handle_cast(
-        {:update, %LogEvent{}, %Source{lock_schema: true}},
-        state
-      ),
-      do: {:noreply, state}
+  def handle_cast({:update, counter, log_event, source}, state) do
+    try do
+      handle_update(log_event, source, state)
+    after
+      checkin_update(counter)
```
ruslandoga, line 92: The decrement is intentionally in `after` so the counter is released only after schema diffing/patch handling finishes, including early returns and errors. This is the piece that makes the value represent actual outstanding work instead of cast latency.

lib/logflare/sources/source/bigquery/schema_builder.ex
```diff
@@ -126,28 +126,34 @@ defmodule Logflare.Sources.Source.BigQuery.SchemaBuilder do
   @spec build_table_schema([map()] | map(), TFS.t()) :: TFS.t()
 
   def build_table_schema(params, %{fields: old_fields}) do
-    protected_keys = Enum.map(initial_table_schema().fields, & &1.name)
+    initial_schema = initial_table_schema()
+    protected_keys = MapSet.new(initial_schema.fields, & &1.name)
+    param_keys = Map.keys(params)
+    param_key_set = MapSet.new(param_keys)
+    old_fields_by_name = Map.new(old_fields, &{&1.name, &1})
```
ruslandoga, line 162: `build_table_schema/2` now indexes old fields and param keys once, avoiding repeated linear scans while preserving the existing field-retention behavior for keys not present in the sampled event.

lib/logflare/sources/source/bigquery/schema_builder.ex
```diff
@@ -242,51 +248,41 @@ defmodule Logflare.Sources.Source.BigQuery.SchemaBuilder do
     Map.has_key?(params, "resource") and Map.has_key?(params, "scope")
   end
 
-  defimpl DeepMerge.Resolver, for: Model.TableFieldSchema do
-    @doc """
-    Implements merge for schema key conflicts.
-    Overwrites fields schemas that are present BOTH in old and new TFS structs and keeps fields schemas present ONLY in old.
-    """
+  defp merge_payload_maps(map, acc) when is_map(map) do
```
ruslandoga, line 277: This replaces `DeepMerge` in the array-of-maps path with a narrow recursive map merge. It avoids protocol dispatch in the hot path while keeping the previous behavior of merging nested map samples from repeated records.

lib/logflare/google/bigquery/schema_utils/schema_utils.ex
```diff
@@ -101,17 +103,16 @@ defmodule Logflare.Google.BigQuery.SchemaUtils do
...
 
-      k =
-        cond do
-          is_atom(k) -> k
-          String.valid?(k) -> String.to_atom(k)
-          true -> decode_until_valid!(k)
-        end
-
-      {k, v}
+      {normalize_typemap_key(k), v}
```
ruslandoga, line 106: Typemap creation no longer atomizes incoming field names. The flattened schema maps already stringify paths, so keeping binary keys here avoids atom-table growth and removes `String.to_atom/1` from the schema hot path.

ruslandoga (commented):
Added inline notes on the schema pressure-control and helper hot-path changes.

ruslandoga:
Follow-up profiling pass pushed in `07c6d38f8`.

What changed:
- Added `bench/schema_update_bench.exs`, an e2e-ish local Schema update benchmark that does `build_table_schema/2` plus the schema equality/update decision, optionally including flatmap generation for the post-patch/persist path.
- Added `PROFILE_AFTER=eprof|tprof|cprof|fprof` support to the schema helper benches.
- Simplified `Schema.same_schemas?/2` to `old_schema == new_schema`. The previous implementation computed two flatmaps and then still required `old_schema == new_schema`, so the flatmap work was redundant on both matching and changed schemas.
- Avoided rebuilding the initial table schema/protected-key set on every `build_table_schema/2` call.

Quick wide-top-level local run after the change:
- `existing payload: build + diff`: ~16.8K ips, avg ~59.6us in the eprof run.
- `one new field: build + diff`: ~18.7K ips, avg ~53.5us in the eprof run.
- `one new field: build + diff + flatmap`: ~12.4K ips, avg ~80.7us in the eprof run.

Profiler notes:
- Before the `same_schemas?/2` simplification, the wide e2e path spent visible time flattening both old and new schemas even when only deciding whether a patch is needed.
- After the simplification, the remaining cost is mostly `SchemaBuilder` traversal/sorting/struct construction, with the flatmap path only showing up when explicitly benchmarking the persist-style `+ flatmap` case.

Commands used:
- `mix test test/logflare/source/bigquery/schema_test.exs test/logflare/google/bigquery/schema_builder_test.exs`
- `SCHEMA_UPDATE_BENCH_SCENARIO=wide_top_level BENCH_TIME=1 BENCH_WARMUP=0.2 BENCH_MEMORY_TIME=0 BENCH_REDUCTION_TIME=0 mix run --no-start bench/schema_update_bench.exs`
- `PROFILE_AFTER=eprof SCHEMA_UPDATE_BENCH_SCENARIO=wide_top_level BENCH_TIME=0.2 BENCH_WARMUP=0 BENCH_MEMORY_TIME=0 BENCH_REDUCTION_TIME=0 mix run --no-start bench/schema_update_bench.exs`

bench/schema_update_bench.exs
```diff
@@ -0,0 +1,151 @@
...
+  alias Logflare.Sources.Source.BigQuery.SchemaBuilder
+
+  @type scenario :: SchemaHelpersBench.scenario()
+  @type update_result :: :same | {:changed, map(), non_neg_integer()}
+
+  @spec run() :: :ok
+  def run do
+    scenarios = selected_scenarios()
+
+    Benchee.run(
```
ruslandoga, line 32: This is the e2e-ish benchmark I used for profiling. It avoids app startup and external BigQuery/Postgres calls, but exercises the local Schema GenServer hot path shape: build a candidate schema, decide whether it changed, and optionally build the flatmap that persist would need.

bench/schema_update_bench.exs
```diff
@@ -0,0 +1,151 @@
...
+      %{
+        "existing payload: build + diff" => fn scenario ->
+          existing_payload_update(scenario)
+        end,
+        "one new field: build + diff" => fn scenario ->
+          new_field_update(scenario)
+        end,
+        "one new field: build + diff + flatmap" => fn scenario ->
```
ruslandoga, line 40: The `+ flatmap` case is intentionally separated because eprof/tprof showed flatmap generation is a distinct cost. On the wide-top-level run, `build + diff` was ~53-60us, while `build + diff + flatmap` was ~80us, so flatmap construction adds roughly 20-30us for that shape.

bench/schema_update_bench.exs
```diff
@@ -0,0 +1,151 @@
...
+          SchemaUtils.bq_schema_to_flat_typemap(new_schema)
+        else
+          %{}
+        end
+
+      {:changed, new_schema, map_size(flatmap)}
+    end
+  end
+
+  defp same_schemas?(old_schema, new_schema) do
```
ruslandoga: This mirrors the production `same_schemas?/2` simplification. eprof/tprof showed the old check spending time in `SchemaUtils.bq_schema_to_flat_typemap/1`, `flatten_node/3`, `to_typemap/2`, and map/list traversal even though the final predicate still required structural equality. That flatmap work was redundant for the equality decision.

bench/support/schema_helpers_bench.ex
```diff
@@ -0,0 +1,412 @@
...
+    end
+  end
+
+  defp benchmark_opts(overrides) do
+    [
+      time: benchmark_seconds("BENCH_TIME", 5.0),
+      warmup: benchmark_seconds("BENCH_WARMUP", 2.0),
+      memory_time: benchmark_seconds("BENCH_MEMORY_TIME", 3.0),
+      reduction_time: benchmark_seconds("BENCH_REDUCTION_TIME", 3.0),
+      profile_after: profile_after(),
```
ruslandoga, line 208: Benchee 1.5 supports `profile_after`, including `:tprof` on this OTP/Elixir combo. Use e.g. `PROFILE_AFTER=eprof ... mix run --no-start bench/schema_update_bench.exs` or `PROFILE_AFTER=tprof ...`; tprof output broadly matched eprof in pointing at SchemaBuilder traversal/sort/merge and flatmap generation.

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -195,13 +208,53 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
...
+
+  defp checkin_update(nil), do: :ok
+  defp checkin_update(counter), do: :atomics.sub(counter, 1, 1)
+
+  defp max_message_queue_len do
+    Application.get_env(:logflare, __MODULE__)[:max_message_queue_len]
+  end
+
+  defp same_schemas?(old_schema, new_schema) do
+    old_schema == new_schema
```
ruslandoga: Profiling found this was the most obvious avoidable work: the previous implementation built old and new flatmaps and then still required `old_schema == new_schema`. This line preserves the actual behavior while removing two complete typemap/flatten traversals from every schema update decision.

lib/logflare/sources/source/bigquery/schema_builder.ex
```diff
@@ -126,60 +152,44 @@ defmodule Logflare.Sources.Source.BigQuery.SchemaBuilder do
   @spec build_table_schema([map()] | map(), TFS.t()) :: TFS.t()
 
   def build_table_schema(params, %{fields: old_fields}) do
```
ruslandoga: After removing the redundant flatmap equality work, eprof/tprof show the remaining local CPU is concentrated here: building field schemas, merging with the old schema, sorting via `deep_sort_by_fields_name/1`, and `Enum.uniq_by/2`. For wide top-level samples, this is now the main CPU area rather than typemap flattening.

lib/logflare/sources/source/bigquery/schema_builder.ex
```diff
@@ -7,6 +7,32 @@ defmodule Logflare.Sources.Source.BigQuery.SchemaBuilder do
   alias Logflare.BigQuery.SchemaTypes
   alias Model.TableFieldSchema, as: TFS
 
+  @initial_table_schema %Model.TableSchema{
```
ruslandoga, line 10: This avoids reconstructing the initial schema fields and protected-key set machinery on each call. It is a small win compared with the flatmap short-circuit, but it removes repeated setup work visible around `initial_table_schema/0`, `MapSet.new/2`, and `MapSet.member?/2` in the profile.

lib/logflare/sources/source/bigquery/schema_builder.ex
```diff
@@ -242,51 +252,41 @@ defmodule Logflare.Sources.Source.BigQuery.SchemaBuilder do
...
-        )
-        when is_list(old_fields)
-        when is_list(new_fields) do
-      # collect all names for new fields schemas
-      new_fields_names = Enum.map(new_fields || [], & &1.name)
+  defp merge_field_schema(nil, %TFS{} = new), do: new
 
-      # filter field schemas that are present only in old table field schema
-      uniq_old_fs = for fs <- old_fields, fs.name not in new_fields_names, do: fs
+  defp merge_field_schema(%TFS{fields: old_fields}, %TFS{fields: new_fields} = new)
```
ruslandoga: The merge path still appears in the profiles (`merge_field_schema/2`, nested `Enum.map/2`, `Map.get/3`, `MapSet.member?/2`). Next possible optimization would be more structural: return `{schema, changed?}` from SchemaBuilder or avoid global sorting/uniq when no schema shape changed.

ruslandoga (commented):
Added a profiling-focused review pass. The short version: before the latest change, profiling showed redundant flatmap generation in the schema equality check; after the change, the remaining local CPU is mostly schema construction/sorting/merge traversal.

ruslandoga:
Profiling summary from the latest `eprof`/`tprof` pass:

What showed up before the latest simplification:
- The e2e-ish `build + diff` path spent visible time in `SchemaUtils.bq_schema_to_flat_typemap/1`, `SchemaUtils.to_typemap/2`, `SchemaUtils.flatten_node/3`, `Enum.map/2`, and map/list traversal.
- That came from `same_schemas?/2` building flatmaps for both old and new schemas before still requiring `old_schema == new_schema`.
- Since the final predicate already required structural equality, the flatmap diff work was redundant. Removing it preserved behavior and cut the local wide-top-level decision path from roughly ~130-170us in earlier smoke runs to ~53-60us in the focused eprof run.

Current wide-top-level eprof run after the simplification:
- `one new field: build + diff`: ~18.7K ips, avg ~53.5us.
- `existing payload: build + diff`: ~16.8K ips, avg ~59.6us.
- `one new field: build + diff + flatmap`: ~12.4K ips, avg ~80.7us.

What remains hot:
- `SchemaBuilder.build_table_schema/2` and helpers dominate the local CPU now.
- The recurring functions in eprof/tprof are `build_fields_schemas/2`, `merge_field_schema/2`, `determine_type/3`, `deep_sort_by_fields_name/1`, `Enum.uniq_by/2`, `Enum.map/2`, `Map.get/3`, and lower-level map/list traversal.
- Flatmap generation is still a real cost, but only in the explicit `+ flatmap` path that models persist-time work, not in the ordinary update decision anymore.

Possible next optimizations, if we want to keep going:
- Have `SchemaBuilder` return `{schema, changed?}` so the caller does not need a second full structural comparison.
- Avoid `deep_sort_by_fields_name/1` when no new field was introduced, or sort only changed subtrees.
- Avoid the final `Enum.uniq_by/2` by constructing top-level fields in a way that guarantees uniqueness.
- Consider a builder representation that tracks old fields by name through recursion, so nested merges avoid rebuilding lookup maps repeatedly.

I added inline comments on the PR around the benchmark scenarios, profiler config, `same_schemas?/2`, and the remaining SchemaBuilder hot paths.

ruslandoga:
Follow-up pushed: `b5e0aa40a` (`Skip schema diff work for unchanged payloads`).

This makes `SchemaBuilder` return `{schema, changed?}` so `Schema` can skip the separate full-schema equality diff for unchanged payloads. When no field changed, the builder returns the existing schema directly and avoids the final `deep_sort_by_fields_name/1` pass. For changed payloads, it also avoids the previous final `Enum.uniq_by/2` by tracking existing field names while constructing the updated field list.

Short branchmarks against the previous PR commit (`BENCH_TIME=2 BENCH_WARMUP=1 BENCH_REDUCTION_TIME=1 mix run --no-start bench/schema_update_bench.exs`):

```text
wide top-level existing payload: 7.61K ips -> 34.63K ips, reductions 5.94K -> 3.65K
wide top-level one new field: 12.81K ips -> 22.11K ips, reductions 5.99K -> 6.02K
otel trace existing payload: 58.26K ips -> 73.86K ips, reductions 2.09K -> 1.41K
otel trace one new field: 21.67K ips -> 65.68K ips, reductions 2.24K -> 2.32K
cloudflare nested one new field: 8.92K ips -> 26.11K ips, reductions 3.99K -> 4.21K
cloudflare nested existing payload: 26.10K ips -> 18.61K ips, reductions 3.95K -> 3.00K
```

The nested no-op wall-time result is mixed/noisy, but it still reduces scheduler work. The more important ingest cases here are no-op wide schemas and changed schemas, where this removes the old sort+compare path and improves ips materially.

Verification: `mix test test/logflare/google/bigquery/schema_builder_test.exs`.

ruslandoga:
Closing in favor of smaller, self-contained patches, starting with https://github.com/Logflare/logflare/pull/3534

PR #3458, opened 2026-05-12
perf(schema): rewrite SchemaUtils.flatten_typemap as direct walk

djwhitt:
## Summary

- `SchemaUtils.flatten_typemap/1` was using `Iteraptor.to_flatmap/1`, which paid a heavy generic-traversal cost (per-key `Integer.parse` for list-vs-map detection, polymorphic `String.Chars` dispatch on every value, key-path concatenation via `Iteraptor.Utils.squeeze`) that wasn't load-bearing for the narrow typemap shape (`%{atom => %{t: type} | %{t: :map, fields: nested}}`).
- Rewritten as a direct tail-recursive walk that emits dot-delimited keys in one pass; the `format_flatmap_field_names/1` helper is no longer needed.
- Adds snapshot/history tooling to `test/profiling/log_event_make_bench.exs` so future perf work can track trends. The history file is seeded with the `origin/main` baseline (in the tooling commit) and the post-change measurement (in the perf commit), so the diff itself documents the win.

## Results — apples-to-apples on `test/profiling/log_event_make_bench.exs` (Apple M5)

| | Before | After | Improvement |
|---|---|---|---|
| **otel trace** wall | 79.4 µs | 38.2 µs | **2.08× faster** |
| **otel trace** memory | 0.29 MB | 58.5 KB | **5.07× less** |
| **edge log** wall | 307.5 µs | 79.9 µs | **3.85× faster** |
| **edge log** memory | 1.29 MB | 145.8 KB | **9.05× less** |
| **edge log** reductions | 137.4 K | 10.43 K | **13.17× less** |

The headline number for sustained ingest is the ~1.14 MB allocation drop per edge log event — pure GC relief.

## Diagnosis

`:tprof` against the realistic edge_log fixture (~80 leaves of mixed-case Cloudflare metadata, Factory-backed source with `validate_schema: true`) showed `BigQuerySchemaChange.validate/2` and its descendants accounting for ~25-30% of `LogEvent.make/2` time. The entry point was `Iteraptor.to_flatmap/1` inside `SchemaUtils.flatten_typemap/1`. The typemap shape is narrow and well-typed, so none of Iteraptor's polymorphism was earning its keep.

## Post-change hot path

Re-running `:tprof` shows Iteraptor gone from the top 20. The new top contributor is `SchemaUtils.to_typemap/1`'s per-key `String.valid?/1` + `String.to_atom/1` (~15-20% combined). Worth a follow-up on its own — also removes a `String.to_atom`-on-user-input memory leak vector.

## Test plan

- [x] CI green
- [x] 11 new unit tests in `test/logflare/google/bigquery/schema_utils_test.exs` cover the contract (nil, empty, leaves, datetime, list-typed leaves, nested maps at multiple depths, BQ-schema typemap shape) — all pass against the new impl after passing against the original Iteraptor impl
- [x] Broader consumer tests pass on the new impl: `source_schemas_test`, `source/bigquery/schema_test`, `lql_test` + `lql/parser_test` + `lql/integration_test`, `log_event_test`, `logs/search_test` (3 pre-existing `:failing`-tagged validator tests fail identically before and after — unrelated)
- [x] Optional: run `MIX_ENV=test mix run test/profiling/log_event_make_bench.exs` locally and verify the delta against the seeded snapshots is within noise (~5-10%)

### DIFF
--- a/lib/logflare/google/bigquery/schema_utils/schema_utils.ex
+++ b/lib/logflare/google/bigquery/schema_utils/schema_utils.ex
@@ -191,16 +191,27 @@ defmodule Logflare.Google.BigQuery.SchemaUtils do
     |> flatten_typemap()
   end
 
+  @spec flatten_typemap(map | nil) :: map
   def flatten_typemap(nil), do: %{}
 
-  def flatten_typemap(%{} = typemap) do
-    for {k, v} <- Iteraptor.to_flatmap(typemap), into: %{} do
-      {format_flatmap_field_names(k), v}
-    end
+  def flatten_typemap(typemap) when is_map(typemap) do
+    do_flatten_typemap(typemap, "", %{})
   end
 
-  defp format_flatmap_field_names(k) do
-    String.trim_trailing(k, ".t")
-    |> String.replace(".fields.", ".")
+  defp do_flatten_typemap(typemap, prefix, acc) do
+    Enum.reduce(typemap, acc, fn {key, value}, acc ->
+      flat_key =
+        if prefix == "", do: to_string(key), else: prefix <> "." <> to_string(key)
+
+      flatten_node(value, flat_key, acc)
+    end)
+  end
+
+  defp flatten_node(%{t: :map, fields: fields}, key, acc) do
+    do_flatten_typemap(fields, key, Map.put(acc, key, :map))
+  end
+
+  defp flatten_node(%{t: type}, key, acc) do
+    Map.put(acc, key, type)
   end
 end
--- a/test/logflare/google/bigquery/schema_utils_test.exs
+++ b/test/logflare/google/bigquery/schema_utils_test.exs
@@ -1,0 +1,125 @@
+defmodule Logflare.Google.BigQuery.SchemaUtilsTest do
+  use ExUnit.Case, async: true
+
+  alias Logflare.Google.BigQuery.SchemaUtils
+
+  describe "flatten_typemap/1" do
+    test "nil input returns empty map" do
+      assert SchemaUtils.flatten_typemap(nil) == %{}
+    end
+
+    test "empty map returns empty map" do
+      assert SchemaUtils.flatten_typemap(%{}) == %{}
+    end
+
+    test "single leaf field" do
+      assert SchemaUtils.flatten_typemap(%{name: %{t: :string}}) == %{"name" => :string}
+    end
+
+    test "multiple leaf fields at top level" do
+      typemap = %{
+        name: %{t: :string},
+        count: %{t: :integer},
+        active: %{t: :boolean}
+      }
+
+      assert SchemaUtils.flatten_typemap(typemap) == %{
+               "name" => :string,
+               "count" => :integer,
+               "active" => :boolean
+             }
+    end
+
+    test "datetime leaf" do
+      assert SchemaUtils.flatten_typemap(%{created_at: %{t: :datetime}}) == %{
+               "created_at" => :datetime
+             }
+    end
+
+    test "list-typed leaf preserves the tuple value" do
+      assert SchemaUtils.flatten_typemap(%{tags: %{t: {:list, :string}}}) == %{
+               "tags" => {:list, :string}
+             }
+    end
+
+    test "nested map emits parent :map entry and child leaf entries" do
+      typemap = %{
+        request: %{
+          t: :map,
+          fields: %{
+            method: %{t: :string},
+            status: %{t: :integer}
+          }
+        }
+      }
+
+      assert SchemaUtils.flatten_typemap(typemap) == %{
+               "request" => :map,
+               "request.method" => :string,
+               "request.status" => :integer
+             }
+    end
+
+    test "deeply nested maps preserve dot-delimited paths" do
+      typemap = %{
+        request: %{
+          t: :map,
+          fields: %{
+            headers: %{
+              t: :map,
+              fields: %{
+                content_type: %{t: :string}
+              }
+            }
+          }
+        }
+      }
+
+      assert SchemaUtils.flatten_typemap(typemap) == %{
+               "request" => :map,
+               "request.headers" => :map,
+               "request.headers.content_type" => :string
+             }
+    end
+
+    test "mixed top-level leaves and nested maps" do
+      typemap = %{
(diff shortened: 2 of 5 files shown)


test/profiling/log_event_make_profile.exs
```diff
@@ -0,0 +1,165 @@
...
+iterations = String.to_integer(System.get_env("ITERATIONS") || "2000")
+type = String.to_atom(System.get_env("TPROF_TYPE") || "time")
+
+for _ <- 1..500, do: LogEvent.make(params, %{source: source})
+
+IO.puts(
+  "\n== :tprof #{type} over #{iterations} iterations of LogEvent.make/2 (#{fixture_name}) ==\n"
+)
+
+Mix.Tasks.Profile.Tprof.profile(
```
ruslandoga, line 158: 👋 @djwhitt 

I wonder if using `profile_after` in Benchee would be able to achieve the same result. AFAIK it already supports tprof too. That way the benchmarking/profiling would be a bit more standardized.

ruslandoga:
Also, a bit off-topic, I recently [started](https://github.com/plausible/ch/pull/325) using https://github.com/benchmark-action/github-action-benchmark in [ch](https://github.com/plausible/ch/blob/25bf2a4b55f638f3a79ff5a977b84b1b76f8fd31/.github/workflows/bench.yml#L67-L81) and I don't particularly like it but it does introduce some structure into benchmarking/history. I wonder if something like that is a long term plan or would snapshotting stay like a script? I think there is also some benefit in re-using Benchee functionality maybe like custom reporters.

lib/logflare/google/bigquery/schema_utils/schema_utils.ex
```diff
@@ -191,16 +191,27 @@ defmodule Logflare.Google.BigQuery.SchemaUtils do
...
-      {format_flatmap_field_names(k), v}
-    end
+  def flatten_typemap(typemap) when is_map(typemap) do
+    do_flatten_typemap(typemap, "", %{})
+  end
+
+  defp do_flatten_typemap(typemap, prefix, acc) do
+    Enum.reduce(typemap, acc, fn {key, value}, acc ->
+      flat_key =
+        if prefix == "", do: to_string(key), else: prefix <> "." <> to_string(key)
```
amokan, line 204: Nit.

You could break this down a bit with pattern matching and kill that if..else. Not a requirement though.

Something like this should work (_not tested_).

```elixir
defp do_flatten_typemap(typemap, prefix, acc) do
  Enum.reduce(typemap, acc, fn {key, value}, acc ->
    flatten_node(value, join_key(prefix, key), acc)
  end)
end

defp join_key("", key), do: to_string(key)
defp join_key(prefix, key), do: prefix <> "." <> to_string(key)
```

amokan (approved):
I think this all looks great @djwhitt 

Glance at the comments from @ruslandoga and I noted a very minor/opinionated nit.

test/logflare/google/bigquery/schema_utils_test.exs
```diff
@@ -0,0 +1,125 @@
...
+  alias Logflare.Google.BigQuery.SchemaUtils
+
+  describe "flatten_typemap/1" do
+    test "nil input returns empty map" do
+      assert SchemaUtils.flatten_typemap(nil) == %{}
+    end
+
+    test "empty map returns empty map" do
+      assert SchemaUtils.flatten_typemap(%{}) == %{}
+    end
```
Ziinc, line 13: can be combined

test/logflare/google/bigquery/schema_utils_test.exs
```diff
@@ -0,0 +1,125 @@
...
+        count: %{t: :integer},
+        active: %{t: :boolean}
+      }
+
+      assert SchemaUtils.flatten_typemap(typemap) == %{
+               "name" => :string,
+               "count" => :integer,
+               "active" => :boolean
+             }
+    end
```
Ziinc, line 31: can be combined to be less verbose.

test/logflare/google/bigquery/schema_utils_test.exs
```diff
@@ -0,0 +1,125 @@
...
+      }
+
+      assert SchemaUtils.flatten_typemap(typemap) == %{
+               "request" => :map,
+               "request.method" => :string,
+               "request.status" => :integer
+             }
+    end
+
+    test "deeply nested maps preserve dot-delimited paths" do
```
Ziinc, line 63: property testing using StreamData would be a nice compresensive way to condense all these tests
https://hexdocs.pm/stream_data/StreamData.html

test/profiling/log_event_make_bench.exs
```diff
@@ -131,53 +143,133 @@ edge_log_params = %{
...
-# otel trace         0.31 MB     ±0.00%        0.31 MB        0.31 MB
-# edge log           1.34 MB     ±0.00%        1.34 MB        1.34 MB
-#
-# Reduction count statistics:
-#
-# Name               average  deviation         median         99th %
-# otel trace         32.78 K     ±0.13%        32.76 K        32.89 K
-# edge log          140.89 K     ±0.00%       140.89 K       140.89 K
+history =
+  if File.exists?(history_path) do
```
Ziinc, line 200: we can move most of this logic to a utility module under test/profiling/utils.ex for example to reuse across benchmarks.

test/profiling/log_event_make_profile.exs
```diff
@@ -0,0 +1,165 @@
...
+
+edge_log_params = %{
+  "event_message" =>
+    "POST | 200 | 3.254.227.51 | 9ca90d4f3f7cc99e | https://example.supabase.co/rest/v1/calls",
+  "id" => "3701391c-dead-405d-9943-61cc7576da87",
+  "identifier" => "example_project",
+  "metadata" => %{
+    "logflare_worker" => %{"worker_id" => "ZZ79SS"},
+    "request" => %{
+      "cf" => %{
```
Ziinc, line 63: note: in prod the cf object has been trimmed down significantly, so this object is not as representative anymore.

Ziinc (approved):
would be good to include ips values as well in benchmark results table for comparison, as pure time value is not as intuitive as to rate comparisons.

PR #3463, opened 2026-05-13
Add syslog adaptor benchmarks

ruslandoga:
Please ignore this PR for now :)

---

Adds benchmark-only syslog profiling scripts under test/profiling.

The benchmarks compare current batcher routing, no-batcher per-message sends, and preformatted batch sends using a local RFC6587 octet-counting TCP sink.

The Broadway benchmark runs through real Broadway processors/batchers, BufferProducer, a local NimblePool TCP sender, and fake cache entries so it can run with mix run --no-start.

Validated with mix format and small smoke runs for both benchmark scripts.

### DIFF
--- a/test/profiling/syslog_broadway_bench.exs
+++ b/test/profiling/syslog_broadway_bench.exs
@@ -1,0 +1,268 @@
+# Benchmark syslog adaptor pipeline shapes end-to-end through Broadway.
+#
+# Run with:
+#   mix run --no-start test/profiling/syslog_broadway_bench.exs
+#
+# Useful env:
+#   SYSLOG_BENCH_MESSAGE_BYTES=200,2000,50000
+#   SYSLOG_BENCH_CONCURRENCY=1,5
+#   SYSLOG_BENCH_BATCH_SIZE=50
+#   SYSLOG_BENCH_EVENTS=5000
+#   SYSLOG_BENCH_PUSH_CHUNK=1000
+#   SYSLOG_BENCH_TIME=5
+#   SYSLOG_BENCH_WARMUP=2
+
+Code.require_file("syslog_bench_support.exs", __DIR__)
+
+defmodule SyslogBroadwayBench.Pipeline do
+  @moduledoc false
+
+  alias Broadway.Message
+  alias Logflare.Backends
+  alias Logflare.Backends.Adaptor.SyslogAdaptor.Syslog
+  alias SyslogBenchSupport.NimbleTcpPool
+
+  def producer(source, backend) do
+    [
+      module:
+        {Backends.BufferProducer,
+         backend_id: backend.id,
+         source_id: source.id,
+         interval: 10,
+         buffer_size: SyslogBenchSupport.events() * 2},
+      transformer: {__MODULE__, :transform, []}
+    ]
+  end
+
+  def processor_opts(concurrency) do
+    [default: [concurrency: concurrency, min_demand: 1]]
+  end
+
+  def batcher_opts(concurrency, batch_size) do
+    [syslog: [concurrency: concurrency, batch_size: batch_size]]
+  end
+
+  def context(pool, backend) do
+    %{pool: pool, backend_id: backend.id}
+  end
+
+  def send_batch(pool, content, messages) do
+    case NimbleTcpPool.send(pool, content) do
+      :ok -> messages
+      {:error, reason} -> fail_batch(messages, reason)
+    end
+  end
+
+  def lookup_backend_config(backend_id) do
+    %{config: config} =
+      Backends.Cache.get_backend(backend_id) || raise "missing backend #{backend_id}"
+
+    config
+  end
+
+  def transform(event, _opts) do
+    %Message{data: event, acknowledger: {__MODULE__, _ref = nil, _meta = []}}
+  end
+
+  def ack(_ack_ref, _successful, _failed), do: :ok
+
+  defp fail_batch(messages, reason) do
+    Enum.map(messages, fn message ->
+      Broadway.Message.failed(message, reason)
+    end)
+  end
+end
+
+defmodule SyslogBroadwayBench.BatchedPipeline do
+  @moduledoc false
+
+  use Broadway
+
+  alias Broadway.Message
+  alias Logflare.Backends.Adaptor.SyslogAdaptor.Syslog
+  alias SyslogBroadwayBench.Pipeline
+
+  @behaviour Broadway.Acknowledger
+
+  def start_link(opts) do
+    Broadway.start_link(__MODULE__,
+      name: Keyword.fetch!(opts, :name),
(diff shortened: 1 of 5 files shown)


ruslandoga:
Benchmark smoke results from this workspace. These used tiny inputs to validate the scripts and the RFC6587 sink wiring, so treat the relative ordering as a sanity check only, not as production throughput data.

Direct formatter + TCP send:

```text
Command:
SYSLOG_BENCH_EVENTS=20 SYSLOG_BENCH_MESSAGE_BYTES=200 SYSLOG_BENCH_BATCH_SIZE=5 SYSLOG_BENCH_TIME=0.1 SYSLOG_BENCH_WARMUP=0 mix run --no-start test/profiling/syslog_send_bench.exs

##### With input 200 byte message #####
Name                                                                          ips        average  deviation         median         99th %
handle_message routes, handle_batch formats+sends batch                    957.01        1.04 ms   +/-260.07%        0.74 ms       27.09 ms
no batcher: handle_message formats+sends each message                      466.68        2.14 ms    +/-64.21%        1.83 ms       10.36 ms
preformat in handle_message, handle_batch sends preformatted frames        275.50        3.63 ms    +/-98.59%        2.17 ms       15.66 ms

Comparison:
handle_message routes, handle_batch formats+sends batch                    957.01
no batcher: handle_message formats+sends each message                      466.68 - 2.05x slower +1.10 ms
preformat in handle_message, handle_batch sends preformatted frames        275.50 - 3.47x slower +2.58 ms
```

Broadway end-to-end:

```text
Command:
SYSLOG_BENCH_EVENTS=20 SYSLOG_BENCH_MESSAGE_BYTES=200 SYSLOG_BENCH_BATCH_SIZE=5 SYSLOG_BENCH_CONCURRENCY=1 SYSLOG_BENCH_PUSH_CHUNK=10 SYSLOG_BENCH_TIME=0.1 SYSLOG_BENCH_WARMUP=0 mix run --no-start test/profiling/syslog_broadway_bench.exs

##### With input 200 byte message #####
Name                                                                                  ips        average  deviation         median         99th %
broadway: no batcher, handle_message formats+sends each message                     80.99       12.35 ms     +/-3.29%       12.17 ms       12.96 ms
broadway: handle_message routes, handle_batch formats+sends batch                   74.04       13.51 ms     +/-6.17%       13.38 ms       15.01 ms
broadway: handle_message preformats, handle_batch sends preformatted frames          6.86      145.79 ms     +/-0.00%      145.79 ms      145.79 ms

Comparison:
broadway: no batcher, handle_message formats+sends each message                     80.99
broadway: handle_message routes, handle_batch formats+sends batch                   74.04 - 1.09x slower +1.16 ms
broadway: handle_message preformats, handle_batch sends preformatted frames          6.86 - 11.81x slower +133.44 ms
```

Interpretation:

The direct send benchmark is mostly isolating formatter cost plus the number of TCP sends. With this tiny input, batching wins clearly because it does fewer `:gen_tcp.send/2` calls and the local sink waits until all expected octet-counted frames are parsed.

The Broadway smoke run includes pipeline startup, BufferProducer polling, processor/batcher handoff, NimblePool checkout, TCP send, and receiver parsing. At this very small event count, fixed Broadway overhead dominates, so the no-batcher path can look slightly faster than the batched path. The preformat variant is much slower here because it pays formatting cost in `handle_message/3` and still goes through batcher coordination before sending.

For useful throughput comparison, run the same scripts with larger inputs, for example:

```bash
SYSLOG_BENCH_MESSAGE_BYTES=200,2000,50000 \
SYSLOG_BENCH_CONCURRENCY=1,5 \
SYSLOG_BENCH_BATCH_SIZE=50 \
SYSLOG_BENCH_EVENTS=5000 \
SYSLOG_BENCH_PUSH_CHUNK=1000 \
mix run --no-start test/profiling/syslog_broadway_bench.exs
```

test/profiling/syslog_broadway_bench.exs
```diff
@@ -0,0 +1,268 @@
...
+  use Broadway
+
+  alias Broadway.Message
+  alias Logflare.Backends.Adaptor.SyslogAdaptor.Syslog
+  alias SyslogBroadwayBench.Pipeline
+
+  @behaviour Broadway.Acknowledger
+
+  def start_link(opts) do
+    Broadway.start_link(__MODULE__,
```
Ziinc, line 163: instead of copying the pipeline, can we make the pipeline parameterized such that we can pass opts to it, see `BigQuery.Pipeline` for how the args are passed for benchmarking

Ziinc (commented):
can be condensed.

ruslandoga:
Codex says:

---

Correction to my earlier smoke-results comment after rebasing onto current `main`:

The earlier Broadway numbers should not be used for a throughput decision. That harness created and stopped the TCP sink, pool, and Broadway pipeline inside every timed iteration, so fixed setup cost dominated the 20-event run.

The revised harness keeps the sink, real `SyslogAdaptor.Pool`, and Broadway pipeline alive for each Benchee scenario. Each iteration arms the sink, pushes messages through `Broadway.DummyProducer`, and stops timing only after the sink parses the expected RFC6587 frame count. Processor and batcher concurrency are now independent, matching production topology.

```bash
SYSLOG_BENCH_MESSAGE_BYTES=200,2000,50000 \
SYSLOG_BENCH_CONCURRENCY=16 \
SYSLOG_BENCH_BATCHER_CONCURRENCY=5 \
SYSLOG_BENCH_BATCH_SIZE=50 \
SYSLOG_BENCH_EVENTS=1000 \
SYSLOG_BENCH_TIME=3 \
SYSLOG_BENCH_WARMUP=1 \
mix run --no-start test/profiling/syslog_broadway_bench.exs
```

Apple M2, 8 cores; 16 Broadway processors, 5 batchers, 50 events/batch. Each iteration sent 1,000 events:

| Payload | Current: format + batch send | Preformat + batch send | No batcher |
| --- | ---: | ---: | ---: |
| 200 B | 286.1k events/s | 140.6k events/s | 90.0k events/s |
| 2 KB | 182.1k events/s | 165.5k events/s | 63.8k events/s |
| 50 KB | 16.35k events/s | 14.87k events/s | 14.06k events/s |

Current batching won every production-topology case. Preformatting was 10–51% slower and had substantially higher variance; removing the batcher was 16–69% slower. This also matters after #3812: unbatched handling would perform one bounded pool checkout and socket send per event instead of once per batch.

I also sampled batch sizes 10, 50, and 100. Size 50 was best for 200 B payloads; sizes 10 and 50 were effectively tied at 50 KB, while 100 was slightly worse. With optional AES-GCM encryption enabled, current batching still beat preformatting by 13–28% across the three payload sizes. There is no broad throughput win here that justifies removing `handle_batch` or preformatting in `handle_message`.

My recommendation is to keep the current production implementation. This PR is still useful as reproducible investigation, but the harness is roughly 700 lines, is not run in CI, and adds a production-only parameterization hook. Unless we expect recurring syslog tuning work, I would preserve these results in the PR and close it rather than merge benchmark maintenance into the codebase.

PR #3465, opened 2026-05-14
Follow-ups to #3458: schema_utils + bench profiling

djwhitt:
## Summary

Addresses review feedback on the already-merged #3458 (`SchemaUtils.flatten_typemap` rewrite). Splits into two areas:

**Schema utils**
- Pattern-match the empty-prefix check out of `do_flatten_typemap`'s `Enum.reduce` callback for clarity.
- Consolidate five overlapping leaf-case tests into one case-based test and add a StreamData property that compares `flatten_typemap/1` against an oracle implementation across random typemap shapes.

**Bench profiling**
- Extract the inline snapshot/history/delta-print block from `test/profiling/log_event_make_bench.exs` into a reusable `Logflare.Profiling` module — each bench script now reduces to `Benchee.run(...) |> Profiling.track(history_path)`.
- Add an ips delta column alongside the existing wall-time delta (display-only; history format unchanged).
- Fold `:tprof` profiling into Benchee's built-in `profile_after` via `PROFILE=1` / `TPROF_TYPE` env knobs, replacing the parallel `log_event_make_profile.exs` script.
- Flag the bench's `cf` fixture as overstating production shape (`botManagement` + `ja4Signals` + `tlsClientAuth` subtrees have been trimmed upstream) — kept as-is so existing history baselines remain comparable.

## Test plan
- [x] `mix test test/logflare/google/bigquery/schema_utils_test.exs` passes (including the new property test).
- [x] `MIX_ENV=test mix run test/profiling/log_event_make_bench.exs` runs and prints the new ips delta column.
- [x] `MIX_ENV=test PROFILE=1 mix run test/profiling/log_event_make_bench.exs` produces tprof output via Benchee.
- [x] `mix lint.all` clean.


Ziinc (approved):

Ziinc:
Needs to pass ci before merge

PR #3469, opened 2026-05-14
feat: team param for query LiveView

msmithstubbs:
Follows #3454 

Closes O11Y-1754

### DIFF
--- a/lib/logflare/teams/team_context.ex
+++ b/lib/logflare/teams/team_context.ex
@@ -143,13 +143,14 @@ defmodule Logflare.Teams.TeamContext do
     |> select([resource_team: team], team.id)
   end
 
-  def resource_team_id_query(LogflareWeb.QueryLive, %{"q" => query}, user)
+  def resource_team_id_query(LogflareWeb.QueryLive, %{"q" => query}, user_or_team_user)
       when is_binary(query) and query != "" do
     with {:ok, formatted} <- SqlFmt.format_query(query),
          {:ok, [source_name | _]} <- Sql.extract_table_names(formatted) do
       Sources.Source
-      |> Teams.filter_by_user_access(user)
+      |> Teams.filter_by_user_access(user_or_team_user)
       |> where([source], source.name == ^source_name)
+      |> limit(2)
       |> select([resource_team: team], team.id)
     else
       _ -> nil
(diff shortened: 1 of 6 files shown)


test/logflare_web/live_views/query_live_test.exs
```diff
@@ -213,4 +222,14 @@ defmodule LogflareWeb.QueryLiveTest do
       refute html =~ "bytes processed"
     end
   end
+
+  defp submit_query_form(view, _conn) do
+    case view |> element("form") |> render_submit(%{}) do
+      {:error, {:live_redirect, _}} ->
+        render(view)
```
Ziinc: need to call follow_redirect() otherwise the redirect is not adhered to

Ziinc (changes requested):

lib/logflare/teams/team_context.ex
```diff
@@ -142,6 +143,19 @@ defmodule Logflare.Teams.TeamContext do
     |> select([resource_team: team], team.id)
   end
 
+  def resource_team_id_query(LogflareWeb.QueryLive, %{"q" => query}, user)
```
Ziinc:
```suggestion
  def resource_team_id_query(LogflareWeb.QueryLive, %{"q" => query}, user_or_team_user)
```

test/logflare_web/live_views/query_live_test.exs
```diff
@@ -145,45 +139,63 @@ defmodule LogflareWeb.QueryLiveTest do
...
 
+    test "updates team param from source in query URL", %{
+      conn: conn,
+      user: user,
+      team: team
+    } do
+      team_owner = insert(:user)
+      other_team = insert(:team, user: team_owner)
+      insert(:team_user, email: user.email, team: other_team)
+      insert(:source, user: team_owner, name: "canonical_source")
```
Ziinc, line 225: What if both teams that the effective user is a team_user of has a source with the same name? Basing it on source name should only be a fallback if there's no team param provided

lib/logflare/teams/team_context.ex
msmithstubbs: Also updated these in #3479 for consistency.

test/logflare_web/live_views/query_live_test.exs
msmithstubbs: Good point. The requirement to override an existing team context when a resource is specified falls down when the resource is ambiguous. 

This was also an issue when running a query, when the LiveView attempts to set a team context based on the submitted query.

Updated AuthLive to only override team context when the team param is missing.  
TeamContext will only change the team context when the query suggests a single team.

If a user has the same-named source in multiple teams, they will have to switch to the appropriate team before querying.

Ziinc (approved):

PR #3471, opened 2026-05-14
feat: team param for key value LiveView

msmithstubbs:
Ensures current team `t=` param is set in request URI.

Follows #3454


Ziinc (changes requested):
lets add a team switching query param test as well

msmithstubbs:
> lets add a team switching query param test as well

Does that mean requesting `/key-values` and having it set the team query param?

Key values are user only, so we'd have to just set the user's home team, right?

Ziinc (approved):

PR #3476, opened 2026-05-14
chore: better logging around potential BQ reservation errors

amokan:
There may be a better way to handle it, but this is my crude attempt to capture _potential_ reservation related errors from BQ.

Sources checked for potential strings:
- https://docs.cloud.google.com/bigquery/docs/reservations-workload-management
- https://docs.cloud.google.com/bigquery/docs/troubleshoot-queries


lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -685,7 +686,19 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
...
+          Logger.warning("Possible BigQuery reservation error",
+            user_id: user.id,
+            project_id: project_id,
+            reservation_opt: Keyword.get(query_opts, :reservation),
+            query_type: Keyword.get(query_opts, :query_type),
+            bq_error_message: decoded["message"]
+          )
+        end
+
         {:error, error}
```
Ziinc, line 695: i think this will break the error handling in the BQ pipeline which relies on it being a Tesla.Env

Ziinc (changes requested):

lib/logflare/backends/adaptor/bigquery_adaptor.ex
amokan: How so? We still return the same thing as before

Ziinc: ah i didn't see the binding at :690. this change will result in duplicate logs getting emitted though, one here and one at pipeline. i think better to conslidate the error logging to here.

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -685,7 +686,19 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
...
-        error = Jason.decode!(body)["error"] |> GenUtils.process_bq_errors(user.id)
+        decoded = Jason.decode!(body)["error"]
+        error = GenUtils.process_bq_errors(decoded, user.id)
+
+        if reservation_error?(decoded) do
+          Logger.warning("Possible BigQuery reservation error",
+            user_id: user.id,
+            project_id: project_id,
+            reservation_opt: Keyword.get(query_opts, :reservation),
+            query_type: Keyword.get(query_opts, :query_type),
```
djwhitt: `query_type: Keyword.get(query_opts, :query_type)` will always be `nil` here. By the time `query_opts` reaches `execute_user_query/5`, it's the output of `build_base_query_opts/2`, which *consumes* `:query_type` to resolve `:reservation` but never re-emits it into the returned keyword list. None of the tests pass `query_type`, so the suite won't catch this. Either thread `query_type` through to this point, or drop the field.

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -685,7 +686,19 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
          })}
 
       {:error, %{body: body}} ->
-        error = Jason.decode!(body)["error"] |> GenUtils.process_bq_errors(user.id)
+        decoded = Jason.decode!(body)["error"]
+        error = GenUtils.process_bq_errors(decoded, user.id)
+
+        if reservation_error?(decoded) do
+          Logger.warning("Possible BigQuery reservation error",
```
djwhitt: Echoing @Ziinc's open comment: the reservation-using callers already log the decoded error. `Logflare.Alerting` (alerting.ex:480, `query_type: :alerts`) does its own `Logger.error("Alert query execution failed with bad response", ...)` after decoding, so a reservation failure on an alert query would emit two lines for one failure. Worth deciding where this logging should live. Since the useful context (`query_type`, the resolved reservation) is actually in scope at the callers, consolidating there — or enriching the caller's existing log rather than adding a second line — may be cleaner than logging in `execute_user_query`. (This also resolves the `query_type`-always-nil comment.)

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -685,7 +686,19 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
...
+            user_id: user.id,
+            project_id: project_id,
+            reservation_opt: Keyword.get(query_opts, :reservation),
```
djwhitt: Minor: `reservation_opt` is the *resolved* reservation (post-`build_base_query_opts`), which for `:search`/`:alerts` is the user's configured reservation rather than anything the caller passed as `:reservation`. The value is the right one to log, but the `_opt` suffix is a little misleading — maybe just `reservation:`.

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -717,4 +730,11 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
       parameterValue: %Value{value: param}
     }
   end
+
+  @spec reservation_error?(error :: map() | any()) :: boolean()
```
djwhitt: Nit: `@spec reservation_error?(error :: map() | any())` — `any()` already subsumes `map()`, so this reduces to `error :: any()`.

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -41,6 +41,7 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
 
   @managed_service_account_partition_count 5
   @service_account_prefix "logflare-managed"
+  @reservation_error_regex ~r/reservation/i
```
djwhitt, line 44: Non-blocking, just flagging: this matches any error message containing "reservation" (e.g. a user query that references a column/table named `reservation` and fails for an unrelated reason). Given the log is explicitly framed as "Possible" and it's only a warning, I think that's an acceptable trade-off — just noting the false-positive surface in case you'd rather anchor it (e.g. require it co-occur with a status like `NOT_FOUND`/`PERMISSION_DENIED`).

lib/logflare/backends/adaptor/bigquery_adaptor.ex
amokan: fixed

lib/logflare/backends/adaptor/bigquery_adaptor.ex
amokan: fixed

lib/logflare/backends/adaptor/bigquery_adaptor.ex
amokan: Changed - my reasoning around `reservation_opt` was to make it clear that was the value of the reservation we _intended_ for this to use rather than to say reservation X had an issue.

I'm sure that makes more sense in my head than it does here.

lib/logflare/backends/adaptor/bigquery_adaptor.ex
amokan: fixed

lib/logflare/backends/adaptor/bigquery_adaptor.ex
amokan: fixed to avoid the duplication

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -685,7 +689,9 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
          })}
 
       {:error, %{body: body}} ->
-        error = Jason.decode!(body)["error"] |> GenUtils.process_bq_errors(user.id)
+        decoded = Jason.decode!(body)["error"]
+        error = GenUtils.process_bq_errors(decoded, user.id)
+        maybe_warn_reservation_error(decoded, user, project_id, query_opts)
```
Ziinc, line 694: Thought about this again, we should log out any query error except syntax/entity kind of user errors, not just for the specific reservation case.

Ziinc: It would let us capture and get alerted on other query issues like job queue maxing and reservation saturation

Ziinc (approved):
Can increase logging coverage to all system backend query errors

lib/logflare/backends/adaptor/bigquery_adaptor.ex
amokan: I've created O11Y-1970 for the sake of a follow-up

PR #3477, opened 2026-05-14
test: stabilize syslog truncation properties and scheduler metrics

djwhitt:
## Summary

Fixes two intermittent CI failures observed on `main`.

### 1. Syslog truncation property tests

The two property tests in `test/logflare/backends/adaptor/syslog_adaptor/syslog_test.exs` (added in #3379) asserted `byte_size(syslog_msg) <= max_bytes` for all `max_bytes` in `1..100_000`, but `SyslogAdaptor.Syslog.format/2` always emits the full RFC5424 headers — when `max_message_bytes` falls below the headers length (~74 bytes for a factory log event), the frame is `headers + empty body` and exceeds `max_bytes`. This is intentional in `lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex` (`allowed_length/3` floors to 0), and is also encoded by the existing non-property test "drops encrypted message when encryption overhead would exceed max_message_bytes".

The properties intermittently failed on `main` when StreamData picked a small `max_bytes` — the failure on main used seed 769458 → `max_bytes=6`.

**Fix:** measure the headers length per log event (via a new `headers_only/1` helper that calls `format` with `max_message_bytes: 0`) and assert `byte_size(syslog_msg) <= max(max_bytes, headers_floor)` in both properties. For the ciphertext property, skip the decryption assertions when the frame is headers-only.

Also added a `@doc` to `Syslog.format/2` so the headers-floor contract is visible to callers without reading the implementation.

### 2. Scheduler metrics test

`test/logflare/telemetry_test.exs:129 "scheduler metrics"` called `Schedulers.async_dispatch_stats/1` (which spawns a supervised task) and then `assert_receive` with the default 100ms timeout. Spawn + 10ms sample + dispatch can exceed 100ms under CI load, producing `Assertion failed, no matching message after 100ms`.

**Fix:** switch to the synchronous `Schedulers.collect_dispatch_stats/1` — same emission logic, no race. Matches the synchronous-call pattern other tests in the same file already use (`Telemetry.process_memory_metrics()`, etc.).

## Test plan

- [x] `mix test test/logflare/backends/adaptor/syslog_adaptor/syslog_test.exs` passes with default seed and with the originally failing seed `769458`.
- [x] `mix test test/logflare/telemetry_test.exs` passes.
- [x] `mix lint.diff` clean.

### DIFF
--- a/lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex
+++ b/lib/logflare/backends/adaptor/syslog_adaptor/syslog.ex
@@ -34,6 +34,13 @@ defmodule Logflare.Backends.Adaptor.SyslogAdaptor.Syslog do
   @tag_bytes 16
   @aes_overhead @iv_bytes + @tag_bytes
 
+  @doc """
+  Formats a log event as an RFC5424 syslog frame with octet-counting framing.
+
+  When `max_message_bytes` is smaller than the headers length, the frame falls
+  back to headers-only and will exceed `max_message_bytes` — RFC5424 headers
+  are never truncated.
+  """
   def format(log_event, config) do
     cipher_key = config[:cipher_key]
     structured_data = config[:structured_data]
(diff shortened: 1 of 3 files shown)


djwhitt:
Superseded by #3478 — branch renamed to better reflect the bundled scope (syslog properties + scheduler metrics flake).

PR #3479, opened 2026-05-15
refactor: resource_team_id_query

msmithstubbs:
Small refactor followup on default team param PRs.

* reduce repetition in `TeamContext`
* consistency in router

Ready to merge after:
* #3469


test/logflare_web/live_views/query_live_test.exs
```diff
@@ -235,6 +234,36 @@ defmodule LogflareWeb.QueryLiveTest do
...
+
+      view
+      |> render_hook("parse-query", %{
+        value: query
+      })
+
+      html = submit_query_form(view, conn)
+
+      assert html =~ "Ran query successfully"
+      assert_patch(view) =~ ~s(t=#{other_team.id})
```
Ziinc, line 264: we should add in a small flash each time we switch teams as well. i think now flashes will stack on top of each other, would be good to have a more robust toast system.

Ziinc (approved):

PR #3484, opened 2026-05-15
refactor: removed nested function calls in SingleTenant module functions

nelsonmestevao:
These are offenses in the latest version of credo.

### DIFF
--- a/test/logflare/single_tenant_test.exs
+++ b/test/logflare/single_tenant_test.exs
@@ -13,6 +13,7 @@ defmodule Logflare.SingleTenantTest do
   alias Logflare.Auth
   alias Logflare.Backends.Backend
   alias Logflare.Backends
+  alias Logflare.Backends.Adaptor.BigQueryAdaptor
 
   describe "single tenant mode using Big Query" do
     TestUtils.setup_single_tenant()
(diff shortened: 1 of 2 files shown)


test/logflare/single_tenant_test.exs
```diff
@@ -84,6 +85,8 @@ defmodule Logflare.SingleTenantTest do
     end
 
     test "Logflare.Application.startup_tasks/0 should insert plan and user" do
+      expect(BigQueryAdaptor, :update_iam_policy, fn -> :ok end)
```
Ziinc, line 88: why is this needed?

nelsonmestevao: To remove the error log about IAM policies 🙏

Ziinc: i think we should stub the underlying library function globally so that we can silence all related errors that get emitted, pretty sure some error logs are getting emitted due to us not stubbing it in the ConnCase/DataCase.
Can you do a followup for this? 🙏

Ziinc: once we have it in place then we won't need this expect anymore.

Ziinc (approved):

Ziinc (approved):

test/logflare/single_tenant_test.exs
nelsonmestevao: Done in https://github.com/Logflare/logflare/pull/3550 :pray:

PR #3489, opened 2026-05-16
perf: rewrite BigQuerySchemaChange.validate as single-pass walk

djwhitt:
## Summary

This is a perf rewrite **and** reactivates a validator that was silently no-oping in production. Both effects are real and intentional — see "Bugs fixed" for the no-op, "Benchmark" for the perf numbers, and "Rollout note" for behavior delta.

Four commits, layered:

1. **Single-pass walk** (`c0b5e70b7`). The old validator built a transient metadata flat typemap (`SchemaUtils.to_typemap/1` → `SchemaUtils.flatten_typemap/1`), `Map.merge`d it against the cached schema flat map purely to detect type conflicts, then threw the result away. New version walks `le.body` once, infers each leaf's type via `SchemaTypes.type_of/1`, looks up the dot-delimited path in the schema flat map, raises on mismatch. No intermediate typemap, no flatten, no throwaway allocation.

2. **Inner-loop tightening** (`e76cd3b39`). `walk_map/3` switched to `:maps.fold/3` with a named `walk_entry/3` callback instead of `Enum.each` + closure — drops the per-call closure allocation and the `Enumerable.Map` protocol dispatch. The empty-container and top-level-skip checks fold into function-head pattern matches on `walk_entry/3`, replacing two predicate calls per kv pair with a single pattern dispatch. Tail-recursive `walk_maps/3` replaces `Enum.each` for the list-of-maps branch.

3. **`join_key` revert** (`8d6838a9a`). The tightening commit also added `is_binary(key)` fastpath clauses to `join_key/2` to bypass the `String.Chars` protocol. tprof revealed those clauses *increased* allocations from 6.53 to 15.33 words/call — adding the guarded clauses appears to disable some compiler specialization around chained `<>` / `to_string`. Reverted those two clauses, kept everything else.

4. **Per-element scalar list validation** (`7c4c80103`). Addresses review feedback: `check_field/3`'s non-map list branch only validated `hd(v)`, so a body like `%{"user" => %{"ids" => [299, "x", 12]}}` against a `{:list, :integer}` schema slipped through. Walk every element. Bounded cost — see benchmarks below.

## Bugs fixed (in commit 1)

- **`validate/2` was a silent no-op in production.** The old code stored the full `%SourceSchema{}` struct in `schema_flatmap` (not its `:schema_flat_map` field), so `Map.merge` was merging the metadata flat map's string keys against the struct's atom keys (`:bigquery_schema`, `:inserted_at`, …). No key ever overlapped, no conflict ever raised. Production type errors were caught only later at BigQuery insert time. The rewrite extracts `:schema_flat_map`, so conflicts now surface as `{:error, _}` at validate time. Bug introduced in `c6551ae0c` (2021-08-07, "Hit the SourceSchema cache again") when the lookup was switched from `bq_schema_to_flat_typemap(schema)` to `SourceSchemas.Cache.get_source_schema_by/1` without extracting the flat-map field.
- **Top-level `"timestamp"` skip is what makes the no-op fix viable.** `LogEvent.mapper/1` injects `"timestamp"` as integer microseconds while the BQ schema types it `:datetime`. Once the validator actually compares again, every event with a cached schema would fail this conflict without an explicit carve-out. `@skip_top_level_keys ~w(timestamp)` plus a top-level-only skip clause handles it — nested `"timestamp"` fields still get validated.
- **Short-circuit empty schemas.** Cold-start sources used to walk the whole body just to merge against an empty map. `check_body(_body, schema_flat_map) when map_size(schema_flat_map) == 0, do: :ok`.

## Rollout note

This turns a dormant validator back on. Sources with type drift that previously slipped through (and either succeeded at BigQuery insert or failed downstream) will now get rejected at validate time with `{:error, "Type error! ..."}`. Worth watching ingestion error rates on rollout — the change is strictly more correct, but a real behavior delta for any source that's been accumulating schema-vs-payload skew.

Per-element list validation also scales linearly (~9 reductions and ~50 B per extra element). Sources sending very long REPEATED scalar lists (hundreds of elements or more) will see the largest cost shift, though absolute numbers remain well below the broken baseline — see benchmark caveats below.

## Tests

**Test-side fix.** The previously-`:failing` test file is un-tagged and runs by default. The original `:failing` assertions were failing because of a test wrapping bug (`valid?(metadata, schema)` instead of `valid?(%{"metadata" => metadata}, schema)`), not a real comparison-logic bug. Fixed the wrapping; assertions pass against the new walk.

Coverage added:

- `validate/2` short-circuits when `validate_schema: false` (no cache access)
- `validate/2` `:ok` for body matching the cached schema (timestamp regression)
- `validate/2` `{:error, _}` on a real leaf conflict
- Nested `"timestamp"` field still gets validated (top-level skip locked in)
- REPEATED RECORD list-of-maps walked entry-by-entry
- REPEATED scalar list walked entry-by-entry (commit 4)
- `DateTime` / `NaiveDateTime` values match `:datetime`

## Benchmark (Apple M5)

**Focused micro-bench** — `test/profiling/bq_schema_change_validate_bench.exs` (new, calls `validate/2` directly against a source with a populated `source_schema` cache, so `walk_map` actually fires). Baseline = `8e03624a4` (current main tip, dormant validator); branch tip = `7c4c80103` (final per-element walk).

| Scenario | wall | mem | reds |
|---|---|---|---|
| scalars (third) | 6.63 → 4.21 µs (**-37%**) | 14.1 → 2.5 KB (**-82%**) | 1255 → 526 (**-58%**) |
| lists (third_with_lists) | 8.13 → 4.92 µs (**-39%**) | 15.4 → 3.6 KB (**-77%**) | 1349 → 616 (**-54%**) |
| empty schema short-circuit | 3.67 → 0.50 µs (**-86%**) | 10.5 → 0.9 KB (**-91%**) | 951 → 134 (**-86%**) |

Net effect: the old code was already paying for a full body traversal + flat typemap build + `Map.merge` allocation on every event — that work just produced an inert result. The new code does a comparable single-pass walk and *actually* surfaces conflicts, while landing strictly cheaper on every metric.

The lists scenario gives back ~5% wall and ~14% mem compared to the head-only-check intermediate state, in exchange for actually validating every element. Cost scales linearly: ~9 reductions and ~50 B per extra list element — the 10-element-per-list bench above is representative of typical payloads; bodies with much longer REPEATED scalar lists will pay proportionally more but remain well below the broken baseline.

**End-to-end** — `test/profiling/log_event_make_bench.exs` (full `LogEvent.make/*` path; validate is one part):

| Scenario | wall | mem | reds |
|---|---|---|---|
| edge log | -25% | -43% | -43% |
| otel trace | -34% | -45% | -42% |

The wider bench's scenarios don't populate a `source_schema`, so they hit the empty-schema short-circuit and the dominant savings come from eliminating the transient flat-map allocation.

Full per-scenario history in `test/profiling/bq_schema_change_validate_bench.history.exs` and `test/profiling/log_event_make_bench.history.exs`.

## Test plan

- [x] `mix test test/logflare/validator/bq_schema_change_validator_test.exs` (8 `valid?/2` tests pass locally; 4 `validate/2` tests fail in `setup` on a stale "Free" plan row, unrelated to validator code — clean test DB expected on CI)
- [x] `mix test` on broader related files (validator, log event, schema utils, source schemas, BQ schema)
- [x] `mix lint.diff` — no new findings
- [x] `mix test.typings` — passes

### DIFF
--- a/lib/logflare/logs/validators/bq_schema_change_validator.ex
+++ b/lib/logflare/logs/validators/bq_schema_change_validator.ex
@@ -96,7 +96,11 @@ defmodule Logflare.Logs.Validators.BigQuerySchemaChange do
   defp walk_maps([], _prefix, _schema_flat_map), do: :ok
 
   defp walk_maps([h | t], prefix, schema_flat_map) do
-    walk_map(h, prefix, schema_flat_map)
+    case is_map(h) do
+      true -> walk_map(h, prefix, schema_flat_map)
+      false -> raise("Type error! Field `#{prefix}` has an unexpected type.")
+    end
+
     walk_maps(t, prefix, schema_flat_map)
   end
 
--- a/lib/logflare/logs/validators/bq_schema_change_validator.ex
+++ b/lib/logflare/logs/validators/bq_schema_change_validator.ex
@@ -43,15 +43,19 @@ defmodule Logflare.Logs.Validators.BigQuerySchemaChange do
   end
 
   defp walk_map(map, prefix, schema_flat_map) do
-    Enum.each(map, fn {k, v} ->
-      if not_empty_container?(v) and not skip?(prefix, k) do
-        check_field(v, join_key(prefix, k), schema_flat_map)
-      end
-    end)
+    :maps.fold(&walk_entry/3, {prefix, schema_flat_map}, map)
   end
 
-  defp skip?("", k), do: k in @skip_top_level_keys
-  defp skip?(_, _), do: false
+  defp walk_entry(_k, [], acc), do: acc
+  defp walk_entry(_k, m, acc) when is_map(m) and map_size(m) == 0, do: acc
+  defp walk_entry(_k, [[]], acc), do: acc
+  defp walk_entry(_k, [m], acc) when is_map(m) and map_size(m) == 0, do: acc
+  defp walk_entry(k, _v, {"", _} = acc) when k in @skip_top_level_keys, do: acc
+
+  defp walk_entry(k, v, {prefix, schema_flat_map} = acc) do
+    check_field(v, join_key(prefix, k), schema_flat_map)
+    acc
+  end
 
   defp check_field(%DateTime{}, key, schema_flat_map),
     do: enforce_type(:datetime, key, schema_flat_map)
@@ -63,7 +67,7 @@ defmodule Logflare.Logs.Validators.BigQuerySchemaChange do
     case hd(v) do
       head when is_map(head) ->
         enforce_type(:map, key, schema_flat_map)
-        Enum.each(v, &walk_map(&1, key, schema_flat_map))
+        walk_maps(v, key, schema_flat_map)
 
       head ->
         enforce_type({:list, SchemaTypes.type_of(head)}, key, schema_flat_map)
@@ -77,6 +81,13 @@ defmodule Logflare.Logs.Validators.BigQuerySchemaChange do
 
   defp check_field(v, key, schema_flat_map),
     do: enforce_type(SchemaTypes.type_of(v), key, schema_flat_map)
+
+  defp walk_maps([], _prefix, _schema_flat_map), do: :ok
+
+  defp walk_maps([h | t], prefix, schema_flat_map) do
+    walk_map(h, prefix, schema_flat_map)
+    walk_maps(t, prefix, schema_flat_map)
+  end
 
   defp enforce_type(incoming, key, schema_flat_map) do
     case Map.fetch(schema_flat_map, key) do
@@ -93,15 +104,8 @@ defmodule Logflare.Logs.Validators.BigQuerySchemaChange do
     end
   end
 
+  defp join_key("", key) when is_binary(key), do: key
   defp join_key("", key), do: to_string(key)
+  defp join_key(prefix, key) when is_binary(key), do: prefix <> "." <> key
   defp join_key(prefix, key), do: prefix <> "." <> to_string(key)
-
-  defp not_empty_container?(value)
-       when value == []
-       when value == %{}
-       when value == [[]]
-       when value == [%{}],
-       do: false
-
-  defp not_empty_container?(_), do: true
 end
--- a/lib/logflare/logs/validators/bq_schema_change_validator.ex
+++ b/lib/logflare/logs/validators/bq_schema_change_validator.ex
@@ -9,11 +9,18 @@ defmodule Logflare.Logs.Validators.BigQuerySchemaChange do
   alias Logflare.SourceSchemas
   alias Logflare.Sources.Source
 
-  # Logflare injects "timestamp" as integer microseconds in LogEvent.mapper/1,
-  # but the BQ schema types it as TIMESTAMP (:datetime). Skip at the top level
-  # so the integer/datetime mismatch doesn't trip validation. "id" and
-  # "event_message" are strings on both sides, so they don't need the skip.
-  @skip_top_level_keys ~w(timestamp)
+  # Some top-level fields arrive as unix-time integers but the BQ schema types
+  # them as TIMESTAMP (:datetime), so a naive type check would always conflict:
+  #
+  #   - "timestamp" — injected as integer microseconds by LogEvent.mapper/1.
+  #   - "start_time" / "end_time" — injected as integer nanoseconds by
+  #     OtelTrace/OtelMetric and promoted to TIMESTAMP for OTEL sources in
+  #     SchemaBuilder.determine_type/3.
+  #
+  # Skip these at the top level only; nested fields with the same names still
+  # get validated. "id" and "event_message" are strings on both sides, so they
+  # don't need the skip.
+  @skip_top_level_keys ~w(timestamp start_time end_time)
 
   @spec validate(LE.t(), Source.t()) :: :ok | {:error, String.t()}
   def validate(%LE{body: _body}, %Source{validate_schema: false}), do: :ok
--- a/lib/logflare/logs/validators/bq_schema_change_validator.ex
+++ b/lib/logflare/logs/validators/bq_schema_change_validator.ex
@@ -1,6 +1,8 @@
 defmodule Logflare.Logs.Validators.BigQuerySchemaChange do
   @moduledoc false
 
+  import Logflare.Utils.Guards, only: [is_empty_map: 1]
+
   alias Logflare.BigQuery.SchemaTypes
   alias Logflare.Google.BigQuery.SchemaUtils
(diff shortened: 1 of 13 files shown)


lib/logflare/logs/validators/bq_schema_change_validator.ex
```diff
@@ -1,69 +1,107 @@
...
+    do: enforce_type(:datetime, key, schema_flat_map)
+
+  defp check_field(v, key, schema_flat_map) when is_list(v) do
+    case hd(v) do
+      head when is_map(head) ->
+        enforce_type(:map, key, schema_flat_map)
+        Enum.each(v, &walk_map(&1, key, schema_flat_map))
+
+      head ->
+        enforce_type({:list, SchemaTypes.type_of(head)}, key, schema_flat_map)
```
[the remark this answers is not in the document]
djwhitt: Fixed in fa074a981793ab4a1f5839ad971bb836735b81a5

test/logflare/google/bigquery/schema_utils_test.exs
```diff
@@ -4,6 +4,62 @@ defmodule Logflare.Google.BigQuery.SchemaUtilsTest do
...
+    end
+
+    test "atom keys are preserved" do
+      assert SchemaUtils.to_typemap(%{name: "x", count: 1}) == %{
+               name: %{t: :string},
+               count: %{t: :integer}
+             }
+    end
+
+    test "nested map keys are atomized at every level" do
```
Ziinc: Should not atomise keys, can lead to memory attacks

Ziinc: Probably want to look at whether we can have a full string map as the type map and if it will affect ingestion. Atoms don't get garbage collected so it's a red flag to dynamically generate it

Ziinc: fine for now since its already being done

lib/logflare/logs/validators/bq_schema_change_validator.ex
```diff
@@ -1,69 +1,113 @@
...
 
-  @spec validate(LE.t(), Source.t()) :: :ok | {:error, String.t()}
-  def validate(%LE{body: _body}, %Source{validate_schema: false}) do
-    :ok
-  end
+  # Logflare injects "timestamp" as integer microseconds in LogEvent.mapper/1,
+  # but the BQ schema types it as TIMESTAMP (:datetime). Skip at the top level
+  # so the integer/datetime mismatch doesn't trip validation. "id" and
+  # "event_message" are strings on both sides, so they don't need the skip.
+  @skip_top_level_keys ~w(timestamp)
```
Ziinc: some other otel-specific fields that are stored as TIMESTAMP are `start_time` and `end_time` whcih are also unix microseconds

test/logflare/google/bigquery/schema_utils_test.exs
```diff
@@ -4,6 +4,62 @@ defmodule Logflare.Google.BigQuery.SchemaUtilsTest do
 
   alias Logflare.Google.BigQuery.SchemaUtils
 
+  describe "to_typemap/1" do
```
Ziinc, line 8: we should also handle error cases, where lists are provided or strings are provided.
there are some error logs in prod for this.

Ziinc (approved):

lib/logflare/logs/validators/bq_schema_change_validator.ex
```diff
@@ -1,69 +1,113 @@
...
+  end
+
+  defp enforce_type(incoming, key, schema_flat_map) do
+    case Map.fetch(schema_flat_map, key) do
+      {:ok, ^incoming} ->
+        :ok
+
+      {:ok, existing} ->
+        raise(
+          "Type error! Field `#{key}` has type of `#{inspect(existing)}`. Incoming metadata has type of `#{inspect(incoming)}`."
```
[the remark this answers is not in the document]
Ziinc: we can just say that there is a type mismatch here, don't need to state the existing type

test/logflare/google/bigquery/schema_utils_test.exs
djwhitt: Two distinct sites:

- Per-event metadata atomization (`to_typemap(body)` on every event) — eliminated by this PR; `validate/2` now consumes only the cached string-keyed `schema_flat_map`.
- Schema build/refresh — `String.to_atom/1` still fires on every BQ field name in `to_typemap/2`. Slower-clocked but the input still derives from `IngestTransformer`-sanitized user keys, so still unbounded across tenants. Real concern.

Tracking the residual as a separate follow-up.

lib/logflare/logs/validators/bq_schema_change_validator.ex
djwhitt: Done in 2f25e3de6. Added both to `@skip_top_level_keys`.

Audited the rest of the codebase for similar TIMESTAMP promotions — `SchemaBuilder.determine_type/3` only promotes integers to TIMESTAMP for `start_time` and `end_time` on OTEL sources, so this should cover it. Tests added: top-level integers don't trip a `:datetime` schema entry, and nested fields with the same names still get validated.

lib/logflare/logs/validators/bq_schema_change_validator.ex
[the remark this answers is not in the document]
djwhitt: Done in `91024edf8`. Reduced to ``"Type error! Field `#{key}` has an unexpected type."`` — dropped both the schema-side type (the leak) and the incoming type (sender already has their own payload to inspect). Updated two nested-field regression tests that asserted on the dropped substrings.

test/logflare/google/bigquery/schema_utils_test.exs
djwhitt: Done in 24cc97ca8. Catch-all returning `%{}` handles the case the prod logs were likely showing: heterogeneous list-of-maps inside metadata. The previous branches both keyed off `hd(v)`, so `[%{}, "x", %{}]` crashed on the recursive `to_typemap("x")` and `["x", %{}]` silently dropped the map. Collapsed into a single list branch that scans for any map element and treats as REPEATED RECORD; non-map elements no-op via `Map.merge(_, %{})`.

Also cdebbbebd — the validator's own `walk_maps` had the same crash on heterogeneous list-of-maps inputs (it would surface as a rescued `BadMapError` rather than a typed error). That's the more directly relevant fix post-this-PR since `validate/2` walks `le.body` directly now and never invokes `to_typemap`. Same `Type error!` shape as the other validation failures.

PR #3491, opened 2026-05-18
fix: avoid stale timestamp in BERT ingestion test

msmithstubbs:
Test failed once the timestamp was more than 72 hours old.


Ziinc:
https://github.com/Logflare/logflare/pull/3493/changes
covered, 🙏

PR #3493, opened 2026-05-18
fix: make test use timestamp relative to now

nelsonmestevao:


test/logflare_web/controllers/log_controller_test.exs
```diff
@@ -482,7 +482,7 @@ defmodule LogflareWeb.LogControllerTest do
               "level" => "info",
               "message" => "bert body source_name test",
               "metadata" => %{},
-              "timestamp" => "2026-05-14T17:00:00Z"
+              "timestamp" => DateTime.to_iso8601(DateTime.utc_now())
```
nelsonmestevao, line 485: @Ziinc/ @amokan  I believe we need a date in the last 72h so I am using now 🙏

Ziinc (approved):

PR #3495, opened 2026-05-18
feat: Use pull mode in OtelExporter

bblaszkow06:
The PR allows testing of dev version of pull mode in otel exporter.


lib/logflare/backends/user_monitoring/ingest_pipeline.ex
```diff
@@ -0,0 +1,64 @@
...
+        concurrency: 1,
+        transformer: {__MODULE__, :transform, []}
+      ],
+      processors: [
+        default: [concurrency: System.schedulers_online()]
+      ],
+      batchers: [
+        default: [
+          concurrency: System.schedulers_online(),
+          batch_size: 10,
```
Ziinc: we can either do it on a per event basis without batchers, or do it with high batch volume. need to benchmark

Ziinc: in any case, 10 is very small, we can do larger batches if going with batcher

Ziinc:
demand should be based on the metric-tag permutation, so one metrric could have a ton of datapoints

Baishan:
Moved to #3546

PR #3499, opened 2026-05-19
 refactor: use format_timestamp/1 in event component

msmithstubbs:
Preparation for O11Y-1594 and #3500 

* Refactors Helpers.format_timestamp to accept custom format
* Remove format_timestamp from LogEventComponents and use Helpers.format_timestamp


Ziinc:
ah seems like i got the order wrong, looks good though 👍

PR #3500, opened 2026-05-19
feat: render unix timestamps in human readable format

msmithstubbs:
Renders timestamp field values in a human readable format.

* Add `FormattedTimestampComponent.formatted_timestamp/1` for rendering formatted timestamps
* `datetime` fields are rendered in the user's timezone, otherwise UTC.
* Tooltip shows timestamp is UTC format for reference 

Follows #3499 

Closes O11Y-1594

CleanShot 2026-05-20 at 10 12 25@2x

CleanShot 2026-05-20 at 10 14 09@2x

### Limitations

* Only fields identified as `datetime` in the source schema are formatted, such as the event timestamp.
* Search event modal currently only shows the search timezone if a user has saved it to their settings:

https://github.com/Logflare/logflare/blob/b8d6cbab1a53c3d8f99f9baef68d7d8ab2a6530f/lib/logflare_web/live/log_event_live/search_log_event_viewer_component.ex#L96-L100 

## Demo

https://github.com/user-attachments/assets/aaa52e61-4956-42a8-9edd-819965423d79

### DIFF
--- a/test/logflare_web/live/search_live/logs_search_lv_test.exs
+++ b/test/logflare_web/live/search_live/logs_search_lv_test.exs
@@ -5,10 +5,13 @@ defmodule LogflareWeb.Source.SearchLVTest do
   import Phoenix.LiveViewTest
 
   alias Ecto.Adapters.SQL
+  alias GoogleApi.BigQuery.V2.Model.TableSchema, as: TS
+  alias GoogleApi.BigQuery.V2.Model.TableFieldSchema, as: TFS
   alias Logflare.Backends
   alias Logflare.Google.BigQuery.SchemaUtils
   alias Logflare.SingleTenant
   alias Logflare.Sources.Source.BigQuery.Schema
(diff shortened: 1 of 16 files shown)


test/logflare_web/live/search_live/logs_search_lv_test.exs
```diff
@@ -1079,23 +1109,56 @@ defmodule LogflareWeb.Source.SearchLVTest do
...
+        }
+        |> then(&SchemaBuilder.build_table_schema(%{"testing" => "string"}, &1))
+
+      response_schema =
+        %TS{
+          fields: [
+            %TFS{name: "deployment_time", type: "TIMESTAMP"}
+          ]
+        }
+        |> then(&SchemaBuilder.build_table_schema(%{"testing" => "string"}, &1))
```
Ziinc, line 1126: i think we can probably move this schema generation stuff to test utils or a separate helper function for adding such timestamp columns to a schema

Ziinc (approved):

PR #3505, opened 2026-05-20
test: add Monaco editor e2e regression coverage

msmithstubbs:
Adds baseline test for Monaco editor in preparation for refactor.

## PR Stack
This is the first in a sequence of PRs that are intended to be merged in order.

See #3525 

or search 🔍  [is:pr O11Y-1543 sort:created-asc ](https://github.com/Logflare/logflare/pulls?q=is%3Apr+O11Y-1543+sort%3Acreated-asc)

Part of O11Y-1684 and O11Y-1543


.github/workflows/elixir-ci.yml
```diff
@@ -160,8 +162,31 @@ jobs:
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ needs.versions.outputs.node }}
+          cache: npm
+          cache-dependency-path: |
+            deps/phoenix_test_jsdom/priv/package-lock.json
+            assets/package-lock.json
+
+      - name: Install phoenix_test_jsdom Node dependencies
+        run: npm ci --prefix deps/phoenix_test_jsdom/priv
```
Ziinc: I think it would make sense for these to be separated out to an elixir-web-ci workflow that only triggers on change in the assets or logflare_web directory.

Ziinc: Otherwise this essentially extends our CI for all changes even if its a minor change under the lib/logflare

Ziinc: Alternatively we can do a specific jsdom tag for selectively running these tests

msmithstubbs: I was thinking with caching the npm install step would have minimal impact.  I'll put it in a separate workflow for now. Would be good to have a strategy for where the jsdom tests fit in between playwright and LIveViewTest.

msmithstubbs: Another thought: we could hard code the [list of events](https://github.com/Ziinc/phoenix_test_jsdom/blob/7f628b1f27b5d7b21dafbab4fd03df0f1b79c28c/lib/phoenix_test_jsdom/fire_event.ex#L67) so it doesn't call node at compile time.

Ziinc: i'll have another look at it next week 🤞 , removing it from compile time should be possible, can likely do it at packaging side...
i've merged in your cookie persistence PR btw will cut a release for it

msmithstubbs: Would you like me to look into doing this? Or could move the JSDOM tests to a parallel PR? Be good to keep the monaco editor refactor moving forward.

Ziinc (approved):

PR #3508, opened 2026-05-21
perf: rework clickhouse pipeline to batch by {event_type, day_bucket} with fresh/stale tiers

amokan:
## Why?

The ClickHouse ingest tables use `PARTITION BY toDate(timestamp)`, but Logflare's ingest pattern accepts timestamps spanning roughly ±72 hours. The current pipeline batches by `event_type` only, so a single 50K-row insert routinely lands across multiple day partitions — forcing one part per partition.

Data captured from a representative full day (_May 19_) on the `otel_logs_*` ingest table:
- Only 47.55% of inserts touched a single partition.
- 99.09% of events went to today's partition.
- But 46% of parts created that day landed in stale partitions as ~190-row slivers — useless to query performance, expensive to write, and a constant source of merge backlog.

The May 14 "by-source-name" materialized view turned this latent inefficiency into an acute one. MV cost scales linearly with partitions touched per insert (_~350ms per partition_):

| Metric                 | Pre-MV   | Post-MV (May 14) | Change     |
| ---------------------- | -------- | ---------------- | ---------- |
| `otel_logs` insert P50 | 441 ms   | 1,351 ms         | 3.06x      |
| `otel_logs` insert P90 | 1,129 ms | 2,859 ms         | 2.53x      |

The MV pushed the partition fan-out cost into the foreground, but the underlying part-dust pattern was already costing us merges and write amplification daily.

## What changed?

The core idea: compute each event's `day_bucket` and `ingest_freshness` once at `LogEvent.make/2`, then route batches by those labels in Broadway so every insert targets a single partition.

**LogEvent.make/2** (_lib/logflare/logs/log_event.ex_)

- Added two schema fields: `:day_bucket` (_integer, days since epoch_) and `:ingest_freshness` (_Ecto.Enum [`:fresh`, `:stale`]_).
- Reordered `make/2` so `TypeDetection.detect/1` runs first, then `mapper/2` threads event_type through `determine_timestamp/2`.
- Moved the trace start_time substitution out of the ClickHouse pipeline and into `determine_timestamp/2` via a tiny `maybe_use_trace_start_time/3` wrapper. Behavior unchanged; just one pass over the params instead of two.
- `make_from_db/2` passes `:log` explicitly to `mapper/2` — DB-loaded events never run freshness/trace logic, and that contract is now visible at the call site.

**Logflare.LogEvent.DayBucket** (_new, lib/logflare/logs/log_event/day_bucket.ex_)

A small GenServer that caches the current UTC day bucket in `:persistent_term` and refreshes shortly after midnight UTC.

The hot path is `:persistent_term.get/1` with a rescue fallback by design, so callers stay correct even before the GenServer seeds the cache and we avoid the heavier `:persistent_term.get/2`.

Public client functions:

- `current/0` — cached read
- `from_microseconds/1` — pure conversion (_accepts any integer; pre-epoch timestamps produce negative buckets_)
- `classify_freshness/1` — `:fresh` if within ±1 day of `current/0`, else `:stale`

**ClickHouse pipeline** (_lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex_)

- Split the single `:ch` batcher into `:ch_fresh` and `:ch_stale`, with different concurrency and flush windows:

|           | concurrency | batch_size | batch_timeout |
|-----------|-------------|------------|---------------|
|`:ch_fresh`| 4           | 75,000     | 5s            |
|`:ch_stale`| 2           | 75,000     | 30s           |

- `handle_message/3` routes by `ingest_freshness` and keys batches by `{event_type, day_bucket}` — every batch now targets a single ClickHouse partition
- Processor concurrency bumped 4 → 6.
- Deleted the in-pipeline `maybe_replace_inferred_timestamp/3` (_moved to `LogEvent.make/2`_).
- Telemetry and OTEL spans gained `batcher` and `day_bucket` attributes.

**DynamicPipeline scaling** (_lib/logflare/backends/adaptor/clickhouse_adaptor.ex_)

- `@scaling_threshold` bumped from 5,000 → 15,000. The old value gated scale-up at ~10% of one old batch (_50k_). With the new topology (_75k batches, 6 processors_), 5k pending was deep below saturation and triggered noisy scale-up/down churn. 15k matches the new per-pipeline drain rate.

## Expected Impact

Based on the captured May 19 data:

- ~8× reduction in total daily parts created across `otel_logs_*` (_3.23M → ~415k_).
- ~350× reduction in stale-partition parts (_eliminating the ~190-row dust pattern_).
- 30–40% lower P90 insert latency on the affected MV path.
- The MV remains incremental - no architecture change, just every MV invocation now runs the cheap single-partition code path.

## Validation

Run this query 24h after deploy ; pre-change baseline was 47.55% one-part. Post-deploy target is 100%:

```sql
SELECT parts_per_insert, count() AS num_inserts,
       round(100 * count() / sum(count()) OVER (), 2) AS pct
FROM (
    SELECT query_id, count() AS parts_per_insert
    FROM clusterAllReplicas('default', system.part_log)
    WHERE event_date = today() - 1
      AND event_type = 'NewPart'
      AND query_id != ''
      AND table = 'otel_logs_e2338e81_aed4_4de0_a5a0_358084c77b85'
    GROUP BY query_id
)
GROUP BY parts_per_insert
ORDER BY parts_per_insert;
```

Also worth watching post-deploy: insert P50/P90 on the `otel_logs_*` MV path (_expect ~30–40% P90 improvement_), merge backlog (_expect noticeable drop_), and the `:ch_stale` batcher's trigger mix in telemetry (`:size` vs `:timeout` — early days will be mostly `:timeout`, that's expected).

## Operational notes

- Stale events sit in the `:ch_stale` Broadway batcher for up to 30s. On pipeline crash, up to 2 × 75k = 150k stale events per pipeline would be dropped. Seems acceptable for stale data but worth noting.
- Memory: at peak each pipeline can hold ~450k events in-flight (_4 × 75k fresh + 2 × 75k stale_). At ~5 KB/event that's ~2.2 GB per pipeline; with up to 8 pipelines per scheduler count, ~17 GB node-wide ceiling. _Should_ fit current machine sizing?
- DayBucket GenServer is in the supervision tree for both test and prod paths. The refresh tick fires once per day; the `:persistent_term.put` triggers a global term-fingerprint update that's negligible at daily frequency.

## Testing

- `mix test test/logflare/log_event_test.exs` — 62 tests, including 4 new ones covering trace start_time substitution (_OTEL span path, start_time_unix_nano path, explicit-timestamp non-override, non-trace isolation_).
- `mix test test/logflare/log_event/day_bucket_test.exs` — new file, 3 doctests + 14 tests covering `current/0` (_cache + fallback_), from_microseconds/1 (_epoch / positive / pre-epoch / non-integer guards_), `classify_freshness/1` (_±1 boundary, far-past, far-future_).
- `mix test test/logflare/backends/adaptor/clickhouse_adaptor/pipeline_test.exs` — 20 tests; `handle_message/3` and `handle_batch/4` assertions rewritten for the new batcher + composite-key shape. The _"inferred timestamp replacement"_ describe block was deleted — that logic moved to `LogEvent.make/2` and is now tested at that layer.

## Benchmark

`MIX_ENV=test mix run test/profiling/log_event_make_bench.exs` against the pre-change M4 baseline: noise-level deltas (_±2-3% wall on most scenarios_).

No measurable per-event regression from the additional day_bucket + ingest_freshness work, since `DayBucket.current/0` is a `:persistent_term.get/1` (_~10–30ns_) and `from_microseconds/1` is a single `div/2`.


Baishan:
The batcher fork is pretty clever! 

` On pipeline crash, up to 2 × 75k = 150k stale events per pipeline get re-pulled from IngestEventQueue`

I think these events are popped from the ets tables so they would technically be lost if the broadway pipeline crashes?

Baishan (approved):
Nice

amokan:
> ` On pipeline crash, up to 2 × 75k = 150k stale events per pipeline get re-pulled from IngestEventQueue`
> 
> I think these events are popped from the ets tables so they would technically be lost if the broadway pipeline crashes?

Thats right. I have updated that note.

Ziinc (approved):

PR #3514, opened 2026-05-21
S3 Adaptor connection testing

bblaszkow06:
Part of O11Y-707

Probes connectivity, credentials, region, and bucket access by writing
a tiny sentinel parquet file at a fixed key in the bucket root using
the same `Explorer.DataFrame.to_parquet/3` path as ingest. Subsequent
probes overwrite the same key, bounding the artifact to one ~1KB file.


lib/logflare/backends/adaptor/s3_adaptor.ex
```diff
@@ -207,4 +226,12 @@ defmodule Logflare.Backends.Adaptor.S3Adaptor do
     |> Atom.to_string()
     |> String.replace("-", "_")
   end
+
+  defp fss_s3_config(backend_config) do
+    [
+      access_key_id: backend_config.access_key_id,
```
bblaszkow06, line 232:
```suggestion
    [
      endpoint: "https://storage.googleapis.com",
      access_key_id: backend_config.access_key_id,
```
FYI this is all I needed to test it with Google Cloud Storage

Ziinc: @bblaszkow06 lets allow endpoint to be customizable in a followup PR

Ziinc (approved):

PR #3517, opened 2026-05-22
feat: multi-source ingestion

Ziinc:
This allows source routing at the payload level, with `__LF_SOURCE` being provided as a special routing key.<br>This allows separate HTTP pipelines to be merged together into a multi-source request, while maintaining backwards compatability with query parameter source routing.

internally, at LogEvent.make/2 nothing changes. All changes are done at the controller level.

Pre-merge todos:

- [X] test locally
- [X] verify performance
- [X] docs update

Closes [O11Y-1838](https://linear.app/supabase/issue/O11Y-1838/support-multi-source-requests-in-ingest-controller)

here i am sending a multi-source payload to all sources named loadfest.

for benchmark of 500 events:

```
##### With input default #####
Name                                                                  ips        average  deviation         median         99th %
VerifyDeclaredSources - multi-source (1 source, 500 events)       16.80 K       59.51 μs    ±99.27%       53.04 μs      121.76 μs
VerifyDeclaredSources - passthrough (no __LF_SOURCE)              12.80 K       78.14 μs    ±34.03%       78.37 μs      122.00 μs

Comparison: 
VerifyDeclaredSources - multi-source (1 source, 500 events)       16.80 K
VerifyDeclaredSources - passthrough (no __LF_SOURCE)              12.80 K - 1.31x slower +18.63 μs

Reduction count statistics:

Name                                                        Reduction count
VerifyDeclaredSources - multi-source (1 source, 500 events)          9.63 K
VerifyDeclaredSources - passthrough (no __LF_SOURCE)              0.00400 K - 0.00x reduction count -9.62100 K

##### With input default #####
Name                                                              ips        average  deviation         median         99th %
Full ingest pipeline - single source (baseline)                1.67 K      598.03 μs  ±3055.95%      244.88 μs     3814.65 μs
Full ingest pipeline - multi source (1 declared source)        1.64 K      610.24 μs   ±621.85%      315.63 μs     2450.91 μs

Comparison: 
Full ingest pipeline - single source (baseline)                1.67 K
Full ingest pipeline - multi source (1 declared source)        1.64 K - 1.02x slower +12.21 μs

Name                                                            average  deviation         median         99th %
Full ingest pipeline - single source (baseline)                 28.36 K     ±0.00%        28.36 K        28.36 K
Full ingest pipeline - multi source (1 declared source)         51.70 K     ±0.00%        51.70 K        51.70 K

Comparison: 
Full ingest pipeline - single source (baseline)                 28.36 K
Full ingest pipeline - multi source (1 declared source)         51.70 K - 1.82x reduction count +23.35 K
```

there is very marginal performance impact, in the grand scheme of things it would speed things up as less http requests are made overall.

https://github.com/user-attachments/assets/90bf5082-f10c-4710-ba12-965a7b2bfe49

### DIFF
--- a/docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
+++ b/docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
@@ -95,6 +95,35 @@ To ingest by batch, send your request with the following JSON body:
 
 Note that if you have multiple sources with the same name, it will result in an error on ingestion and the log event will be discarded.
 
+### Multi-source batching (`__LF_SOURCE`)
+
+A single ingestion request can write different events from the same payload into **different sources** by setting the reserved field `__LF_SOURCE` on each event you want to route explicitly.
+
+- **Value**: The **source UUID** (the same identifier you use for the `source` query parameter or in the dashboard). It must be a valid UUID string.
+
+Example — two sources in one `batch`:
+
+```json
+{
+  "batch": [
+    {
+      "__LF_SOURCE": "9dd9a6f6-8e9b-4fa4-b682-4f2f5cd99da3",
+      "message": "to source A",
+    },
+    {
+      "__LF_SOURCE": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
+      "message": "to source B",
+    }
+  ]
+}
+```
+
+The same routing applies when POSTing a **bare JSON array** (array root with `application/json`), which maps to `_json`-style ingestion.
+
+:::note
+In multi-source batches, events that omit `__LF_SOURCE` fall back to the source from the **`source`** or **`source_name`** query string. Omitting both per-event routing and a query-parameter default returns **406** with ingestion errors.
+:::
+
 ## Pipeline Transformations
 
 At ingestion, the pipeline will perform the following in sequence:
--- a/lib/logflare_web/controllers/log_controller.ex
+++ b/lib/logflare_web/controllers/log_controller.ex
@@ -68,25 +68,87 @@ defmodule LogflareWeb.LogController do
     }
   )
 
-  def create(%{assigns: %{source: source}} = conn, %{"batch" => batch}) when is_list(batch) do
-    batch
-    |> Processor.ingest(Logs.Raw, source)
-    |> handle(conn)
+  def create(conn, %{"batch" => batch}) when is_list(batch) do
+    create_dispatch(conn, batch)
   end
 
-  def create(%{assigns: %{source: source}} = conn, %{"_json" => batch})
-      when is_list(batch) do
-    batch
-    |> Processor.ingest(Logs.Raw, source)
-    |> handle(conn)
+  def create(conn, %{"_json" => batch}) when is_list(batch) do
+    create_dispatch(conn, batch)
   end
 
-  def create(%{assigns: %{source: source}} = conn, _slog_params) do
-    conn.body_params
-    |> Map.drop(~w[timestamp id])
-    |> List.wrap()
-    |> Processor.ingest(Logs.Raw, source)
-    |> handle(conn)
+  def create(conn, _slog_params) do
+    event = Map.drop(conn.body_params, ~w[timestamp id])
+    create_dispatch(conn, [event])
+  end
+
+  @lf_source_key "__LF_SOURCE"
+
+  defp create_dispatch(conn, events) when is_list(events) do
+    default_source = Map.get(conn.assigns, :source)
+    multi_source? = Map.has_key?(conn.assigns, :declared_sources)
+
+    if not multi_source? and default_source != nil do
+      events
+      |> Processor.ingest(Logs.Raw, default_source)
+      |> handle(conn)
+    else
+      declared = Map.get(conn.assigns, :declared_sources, %{})
+      {grouped, errors} = group_events_by_source(events, declared, default_source)
+
+      results =
+        for {source, batch} <- grouped do
+          Processor.ingest(batch, Logs.Raw, source)
+        end
+
+      results
+      |> aggregate_results(errors)
+      |> handle(conn)
+    end
+  end
+
+  defp group_events_by_source(events, declared, default_source) do
+    {groups, errors} =
+      Enum.reduce(events, {%{}, []}, fn event, {groups, errors} ->
+        token = event_source_token(event)
+        stripped = Map.drop(event, [@lf_source_key, :"#{@lf_source_key}"])
+
+        cond do
+          is_binary(token) and is_map_key(declared, token) ->
+            source = Map.fetch!(declared, token)
+            {Map.update(groups, source, [stripped], &[stripped | &1]), errors}
+
+          is_binary(token) ->
+            {groups, ["event references unknown or unauthorized source #{token}" | errors]}
+
+          token != nil ->
+            {groups, ["invalid __LF_SOURCE value" | errors]}
+
+          default_source != nil ->
+            {Map.update(groups, default_source, [stripped], &[stripped | &1]), errors}
+
+          true ->
+            {groups, ["event missing __LF_SOURCE and no source query param" | errors]}
+        end
+      end)
+
+    grouped = for {source, batch} <- groups, do: {source, Enum.reverse(batch)}
+    {grouped, Enum.reverse(errors)}
+  end
+
+  defp event_source_token(event) when is_map(event) do
+    Map.get(event, @lf_source_key) || Map.get(event, :"#{@lf_source_key}")
+  end
+
+  defp aggregate_results(results, errors) do
+    {count, all_errors} =
+      Enum.reduce(results, {0, errors}, fn
+        {:ok, n}, {acc, errs} -> {acc + n, errs}
+        :ok, {acc, errs} -> {acc, errs}
+        {:error, more}, {acc, errs} when is_list(more) -> {acc, errs ++ more}
+        {:error, err}, {acc, errs} -> {acc, errs ++ [err]}
+      end)
+
+    if all_errors == [], do: {:ok, count}, else: {:error, all_errors}
   end
 
   operation :cloudflare, false
--- a/lib/logflare_web/controllers/log_controller.ex
+++ b/lib/logflare_web/controllers/log_controller.ex
@@ -137,7 +137,7 @@ defmodule LogflareWeb.LogController do
   end
 
   defp event_source_token(%{@lf_source_key => token}), do: token
-  defp event_source_token(%{:"__LF_SOURCE" => token}), do: token
+  defp event_source_token(%{:__LF_SOURCE => token}), do: token
   defp event_source_token(_), do: nil
 
   defp aggregate_results(results, errors) do
--- a/lib/logflare_web/controllers/log_controller.ex
+++ b/lib/logflare_web/controllers/log_controller.ex
@@ -108,45 +108,49 @@ defmodule LogflareWeb.LogController do
 
   defp group_events_by_source(events, declared, default_source) do
     {groups, errors} =
-      Enum.reduce(events, {%{}, []}, fn event, {groups, errors} ->
-        token = event_source_token(event)
-        stripped = Map.drop(event, [@lf_source_key, :"#{@lf_source_key}"])
+      for event <- events, reduce: {%{}, []} do
+        {groups, errors} ->
+          token = event_source_token(event)
+          stripped = Map.drop(event, [@lf_source_key, :"#{@lf_source_key}"])
 
-        cond do
-          is_binary(token) and is_map_key(declared, token) ->
-            source = Map.fetch!(declared, token)
-            {Map.update(groups, source, [stripped], &[stripped | &1]), errors}
+          cond do
+            is_binary(token) and is_map_key(declared, token) ->
+              source = Map.fetch!(declared, token)
+              {Map.update(groups, source, [stripped], &[stripped | &1]), errors}
 
-          is_binary(token) ->
-            {groups, ["event references unknown or unauthorized source #{token}" | errors]}
+            is_binary(token) ->
+              {groups, ["event references unknown or unauthorized source #{token}" | errors]}
 
-          token != nil ->
-            {groups, ["invalid __LF_SOURCE value" | errors]}
+            token != nil ->
+              {groups, ["invalid __LF_SOURCE value" | errors]}
 
-          default_source != nil ->
-            {Map.update(groups, default_source, [stripped], &[stripped | &1]), errors}
+            default_source != nil ->
+              {Map.update(groups, default_source, [stripped], &[stripped | &1]), errors}
 
-          true ->
-            {groups, ["event missing __LF_SOURCE and no source query param" | errors]}
-        end
-      end)
+            true ->
+              {groups, ["event missing __LF_SOURCE and no source query param" | errors]}
+          end
+      end
 
     grouped = for {source, batch} <- groups, do: {source, Enum.reverse(batch)}
     {grouped, Enum.reverse(errors)}
   end
 
-  defp event_source_token(event) when is_map(event) do
-    Map.get(event, @lf_source_key) || Map.get(event, :"#{@lf_source_key}")
-  end
+  defp event_source_token(%{@lf_source_key => token}), do: token
+  defp event_source_token(%{:"__LF_SOURCE" => token}), do: token
+  defp event_source_token(_), do: nil
 
   defp aggregate_results(results, errors) do
     {count, all_errors} =
-      Enum.reduce(results, {0, errors}, fn
-        {:ok, n}, {acc, errs} -> {acc + n, errs}
-        :ok, {acc, errs} -> {acc, errs}
-        {:error, more}, {acc, errs} when is_list(more) -> {acc, errs ++ more}
-        {:error, err}, {acc, errs} -> {acc, errs ++ [err]}
-      end)
+      for result <- results, reduce: {0, errors} do
+        {acc, errs} ->
+          case result do
+            {:ok, n} -> {acc + n, errs}
+            :ok -> {acc, errs}
+            {:error, more} when is_list(more) -> {acc, errs ++ more}
+            {:error, err} -> {acc, errs ++ [err]}
+          end
+      end
 
     if all_errors == [], do: {:ok, count}, else: {:error, all_errors}
   end
--- a/lib/logflare_web/controllers/log_controller.ex
+++ b/lib/logflare_web/controllers/log_controller.ex
@@ -147,8 +147,8 @@ defmodule LogflareWeb.LogController do
           case result do
             {:ok, n} -> {acc + n, errs}
             :ok -> {acc, errs}
-            {:error, more} when is_list(more) -> {acc, errs ++ more}
-            {:error, err} -> {acc, errs ++ [err]}
+            {:error, more} when is_list(more) -> {acc, Enum.reverse(more) ++ errs}
+            {:error, err} -> {acc, [err | errs]}
           end
       end
 
--- a/lib/logflare_web/controllers/plugs/buffer_limiter.ex
+++ b/lib/logflare_web/controllers/plugs/buffer_limiter.ex
@@ -23,6 +23,11 @@ defmodule LogflareWeb.Plugs.BufferLimiter do
   Checks buffer capacity and applies rate limiting based on source configuration.
   """
   @spec call(Plug.Conn.t(), opts()) :: Plug.Conn.t()
+  def call(%{assigns: %{declared_sources: declared}} = conn, _opts)
+      when map_size(declared) > 0 do
+    conn
+  end
+
   def call(%{assigns: %{source: %Source{} = source}} = conn, _opts) do
     if SystemCache.memory_utilization() >= @memory_limit or
          Backends.cached_local_pending_buffer_full?(source) do
--- a/lib/logflare_web/controllers/plugs/buffer_limiter.ex
+++ b/lib/logflare_web/controllers/plugs/buffer_limiter.ex
@@ -28,13 +28,15 @@ defmodule LogflareWeb.Plugs.BufferLimiter do
     if SystemCache.memory_utilization() >= @memory_limit do
       handle_buffer_full(conn, nil)
     else
-      Enum.reduce_while(declared, conn, fn {_token, source}, acc ->
-        if Backends.cached_local_pending_buffer_full?(source) do
-          {:halt, handle_buffer_full(acc, source)}
-        else
-          {:cont, acc}
-        end
-      end)
+      Enum.reduce_while(declared, conn, &reduce_declared_source/2)
+    end
+  end
+
+  defp reduce_declared_source({_token, source}, conn) do
+    if Backends.cached_local_pending_buffer_full?(source) do
+      {:halt, handle_buffer_full(conn, source)}
+    else
+      {:cont, conn}
     end
   end
 
--- a/lib/logflare_web/controllers/plugs/buffer_limiter.ex
+++ b/lib/logflare_web/controllers/plugs/buffer_limiter.ex
@@ -25,21 +25,35 @@ defmodule LogflareWeb.Plugs.BufferLimiter do
   @spec call(Plug.Conn.t(), opts()) :: Plug.Conn.t()
   def call(%{assigns: %{declared_sources: declared}} = conn, _opts)
       when map_size(declared) > 0 do
-    conn
+    if SystemCache.memory_utilization() >= @memory_limit do
+      handle_buffer_full(conn, nil)
+    else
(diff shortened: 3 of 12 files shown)


lib/logflare_web/controllers/plugs/verify_declared_sources.ex
```diff
@@ -0,0 +1,99 @@
...
+  alias LogflareWeb.Plugs.VerifyResourceAccess
+
+  @lf_source_key "__LF_SOURCE"
+
+  def init(_opts), do: nil
+
+  def call(%{assigns: %{resource_type: :source, user: %User{} = user}} = conn, _opts) do
+    case extract_events(conn.body_params) do
+      [first | _] = events
+      when is_map(first) and (is_map_key(first, :__LF_SOURCE) or is_map_key(first, "__LF_SOURCE")) ->
```
djwhitt, line 32: Could we make this more explicit by using a header to identify these? I worry about mistakes with the more implicit check on the first event.

lib/logflare_web/controllers/plugs/buffer_limiter.ex
```diff
@@ -20,6 +20,11 @@ defmodule LogflareWeb.Plugs.BufferLimiter do
   Checks buffer capacity and applies rate limiting based on source configuration.
   """
   @spec call(Plug.Conn.t(), opts()) :: Plug.Conn.t()
+  def call(%{assigns: %{declared_sources: declared}} = conn, _opts)
+      when map_size(declared) > 0 do
+    conn
```
djwhitt, line 28: Is the limiter bypass here intentional?

test/logflare_web/controllers/log_controller_test.exs
```diff
@@ -611,6 +611,181 @@ defmodule LogflareWeb.LogControllerTest do
...
+          Routes.log_path(conn, :create),
+          %{
+            "batch" => [
+              %{"__LF_SOURCE" => Atom.to_string(source_a.token), "event_message" => "with"},
+              %{"event_message" => "without"}
+            ]
+          }
+        )
+
+      assert json_response(conn, 406)
```
djwhitt, line 790: Is 406 right? Seems more like 400.

lib/logflare_web/controllers/log_controller.ex
```diff
@@ -66,25 +66,91 @@ defmodule LogflareWeb.LogController do
...
+
+  defp aggregate_results(results, errors) do
+    {count, all_errors} =
+      for result <- results, reduce: {0, errors} do
+        {acc, errs} ->
+          case result do
+            {:ok, n} -> {acc + n, errs}
+            :ok -> {acc, errs}
+            {:error, more} when is_list(more) -> {acc, errs ++ more}
+            {:error, err} -> {acc, errs ++ [err]}
```
djwhitt, line 151: Nit: pretty sure this O(N) in list length. Could prepend and reverse after.

lib/logflare_web/controllers/plugs/verify_declared_sources.ex
Ziinc: i think this is fine, as if any of the events do not have source declared, then the request will error out when we resolve the declared sources and no default source is provided

djwhitt: Makes sense, I didn't have the full context on this work when I made the comment too. I realized later the headers thing likely wouldn't work with Vector.

djwhitt:
@Ziinc Sorry, I saw you were working on this so I asked my 🤖 to take another look to see which of my earlier comments had been addressed so I could "resolve" them and it interpreted that as permission to push to the PR 🫠. I didn't really want to take time right now to see if what it had done was sensible, so I force pushed back to your last commit.

amokan (commented):
🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Verdict: don't approve yet.** 7 blocking findings, 5 minor, left inline.
>
> The blocking set clusters into three real problems:
>
> - **Events can bypass rate limiting and backpressure entirely.** `RateLimiter` and `BufferLimiter` only iterate `declared_sources`, but untagged events in the same request still go to `conn.assigns.source`, which neither plug checks in multi-source mode.
> - **`__LF_SOURCE` reaches the source schema.** The single-source fast path never strips it, so one request adds a permanent column to a source's BigQuery/ClickHouse schema.
> - **Two unscoped behaviour changes.** `VerifyDeclaredSources` is added to the entire `:ingest_api` pipeline though only `create/2` reads it (new 401s on ~20 other actions), and `handle({:error, _})` flips 406 → 400 for every ingest endpoint.
>
> Plus: multi-source mode keys off the *first* event only, and a partially-ingested batch returns 400, so a normal client retry double-writes.
>
> Line numbers are against `fe66e1c`. Each comment states the failure and proposes fixes with trade-offs — I have not applied any of them.

lib/logflare_web/controllers/plugs/verify_declared_sources.ex
amokan, line 32: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Blocking: multi-source mode is decided by the first event only.**
>
> This guard inspects `first` and nothing else, so a batch whose first event happens to lack `__LF_SOURCE` makes the whole plug a no-op — `conn` is returned unchanged, `declared_sources` is never assigned, and every `__LF_SOURCE`-tagged event behind it is written to the query-param source instead of its declared one. The stated use case is merging separate HTTP pipelines into one request, where event ordering is exactly what you cannot rely on.
>
> **Fix** — scan the batch rather than its head:
>
> ```elixir
> def call(%{assigns: %{resource_type: :source, user: %User{} = user}} = conn, _opts) do
>   events = extract_events(conn.body_params)
>
>   if Enum.any?(events, &declares_source?/1) do
>     resolve_and_verify(conn, events, user)
>   else
>     conn
>   end
> end
>
> defp declares_source?(event) when is_map(event),
>   do: is_map_key(event, :__LF_SOURCE) or is_map_key(event, "__LF_SOURCE")
>
> defp declares_source?(_), do: false
> ```
>
> If walking a 500-event batch twice is a concern on this path, bound it (`Enum.take(events, 50)`) — but then the bound needs documenting, because it reintroduces a smaller version of the same silent mis-route. Scanning the whole list is O(n) over a list already fully materialised by the JSON parser, so I would just do that.

lib/logflare_web/controllers/log_controller.ex
```diff
@@ -64,29 +65,96 @@ defmodule LogflareWeb.LogController do
...
+
+  @lf_source_key "__LF_SOURCE"
+
+  defp create_dispatch(conn, events) when is_list(events) do
+    default_source = Map.get(conn.assigns, :source)
+    multi_source? = Map.has_key?(conn.assigns, :declared_sources)
+
+    if not multi_source? and default_source != nil do
+      events
+      |> Processor.ingest(Logs.Raw, default_source)
```
amokan, line 94: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Blocking: the single-source fast path never strips `__LF_SOURCE`, so the key lands in the source schema.**
>
> `Map.drop(event, [@lf_source_key, :"#{@lf_source_key}"])` only happens inside `group_events_by_source/3` (line 116). Any request that skips the multi-source branch — `declared_sources` unassigned because the first event was untagged (see the `verify_declared_sources.ex` comment), or any of the ~20 sibling ingest actions — pipes events into `Processor.ingest` with the key intact. One such request permanently adds a `__LF_SOURCE` string column to that source's BigQuery/ClickHouse schema, and schema fields are not removable after the fact.
>
> **Fix** — strip unconditionally before dispatching, so neither branch can leak it:
>
> ```elixir
> defp create_dispatch(conn, events) when is_list(events) do
>   events = Enum.map(events, &Map.drop(&1, [@lf_source_key, :"#{@lf_source_key}"]))
>   ...
> ```
>
> and drop the now-redundant `stripped = ...` in `group_events_by_source/3`. The alternative — repeating the `Map.drop` in this branch — works but leaves two places to keep in sync, and this bug is what that duplication already cost.

lib/logflare_web/router.ex
```diff
@@ -107,6 +107,7 @@ defmodule LogflareWeb.Router do
     )
 
     plug(LogflareWeb.Plugs.FetchResource)
+    plug(LogflareWeb.Plugs.VerifyDeclaredSources)
```
amokan, line 110: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Blocking: this plug runs for the whole `:ingest_api` pipeline, but only `LogController.create/2` honours the assign.**
>
> Roughly 20 other actions route through `:ingest_api` — `generic_json`, `vector`, `cloudflare`, `vercel_ingest`, `syslog`, `elixir_logger`, `create_with_typecasts`, `browser_reports` and friends — and all of them still match `%{assigns: %{source: source}}`, ignoring `declared_sources` entirely. Two concrete regressions:
>
> 1. A payload to `/api/logs/json` whose first event carries a `__LF_SOURCE` UUID the caller does not own now returns **401** on a request that previously succeeded. `__LF_SOURCE` was just an ordinary field to those endpoints.
> 2. If the UUID *is* owned, `declared_sources` gets assigned, so `RateLimiter` and `BufferLimiter` switch to their multi-source clauses and stop guarding the source the controller actually ingests to.
>
> **Fix** — scope it to the one action that reads it:
>
> ```elixir
> plug(LogflareWeb.Plugs.VerifyDeclaredSources when action == :create)
> ```
>
> `plug/2` with a guard is only available inside a controller, so from the router either move this plug into `LogController` (alongside the other action-scoped plugs) or gate it internally on `conn.private[:phoenix_action] == :create`. Moving it into the controller is cleaner — the plug is a `LogController` concern and nothing else consults its output.

lib/logflare_web/controllers/plugs/rate_limiter.ex
```diff
@@ -10,7 +10,39 @@ defmodule LogflareWeb.Plugs.RateLimiter do
...
 
-  def call(%{assigns: %{user: user, source: source, plan: plan}} = conn, _opts \\ []) do
+  def call(conn, opts \\ [])
+
+  def call(
+        %{assigns: %{user: user, plan: plan, declared_sources: declared}} = conn,
+        _opts
+      )
+      when map_size(declared) > 0 do
+    Enum.reduce_while(declared, conn, fn {_token, source}, acc ->
```
amokan, line 20: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Blocking: in multi-source mode the query-param source is never rate-limited.**
>
> This reduce iterates `declared` only, but `create_dispatch/2` still routes every *untagged* event to `conn.assigns.source` (`log_controller.ex:129-130`). So a request carrying one tagged event plus 499 untagged ones passes rate limiting for the tagged source and then writes 499 events to the default source with no quota check at all. That is a straightforward way to bypass plan limits.
>
> **Fix** — verify the default source too when it is present:
>
> ```elixir
> def call(%{assigns: %{user: user, plan: plan, declared_sources: declared}} = conn, _opts)
>     when map_size(declared) > 0 do
>   sources =
>     declared
>     |> Map.values()
>     |> then(fn srcs ->
>       case Map.get(conn.assigns, :source) do
>         nil -> srcs
>         source -> [source | srcs]
>       end
>     end)
>     |> Enum.uniq_by(& &1.id)
>
>   Enum.reduce_while(sources, conn, fn source, acc -> ... end)
> end
> ```
>
> `uniq_by(& &1.id)` matters because `?source=<uuid>` and `__LF_SOURCE: <same uuid>` produce two distinct `%Source{}` structs — same root cause as the grouping-key comment on `log_controller.ex:121`.
>
> The alternative is to make the default source unavailable in multi-source mode (reject untagged events outright when `declared_sources` is set), which is a cleaner contract but a bigger behavioural decision than a limiter fix.

lib/logflare_web/controllers/plugs/buffer_limiter.ex
```diff
@@ -23,18 +23,39 @@ defmodule LogflareWeb.Plugs.BufferLimiter do
   Checks buffer capacity and applies rate limiting based on source configuration.
   """
   @spec call(Plug.Conn.t(), opts()) :: Plug.Conn.t()
+  def call(%{assigns: %{declared_sources: declared}} = conn, _opts)
+      when map_size(declared) > 0 do
+    if SystemCache.memory_utilization() >= @memory_limit do
+      handle_buffer_full(conn, nil)
+    else
+      Enum.reduce_while(declared, conn, &reduce_declared_source/2)
```
amokan, line 31: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Blocking: same gap as `RateLimiter` — the default source's buffer is never checked.**
>
> `reduce_declared_source/2` calls `cached_local_pending_buffer_full?/1` for declared sources only. Untagged events in the same request still go to `conn.assigns.source` (`log_controller.ex:129-130`), so they keep flowing into a source that is already backed up — precisely the overload this plug exists to prevent, and the memory-utilisation check above it won't catch a single-source backlog.
>
> **Fix** — mirror whatever shape the `RateLimiter` fix takes:
>
> ```elixir
> def call(%{assigns: %{declared_sources: declared}} = conn, _opts)
>     when map_size(declared) > 0 do
>   if SystemCache.memory_utilization() >= @memory_limit do
>     handle_buffer_full(conn, nil)
>   else
>     declared
>     |> Map.values()
>     |> then(fn srcs ->
>       case Map.get(conn.assigns, :source) do
>         nil -> srcs
>         source -> [source | srcs]
>       end
>     end)
>     |> Enum.uniq_by(& &1.id)
>     |> Enum.reduce_while(conn, &reduce_declared_source/2)
>   end
> end
> ```
>
> `reduce_declared_source/2` currently destructures `{_token, source}`, so it needs its head changed to take a bare source. Worth extracting the "sources this request will actually write to" computation into a shared helper the two plugs call, since getting it right in one and not the other is how this pair of bugs happened.

lib/logflare_web/controllers/log_controller.ex
```diff
@@ -64,29 +65,96 @@ defmodule LogflareWeb.LogController do
...
+      |> handle(conn)
+    else
+      declared = Map.get(conn.assigns, :declared_sources, %{})
+      {grouped, errors} = group_events_by_source(events, declared, default_source)
+
+      results =
+        for {source, batch} <- grouped do
+          Processor.ingest(batch, Logs.Raw, source)
```
amokan, line 102: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Blocking: partial ingestion is reported as total failure, so client retries duplicate data.**
>
> Every successfully-grouped batch is ingested here, and only then does `aggregate_results/2` (line 106) fold in the grouping errors and produce `{:error, _}` → HTTP 400. A 500-event batch containing one bad `__LF_SOURCE` therefore writes the other 499 events and *still* returns 400. Any client following normal retry-on-4xx semantics resends the whole batch and double-writes those 499.
>
> **Fix** — validate before any side effect:
>
> ```elixir
> case group_events_by_source(events, declared, default_source) do
>   {grouped, []} ->
>     grouped
>     |> Enum.map(fn {source, batch} -> Processor.ingest(batch, Logs.Raw, source) end)
>     |> aggregate_results([])
>     |> handle(conn)
>
>   {_grouped, errors} ->
>     handle({:error, errors}, conn)
> end
> ```
>
> All-or-nothing is the right default for a retryable endpoint: the client's retry becomes safe, and a malformed event is a caller bug they should fix rather than a partial write they have to reconcile.
>
> If you would rather not reject a 500-event batch over one bad row, the other option is to keep ingesting the good ones and report **207** with a per-event error list — but that only helps clients that actually parse the body, and it needs documenting in `docs/docs.logflare.com/docs/concepts/ingestion/index.mdx` before it is worth the complexity. What is not viable is the current combination: partial write plus a status that invites a full resend.

lib/logflare_web/controllers/log_controller.ex
```diff
@@ -244,7 +312,7 @@ defmodule LogflareWeb.LogController do
 
   defp handle({:error, errors}, conn) do
     conn
-    |> put_status(406)
+    |> put_status(400)
```
amokan, line 315: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Blocking: this changes 406 → 400 for every ingest path, not just multi-source.**
>
> `handle/2` is the shared terminal for `cloudflare`, `syslog`, `vercel_ingest`, `generic_json`, `create_with_typecasts` and the rest — they all funnel through this clause (e.g. line 307). On `main` this is `put_status(406)` (`main:245-250`), and it has been 406 for the life of the endpoint. Any client branching on 406 silently changes behaviour with no mention in the PR description or docs.
>
> **Fix** — return the new status only from the path that needs it and leave the shared clause alone:
>
> ```elixir
> defp handle({:error, errors}, conn) do
>   conn
>   |> put_status(406)
>   ...
> end
>
> defp handle({:error, :invalid_sources, errors}, conn) do
>   conn
>   |> put_status(400)
>   ...
> end
> ```
>
> with `aggregate_results/2` tagging grouping failures as `{:error, :invalid_sources, errors}`.
>
> If the intent really is to correct 406 across the board — 400 *is* the more defensible status for a malformed body — then it should be its own commit with a changelog note and a docs update, not a side effect of the multi-source work, so it can be reverted independently if a client breaks.

lib/logflare_web/controllers/plugs/rate_limiter.ex
```diff
@@ -10,7 +10,39 @@ defmodule LogflareWeb.Plugs.RateLimiter do
...
+      case Users.API.verify_api_rates_quotas(%{
+             user: user,
+             source: source,
+             plan: plan,
+             type: {:api_call, :logs_post}
+           }) do
+        {:ok, _} ->
+          {:cont, acc}
```
amokan, line 28: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Non-blocking: the multi-source success path drops the `x-rate-limit-*` response headers.**
>
> The single-source clause always calls `put_x_rate_limit_headers/2` on `{:ok, metrics}` (line 52-ish), but this branch returns `{:cont, acc}` and discards `metrics`. Clients that self-throttle on `x-rate-limit-source_remaining` get no signal on multi-source requests and will only discover the limit by taking a 429.
>
> **Fix** — carry the metrics through. With several sources in play the honest answer is the tightest one:
>
> ```elixir
> {:ok, metrics} ->
>   {:cont, put_min_rate_limit_headers(acc, metrics)}
> ```
>
> where the helper keeps whichever metrics have the lowest remaining count, so the headers describe the source that will throttle first. Simply calling `put_x_rate_limit_headers/2` on each iteration also works — last writer wins — but then the reported numbers depend on map iteration order, which is worse than not reporting at all.

lib/logflare_web/controllers/log_controller.ex
```diff
@@ -64,29 +65,96 @@ defmodule LogflareWeb.LogController do
...
+    {groups, errors} =
+      for event <- events, reduce: {%{}, []} do
+        {groups, errors} ->
+          token = event_source_token(event)
+          stripped = Map.drop(event, [@lf_source_key, :"#{@lf_source_key}"])
+
+          cond do
+            is_binary(token) and is_map_key(declared, token) ->
+              source = Map.fetch!(declared, token)
+              {Map.update(groups, source, [stripped], &[stripped | &1]), errors}
```
amokan, line 121: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Non-blocking: grouping is keyed on the `%Source{}` struct, which is not a stable identity.**
>
> `Sources.refresh_source_metrics_for_ingest/1` stamps `metrics.avg` read from `PubSubRates.Cache` at call time — once in `FetchResource` and again in `VerifyDeclaredSources`. If a request supplies both `?source=<uuid>` and `__LF_SOURCE: <same uuid>` and the cached rate ticks between those two reads, the two structs are unequal, `Map.update/4` creates two entries, and the same source gets two separate `Processor.ingest` calls — two pipeline dispatches and double telemetry spans for one logical target.
>
> **Fix** — key on the id and carry the struct in the value:
>
> ```elixir
> {Map.update(groups, source.id, {source, [stripped]}, fn {s, batch} -> {s, [stripped | batch]} end),
>  errors}
> ```
>
> then `for {_id, {source, batch}} <- groups, do: {source, Enum.reverse(batch)}` at line 137.
>
> Alternatively, strip the volatile field before using the struct as a key (`Map.put(source, :metrics, nil)`), but that couples this function to which fields happen to be volatile today — `source.id` is the actual identity and won't drift.

test/profiling/multi_source_ingest_bench.exs
```diff
@@ -0,0 +1,118 @@
...
+# Mimic stubs would not apply. Global mode mirrors ExUnit setups that stub GenServers.
+Mimic.set_mimic_global()
+
+Mimic.stub(Goth)
+
+Mimic.stub(Logflare.Sources.Source.Data, :get_log_count, fn _, _ ->
+  0
+end)
+
+Mimic.stub(Logflare.Backends, :ingest_logs, fn events, _ ->
```
amokan, line 26: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Non-blocking: this stub has the wrong arity, so the "full ingest pipeline" numbers measure real ingestion.**
>
> `Logflare.Backends.ingest_logs/4` is defined with default args (`lib/logflare/backends.ex:633`), and `Logflare.Logs.Processor` calls the 4-arity clause: `Backends.ingest_logs(batch, source, nil, true)` (`lib/logflare/logs/processor.ex:51`). Stubbing `fn events, _ -> ... end` replaces `ingest_logs/2`, which nothing on this path calls — Mimic matches on arity, so the real function runs and the benchmark silently exercises the unstubbed backend. The numbers in the PR description are not measuring what the label says.
>
> **Fix:**
>
> ```elixir
> Mimic.stub(Logflare.Backends, :ingest_logs, fn events, _source, _backend, _allow_spooling ->
>   {:ok, length(events)}
> end)
> ```
>
> Worth re-running and updating the PR description afterwards, since the point of the benchmark is to show multi-source grouping overhead in isolation and the current figures include whatever the backend did.

docs/docs.logflare.com/docs/concepts/ingestion/index.mdx
```diff
@@ -95,6 +95,35 @@ To ingest by batch, send your request with the following JSON body:
...
+- **Value**: The **source UUID** (the same identifier you use for the `source` query parameter or in the dashboard). It must be a valid UUID string.
+
+Example — two sources in one `batch`:
+
+```json
+{
+  "batch": [
+    {
+      "__LF_SOURCE": "9dd9a6f6-8e9b-4fa4-b682-4f2f5cd99da3",
+      "message": "to source A",
```
amokan, line 111: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Non-blocking: this example is not valid JSON, and the documented status code below is wrong.**
>
> Two problems in this block:
>
> 1. Both `"message"` values are followed by a trailing comma, so copy-pasting the example straight into a request body produces a parse error. Remove the trailing commas.
> 2. Line 124 states that omitting both per-event routing and a query-param default "returns **400**". It returns **401** — `VerifyResourceAccess` halts before the controller is reached, which is what this PR's own test asserts (`test/logflare_web/controllers/log_controller_test.exs`, "batch with no `__LF_SOURCE` and no query-param source returns 401").
>
> Also worth documenting here: multi-source mode is currently triggered by the **first** event in the batch only (see the `verify_declared_sources.ex` comment). If that behaviour is fixed as suggested, no doc note is needed; if it is kept deliberately, it must be stated, because it means a batch's routing depends on element order.

lib/logflare_web/controllers/plugs/verify_resource_access.ex
```diff
@@ -63,13 +63,36 @@ defmodule LogflareWeb.Plugs.VerifyResourceAccess do
...
   end
 
   # no resource is set, passthrough
   def call(conn, _), do: conn
 
+  @doc """
+  Checks that a user owns a source and (if an access token is present) that
+  the token's scopes cover ingestion for that source.
+  """
+  @spec verify_source_access(Source.t(), User.t(), String.t() | nil) :: boolean()
```
amokan, line 88: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Non-blocking: this `@spec` contradicts every caller.**
>
> The third parameter is typed `String.t() | nil`, but callers pass an `%OauthAccessToken{}` struct: `VerifyDeclaredSources` forwards `conn.assigns.access_token`, which `verify_api_access.ex:64` assigns as the token struct, and `check_resource/2` → `Auth.check_scopes/2` expects exactly that struct. The spec is a dialyzer contract violation and will mislead the next caller into passing a string.
>
> **Fix:**
>
> ```elixir
> @spec verify_source_access(Source.t(), User.t(), OauthAccessToken.t() | nil) :: boolean()
> ```
>
> with `alias Logflare.Auth.OauthAccessToken` at the top of the module.

PR #3519, opened 2026-05-22
perf: relax the max memory health check

Ziinc:
This PR relaxes the max memory threshold the load balancer will pull off the instance due to failing health checks. this helps to accomomdate memory spikes.
And if memory does not recover, then instance group will repair the instance.


Baishan (approved):
Does this affect scaling out or just health check?

Ziinc:
it affects just the health checks

PR #3524, opened 2026-05-25
refactor: extract email notification helpers from nested code

nelsonmestevao:


lib/logflare/sources/source/email_notification_server.ex
```diff
@@ -69,6 +48,42 @@ defmodule Logflare.Sources.Source.EmailNotificationServer do
...
   end
 
+  defp deliver_notifications(source, rate) do
+    notify_user(source, rate)
+    notify_other_emails(source, rate)
+    notify_team_users(source, rate)
+  end
+
+  defp notify_user(source, rate) do
+    if source.notifications.user_email_notifications do
```
Ziinc, line 58: we can use a match for this

lib/logflare/sources/source/email_notification_server.ex
```diff
@@ -69,6 +48,42 @@ defmodule Logflare.Sources.Source.EmailNotificationServer do
...
+      user = Users.Cache.get(source.user_id)
+      AccountEmail.source_notification(user, rate, source) |> Mailer.deliver()
+    end
+  end
+
+  defp notify_other_emails(source, rate) do
+    stranger_emails = source.notifications.other_email_notifications
```
Ziinc: match would be cleaner

lib/logflare/sources/source/email_notification_server.ex
```diff
@@ -69,6 +48,42 @@ defmodule Logflare.Sources.Source.EmailNotificationServer do
...
+
+    if stranger_emails do
+      stranger_emails
+      |> String.split(",")
+      |> Enum.each(&deliver_to_other(&1, rate, source))
+    end
+  end
+
+  defp deliver_to_other(email, rate, source) do
```
Ziinc: if this is only used in one location, then we don't need to extract it out

Ziinc (changes requested):

lib/logflare/sources/source/email_notification_server.ex
nelsonmestevao: I am not sure what you mean. On the function signature like so?


```suggestion
  defp notify_user(source = %{notifications: %{user_email_notifications: true}}, rate) do
```

Ziinc: yup that's what i mean, but just a small nit

Ziinc (approved):

PR #3525, opened 2026-05-26
feat: use new code editor component in QueryLive

msmithstubbs:
Replace the LiveMonacoEditor component in QueryLive with our own component wrapper.

Key differences:
* A simpler functional component instead of a live component
* The editor does not push events to LiveView. Instead, it syncs content with a hidden field that uses regular `phx-*` bindings to interact with the server.
* Expanded JavaScript unit testing of the hook. With no need for a live view process, these tests focus on the DOM interaction.  
* Monaco JavaScript dependencies vendored directly into the repository
* Update Monaco editor to latest version 0.55.1

CleanShot 2026-05-29 at 15 03 58@2x

https://github.com/user-attachments/assets/2f412e08-ad24-4bd1-8fd9-89e807033729

## Monaco editor is independent of LiveView

Untitled-2026-05-26-0715

The key difference in this architecture is that the Monaco editor doesn't send update events to the live view process. 

The Monaco editor updates a hidden textarea field and triggers a form-change event to the live view, matching the standard form-backed LiveView setup. The LiveView can update the form, and the Monaco editor will receive changes from the form field itself.

This simplifies the implementation and makes it much easier to test: we can test the Monaco DOM interactions with JS unit tests, and form interactions in LiveViewTest. 

## PR Stack
This is part of a sequence of PRs that are intended to be merged in order:

* #3505  
* #3525
* Isolated changes, can be merged in any order after the above
  * #3528 
  * #3529 
  * #3527 
* #3530  

or search 🔍  [is:pr O11Y-1543 sort:created-asc ](https://github.com/Logflare/logflare/pulls?q=is%3Apr+O11Y-1543+sort%3Acreated-asc)

Part of O11Y-1684 and O11Y-1543

### DIFF
--- a/assets/js/monaco_hook.js
+++ b/assets/js/monaco_hook.js
@@ -1,0 +1,212 @@
+import * as monaco from "monaco-editor/esm/vs/editor/edcore.main.js";
+import "monaco-editor/esm/vs/basic-languages/sql/sql.contribution.js";
+import {
+  registerLqlLanguage,
+  registerLqlCompletionProvider,
+} from "./lql_language";
+
+import { theme } from "./monaco_editor_theme";
+
+const minEditorHeight = 100;
+
+const MonacoHook = {
+  mounted() {
+    window.MonacoEnvironment = {
+      getWorkerUrl() {
+        return "/js/monaco_editor_worker.js";
+      },
+    };
+    window.monaco = monaco;
+
+    this.editorContainer = this.el.querySelector("[data-editor-container]");
+    this.formField = this.el.querySelector("[data-editor-input]");
+    this.loadDataset();
+    this.ignoreEditorChange = false;
+    this.disposables = [];
+
+    monaco.editor.defineTheme("default", theme);
+
+    const { minHeight: _minHeight, ...configuredOptions } = this.options;
+
+    this.editor = monaco.editor.create(this.editorContainer, {
+      ...configuredOptions,
+      language: this.language,
+      value: this.formField.value,
+    });
+
+    this.editorContainer.style.minHeight = `${this.minEditorHeight}px`;
+    this.resizeToContent();
+
+    if (this.language === "lql") {
+      this.registerLqlLanguage();
+      this.registerLqlCompletions();
+      this.registerLqlSubmitCommand();
+    } else {
+      this.registerConfiguredCompletions();
+    }
+
+    this.disposables.push(
+      this.editor.onDidChangeModelContent(() => {
+        if (this.ignoreEditorChange) return;
+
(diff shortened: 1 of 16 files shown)


lib/logflare_web/live/query_live.ex
msmithstubbs: Kept endpoints and alerts in assigns to match the current implementation but could just fetch  them from Cache if it is fast enough.

assets/js/monaco_hook.js
```diff
@@ -0,0 +1,212 @@
...
+  registerConfiguredCompletions() {
+    const editor = this.editor;
+
+    if (this.completions.length === 0) return;
+
+    const suggestions = this.completions;
+
+    this.disposables.push(
+      monaco.languages.registerCompletionItemProvider(this.language, {
+        provideCompletionItems(model, position) {
```
Ziinc: A separate arrow function passed to the obj argument would have less nesting and be cleaner

assets/js/monaco_hook.js
```diff
@@ -0,0 +1,212 @@
...
+      monaco.KeyCode.Enter,
+      () => {
+        this.syncFormField(this.editor.getValue());
+        this.submitClosestForm();
+      },
+      "!suggestWidgetVisible",
+    );
+  },
+
+  syncFormField(value) {
```
Ziinc, line 198: Adding some inline docs for these hook functions would be good for maintenance

Ziinc (changes requested):

assets/package.json
```diff
@@ -19,6 +19,7 @@
     "lodash": "^4.18.1",
     "luxon": "^3.2.1",
     "moment": "^2.29.4",
+    "monaco-editor": "^0.55.1",
```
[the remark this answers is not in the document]
msmithstubbs: 1. Tried bumping monaco to 0.56 but it also includes a dompurify warning.
2. We can't use `overrides` because monaco vendors dompurify.
3. The advisory states: 
>  Attackers can include payloads like `</noscript>` in attribute values to execute JavaScript when sanitized output is placed inside these unprotected rawtext contexts. 
[source](https://github.com/advisories/GHSA-v2wj-7wpq-c8vv)

Since we use a div, and none of the affected elements, I don't think this is a problem. 

@Ziinc please advise?

assets/package.json
```diff
@@ -19,6 +19,7 @@
     "lodash": "^4.18.1",
     "luxon": "^3.2.1",
     "moment": "^2.29.4",
+    "monaco-editor": "^0.56.0",
```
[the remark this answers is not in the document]
Ziinc: Hmm looks legit

PR #3534, opened 2026-05-28
refactor: Remove key atomization in `SchemaUtils.to_typemap/2`

ruslandoga:
Sort of related: https://erlef.org/blog/eef/atom-exhaustion -- but I'll also check if it helps perf.

<details><summary>Perf</summary>

```
Delta vs 7b2d6c250 (main-plan-update):
  noop / edge log
    ips:  10.5K ips → 34.2K ips  (+226.1%)
    wall: 67.13 → 26.58 µs  (-60.4%)
    mem:  37.7 KB → 35.7 KB  (-5.3%)
    reds: 3.5K → 2.9K  (-15.8%)
  noop / lists
    ips:  40.8K ips → 120.6K ips  (+195.4%)
    wall: 11.96 → 7.42 µs  (-38.0%)
    mem:  13.9 KB → 13.4 KB  (-3.7%)
    reds: 1.6K → 1.3K  (-14.1%)
  noop / scalars
    ips:  40.7K ips → 83.2K ips  (+104.3%)
    wall: 16.79 → 10.54 µs  (-37.2%)
    mem:  17.7 KB → 17.2 KB  (-2.9%)
    reds: 1.8K → 1.5K  (-16.4%)
  update / edge log
    ips:  16.2K ips → 29.9K ips  (+84.8%)
    wall: 49.13 → 32.08 µs  (-34.7%)
    mem:  42.3 KB → 40.7 KB  (-3.8%)
    reds: 3.8K → 3.2K  (-15.7%)
  update / lists
    ips:  56.6K ips → 109.1K ips  (+92.8%)
    wall: 12.08 → 8.17 µs  (-32.4%)
    mem:  15.1 KB → 14.5 KB  (-3.6%)
    reds: 1.7K → 1.4K  (-13.8%)
  update / scalars
    ips:  55.5K ips → 75.0K ips  (+35.3%)
    wall: 15.46 → 11.83 µs  (-23.5%)
    mem:  18.9 KB → 18.5 KB  (-2.3%)
    reds: 1.9K → 1.6K  (-16.7%)
```

</details>

### DIFF
--- a/lib/logflare/google/bigquery/schema_utils/schema_utils.ex
+++ b/lib/logflare/google/bigquery/schema_utils/schema_utils.ex
@@ -72,7 +72,9 @@ defmodule Logflare.Google.BigQuery.SchemaUtils do
 
   def struct_to_map(struct), do: struct |> Poison.encode!() |> Poison.decode!()
 
-  @spec to_typemap(TS.t() | [TFS.t()], from: :bigquery_schema) :: %{required(atom) => map | atom}
+  @spec to_typemap(TS.t() | [TFS.t()], from: :bigquery_schema) :: %{
+          required(String.t()) => map | atom
+        }
   def to_typemap(%TS{fields: fields} = schema, from: :bigquery_schema) when is_map(schema) do
     to_typemap(fields, from: :bigquery_schema)
   end
@@ -80,7 +82,7 @@ defmodule Logflare.Google.BigQuery.SchemaUtils do
   def to_typemap(fields, from: :bigquery_schema) when is_list(fields) do
     Map.new(fields, fn
       %TFS{fields: fields, name: n, type: t, mode: mode} ->
-        k = String.to_atom(n)
+        n = to_string(n)
 
         v =
           cond do
@@ -108,7 +110,7 @@ defmodule Logflare.Google.BigQuery.SchemaUtils do
             v
           end
 
-        {k, v}
+        {n, v}
     end)
   end
 
--- a/lib/logflare/source_schemas.ex
+++ b/lib/logflare/source_schemas.ex
@@ -161,22 +161,22 @@ defmodule Logflare.SourceSchemas do
   end
 
   defp typemap_to_json_schema({key, %{fields: fields, t: :map}}) do
-    {Atom.to_string(key), typemap_to_json_schema(fields)}
+    {to_string(key), typemap_to_json_schema(fields)}
   end
 
   defp typemap_to_json_schema({key, %{t: {:list, type}}}) do
-    {Atom.to_string(key), %{"type" => "array", "items" => %{"type" => Atom.to_string(type)}}}
+    {to_string(key), %{"type" => "array", "items" => %{"type" => Atom.to_string(type)}}}
   end
 
   defp typemap_to_json_schema({key, %{t: :datetime}}),
-    do: {Atom.to_string(key), %{"type" => "number"}}
+    do: {to_string(key), %{"type" => "number"}}
 
   defp typemap_to_json_schema({key, %{t: :integer}}),
-    do: {Atom.to_string(key), %{"type" => "number"}}
+    do: {to_string(key), %{"type" => "number"}}
 
   defp typemap_to_json_schema({key, %{t: :float}}),
-    do: {Atom.to_string(key), %{"type" => "number"}}
+    do: {to_string(key), %{"type" => "number"}}
 
   defp typemap_to_json_schema({key, %{t: type}}),
-    do: {Atom.to_string(key), %{"type" => Atom.to_string(type)}}
+    do: {to_string(key), %{"type" => Atom.to_string(type)}}
 end
--- a/test/logflare/google/bigquery/schema_utils_test.exs
+++ b/test/logflare/google/bigquery/schema_utils_test.exs
@@ -5,6 +5,38 @@ defmodule Logflare.Google.BigQuery.SchemaUtilsTest do
   alias Logflare.Google.BigQuery.SchemaUtils
   alias Logflare.TestUtils
 
+  describe "to_typemap/2" do
+    test "BigQuery schema field names are preserved as strings" do
+      assert %{"user" => %{"address" => %{"city" => "Dublin"}}, "tags" => ["a", "b"]}
+             |> TestUtils.build_bq_schema()
+             |> SchemaUtils.to_typemap(from: :bigquery_schema) == %{
+               "event_message" => %{t: :string},
+               "id" => %{t: :string},
+               "tags" => %{t: {:list, :string}},
+               "timestamp" => %{t: :datetime},
+               "user" => %{
+                 fields: %{"address" => %{fields: %{"city" => %{t: :string}}, t: :map}},
+                 t: :map
+               }
+             }
+    end
+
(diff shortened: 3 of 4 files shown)


test/logflare/google/bigquery/schema_utils_test.exs
```diff
@@ -32,36 +32,43 @@ defmodule Logflare.Google.BigQuery.SchemaUtilsTest do
...
 
-    test "arbitrary high bytes fall back to Latin1 and produce an atom" do
+    test "arbitrary high bytes fall back to Latin1 and produce a string" do
       raw = <<0xFF, 0xFE, 0xFD>>
       refute String.valid?(raw)
-      typemap = SchemaUtils.to_typemap(%{raw => "x"})
-      [normalized_key] = Map.keys(typemap)
-      assert is_atom(normalized_key)
-      assert String.valid?(Atom.to_string(normalized_key))
+      assert SchemaUtils.to_typemap(%{raw => "x"}) == %{"yþy" => %{t: :string}}
```
ruslandoga: Off-topic, but is this actually the desired behaviour? The produced string is very different.

```elixir
iex(1)> "yþy" <> <<0>>
#==> <<121, 195, 190, 121, 0>>
iex(2)> <<0xFF, 0xFE, 0xFD>>
#==> <<255, 254, 253>>
```

djwhitt:
I think it's worth cleaning up the dead `to_typemap/1` metadata path as part of this PR.

- The metadata variant (lines 75-118) and `decode_until_valid!/2` are unreachable from production code. The only caller that fed user-supplied keys into `to_typemap/1` was `BigQuerySchemaChange.validate`, and #3489 rewrote that to walk the body directly with `to_string/1` — no intermediate typemap, no atomization, no canonical form needed for keys. Invalid-UTF-8 keys in log payloads now just miss the `schema_flat_map` lookup and pass through as unknown fields, which is the same effective outcome the old code produced.
- A `git grep to_typemap` across `lib/` and `test/` confirms zero non-test callers of the metadata variant; the only consumer is `schema_utils_test.exs` itself.
- The new "yþy" assertion in this PR is documenting behavior of dead code.

Brief history for context on why the path is there in the first place:

- `85dfb6972` (Jan 2020) introduced `to_typemap/1` with `String.to_existing_atom`, assuming field-name atoms were already minted earlier in the pipeline.
- `9d7470435` (Sep 2020, "fix the schema utils to not existing atom bug") swapped that for `String.to_atom` after the assumption broke on novel payload fields — trading the crash for unbounded atom growth.
- `e08caec35` (Dec 2020, "fix the utf8 encoding bug") added `decode_until_valid!/2` after customers sent Latin1-encoded field names (the fixture in that commit is the Portuguese "Código de Rastreio" mojibake). `String.to_atom` requires valid UTF-8, so the decoder fallback + `Unicode.unaccent` existed solely to produce a string the atomizer would accept.
- `62aaa6ef0` (#3489, May 2026) rewrote the validator to skip the intermediate typemap entirely. That severed the only production call site that fed user-supplied keys into `to_typemap/1`, which means the whole "make an atom out of arbitrary bytes" chain no longer has a job to do.

Suggested steps to do this:

- Delete the `to_typemap/1` metadata clauses in `schema_utils.ex` (lines 75-118 on this branch).
- Delete `decode_until_valid!/2` (lines 120-136).
- Delete `not_empty_container?/1` (lines 138-148) if there are no other callers — looks like there aren't.
- In `schema_utils_test.exs`, drop the `describe "to_typemap/1"` block's metadata-input cases. Keep the BigQuery-schema test ("BigQuery schema field names are preserved as strings") since it covers the live variant.

ruslandoga:
I've opened a separate PR to remove `to_typemap/1`: https://github.com/Logflare/logflare/pull/3539

This PR will focus on removing atomization from `to_typemap/2`.

test/logflare/google/bigquery/schema_utils_test.exs
```diff
@@ -151,6 +151,38 @@ defmodule Logflare.Google.BigQuery.SchemaUtilsTest do
...
+               "tags" => %{t: {:list, :string}},
+               "timestamp" => %{t: :datetime},
+               "user" => %{
+                 fields: %{"address" => %{fields: %{"city" => %{t: :string}}, t: :map}},
+                 t: :map
+               }
+             }
+    end
+
+    test "atom BigQuery schema field names are converted to strings" do
```
ruslandoga, line 24: Not sure if it's possible for BigQuery schema to have atom names. It's not possible according to the struct spec https://hexdocs.pm/google_api_big_query/0.88.0/GoogleApi.BigQuery.V2.Model.TableFieldSchema.html#t:t/0, but I am not sure what might be happening to it before it reaches this function :)

lib/logflare/source_schemas.ex
```diff
@@ -152,22 +152,22 @@ defmodule Logflare.SourceSchemas do
   end
 
   defp typemap_to_json_schema({key, %{fields: fields, t: :map}}) do
-    {Atom.to_string(key), typemap_to_json_schema(fields)}
+    {to_string(key), typemap_to_json_schema(fields)}
```
ruslandoga, line 164: These `to_string/1` calls shouldn't be necessary, since field names are supposed to be strings already according to the spec https://hexdocs.pm/google_api_big_query/0.88.0/GoogleApi.BigQuery.V2.Model.TableFieldSchema.html#t:t/0. But I'm adding it just in case.

lib/logflare/google/bigquery/schema_utils/schema_utils.ex
```diff
@@ -147,15 +147,15 @@ defmodule Logflare.Google.BigQuery.SchemaUtils do
...
+  @spec to_typemap(TS.t() | list(TS.t()), keyword) :: %{required(String.t()) => map | atom}
   def to_typemap(%TS{fields: fields} = schema, from: :bigquery_schema) when is_map(schema) do
     to_typemap(fields, from: :bigquery_schema)
   end
 
   def to_typemap(fields, from: :bigquery_schema) when is_list(fields) do
     Map.new(fields, fn
       %TFS{fields: fields, name: n, type: t, mode: mode} ->
-        k = String.to_atom(n)
+        n = to_string(n)
```
ruslandoga, line 85: Same as https://github.com/Logflare/logflare/pull/3534/changes#r3341313425

ruslandoga:
@djwhitt I think this is ready for your review!

ruslandoga:
This is a small change but it could lead to silent data corruption if some execution path still uses atoms. I wonder how we should go about trying it out? And I wonder if https://github.com/Logflare/logflare/pull/3543 is related. CC @Ziinc

djwhitt (approved):
Approving. Audited every consumer of `to_typemap/2`'s output: the only place doing atom-specific key access was `SourceSchemas.typemap_to_json_schema/1` (`Atom.to_string(key)`), and this PR fixes exactly those clauses with `to_string/1`. Everything else either uses `to_string/1` already or operates on the flat typemap, which was always string-keyed.

On the silent-corruption concern: #3543's `enabled: false` flag keeps the newly-working schema verification dormant in prod, so even an overlooked atom path can't reject/corrupt events until that flag is flipped. Nicely de-risked alongside the #3539 dead-code removal.

PR #3536, opened 2026-05-28
fix: split ingest_logs telemetry counters per outcome

djwhitt:
## Summary

- The three `logflare.logs.ingest_logs.*` counters in `lib/telemetry.ex` had no explicit `event_name:`, so `Telemetry.Metrics.counter/2` collapsed them all onto `[:logflare, :logs, :ingest_logs]` and every emission incremented all three. Give each its own event name and emit reason-specific events from `do_telemetry/2`.
- Re-instrument `buffer_full` in `LogflareWeb.Plugs.BufferLimiter` alongside the 429 path. Its previous emission site was removed when `Logflare.Logs.Logs` was deleted, leaving the counter receiving only spurious increments from the shared-event bug.
- Split the unsuffixed `drop` counter into `drop_lql`, `drop_stale`, and `drop_future`. These are three distinct operational signals — user-configured filtering vs. late/replayed data vs. client clock skew — that `split_valid_events/2` was already differentiating in its warning log but folding into a single counter.

Downstream metrics consumers keyed on `logflare.logs.ingest_logs.*` will need to be updated: outcome counters now move independently, and the unsuffixed `drop` counter is replaced by the three per-reason counters (their sum equals what the old `drop` counter would have reported, had it not been broken).

## Test plan

- [x] `mix test test/logflare/backends_test.exs`
- [x] `mix test test/logflare_web/controllers/plugs/buffer_limiter_test.exs`
- [x] `mix lint.diff`
- [x] `mix test.typings`
- [ ] After deploy, confirm the three outcome counters diverge and `drop_lql + drop_stale + drop_future` matches the previously-reported drop total.

### DIFF
--- a/test/logflare/backends_test.exs
+++ b/test/logflare/backends_test.exs
@@ -867,11 +867,46 @@ defmodule Logflare.BackendsTest do
       start_supervised!({SourceSup, source})
       :timer.sleep(1000)
 
+      TestUtils.attach_forwarder([:logflare, :logs, :ingest_logs, :drop_lql])
+
       le = build(:log_event, message: "testing 123", source: source)
 
       assert {:ok, 0} = Backends.ingest_logs([le], source)
       assert [] = Backends.list_recent_logs_local(source)
+
+      source_id = source.id
+      source_token = source.token
+
+      assert_receive {:telemetry_event, [:logflare, :logs, :ingest_logs, :drop_lql], %{count: 1},
+                      %{source_id: ^source_id, source_token: ^source_token}}
+
       :timer.sleep(1000)
+    end
+
+    test "emits rejected telemetry for events with pipeline_error", %{user: user} do
+      source = insert(:source, user: user)
+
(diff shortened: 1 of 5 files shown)


test/logflare/backends_test.exs
```diff
@@ -867,13 +867,48 @@ defmodule Logflare.BackendsTest do
...
+
       :timer.sleep(1000)
     end
 
+    test "emits rejected telemetry for events with pipeline_error", %{user: user} do
+      source = insert(:source, user: user)
+
+      TestUtils.attach_forwarder([:logflare, :logs, :ingest_logs, :rejected])
+
+      le = %Logflare.LogEvent{
```
amokan: Nit. Could alias `Logflare.LogEvent` and `Logflare.LogEvent.PipelineError`

amokan (approved):
Looks good!

Only other thing I think we'd want to maybe add is a test for the batch aggregation since every test is currently doing a `count: 1`.

PR #3537, opened 2026-05-28
perf: use Enum.flat_map in OTel log and trace converters

djwhitt:
## Summary

- `Logs.OtelLog.handle_batch/2` and `Logs.OtelTrace.handle_batch/2` were walking `resource → scope → record/span` with `Enum.map(...) |> List.flatten()`, allocating deeply nested intermediate lists and then re-walking them to flatten. Replaced with `Enum.flat_map` at each intermediate level, mirroring the pattern already used by `Logs.OtelMetric`.
- Output shape is identical — pure refactor.
- Added unit tests in both `otel_log_test.exs` and `otel_trace_test.exs` that build a 2×2×2 (resources × scopes × records/spans) fixture and assert the returned batch length, guarding the flatten invariant that the existing single-record fixtures could not catch.

## Test plan

- [x] `mix test test/logflare/logs/otel_log_test.exs test/logflare/logs/otel_trace_test.exs` — 11 tests pass (4 OtelLog + 7 OtelTrace, including the 2 new ones)
- [x] New tests pass against the pre-refactor code (baseline) and the post-refactor code (proves equivalence)
- [x] `mix lint.diff` — clean

### DIFF
--- a/lib/logflare/logs/otel_log.ex
+++ b/lib/logflare/logs/otel_log.ex
@@ -21,14 +21,12 @@ defmodule Logflare.Logs.OtelLog do
   @behaviour Logflare.Logs.Processor
 
   def handle_batch(resource_logs, _source) when is_list(resource_logs) do
-    resource_logs
-    |> Enum.map(&handle_resource_logs/1)
-    |> List.flatten()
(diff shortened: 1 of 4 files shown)


amokan (approved):
Good work 👍

PR #3538, opened 2026-05-28
test: cover billing edit LiveView payment method actions

laraujo7:


djwhitt (approved):
Solid, accurate test coverage for the payment-method flows — happy paths plus the error branches (stripe failure, last-method guard) and a cross-account authz case. Spot-checked the assertions against the component and `Logflare.Billing`; they match real behavior. The `async: false` switch is required here (global Mimic for the cross-process Stripe expectations via DataCase.setup_mocking/1), not a regression.

Non-blocking follow-up: `sync` and `make-default` only have happy-path coverage, and unlike `delete`/`save` their `with` chains have no `else` clause — a Stripe error there would crash the LiveView rather than flash. Worth a future PR.

PR #3539, opened 2026-05-28
chore: Remove dead schema typemap conversion

ruslandoga:
Context: https://github.com/Logflare/logflare/pull/3534#issuecomment-4564137042


ruslandoga:
@djwhitt I think this is ready for your review!

lib/logflare/google/bigquery/schema_utils/schema_utils.ex
```diff
@@ -12,7 +12,7 @@ defmodule Logflare.Google.BigQuery.SchemaUtils do
   Transform BigQuery query response into a map.
   via https://stackoverflow.com/questions/53913182/decoding-googleapi-bigquery-object-elixir
   """
-  @spec deep_sort_by_fields_name(TFS.t()) :: TFS.t()
+  @spec deep_sort_by_fields_name(TS.t() | TFS.t()) :: TS.t() | TFS.t()
```
ruslandoga, line 15: Previous spec was incorrect but it wasn't being caught since `to_typemap/1` was erasing the type info.

lib/logflare/sources/source/bigquery/schema_builder.ex
```diff
@@ -123,7 +124,7 @@ defmodule Logflare.Sources.Source.BigQuery.SchemaBuilder do
 
 
   """
-  @spec build_table_schema([map()] | map(), TFS.t()) :: TFS.t()
+  @spec build_table_schema([map()] | map(), TS.t()) :: TS.t()
```
ruslandoga, line 127: Same as https://github.com/Logflare/logflare/pull/3539/changes#r3341185512

test/logflare/google/bigquery/schema_utils_test.exs
```diff
@@ -5,152 +5,6 @@ defmodule Logflare.Google.BigQuery.SchemaUtilsTest do
   alias Logflare.Google.BigQuery.SchemaUtils
   alias Logflare.TestUtils
 
-  describe "to_typemap/1" do
```
ruslandoga, line 8: Not sure if any of these tests should be rewritten with `to_typemap/2`.

ruslandoga:
I see a failing test

```

  1) test format_schema/3 json schema returns a valid empty result when bigquery_schema is non-TS data (Logflare.SourceSchemasTest)
Error:      test/logflare/source_schemas_test.exs:186
     ** (FunctionClauseError) no function clause matching in Logflare.Google.BigQuery.SchemaUtils.Mimic.Original.Module.to_typemap/2

     The following arguments were given to Logflare.Google.BigQuery.SchemaUtils.Mimic.Original.Module.to_typemap/2:

         # 1
         "not a TS struct"

         # 2
         [from: :bigquery_schema]

     code: assert SourceSchemas.format_schema(schema, :json_schema) == %{
     stacktrace:
       lib/logflare/google/bigquery/schema_utils/schema_utils.ex:76: Logflare.Google.BigQuery.SchemaUtils.Mimic.Original.Module.to_typemap/2
       (logflare 1.43.1) lib/logflare/source_schemas.ex:136: Logflare.SourceSchemas.format_schema/3
       test/logflare/source_schemas_test.exs:189: (test)
```

I'll check if it's actually reachable and if so, restore the previous empty map fallback.

---

Seems like it's reachable, I restored the fallback, but made it local to `format_schema/3` in https://github.com/Logflare/logflare/pull/3539/changes/0058511ca0c4b7ad9613d60b8d58717b93b1dbb0

lib/logflare/source_schemas.ex
```diff
@@ -132,8 +132,17 @@ defmodule Logflare.SourceSchemas do
...
 
   def format_schema(%SourceSchema{bigquery_schema: bq_schema}, :json_schema, to_merge) do
-    bq_schema
-    |> SchemaUtils.to_typemap()
+    typemap =
+      case bq_schema do
+        %GoogleApi.BigQuery.V2.Model.TableSchema{} = schema ->
+          SchemaUtils.to_typemap(schema, from: :bigquery_schema)
+
+        # Ecto.Term can contain malformed persisted or cached legacy values.
```
ruslandoga:
```suggestion
        # fallback to empty map in case of malformed persisted or cached legacy values.
```

djwhitt (approved):
Approving — clean, well-justified dead-code removal.

Verified:
- `git grep to_typemap` across the branch shows only the 2-arg `from: :bigquery_schema` form; `not_empty_container?/1` and `decode_until_valid!/2` have zero remaining references.
- The live `to_typemap/2` has no nil clause, but no caller passes nil — `source_schemas.ex` guards with a `case`, `bq_schema_to_flat_typemap(nil)` short-circuits, and all three `count_fields/1` sites pass a real `%TS{}` (a nil would already crash earlier in `deep_sort_by_fields_name/1`).
- Spec corrections are accurate: `build_table_schema/2` does return `initial_table_schema()` (`%Model.TableSchema{}`), and `deep_sort_by_fields_name` legitimately handles both TS and TFS.
- Live-variant coverage is retained transitively via the `bq_schema_to_flat_typemap/1` tests plus the new non-TS fallback test; the removed `describe "to_typemap/1"` block only exercised the dead metadata path.

CI green (Dialyzer + full suite). Two optional, non-blocking notes: (1) the `format_schema/3` `_ -> %{}` fallback is actually a small improvement over the old behavior, which would have raised on a nil `bigquery_schema`; (2) a direct unit test for the `from: :bigquery_schema` variant would make the string-key contract explicit, though indirect coverage is adequate.

PR #3546, opened 2026-06-02
feat: metrics pull clean

Baishan:
Switch to using the metric pull producer.


lib/telemetry.ex
```diff
@@ -94,12 +94,19 @@ defmodule Logflare.Telemetry do
...
-      # distribution("logflare.repo.query.decode_time", unit: {:native, :millisecond}),
+      # decode_time and queue_time/idle_time can be nil for certain query types
+      # (e.g. Oban internal queries that bypass the pool). OtelMetricExporter crashes on
+      # round(nil) when converting native time units, detaching the whole event handler.
+      # Default nil to 0: semantically correct (nil queue_time = no pool wait).
       distribution("logflare.repo.query.query_time", unit: {:native, :millisecond}),
-      distribution("logflare.repo.query.queue_time", unit: {:native, :millisecond}),
-      distribution("logflare.repo.query.idle_time", unit: {:native, :millisecond})
+      distribution("logflare.repo.query.queue_time",
+        measurement: fn m -> m[:queue_time] || 0 end,
```
Ziinc, line 103: Hmm we shouldn't need to do this here if we fix the 0 rounding error in the exporter

Ziinc (approved):

PR #3547, opened 2026-06-02
perf: Optimize RulesTree for many exact-matching rules

bblaszkow06:
Introduces `eq_index` into the tree that accumulates all the rules that expect an exact value at some path.

It also skips the tuple with a bitmask for rules that only have one filter.

Here are the gains in the benchmark included with this PR:
| Scenario | ips before | ips after | speedup | wall_med Δ |
|---|---:|---:|---:|---:|
| **drain 1000 match \| edge** | 89.3K | 6.01M | **67.3×** | 10.67 → 0.13 µs (-99%) |
| **drain 1000 match \| otel** | 89.9K | 2.24M | **24.9×** | 11.08 → 0.42 µs (-96%) |
| drain 100 miss \| edge | 861K | 7.58M | 8.8× | 1.17 → 0.13 µs (-89%) |
| drain 100 match \| edge | 853K | 6.30M | 7.4× | 1.17 → 0.13 µs (-89%) |
| drain 100 miss \| otel | 857K | 5.88M | 6.9× | 1.17 → 0.13 µs (-89%) |
| drain 100 match \| otel | 689K | 2.22M | 3.2× | 1.42 → 0.42 µs (-70%) |
| empty \| edge | 18.7M | 46.1M | 2.5× | 0.04 → 0.02 µs (-50%) |
| empty \| otel | 15.7M | 17.8M | 1.1× | flat |
| mixed 100 \| otel | 185K | 219K | 1.18× | 5.38 → 4.46 µs (-17%) |
| mixed 100 \| edge | 188K | 219K | 1.16× | 5.13 → 4.42 µs (-14%) |
| scattered 100 \| otel | 173K | 178K | 1.03× | -2% |
| regex only 100 \| edge | 33.0K | 33.5K | 1.01× | flat |
| regex only 100 \| otel | 33.2K | 33.2K | 1.00× | flat |
| scattered 100 \| edge | 244K | 240K | 0.98× | flat |

During implementation, I've also tested a variant using Event's `flattened_body` (and keeping exact match filters in a map next to the RulesTree). It was 2-10x times faster on the scattered scenario, but the same or slower (up to 25%) on the drain scenario, described in the Linear ticket as dominating.


djwhitt (approved):
Two non-blocking notes:
- The bare-int single-filter path adds a second target representation for a fairly modest secondary gain (~11% on top of eq_index itself). Reasonable, but it's the kind of complexity worth a comment for the next reader.
- Tightening `entry()` from `{operator(), [route()]}` to `{operator(), route()}` is a good correction of a pre-existing type inaccuracy, not just churn.

PR #3550, opened 2026-06-02
test: silence BigQuery IAM/schema noise from background mocks

nelsonmestevao:
Tests were logging (and occasionally failing on) Mimic errors caused by BigQuery calls. By globally stubbing `BigQueryAdaptor` calls to update_iam_policy/0,1 and patch_dataset_access/1 we reduce the noise.

### DIFF
--- a/test/logflare/source/bigquery/schema_test.exs
+++ b/test/logflare/source/bigquery/schema_test.exs
@@ -19,6 +19,10 @@ defmodule Logflare.Sources.Source.BigQuery.SchemaTest do
   end
 
   test "updates correctly" do
+    prev_config = Application.get_env(:logflare, Schema)
+    Application.put_env(:logflare, Schema, updates_per_minute: 1)
+    on_exit(fn -> Application.put_env(:logflare, Schema, prev_config) end)
+
     user = insert(:user)
     source = insert(:source, user: user)
     schema = TestUtils.default_bq_schema()
--- a/test/support/data_case.ex
+++ b/test/support/data_case.ex
@@ -20,20 +20,23 @@ defmodule Logflare.DataCase do
 
   using do
     quote do
-      alias Logflare.Repo
-      alias Logflare.TestUtils
-      alias Logflare.TestUtilsGrpc
-      alias Logflare.Backends.Adaptor.ClickHouseAdaptor
-      alias Logflare.Backends.IngestEventQueue
-      alias Logflare.PubSubRates
-      require TestUtils
+      use Mimic
+
+      require Logflare.TestUtils
 
       import Ecto
       import Ecto.Changeset
       import Ecto.Query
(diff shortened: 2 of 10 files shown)


test/logflare/source/bigquery/schema_test.exs
```diff
@@ -19,6 +19,10 @@ defmodule Logflare.Sources.Source.BigQuery.SchemaTest do
   end
 
   test "updates correctly" do
+    prev_config = Application.get_env(:logflare, Schema)
+    Application.put_env(:logflare, Schema, updates_per_minute: 1)
+    on_exit(fn -> Application.put_env(:logflare, Schema, prev_config) end)
```
nelsonmestevao, line 24: Force `updates_per_minute: 1` so the Schema throttle actually engages. With the test config's 900_000, the throttle window is ~0ms, so the second update raced past `expect(:bigquery_tables_patch, 1)` and crashed the GenServer. This way we avoid flakyness introduced by the crash.

test/support/data_case.ex
```diff
@@ -44,6 +47,10 @@ defmodule Logflare.DataCase do
           func.()
         end)
 
+        stub(BigQueryAdaptor, :update_iam_policy, fn -> :ok end)
```
amokan, line 50: Robot reviewer callout 🤖

> The blanket update_iam_policy stub breaks 4 existing tests that exercise the real code path (CI-confirmed). These tests were not given the call_original escape hatch that the controller tests got, so the stub swallows the call before it reaches the GoogleApi mocks, and their expects fail at verify_on_exit!. Failing in this PR's CI run: test/logflare/users/users_test.exs:47 (delete_user/1), test/logflare/partners_test.exs:80, and test/logflare/backends/bigquery_adaptor_test.exs:501 and :568 — the latter two are the unit tests of update_iam_policy/0 itself, which now test the stub rather than the adaptor.

amokan (changes requested):
I _think_ we should keep the shared stubs, but prob need to revisit some of the expects we removed from things like `users_test.exs` and `single_tenant_test.exs` unless I'm missing something bigger picture as these changes seem to cost us some coverage. Totally a fan of reducing noise, but my hope is that we can do so without sacrificing legit test cases 🙏 

Part of me also wonders if the blanket stub should be in the data case at all or if it would be better suited as a narrower opt-in for the test files that produce the most noise.

test/support/data_case.ex
nelsonmestevao: I fixed those in last push! :pray: 

The idea is to never think about the update_iam_policy noise in the logs and only do the call_original when we really are testing these.

cc/ @Ziinc that requested this https://github.com/Logflare/logflare/pull/3484#discussion_r3326043039

PR #3552, opened 2026-06-02
feat: Backends return Backends.QueryError struct for errors

msmithstubbs:
Add a standard `%QueryError{}` struct to be returned by backend adaptors on an unsuccessful query.  This allows flexibility on which error messages are logged and which are show to end users.

* Adds `%QueryError{}` to normalise the shape of errors returned by backend adaptors
* Updates BigQuery, Postgres, and Clickhouse adaptors to use QueryError when returning an error
* Adaptors set `QueryError.kind` based on the backend error message:
  * A missing field error would be `:invalid_query`
  * Other errors are `:backend_query` or `:connection_error`
* QueryError.log will log errors, excluding `:invalid_query`
  * User interface layer is not longer responsible for logging backend errors, but may still log user query errors
  * If a user has system_monitoring enabled the logged error messages are included
* User facing errors only include the backend error message for `:invalid_query`

CleanShot 2026-06-04 at 16 29 59@2x

### EndpointsLive

Slice

### QueryLive

CleanShot 2026-06-17 at 11 04 18@2x

### AlertsLive

Slice 2

Closes O11Y-1819 and O11Y-1970

### DIFF
--- a/lib/logflare/alerting.ex
+++ b/lib/logflare/alerting.ex
@@ -15,10 +15,10 @@ defmodule Logflare.Alerting do
   alias Logflare.Backends.Adaptor.SlackAdaptor
   alias Logflare.Backends.Adaptor.WebhookAdaptor
   alias Logflare.Backends.Adaptor.QueryResult
+  alias Logflare.Backends.QueryError
   alias Logflare.Cluster
   alias Logflare.Endpoints
   alias Logflare.Google.BigQuery.GCPConfig
-  alias Logflare.Google.BigQuery.GenUtils
   alias Logflare.Repo
   alias Logflare.Teams
   alias Logflare.TeamUsers.TeamUser
@@ -467,23 +467,14 @@ defmodule Logflare.Alerting do
            ) do
       {:ok, result}
     else
-      {:error, %Tesla.Env{body: body}} ->
-        decoded = Jason.decode!(body)["error"]
-
-        error =
-          decoded
-          |> GenUtils.process_bq_errors(alert_query.user_id)
-          |> case do
-            %{"message" => msg} -> msg
-            other -> other
-          end
-
+      {:error, %QueryError{} = error} ->
         Logger.error("Alert query execution failed with bad response",
           user_id: alert_query.user_id,
           alert_query_id: alert_query.id,
           alert_name: alert_query.name,
-          possible_reservation_error: BigQueryAdaptor.reservation_error?(decoded),
-          error_string: inspect(error)
+          possible_reservation_error: BigQueryAdaptor.reservation_error?(error.raw_error),
+          error_code: error.code,
+          error_string: inspect(error.message)
         )
 
         {:error, error}
--- a/lib/logflare/alerting.ex
+++ b/lib/logflare/alerting.ex
@@ -15,7 +15,6 @@ defmodule Logflare.Alerting do
   alias Logflare.Backends.Adaptor.SlackAdaptor
   alias Logflare.Backends.Adaptor.WebhookAdaptor
   alias Logflare.Backends.Adaptor.QueryResult
-  alias Logflare.Backends.QueryError
   alias Logflare.Cluster
   alias Logflare.Endpoints
   alias Logflare.Google.BigQuery.GCPConfig
@@ -23,6 +22,7 @@ defmodule Logflare.Alerting do
   alias Logflare.Teams
   alias Logflare.TeamUsers.TeamUser
   alias Logflare.User
+  alias Logflare.Utils.LoggerMetadata
 
   require Logger
   require OpenTelemetry.Tracer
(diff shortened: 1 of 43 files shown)


lib/logflare_web/query_error_helpers.ex
```diff
@@ -0,0 +1,126 @@
...
+      ...>   message: ~s|ERROR 42703 (undefined_column) column "notthere" does not exist\\n\\n    query: select notthere|,
+      ...>   code: :invalid_query,
+      ...>   raw_error: %Postgrex.Error{message: ~s|column "notthere" does not exist|},
+      ...>   backend: Logflare.Backends.Adaptor.PostgresAdaptor
+      ...> }
+      iex> LogflareWeb.QueryErrorHelpers.query_error_message(error)
+      ~s(Field "notthere" does not exist.)
+  """
+  @spec query_error_message(QueryError.t()) :: String.t() | nil
+  def query_error_message(%QueryError{
```
msmithstubbs: Error messages could be in the respective backend adaptors but thought it better to keep this in the UI layer.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -328,28 +329,48 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
+      backend: Logflare.Backends.Adaptor.ClickHouseAdaptor
+    }
+  end
+
+  defp to_query_error(%DBConnection.ConnectionError{message: message} = error) do
+    %QueryError{
+      message: message,
+      code: :connection_error,
+      raw_error: error,
+      backend: Logflare.Backends.Adaptor.ClickHouseAdaptor
```
```suggestion
      backend: __MODULE__
```

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -684,15 +685,31 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
...
-        {:error, err}
+  defp to_query_error(%{body: body}, user_id) do
+    with %{"error" => raw_error} <- Jason.decode!(body),
+         %{"message" => message} = processed_error <-
+           GenUtils.process_bq_errors(raw_error, user_id) do
+      %QueryError{
+        message: message,
+        code: :invalid_query,
+        raw_error: processed_error,
+        backend: Logflare.Backends.Adaptor.BigQueryAdaptor
```
Ziinc:
```suggestion
        backend: __MODULE__
```

lib/logflare/backends/adaptor/postgres_adaptor.ex
```diff
@@ -212,6 +224,34 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor do
     Map.put(config, :url, updated)
   end
 
+  @spec to_query_error(term()) :: QueryError.t()
+  defp to_query_error(%Postgrex.Error{} = error) do
+    %QueryError{
+      message: Exception.message(error),
+      code: :invalid_query,
+      raw_error: error,
+      backend: Logflare.Backends.Adaptor.PostgresAdaptor
```
Ziinc:
```suggestion
      backend: __MODULE__
```

lib/logflare/backends/adaptor/postgres_adaptor.ex
```diff
@@ -212,6 +224,34 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor do
...
+      message: message,
+      code: :connection_error,
+      raw_error: error,
+      backend: Logflare.Backends.Adaptor.PostgresAdaptor
+    }
+  end
+
+  defp to_query_error(%Ecto.QueryError{message: message} = error) do
+    %QueryError{
+      message: message,
```
Ziinc: i think we should restrict what is surfaced to users, in the event that sensitive values get surfaced. things like invalid columns etc are columns are fine, probably can match  on the message prefix and give a more generic message for the rest.

Ziinc: This applies for all backends, not just for postgres

Ziinc (changes requested):
needs some adjustment for surfacing of errors.
need some distinction between error messages:
- end consumer errors - needs to be generic and shouldn't leak information like underlying backend errors
- operator user  - should surface underlying backend errors, but logged out to Logger and surfaced to user via `system.logs` with the `user_id` metadata key set. 
- admin user - surfaced to admin via Logger, such as connection pool errors, which might not have a `user_id` associated, and needs the most detail, on par with operator user

msmithstubbs:
Make sense. 

> operator user - should surface underlying backend errors, but logged out to Logger and surfaced to user via system.logs with the user_id metadata key set.

How do we identify an operator user?

msmithstubbs:
Revised approach: 
1. Backend adaptors call Logger for any errors and return QueryError struct
2. Drop `message` attr from QueryError
3. UI layer doesn't log error failures, but handles taking a QueryError struct and determining correct message to show

msmithstubbs:
@Ziinc 
> * end consumer errors - needs to be generic and shouldn't leak information like underlying backend errors

Does this include API consumers? Example:

https://github.com/Logflare/logflare/blob/b1a389c4bbf26930968023b5479113a826f25369/test/logflare_web/controllers/endpoints_controller_test.exs#L266-L267

Should this just return a generic error message?

lib/logflare/backends/adaptor/postgres_adaptor.ex
```diff
@@ -212,6 +224,34 @@ defmodule Logflare.Backends.Adaptor.PostgresAdaptor do
     Map.put(config, :url, updated)
   end
 
+  @spec to_query_error(term()) :: QueryError.t()
+  defp to_query_error(%Postgrex.Error{} = error) do
+    %QueryError{
+      message: Exception.message(error),
```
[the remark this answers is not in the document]
msmithstubbs: Removed `QueryError.message` attr

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -328,28 +329,48 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
-        {:error, _} ->
-          {:error, "Error executing ClickHouse query"}
+        {:error, error} ->
+          {:error, to_query_error(error)}
       end
     end
   end
 
+  @spec to_query_error(term()) :: QueryError.t()
+  defp to_query_error(%Ch.Error{message: message} = error) when is_non_empty_binary(message) do
```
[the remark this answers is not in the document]
msmithstubbs: Removed `QueryError.message` attr

lib/logflare/backends/adaptor/bigquery_adaptor.ex
```diff
@@ -704,20 +705,80 @@ defmodule Logflare.Backends.Adaptor.BigQueryAdaptor do
...
+      {:error, error} ->
+        query_error =
+          error
+          |> to_query_error(user.id)
+          |> QueryError.log(
+            user_id: user.id,
+            bigquery_project_id: project_id
+          )
+
+        maybe_warn_reservation_error(query_error, user, project_id, query_opts)
```
msmithstubbs, line 717: This will be logged already by QueryError.log - do we need the specific wording of maybe_warn_reservation_error ? If not, we can remove it.

lib/logflare/alerting.ex
```diff
@@ -431,72 +431,51 @@ defmodule Logflare.Alerting do
...
-        Logger.error("Alert query execution failed with an unknown error",
-          user_id: alert_query.user_id,
-          alert_query_id: alert_query.id,
-          alert_name: alert_query.name,
-          error_string: inspect(error)
-        )
-
-        {:error, error}
-    end
+    LoggerMetadata.with_metadata(alert_query_logger_metadata(alert_query), fn ->
```
msmithstubbs, line 434: Logging now handled by backend adaptor, but note message ("Alert query execution failed with bad response") will have changed.

lib/logflare/backends/adaptor/bigquery_adaptor.ex
Ziinc: we can remove the duplication in a follow up PR

lib/logflare_web/controllers/api/backend_controller.ex
```diff
@@ -117,9 +117,9 @@ defmodule LogflareWeb.Api.BackendController do
           conn
           |> json(%{connected?: true})
 
-        {:error, reason} ->
+        {:error, _reason} ->
           conn
-          |> json(%{connected?: false, reason: reason})
+          |> json(%{connected?: false})
```
Ziinc, line 122: @msmithstubbs i think for the test connection pathway, we should still show the reason, otherwise users will not know why their connection is failing when adding a backend

lib/logflare/utils/logger_metadata.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Logflare.Utils.LoggerMetadata do
+  @moduledoc false
+
+  @spec with_metadata(Keyword.t(), (-> term())) :: term()
```
Ziinc, line 4: needs `@doc`

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -414,6 +405,43 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
+  @spec ch_query_error_kind(term()) :: QueryError.kind()
+  defp ch_query_error_kind(%Ch.Error{code: code}) when code in [47, 62], do: :invalid_query
+
+  defp ch_query_error_kind(%Ch.Error{message: message}) when is_binary(message) do
+    if message =~ "UNKNOWN_IDENTIFIER" or message =~ "SYNTAX_ERROR" do
+      :invalid_query
+    else
+      :backend_error
+    end
+  end
```
Ziinc, line 432: we will need to support more official [error codes](https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/ErrorCodes.cpp) that make sense so that there is better DX when people are crafting queries

Ziinc: we'll need to narrow them down to error codes relevant to general analytical queries

Ziinc (approved):

lib/logflare_web/controllers/api/backend_controller.ex
Ziinc: https://github.com/Logflare/logflare/pull/3624

msmithstubbs: Agreed - figured I'd drop it per the review feedback and pick it up in review.

lib/logflare/backends/adaptor/bigquery_adaptor.ex
msmithstubbs: #3627

lib/logflare/utils/logger_metadata.ex
msmithstubbs: #3628

PR #3557, opened 2026-06-04
fix+perf: removes DateTime usage from BQ hot code path, unix ns handling

Ziinc:
This PR properly handles unix nanosecond from recognized otel fields, while maintaining and refining the schema detection logic for otel metrics.
Fixes issue where schema detects timestamp fields as integer, and ingested fields are not correctly set to fractional second.
Fractional second ingestion for timestamp field allows us to remove the DateTime from hot code path, allowing some speedup.

After:
CleanShot 2026-06-04 at 15 25 32@2x


lib/logflare/google/bigquery/event_utils.ex
```diff
@@ -7,19 +7,44 @@ defmodule Logflare.Google.BigQuery.EventUtils do
...
     |> Map.put("event_message", body["event_message"])
+    |> convert_to_seconds()
+  end
+
+  @doc """
+  Recursively converts nanosecond/microsecond timestamps to seconds
+  https://docs.cloud.google.com/bigquery/docs/streaming-data-into-bigquery#send_date_and_time_data
+  """
+  @spec convert_to_seconds(map()) :: map()
+  def convert_to_seconds(%{"start_time" => ts} = data) when ts > 1_000_000_000_000_000_000 do
```
josevalim: I feel a bit uneasy about the recursive approach here because we are also recursive on the value itself. Perhaps something like this would be clearer:

```
  def convert_to_seconds(data) do
    data
    |> convert_to_seconds("start_time")
    |> convert_to_seconds("end_time")
    |> convert_to_seconds("timestamp_time")
  end

  defp convert_to_seconds(data, field) do
    case data do
      %{^field => ts} when ts > 1_000_000_000_000_000_000 ->
        %{data | field => div(ts, 1_000_000_000)}

      %{} ->
        data
    end
  end
```

But ideally you would call this explicitly on the cases you know it is in nanoseconds, so you can remove the guard. Thoughts?

josevalim (commented):
I added one comment. Big :+1: for keeping it as integers/floats given it is mostly used as a timestamp.

lib/logflare/google/bigquery/event_utils.ex
djwhitt: Suggestion on `convert_to_seconds/1`: flatten it to avoid recursion. Today each clause converts one field and recurses from the top, so a metric with all of `start_time`/`end_time`/`timestamp` makes ~4 passes re-checking already-converted fields. Touching each field once is faster:

```elixir
def convert_to_seconds(data) do
  data
  |> ns_to_seconds("start_time")
  |> ns_to_seconds("end_time")
  |> timestamp_to_seconds()
end

defp ns_to_seconds(data, key) do
  case data do
    %{^key => ts} when ts > 1_000_000_000_000_000_000 -> %{data | key => ts / 1_000_000_000}
    _ -> data
  end
end

defp timestamp_to_seconds(%{"timestamp" => ts} = data) when ts > 1_000_000_000_000_000_000,
  do: %{data | "timestamp" => ts / 1_000_000_000}
defp timestamp_to_seconds(%{"timestamp" => ts} = data) when ts > 1_000_000_000_000,
  do: %{data | "timestamp" => ts / 1_000_000}
defp timestamp_to_seconds(data), do: data
```

Output is identical across the realistic domain (integer ns/µs, the non-ns 16-digit case, floats); the only divergence is on impossible inputs (a `timestamp` ≥ ~10²⁷) where the recursion double-divides — arguably a latent bug, not worth preserving.

Rough local numbers (OTP 27, best-of-3 over 5M iters), relative only — tiny next to the `DateTime` removal:

| input | recursive | flat |
|---|---|---|
| plain log (`timestamp` µs only) | ~128 ns/op | ~70 ns/op |
| metric (all 3 ns fields) | ~239 ns/op | ~126 ns/op |

(`:math.pow(1000, 3)`/`:math.pow(1000, 2)` are fine as-is — the compiler constant-folds them to `1.0e9`/`1.0e6`; used plain integers above just for readability.)

lib/logflare/sources/source/bigquery/schema_builder.ex
```diff
@@ -240,7 +241,7 @@ defmodule Logflare.Sources.Source.BigQuery.SchemaBuilder do
   end
 
   defp otel_data?(params) do
-    Map.has_key?(params, "resource") and Map.has_key?(params, "scope")
+    TypeDetection.detect(params) in [:metric, :trace]
```
djwhitt: This splits "is `start_time`/`end_time` a timestamp" into two independent predicates: schema typing uses `TypeDetection.detect/1` here, on the un-converted `log_event.body`; value conversion (`convert_to_seconds/1`) uses field name + magnitude on a separate local copy in `le_to_bq_row/1`. When they disagree you reintroduce the integer/float mismatch this PR fixes — e.g. a plain log with an integer `start_time` in the ns range gets converted to a float, but `detect/1` calls it `:log`, so the column stays `INTEGER`. The old `resource` + `scope` gate applied to both paths.

Worth sharing one definition of "timestamp field" across both paths?

Ziinc: can do in followup

Ziinc: Addressed in this PR

lib/logflare/google/bigquery/event_utils.ex
```diff
@@ -3,23 +3,58 @@ defmodule Logflare.Google.BigQuery.EventUtils do
...
+    data
+    |> timestamp_to_seconds()
+    |> ns_to_seconds("start_time")
+    |> ns_to_seconds("end_time")
   end
 
+  defp ns_to_seconds(data, field) do
+    case data do
+      %{^field => ts} when is_integer(ts) and ts > @ns_threshold ->
+        %{data | field => ts / :math.pow(1000, 3)}
```
Baishan: Could precompute the :map.pow as constants

Baishan (approved):

lib/logflare/google/bigquery/event_utils.ex
Ziinc: done

PR #3558, opened 2026-06-04
chore: update logs playwright tests

Ziinc:
Updates playwright tests for e2e against supabase self-hosted due to logs components getting split out to separate compose file.

https://github.com/supabase/supabase/blob/master/docker/docker-compose.logs.yml

and also uses run.sh to add the separate override to the config file
https://github.com/supabase/supabase/blob/master/docker/run.sh

### DIFF
--- a/test/e2e/supabase/setup-supabase-services.sh
+++ b/test/e2e/supabase/setup-supabase-services.sh
@@ -50,10 +50,20 @@ fi
 cp .env.example .env
 endgroup
 
-log "add logs config..."
-sh .run.sh config add logs
+cd ../..
 
+[ ! "$GITHUB_ACTIONS" = "true" ] && log "Build logflare image..." && compose build analytics
+
+log "Pulling docker images..."
+compose pull
 endgroup
+
+log "Starting Supabase stack (inside docker containers)..."
+if ! compose up -d --wait --wait-timeout 180; then
+  if [ "$GITHUB_ACTIONS" = "true" ]; then
+    endgroup
+    echo -n "::group::"
+  fi
 
   exited=$(compose ps --all --format '{{.Service}} {{.State}}' | awk '$2 == "exited" || $2 == "dead" {print $1}')
 
--- a/test/e2e/supabase/setup-supabase-services.sh
+++ b/test/e2e/supabase/setup-supabase-services.sh
@@ -50,20 +50,10 @@ fi
 cp .env.example .env
 endgroup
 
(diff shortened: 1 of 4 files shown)


test/e2e/supabase/setup-supabase-services.sh
nelsonmestevao, line 1: These should change, right?

djwhitt:
Pushed two commits on top of this branch. Quick rundown of what and why:

**ffa452f38 — fix(e2e): keep stack startup when layering split-out logs compose file**

The `bin/compose` change here is the right fix — Supabase moved `analytics` (Logflare) and `vector` out of `docker-compose.yml` into `docker-compose.logs.yml`, and layering that override in via `-f` is what makes those services available to the e2e stack again. Confirmed locally: the stack comes up healthy with both services present and all 9 Logflare sources seeded.

I reverted the `setup-supabase-services.sh` changes, though. As written the script did not run: the edit left an unbalanced `fi` (so `bash -n` fails to parse it), and it dropped the `compose up -d --wait` that actually starts the stack. The `sh .run.sh config add logs` line also would not have helped — the file is `run.sh` (the leading dot fails under `set -e`), and `config add` only rewrites `COMPOSE_FILE=` in `.env`; it does not start anything. On top of that, `COMPOSE_FILE` is ignored whenever `docker compose` is invoked with explicit `-f` flags (which `bin/compose` does), so the override would not take effect regardless. And `run.sh config add` cannot register our e2e override anyway, since `docker-compose.e2e.yml` lives outside the `docker/` dir. So the explicit `-f` layering in `bin/compose` is sufficient on its own, and reverting the script restores the working build/pull/up flow.

**49ba0311b — fix(e2e): hold realtime channel open so "Billing metrics" log is emitted**

Separate, pre-existing flake I hit while validating the above. The realtime logs test asserts on realtime's periodic "Billing metrics" telemetry log, which its PromEx tenant poller emits every ~5s — but only while a tenant has a live connection. The setup subscribed a channel for just 2s and removed it before polling `realtime_logs` for up to 180s, so the short window rarely overlapped a poll tick and usually produced zero emissions. This change keeps the channel subscribed through the ingestion polling (torn down after the `Promise.all` resolves). It was likely masked before by CI's `retries: 2` plus the occasional lucky tick.

Verified locally against a full stack: the chromium logs suite goes from 4 failed / 3 skipped to **9 passed**.

djwhitt (approved):

PR #3559, opened 2026-06-04
fix(webhook): restore redacted secret headers on config update

jordienr:
## Problem

`redact_config/1` masks secret webhook headers (e.g. `Authorization`) as the literal string `"REDACTED"` when a backend is serialized, so API clients never receive the real value.

On update, the client echoes that sentinel back for unchanged headers. Because `cast_config/2` casts `headers` as a plain `:map`, the whole map is replaced with whatever is submitted — there is no within-map merge and no un-redaction. The result, depending on the client:

- echo `"REDACTED"` back → the stored secret is overwritten with the literal string `"REDACTED"`, silently breaking auth.
- strip the sentinel before sending → the header is dropped entirely.

Either way, editing any other header (or adding a new one) destroys the existing `Authorization` secret on a webhook / log drain destination.

## Fix

`cast_config/2` now un-redacts after casting: for each submitted header still set to the `"REDACTED"` sentinel, the existing stored value is swapped back in (and dropped if there is nothing to restore). This makes redaction symmetric with `redact_config/1`:

- unchanged secret (sentinel submitted) → original value preserved
- new value typed in → applied
- header added → added, secret preserved
- empty map submitted → headers cleared
- headers omitted → existing kept (unchanged)

## Tests

Added a `cast_config/2 header redaction round-trip` describe block covering the cases above.

Note: opened as a draft — I don't have an Elixir toolchain on this machine to run the suite locally, so leaving it for CI to validate.

### DIFF
--- a/lib/logflare/backends/adaptor/webhook_adaptor.ex
+++ b/lib/logflare/backends/adaptor/webhook_adaptor.ex
@@ -64,7 +64,7 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
   # still set to the sentinel we swap in the existing stored value (dropping it if
   # there is nothing to restore). Headers the user actually changed pass through.
   defp unredact_headers(changeset, existing_config) do
-    with headers when not is_nil(changeset) <- Ecto.Changeset.get_change(changeset, :headers),
+    with headers when not is_nil(headers) <- Ecto.Changeset.get_change(changeset, :headers),
          existing_headers <-
            Map.get(existing_config, :headers) || Map.get(existing_config, "headers") || %{} do
       restored =
--- a/lib/logflare/backends/adaptor/webhook_adaptor.ex
+++ b/lib/logflare/backends/adaptor/webhook_adaptor.ex
@@ -26,6 +26,9 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
 
   @behaviour Logflare.Backends.Adaptor
 
+  # Sentinel value substituted for secret header values by redact_config/1.
+  @redacted_value "REDACTED"
+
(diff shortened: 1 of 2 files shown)


lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -47,10 +50,45 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
...
+  # still set to the sentinel we swap in the existing stored value (dropping it if
+  # there is nothing to restore). Headers the user actually changed pass through.
+  defp unredact_headers(changeset, existing_config) do
+    case Ecto.Changeset.get_change(changeset, :headers) do
+      nil ->
+        changeset
+
+      headers ->
+        existing_headers =
+          Map.get(existing_config, :headers) || Map.get(existing_config, "headers") || %{}
```
Baishan: ` Map.get(existing_config, :headers, %{})`

Baishan (approved):
LGTM

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -47,10 +50,45 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
...
+
+        restored =
+          headers
+          |> Enum.reduce(%{}, fn
+            {key, @redacted_value}, acc ->
+              case Map.get(existing_headers, key) do
+                nil -> acc
```
Baishan: might be worth failing validation here? This would mean its redacted but it can't find the unredacted value and will omit the value?

Baishan:
Can run `mix format` to pass the ci check

PR #3562, opened 2026-06-04
Feat/metrics puller

Baishan:
- Reapply metrics puller.
- User new output format.

- [x] Update deps


lib/logflare/backends/user_monitoring.ex
```diff
@@ -56,6 +35,37 @@ defmodule Logflare.Backends.UserMonitoring do
...
+
+    exporter_spec =
+      {OtelMetricExporter,
+       [
+         name: @store_name,
+         metrics: metrics(),
+         pull_mode: true,
+         export_period: export_period,
+         extract_tags: &__MODULE__.extract_tags/2,
+         hibernate_after: 5_000,
```
Baishan, line 55: do we still need hibernate and fullsweep?

Ziinc, line 55: If we aren't mutating the genserver state then should be fine to remove/bump higher. Hibernate would still be good for situation where table empties faster than the next rotation comes.

Are we continuously rotating each time the table empties? Or is the rotation only done after the next min export timeout is reached?

lib/logflare/backends/user_monitoring/ingest_pipeline.ex
```diff
@@ -0,0 +1,72 @@
...
+        transformer: {__MODULE__, :transform, []},
+        concurrency: 1
+      ],
+      processors: [
+        default: [concurrency: System.schedulers_online()]
+      ],
+      batchers: [
+        default: [
+          concurrency: System.schedulers_online(),
+          batch_size: 200,
```
Ziinc: Batch sizes of 500-750 should be fine too

lib/telemetry.ex
```diff
@@ -94,12 +94,19 @@ defmodule Logflare.Telemetry do
...
-      # distribution("logflare.repo.query.decode_time", unit: {:native, :millisecond}),
+      # decode_time and queue_time/idle_time can be nil for certain query types
+      # (e.g. Oban internal queries that bypass the pool). OtelMetricExporter crashes on
+      # round(nil) when converting native time units, detaching the whole event handler.
+      # Default nil to 0: semantically correct (nil queue_time = no pool wait).
       distribution("logflare.repo.query.query_time", unit: {:native, :millisecond}),
-      distribution("logflare.repo.query.queue_time", unit: {:native, :millisecond}),
-      distribution("logflare.repo.query.idle_time", unit: {:native, :millisecond})
+      distribution("logflare.repo.query.queue_time",
+        measurement: fn m -> m[:queue_time] || 0 end,
```
Ziinc, line 103: Didn't we fix the issue on nil measurements?

Ziinc (approved):

lib/logflare/backends/user_monitoring.ex
Baishan: it rotates after each export timeout. I'm only emptying the previous generation (the ones before current)

lib/logflare/backends/user_monitoring/ingest_pipeline.ex
Baishan: I think this needs to be much higher 10000. Each element is a datapoint now

lib/telemetry.ex
Baishan: No can do it separately, im not sure where it comes from.

PR #3567, opened 2026-06-05
fix(http): prevent duplicate transport-owned headers

djwhitt:
## Problem

When an HTTP backend is configured with a user-supplied `Content-Type` header, requests went out with **two** `Content-Type` headers (e.g. `application/jsonapplication/json`). Tesla middleware set the header by appending, and `Tesla.put_headers/2` appends rather than de-duplicates, so the user's header survived alongside the transport's. Receivers concatenated the two values, failed to parse the body, and delivered an empty body.

Two paths exhibited this:

- **Webhook-derived backends** (Datadog, Loki, Elastic, Incident.io, …): `Tesla.Middleware.JSON` appends `content-type` when it encodes the body.
- **OTLP backends**: `OtlpAdaptor` passes user-configured headers alongside `ProtobufFormatter`, which always sets `content-type` to `application/x-protobuf`. A user-supplied `content-type` survived next to the formatter's.

## Fix

Give transport-owned headers a single source rather than de-duplicating by header order (which is non-deterministic).

Shared plumbing:

- New `Logflare.Backends.Adaptor.HttpBased.Headers` with `drop_reserved/2` (case-insensitive strip) and `normalize_keys/1`.

Webhook path:

- `WebhookAdaptor.Client.send/1` strips the reserved set per request: `content-type` only when the JSON middleware will actually encode the body — binary payloads (e.g. NDJSON) keep a custom content-type — and `content-encoding` when gzip is enabled. `authorization` is intentionally not reserved here, since several adaptors set it directly.
- `WebhookAdaptor.cast_config/2` lowercases header keys so stored config cannot hold case-variant duplicates of the same header.
- The redacted-secret restore path looks up existing values by normalized key, so a stored secret is not dropped when its casing differs from the submitted (possibly normalized) key.

HTTP-based / OTLP path:

- `HttpBased.Client.new/1` strips `content-encoding` (gzip) and `authorization` (token/basic auth).
- The body is not known at client-build time, so the client cannot infer whether the JSON middleware will own `content-type`. Instead, a formatter that sets `content-type` itself declares ownership via an optional `reserved_headers/0`, and the client drops those names from user headers regardless of body. `OtlpAdaptor.ProtobufFormatter` and `SentryAdaptor.EnvelopeBuilder` declare `["content-type"]`.
- `OtlpAdaptor.cast_config/2` lowercases header keys, matching the webhook path.

### DIFF
--- a/lib/logflare/backends/adaptor/http_based/client.ex
+++ b/lib/logflare/backends/adaptor/http_based/client.ex
@@ -6,6 +6,7 @@ defmodule Logflare.Backends.Adaptor.HttpBased.Client do
 
   alias Logflare.LogEvent
   alias Logflare.Backends.Adaptor.HttpBased.EgressTracer
+  alias Logflare.Backends.Adaptor.HttpBased.Headers
   alias Logflare.Backends.Adaptor.HttpBased.LogEventTransformer
   alias Logflare.Backends.Backend
   alias Logflare.Sources.Source
@@ -50,6 +51,8 @@ defmodule Logflare.Backends.Adaptor.HttpBased.Client do
   """
   @spec new(opts()) :: t()
   def new(opts \\ []) do
+    headers = Headers.drop_reserved(opts[:headers] || %{}, reserved_header_names(opts))
+
     Tesla.client(
       [
         Tesla.Middleware.Telemetry,
@@ -57,7 +60,7 @@ defmodule Logflare.Backends.Adaptor.HttpBased.Client do
         opts[:query] && {Tesla.Middleware.Query, opts[:query]},
         opts[:token] && {Tesla.Middleware.BearerAuth, token: opts[:token]},
         opts[:basic_auth] && {Tesla.Middleware.BasicAuth, opts[:basic_auth]},
-        headers_middleware(opts[:headers]),
+        headers_middleware(headers),
         Keyword.get(opts, :formatter, LogEventTransformer),
         Keyword.get(opts, :json, true) && Tesla.Middleware.JSON,
         opts[:gzip] && {Tesla.Middleware.CompressRequest, format: "gzip"},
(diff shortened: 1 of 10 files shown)


lib/logflare/backends/adaptor/http_based/client.ex
```diff
@@ -68,6 +71,35 @@ defmodule Logflare.Backends.Adaptor.HttpBased.Client do
...
+  # Header names the middleware will set, so they must be dropped from
+  # user-supplied headers to avoid duplicates (see `Headers.drop_reserved/2`).
+  #
+  # `content-type` cannot be inferred from `Tesla.Middleware.JSON` here: the request
+  # body is not available at client-build time, so we cannot tell whether the JSON
+  # middleware will encode it and own the header. Instead, a formatter that sets
+  # `content-type` itself (e.g. the protobuf/envelope formatters) declares ownership
+  # via an optional `reserved_headers/0`, and we drop those names regardless of body.
+  @spec reserved_header_names(opts()) :: [String.t()]
+  defp reserved_header_names(opts) do
```
amokan, line 92: :robot:

> reserved_header_names/1 in http_based/client.ex deliberately omits content-type for the default JSON path, delegating to the formatter's reserved_headers/0. Today that's fine because no consumer uses the default LogEventTransformer formatter and forwards user headers. But this is the one place the original bug could silently reappear: a future adaptor with json: true (default), the default formatter, and user-supplied headers containing content-type would emit a duplicate again. The comment explains the why well, but consider adding a regression test asserting that path stays safe (or a note in the Adaptor behaviour docs) so the invariant "forward user headers ⇒ your formatter must declare content-type" is discoverable, not tribal.

amokan (approved):
Just one 🤖 comment that isn't terrible - but not a hard block there by any means. Everything else looks good!

PR #3569, opened 2026-06-08
perf: pass event ids through broadway pipeline instead of full structs

Baishan:
We can optimize memory usage by not copying events from ets into the broadway pipeline. This results in copies from ets to process heap for items queued up in the buffers. 

This pr optimizes this by passing along ids and dereferencing them only when we need to handle a batch. 
During testing this resulted in a 20% reduction in process memory. 

- Mark events as :processing instead of :ingested to prevent queue janitor from purging them.
- Events are marked as :ingested or deleted when acked depending on metrics.
- Ets lookup handles inside handle_batch
- Queue janitor takes snapshots of processing ids every 10 seconds to purge any that fail to process.

Closes O11Y-1968


lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -135,14 +136,24 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
 
       source ->
-        for %{data: le} <- successful do
-          # emit telemetry on event
-          emit_event_telemetry(queue, source, le, backend_metadata)
+        metrics = Sources.get_source_metrics_for_ingest(source.token)
+
+        for %{data: {id, tid}} <- successful do
+          case :ets.lookup(tid, id) do
+            [{^id, _status, le}] -> emit_event_telemetry(queue, source, le, backend_metadata)
```
Baishan: Can optimize this by moving the calculation to handle_batch and passing along the message sizes. I'd prefer to keep this as a minor followup if theres no issues.

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -470,6 +470,38 @@ defmodule Logflare.Backends.IngestEventQueue do
     end
   end
 
+  @doc """
+  Takes pending item IDs from a given table, marking them as `:ingested` in-place.
```
amokan:
```suggestion
  Takes pending item IDs from a given table, marking them as `:processing` in-place.
```

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -220,6 +240,21 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
+
+  defp fetch_events_from_messages(messages, context) do
+    Enum.flat_map(messages, fn
+      %{data: {id, tid}} ->
+        case :ets.lookup(tid, id) do
+          [{^id, _status, log_event}] -> [process_data(log_event, context)]
+          [] -> []
+        end
+    end)
+  end
```
amokan, line 303:
```suggestion
  defp fetch_events_from_messages(messages) do
    Enum.flat_map(messages, fn
      %{data: {id, tid}} ->
        case :ets.lookup(tid, id) do
          [{^id, _status, log_event}] -> [log_event]
          [] -> []
        end
    end)
  end
```

Calling `process_data/2` in here means we are potentially calling the `Schema.update` twice (_once at `handle_message` and again in `handle_batch`_), right?

Seems like we could bypass that and just hit ETS here 🤷‍♂️

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -189,9 +207,11 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
     } do
       source = Sources.Cache.get_by_id(context.source_id)
 
+      # Fetch full LogEvents from ETS — events are still present (marked :processing)
+      log_events = fetch_events_from_messages(messages, context)
```
amokan: While it _shouldn't_ happen, it may worth changing this return type to a tuple like `{log_events, missing_event_ids}` where we would return the event ids for anything not found in ETS for whatever reason. If `missing_event_ids` was not empty - we could emit a Logger warning and/or a telemetry counter that indicated how many ETS lookups missed or something.

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -503,7 +552,9 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
     }
   end
 
-  def message_size(data) do
-    :erlang.external_size(data)
-  end
+  # For {id, tid} pointers the actual payload size is unknown until ETS lookup at batch time.
+  # Return 0 so the count limit (@max_batch_size) governs batch splits instead.
+  def message_size({_id, _tid}), do: 0
```
amokan: Wondering if we should move the byte cap to the batching side after we call `fetch_events_from_messages`? Could always chunk the messages there into sub-batches under `@max_batch_length` and issue one insert per chunk.

Another thought is that we could toss the message size into ETS as a 4th tuple element potentially.

amokan (changes requested):
So glad to see this coming together. I think we're close, but have some notes I added throughout the PR as well as a few general bullet points below.

- Would be good to get a test in that exercises that `avg > 100` logic
- Didn't notice any `id_passing: true` path logic around `BufferProducer` within the tests. Could be missing something though.
- I'd personally lean towards getting the cleanup logic in here around stale/orphaned `:processing` items. If others are ok with a follow-up PR as you call out, I can live with that.

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -135,14 +136,24 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
+        metrics = Sources.get_source_metrics_for_ingest(source.token)
+
+        for %{data: {id, tid}} <- successful do
+          case :ets.lookup(tid, id) do
+            [{^id, _status, le}] -> emit_event_telemetry(queue, source, le, backend_metadata)
+            [] -> :ok
+          end
+
+          if metrics.avg > 100 do
+            :ets.delete(tid, id)
```
djwhitt: Minor robustness for crash recovery: these write straight to `tid` via `:ets.update_element` / `:ets.delete`. If the owning producer has died — taking its table with it — before an in-flight message acks, `:ets.update_element` on a missing table raises `ArgumentError`. The old `IngestEventQueue.delete/2` path was `get_tid`-guarded and returned `{:error, :not_initialized}` instead. A table-exists guard (or going through the queue API) would keep a producer crash from turning into a raising acknowledger. (A missing key on a live table is fine — no raise.)

test/logflare/bigquery/pipeline_test.exs
```diff
@@ -45,14 +51,20 @@ defmodule Logflare.BigQuery.PipelineTest do
...
+      message = Pipeline.transform({le.id, tid}, ref: ref)
       {mod, ref, _data} = message.acknowledger
-      assert IngestEventQueue.get_table_size(sid_bid_pid) == 0
+      assert IngestEventQueue.get_table_size(sid_bid_pid) == 1
 
       mod.ack(ref, [], [message])
 
-      assert IngestEventQueue.get_table_size(sid_bid_pid) == 0
+      # Event is NOT requeued (retries == max_retries); remains as :ingested in ETS
+      assert IngestEventQueue.total_pending(sid_bid_pid) == 0
```
djwhitt, line 69: The comment says the event "remains as :ingested in ETS", but with `retries == max_retries` the failed path hits `:ets.delete(tid, id)` in `maybe_requeue_failed/3` — the row is deleted, not left `:ingested`. This assertion was `get_table_size(sid_bid_pid) == 0` before this PR, which correctly captured the deletion; switching it to `total_pending == 0` passes whether the row is deleted or stranded non-pending, so it no longer pins the behavior down. Suggest restoring `get_table_size(sid_bid_pid) == 0` and fixing the comment to say the event is deleted.

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: Oh no, I removed it from handle_message on a different branch and forgot to apply it here. 🤦

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -470,6 +470,38 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+    ms =
+      Ex2ms.fun do
+        {event_id, :pending, _event} -> event_id
+      end
+
+    with tid when tid != nil <- get_tid(sid_bid_pid),
+         size when is_integer(size) <- :ets.info(tid, :size),
+         {taken_ids, _cont} <- :ets.select(tid, ms, min(n, max(size, 1))) do
+      for id <- taken_ids do
+        :ets.update_element(tid, id, {2, :processing})
```
Ziinc, line 516: this means that we're introducing a 3rd state, correct? QueueJanitor needs to be aware of this new state, as it measures buffer length based on `:pending` state only.

Ziinc: need to consider what is going to happen to these in failure mode where high memory pressure requires load shedding, do we drop events in processing state last?

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: Fixed

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: Moved these to IngestEventQueue helpers with rescue and telemetry.

test/logflare/bigquery/pipeline_test.exs
Baishan: Fixed

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: Added telemetry to log these missing ones.

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: Chunking in handle_batch now. I'm passing along the sizes now to prevent recalculating them in the ack.

Baishan:
> So glad to see this coming together. I think we're close, but have some notes I added throughout the PR as well as a few general bullet points below.
> 
> * Would be good to get a test in that exercises that `avg > 100` logic
> * Didn't notice any `id_passing: true` path logic around `BufferProducer` within the tests. Could be missing something though.
> * I'd personally lean towards getting the cleanup logic in here around stale/orphaned `:processing` items. If others are ok with a follow-up PR as you call out, I can live with that.

Queue janitor will snapshot :processing ids every 10 seconds now and purge any that haven't been deleted in that 10 second window. 

Added more tests to around the id_passing in buffer producer and more explicit tests for handle_batch in pipeline_test.

lib/logflare/backends/ingest_event_queue/queue_janitor.ex
```diff
@@ -117,6 +127,88 @@ defmodule Logflare.Backends.IngestEventQueue.QueueJanitor do
...
+  end
+
+  defp act_on_stale_event(tid, id) do
+    case :ets.lookup(tid, id) do
+      [{^id, :processing, %{retries: retries}}] when retries >= @max_stale_retries - 1 ->
+        :ets.delete(tid, id)
+        :drop
+
+      [{^id, :processing, %{retries: retries} = le}] ->
+        :ets.insert(tid, {id, :pending, %{le | retries: (retries || 0) + 1}})
```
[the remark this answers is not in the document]
amokan: @Baishan this one seems fairly legit. And [select_replace/2](https://www.erlang.org/doc/apps/stdlib/ets.html#select_replace/2) seems like a reasonable change to cover this one.

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -426,6 +426,27 @@ defmodule Logflare.Backends.IngestEventQueue do
   """
   @spec total_pending(source_backend()) :: integer()
   @spec total_pending(source_backend_pid()) :: integer() | {:error, :not_initialized}
+  @doc """
```
amokan: This whole section (doc/spec/def) should be moved out of the `total_pending` block

lib/logflare/backends/ingest_event_queue/queue_janitor.ex
```diff
@@ -117,6 +127,88 @@ defmodule Logflare.Backends.IngestEventQueue.QueueJanitor do
...
+            :skip -> {resets, drops}
+          end
+        end)
+    end
+  end
+
+  defp act_on_stale_event(tid, id) do
+    case :ets.lookup(tid, id) do
+      [{^id, :processing, %{retries: retries}}] when retries >= @max_stale_retries - 1 ->
+        :ets.delete(tid, id)
```
amokan: Clanker is having a tantrum about not using the more robust `delete_id/2` function you created here :robot:

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -436,16 +525,29 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
 
   # Requeue failed events if the number of previous retries is less than @max_retries
   defp maybe_requeue_failed(_, [], _), do: :ok
-  defp maybe_requeue_failed(_, _, %{max_retries: 0}), do: :ok
+
+  defp maybe_requeue_failed(_, failed, %{max_retries: 0}) do
+    for %{data: {id, tid, _size}} <- failed, do: :ets.delete(tid, id)
```
amokan: _Maybe_ use your `delete_id/2` function here?

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -607,6 +660,71 @@ defmodule Logflare.Backends.IngestEventQueue do
     end
   end
 
+  @doc """
+  Safely deletes a single event by id from a tid, ignoring stale (deleted) tables.
+  Emits telemetry if the table no longer exists.
+  """
+  @spec delete_id(:ets.tid(), term()) :: :ok
```
amokan, line 658: These functions could potentially be moved to some general utils module later on (_and made a little bit more generic like `EtsUtils.safe_delete/2` and `EtsUtils.safe_update/3`_) - lets not sweat that now and slow you down, but we could always take an arg for the telemetry keys or something 🤷‍♂️

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -436,16 +525,29 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
-      |> Enum.map(fn %{data: %LE{} = event} ->
-        %LE{event | retries: (event.retries || 0) + 1}
+    to_requeue =
+      Enum.reduce(failed, [], fn %{data: {id, tid, _size}}, acc ->
+        case :ets.lookup(tid, id) do
+          [{^id, _status, %LE{retries: retries} = le} | _] when retries < max_retries ->
+            [%LE{le | retries: (retries || 0) + 1} | acc]
+
+          [{^id, _status, _le} | _] ->
+            :ets.delete(tid, id)
```
amokan: Same scenario where your delete helper function could make this a bit more resilient.

lib/logflare/backends/ingest_event_queue/queue_janitor.ex
```diff
@@ -117,6 +127,88 @@ defmodule Logflare.Backends.IngestEventQueue.QueueJanitor do
     end
   end
 
+  # Snapshot-based stale :processing cleanup.
+  # On each run, get current :processing IDs per queue. Any IDs that were also
+  # :processing on the previous run are stuck (batcher crash, kill signal, etc.).
+  # Stuck events have their retries incremented; those exceeding @max_stale_retries
+  # are dropped, the rest are reset to :pending.
```
amokan:
```suggestion
  # Stuck events are reset to :pending and retried
  # Dropped once they've been stale @max_stale_retries times
```

Idk. Something to better articulate we try _up to_ `@max_stale_retries` rather than exceeding it.

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -478,32 +580,4 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
   end
 
   # https://hexdocs.pm/broadway/Broadway.html#start_link/2
```
amokan:
```suggestion
```

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -607,6 +660,71 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+      _ -> []
+    end
+  end
+
+  @doc """
+  Resets a given list of event IDs from `:processing` back to `:pending` in a queue.
+  Used by QueueJanitor to recover stale in-flight events.
+  """
+  @spec reset_processing(table_key() | consolidated_table_key(), [term()]) :: :ok
+  def reset_processing(key, ids) do
```
amokan: Ensure this is still being used

lib/logflare/sources/source/bigquery/pipeline.ex
amokan: Pretty sure this is dead code now too

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -189,28 +206,56 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
         end
-      end
+      end)
+
+      messages_with_sizes =
+        Enum.map(triples, fn {%{data: {id, tid}} = msg, _le, size} ->
+          Message.update_data(msg, fn _ -> {id, tid, size} end)
+        end)
+
+      messages_with_sizes
```
amokan: Really hate leaving you with nonsense like this - but double-check this complaint 🤖🤛

> Missing-ID messages silently dropped from the batch — pipeline.ex:206-258 + the vendored fork at deps/broadway/.../batch_processor_stage.ex:76 (verified)
  fetch_events_from_messages/2 splits into {triples, missing}; handle_batch returns only triples. Missing messages are neither in successful nor failed, so
  they're never acked and never retried via maybe_requeue_failed. This only "works" because the fork softened the returned != size guard from raise to
  Logger.error. Relying on a soft log deep in a vendored dep to preserve the batch invariant is fragile. The janitor may recover these (resets
  :processing→:pending), but the warning carries no source_id/IDs in the message body. Recommend: route missing IDs through the failed path (explicit retry) and
  make the warning Sentry-trackable with source_id + bounded ID sample.

amokan (dismissed):
Close 🙏

Be sure to check the automated review comment about the potential race condition in `act_on_stale_event/2`.

Also see that credo is barfing on nested depth in `process_stale_events` (`queue_janitor.ex:174`) and the good ole "_Prefer using an implicit `try` rather than explicit `try`_" for a couple lines in `ingest_event_queue.ex` - I see no issue in using the inline annotations to ignore the try situation.

Also worth bumping the `VERSION` file in this branch to `1.44.0` so you're clear to cut a release after merging and CI clears on main 👍

lib/logflare/backends/ingest_event_queue/queue_janitor.ex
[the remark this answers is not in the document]
Baishan: I've updated this with select_replace. I think this is unlikely to happen but if it does there is still a race condition thats more likely. Queuejanitor can update a :processing to :pending between sending the bigquery event and ack and producer could pick it up again. In order for this to happen an event would need to be in the pipeline for 10-20 seconds without being acked. 

Are bigquery inserts idempotent if they have the same id? I think this is very unlikely but not impossible for cases where events go stale but the pipeline does not crash. We will have telemetry to check if queue janitor is finding stale events and see how likely it is that could be happening. The only robust way I can think of to prevent duplicate inserts is to either never retry stale events (worse imo) or to ensure bigquery prevents duplicate ids. Whenever the pipeline crashes there will be events between ack and handle_batch that will be sent again.

lib/logflare/backends/ingest_event_queue.ex
Baishan: Fixed

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: Done

lib/logflare/backends/ingest_event_queue.ex
Baishan: Agreed, we can refactor these later.

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: These should all be updated now

lib/logflare/backends/ingest_event_queue.ex
Baishan: Removed

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: Passing these through failed so they won't end up starving demand.

Baishan:
I won't bump the version in this. I want to let it run on test for a while and check the telemetry. I can do a followup release if it looks good.

lib/logflare/backends/ingest_event_queue.ex
Ziinc: or not at all and hope for the best?

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -72,10 +73,8 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
           batchers: [
             bq: [
               concurrency: 16,
-              batch_size: bq_batch_size_splitter(),
-              batch_timeout: 1_500,
-              # must be set when using custom batch_size splitter
-              max_demand: @max_batch_size
+              batch_size: @max_batch_size,
```
Ziinc: batch size splitter is needed so that we can keep batch sizes smaller than the max gcp payload limit

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -476,34 +582,4 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
 
     :telemetry.execute([:logflare, :backends, :ingest], metrics, metadata)
   end
-
-  # https://hexdocs.pm/broadway/Broadway.html#start_link/2
-  # split batch sizes based on json size
-  # ensure that we are well below the 10MB limit
-  def bq_batch_size_splitter do
```
Ziinc, line 483: this is needed

Ziinc: would need to pop the event size together with the id

Ziinc (changes requested):

scripts/monitor.iex.exs
```diff
@@ -0,0 +1,125 @@
+import_if_available(Logflare.Utils.Debugging)
```
Ziinc, line 1: might be worth shifting it to the Debugging module

lib/telemetry.ex
```diff
@@ -267,6 +267,25 @@ defmodule Logflare.Telemetry do
...
         unit: {:native, :millisecond},
         description: "Latency of processing an incoming cache gossip cast"
+      ),
+      counter("logflare.ingest_event_queue.stale_table.count",
+        event_name: [:logflare, :ingest_event_queue, :stale_table],
+        description: "Count of ack operations on a stale (already deleted) ETS table"
+      ),
+      sum("logflare.ingest_event_queue.missing_ids.count",
+        event_name: [:logflare, :ingest_event_queue, :missing_ids],
+        tags: [:source_id],
```
Ziinc: do we need to differentiate by different sources? seems like it will result in a lot of noise.

lib/telemetry.ex
```diff
@@ -267,6 +267,25 @@ defmodule Logflare.Telemetry do
...
+        description: "Count of event IDs not found in ETS during handle_batch fetch"
+      ),
+      sum("logflare.ingest_event_queue.stale_processing.reset",
+        event_name: [:logflare, :ingest_event_queue, :stale_processing],
+        tags: [:source_id],
+        description: "Count of stale :processing events reset to :pending by QueueJanitor"
+      ),
+      sum("logflare.ingest_event_queue.stale_processing.dropped",
+        event_name: [:logflare, :ingest_event_queue, :stale_processing],
+        tags: [:source_id],
```
Ziinc: likewise, i think we can remove the tags

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: They are chunked based on size in handle_batch

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: We can refactor it like that but whats the benefit over chunking them in handle_batch?

lib/logflare/backends/ingest_event_queue.ex
Baishan: Yes its a 3rd state. I don't think we should drop :processing as it would be akin to dropping events that were previously popped to the pipeline already (which we dont currently do?). :processing is bounded by batch sizes in the pipeline.

Baishan: Event flow will be :pending -> :processing (in broadway) -> :ingested/:deleted

lib/telemetry.ex
Baishan: Removed the tags. These should never be hit unless we start shedding :processing events

lib/telemetry.ex
Baishan: Removed

lib/logflare/sources/source/bigquery/pipeline.ex
Baishan: Unless we store the size during ingest it would require calculating this in take_pending_ids which means copying the logevent into process memory early on which will likely negate the benefits of id passing.

Doing it in handle_batch means we only need to copy the log event once in the pipeline and can be further optimized to lazily fetch them from ets when building the bigquery payload.

Ziinc: doing so reduces the efficiency of each outbound request (the last request chunk would not be full), and if a source has a lot of large events then one batch could result in many requests being fired. it also implies that outbound requests per pipeline is not capped by batcher count, so one pipeline could be firing off many requests and lower efficiency. 
it also implies that a multi-request batch would be processed slower.
doing the calculation at buffer producer would be a simple fix

djwhitt: > Unless we store the size during ingest

I mean... could we? Seems likely worth it for the outbound request efficiency.

Baishan: I'm testing putting this in buffer producer and doing the size calc in a map to prevent loading all the events in one go. In order to have more than one chunk we would need 500 12kb events. Do we know if that happens?

Baishan: `I mean... could we? Seems likely worth it for the outbound request efficiency.`

We could and should but Id rather not change the format of the events in ets as it will require a larger refactor. Maybe in a followup when we implement id passing to ch

Baishan: I've moved the size calc into buffer producer. It will fetch ids and then map over each id to get the size to prevent a big heap allocation from fetching all the events. 
This still has a 10% increase in process memory vs doing it in handle_batch (fluctuates between reach run so probably not a big deal).

Ziinc: We could just add an extra field in the queue that will let us either pull it at the same time or look it up when needed, then it doesn't have to be pulled twice (once a buffer producer and once at pipeline), just a thought

Baishan: Moved size calc to add_to_table and keep it as a 4th tuple in ets.

Ziinc (approved):

PR #3571, opened 2026-06-08
perf: backpressure high memory

Ziinc:
BufferLimiter now returns 429s at high node memory above 85% for proper backpressure.
It uses a cachex-based warming strategy to ensure that the cache is always hit instead of blocking requests on cache miss. Indicates to clients to retry. Due to warming behaviour, high memory backoff would last for 3s.

Closes O11Y-1884


test/logflare/system_cache_test.exs
```diff
@@ -0,0 +1,22 @@
...
+    Cachex.clear(Logflare.SystemCache)
+    :ok
+  end
+
+  describe "memory_utilization/0" do
+    test "caches the result within TTL" do
+      Logflare.System
+      |> expect(:memory_utilization, 1, fn -> 0.42 end)
+
+      assert Logflare.SystemCache.memory_utilization() == 0.42
```
amokan: Nit - alias `Logflare.SystemCache`

test/logflare_web/controllers/plugs/buffer_limiter_test.exs
```diff
@@ -14,6 +14,31 @@ defmodule LogflareWeb.Plugs.BufferLimiterTest do
     {:ok, conn: conn, source: source, table_key: table_key}
   end
 
+  test "returns 429 when memory utilization is at or over 85%", %{conn: conn, source: source} do
+    Logflare.SystemCache
```
amokan: Nit. alias `Logflare.SystemCache`

lib/logflare/system_cache/warmer.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Logflare.SystemCache.Warmer do
+  @moduledoc false
+
+  use Cachex.Warmer
+
+  @impl true
+  def execute(_state) do
+    {:ok, [{:memory_utilization, Logflare.System.memory_utilization()}]}
+  end
```
amokan, line 15:
```suggestion
  @impl true
  def execute(_state) do
    {:ok, [{:memory_utilization, Logflare.System.memory_utilization()}]}
  rescue
    e ->
      Logger.warning("SystemCache warmer failed: #{inspect(e)}")
      :ignore
  end
```

lib/logflare/system_cache.ex
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  @spec memory_utilization() :: float()
+  def memory_utilization do
+    case Cachex.fetch(@cache, :memory_utilization, fn _ ->
+           {:commit, Logflare.System.memory_utilization()}
+         end) do
+      {:ok, value} -> value
+      {:commit, value} -> value
+      {:error, _} -> Logflare.System.memory_utilization()
```
amokan: May be worth adding some Logging here to avoid a case where we are not aware the cache is erroring on every call or something.

amokan (approved):
Looks good! I think we should tighten up that error case in the `SystemCache.memory_utilization/0` call with some Logging so we're not blind - but I could be over-indexing on that situation.

lib/logflare_web/controllers/plugs/buffer_limiter.ex
```diff
@@ -21,7 +24,8 @@ defmodule LogflareWeb.Plugs.BufferLimiter do
   """
   @spec call(Plug.Conn.t(), opts()) :: Plug.Conn.t()
   def call(%{assigns: %{source: %Source{} = source}} = conn, _opts) do
-    if Backends.cached_local_pending_buffer_full?(source) do
+    if SystemCache.memory_utilization() >= @memory_limit or
```
djwhitt, line 27: Should this 429 include `Retry-After: 3`? That matches the warmer interval, so clients avoid retrying against the same cached high-memory reading.

djwhitt (commented):
One small API/backpressure note.

PR #3576, opened 2026-06-09
O11Y-982: Move non-interactive BigQuery queries to batch jobs

djwhitt:
## Summary
- add a BigQuery batch query path using jobs.insert with BATCH priority and polling via getQueryResults
- route alert queries, endpoint proactive refreshes, and UI tailing polling queries through batch jobs
- update tests for batch execution paths

## Tests
- mix test test/logflare/bigquery/bq_repo_test.exs test/logflare/alerting_test.exs test/logflare/alerting/alert_worker_test.exs test/logflare/backends/adaptor/incidentio_adaptor_test.exs test/logflare/logs/search_operations_test.exs test/logflare/endpoints/cache_test.exs

Linear: O11Y-982

### DIFF
--- a/lib/logflare/alerting.ex
+++ b/lib/logflare/alerting.ex
@@ -463,6 +463,7 @@ defmodule Logflare.Alerting do
              labels: %{
                "alert_id" => alert_query.id
              },
+             job_priority: :batch,
              query_type: :alerts
            ) do
       {:ok, result}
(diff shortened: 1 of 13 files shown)


djwhitt:
Not sure this is worth it given current work to reduce BQ queries. It both comes with some risk and needs more testing.

PR #3578, opened 2026-06-09
feat: adds support for otel `kind`

chasers:
- added string to string serializer in rust mapper
- rust mapper enforces low cardinality of span_kind
- encodes `:SPAN_KIND_SERVER` to `Server` et al in Logs.OtelTrace module
- encoding it here so `kind` value is the same in BigQuery and Clickhouse (more human readable)
- confirmed HyperDX does expect `Server` et al for values for service mapping view

Closes O11Y-1978


amokan (approved):
Looks good!

PR #3587, opened 2026-06-11
feat: add metadata to endpoints list

msmithstubbs:
Adds SQL dialect and backend, if set, to endpoints list.

CleanShot 2026-06-11 at 16 09 31@2x

Long backend name is truncated: 

CleanShot 2026-06-11 at 15 57 04@2x

## Alternate layout

If we want to manage the height of this page we could put them inline after caches:

CleanShot 2026-06-11 at 16 22 39@2x

Closes O11Y-1971


Ziinc:
@msmithstubbs I think inline before caches would be a nicer touch and layout

msmithstubbs:
Moved to before 'caches'

CleanShot 2026-06-15 at 09 32 38@2x

Ziinc (approved):

PR #3588, opened 2026-06-11
chore: allow publishing package

staaldraad:


.github/workflows/docker-retag-versioned.yml
jgoux, line 83: While you're at it, would you mind adding the `--prefer-index=false \` option to this command like I did [here](https://github.com/supabase/cli/pull/5542/changes/820f408ebd6bed18051bf13dc4acf4d78687315c)? 😄

Ziinc (approved):

PR #3592, opened 2026-06-12
feat: add LRU-aware cache warming for KeyValues

bblaszkow06:
Track key-value cache hits via a dual-buffer ETS UsageTracker GenServer that periodically upserts a `key_value_usages` table. On startup, CacheWarmer now warms only the top-N most-recently-used entries (ordered by last_used_at, falling back to updated_at) instead of streaming the full table.


priv/repo/migrations/20260609000000_create_key_value_usages.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Logflare.Repo.Migrations.CreateKeyValueUsages do
+  use Ecto.Migration
+
+  def change do
+    create table(:key_value_usages) do
+      add :key_value_id, references(:key_values, on_delete: :delete_all), null: false
+      add :last_used_at, :utc_datetime_usec, null: false, default: fragment("now()")
```
Ziinc, line 7: hmm this also implies that if all 1 mil records are in use, then 1 mil records will also be created in this table. why would we want a separate table as compared to having these two fields directly on the `:key_values` table?

Ziinc: I'm guessing that the separate table means that the table isn't tracked it isn't tracked as part of ContextCache since it is not included in the published tables, so bumping it would not trigger cache invalidations.
should make this clear in the KeyValuesUsage module

Ziinc: should probably have some periodic truncation logic to keep the size of this table down

lib/logflare/key_values/usage_tracker.ex
```diff
@@ -0,0 +1,110 @@
+defmodule Logflare.KeyValues.UsageTracker do
```
Ziinc: instead of an ets-based tracker that gets routed through a genserver, a simpler and good enough solution would be periodically streamed from the KeyValues.Cache the Sources.recent_events_touch/1, and triggered with Oban via an Oban worker. 
Since LRU cache warming is already limited to 500k results, we don't need 100% accuracy and only need a good-enough sort order during warming.
Furthermore, this would eat into memory, which the KeyValues would already be consuming quite a bit of

Ziinc: we'll need to balance "correctness" for performance in this case. we should defer more accurate LRU tracking until the need arises, this would be too much overhead and overengineering for improved cache warming.

Ziinc (changes requested):

PR #3595, opened 2026-06-15
feat(telemetry): emit build commit SHA as OTel resource attribute

djwhitt:
## What

Emit the build's git commit SHA as a `service.commit` OTel resource attribute on the metrics exporter, so it lands in BigQuery as `resource._service_commit`.

## Why

OTel telemetry can already report `resource._service_version` (e.g. `1.44.3`), read from the compiled `VERSION` file. That's bumped only at release time, so it can't distinguish which commit a continuously-deployed build is actually running — many distinct commits report the same semver. Today the SHA exists only as a Docker image tag in CI and never reaches the running app. This change gives per-build identity.

## Changes

- **`lib/telemetry.ex`** — add a `commit` entry to the metrics exporter `:resource` `service` map, populated from `LOGFLARE_COMMIT_SHA`. The key is only added when the env var is set (see Notes).
- **`Dockerfile`** — `ARG COMMIT_SHA` → `ENV LOGFLARE_COMMIT_SHA` in the runner stage.
- **`.github/workflows/docker-ci-amd-and-arm.yml`** — pass the already-computed `short_sha` as the `COMMIT_SHA` build-arg.

Cloud Build's `secret_setup.Dockerfile` is `FROM` the app image and inherits the `ENV`; versioned/prod images are retags of the same dev image — so the SHA propagates everywhere with no Cloud Build changes.

## Notes

- **Unset is safe.** When `LOGFLARE_COMMIT_SHA` is absent (local/ad-hoc builds), the `commit` key is omitted from the resource `service` map entirely, rather than emitted as an empty string. In OTel terms an unknown attribute should be absent, not present-but-empty — so downstream queries see the attribute missing rather than `''`. CI always sets it, so real deployments are unaffected.

### DIFF
--- a/lib/telemetry.ex
+++ b/lib/telemetry.ex
@@ -3,6 +3,7 @@ defmodule Logflare.Telemetry do
 
   import Telemetry.Metrics
   import Logflare.Utils, only: [ets_info: 1]
+  import Logflare.Utils.Guards, only: [is_non_empty_binary: 1]
 
   def start_link(arg), do: Supervisor.start_link(__MODULE__, arg, name: __MODULE__)
 
@@ -33,12 +34,7 @@ defmodule Logflare.Telemetry do
 
     otel_exporter =
       if Application.get_env(:logflare, :opentelemetry_enabled?) do
-        service =
-          %{
-            name: "Logflare",
-            version: Application.spec(:logflare, :vsn) |> to_string()
-          }
-          |> maybe_put_commit(System.get_env("LOGFLARE_COMMIT_SHA"))
+        service = service_attributes(System.get_env("LOGFLARE_COMMIT_SHA"))
 
         otel_exporter_opts =
           Application.get_all_env(:opentelemetry_exporter)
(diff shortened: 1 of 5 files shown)


lib/telemetry.ex
```diff
@@ -33,15 +33,19 @@ defmodule Logflare.Telemetry do
 
     otel_exporter =
       if Application.get_env(:logflare, :opentelemetry_enabled?) do
+        service =
+          %{
+            name: "Logflare",
+            version: Application.spec(:logflare, :vsn) |> to_string()
+          }
+          |> maybe_put_commit(System.get_env("LOGFLARE_COMMIT_SHA"))
```
chasers: We can test this stuff, might as well

chasers (changes requested):

lib/telemetry.ex
djwhitt: Extracted the commit-building logic into a public `service_attributes/1` and covered the set / nil / empty / whitespace / trim cases in `telemetry_test.exs`. Also fixed the empty-string case Copilot flagged (an unset Docker build-arg expands to `""`, which would otherwise emit `service.commit: ""`).

djwhitt: Added an end-to-end test in `test/logflare/telemetry_otel_export_test.exs` — it starts `OtelMetricExporter` with an `export_callback`, forces a sync export, and asserts the emitted resource actually carries `service.commit` with the right value (and omits it when no SHA is set). Extracted `Telemetry.resource/0` so the test exercises the same resource wiring `init/1` uses. Kept the small `service_attributes/1` unit tests for the trim/empty edge cases too.

djwhitt: One note on why the export test is its own file rather than folded into `telemetry_test.exs`: it mutates global state via `System.put_env`/`delete_env`, so it has to be `async: false`. `telemetry_test.exs` is `async: true`, and merging them would force that whole file serial — needlessly slowing the existing parallel unit tests. The exporter itself is name-scoped (ETS + persistent_term), so it isn't the blocker; the env var is. Keeping it separate preserves parallelism for everything else.

chasers (approved):

lib/telemetry.ex
chasers: ✅ assuming the tests pass

PR #3603, opened 2026-06-17
fix: deleting partner from Admin

msmithstubbs:
* add cascade delete to partner_users and users.partner_id

`partner_users` appears to be an legacy table. Maybe it can be dropped altogether?

https://github.com/user-attachments/assets/c169b47e-314a-48ed-b633-f200f5abe9ca

Fixes O11Y-1855


Ziinc (approved):
We'll remove partner functionality in a separate cleanup.

PR #3610, opened 2026-06-18
perf: timeouts tweaking

Ziinc:
- tweaks timeouts for Ch to 60s
- bumps queue_target to 5s to handle load spikes
- logs slow queue checkouts for early warning signals.

### DIFF
--- a/lib/logflare/endpoints/results_cache.ex
+++ b/lib/logflare/endpoints/results_cache.ex
@@ -43,7 +43,7 @@ defmodule Logflare.Endpoints.ResultsCache do
   We have a %GoogleApi.BigQuery.V2.Model.ErrorProto{} model but it's missing fields we see in error responses.
   """
   def query(cache) when is_pid(cache) do
-    GenServer.call(cache, :query, 30_000)
+    GenServer.call(cache, :query, 60_000)
   catch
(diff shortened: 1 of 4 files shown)


Baishan (approved):
Might be spammy? Do we output a distribution so we can see a histogram of checkout times ?

chasers (approved):

lib/logflare/endpoints/results_cache.ex
```diff
@@ -43,7 +43,7 @@ defmodule Logflare.Endpoints.ResultsCache do
   We have a %GoogleApi.BigQuery.V2.Model.ErrorProto{} model but it's missing fields we see in error responses.
   """
   def query(cache) when is_pid(cache) do
-    GenServer.call(cache, :query, 30_000)
+    GenServer.call(cache, :query, 60_000)
```
djwhitt: Wondering about this being the same as the `Ch.query` timeout. Will that make errors non-deterministic?

Ziinc: hmm possible, i'll bump the gensever timeout higher

Ziinc:
> Might be spammy? Do we output a distribution so we can see a histogram of checkout times ?

not yet, though worthwhile to do

PR #3616, opened 2026-06-20
perf(ingest): update_counter claim + BigQuery pipeline allocation cleanups

djwhitt:
## Summary

Two related ingest-path performance changes:

1. **`IngestEventQueue.take_pending_ids/2`** — replace the per-id `:ets.select_replace` claim with an atomic claim counter (`update_counter`). ~4.9x faster claims, same multi-consumer safety.
2. **BigQuery Broadway pipeline** — remove several sources of unnecessary per-batch and per-event allocation in `handle_batch/4`, the streaming insert path, and `ack/3`.

> ⚠️ **Draft / prototype.** Opening for design feedback. See "Status & open questions" before reviewing for merge.

---

## Part 1 — `take_pending_ids/2` claim via update_counter

`take_pending_ids/2` is the BigQuery pipeline's claim hot path. It claimed each event with a per-id `:ets.select_replace`, which **compiles a match spec per id** in the BIF layer — benchmarked as the dominant cost. This swaps that for an atomic claim counter, keeping the multi-consumer safety that #3609 added while running **~4.9x faster** (652 → 132 µs per 1000-event batch).

### Approach

The row tuple gains a 5th integer field — `{event_id, status, event, size, claim}`. A claim succeeds when `:ets.update_counter/3` transitions `claim` from `0 → 1`:

- **Multi-consumer safe** — only the `0 → 1` winner takes the event; a `2+` result means another consumer (or a `write_concurrency` resize-duplicate within the same select) already claimed it. Same guarantee as `select_replace`.
- **Cheap** — a direct integer op, no per-id match-spec compilation, and no `uniq_by` dedup pass (duplicates lose the CAS naturally).
- **Re-insertion resets the counter to 0**, so both requeue paths keep events claimable: `BufferProducer` failure requeue (via `add_to_table`) and the `QueueJanitor` stale `:processing → :pending` reset.

Also drops a redundant `:ets.info(:size)` read in `take_pending_ids` (`get_tid/1` already confirms the table is live).

### Why not the alternatives

| Strategy (1000-event claim) | avg | multi-consumer safe? |
|---|---|---|
| **update_counter (this PR)** | **132 µs** | ✅ |
| update_element + uniq_by | 191 µs | ❌ single-consumer only |
| select_replace per-id (prior) | 652 µs | ✅ |
| single compiled select_replace | 3519 µs | ✅ but ETS scans an N-clause spec |

`update_element` is faster than the old code but only safe if each queue has exactly one consumer. A single multi-clause `select_replace` (one compilation) is **rejected** — ETS does not index an N-clause literal-key spec into N keyed lookups; it scans, making it ~26x slower. See the benchmark file's recorded history for details.

### Blast radius

The tuple shape is a shared contract. Updated:

- `ingest_event_queue.ex` — construction + all match specs, new `claim_pending/2` helper.
- `queue_janitor.ex` — stale-reset path (**resets `claim` to 0** so requeued events stay claimable).
- `bigquery/pipeline.ex` — external raw `:ets.lookup` patterns.
- Tests updated to the 5-tuple; new `take_pending_ids/2` describe block incl. a concurrent-claim race test.

---

## Part 2 — BigQuery pipeline allocation cleanups

Per-batch and per-event allocations removed from the BQ Broadway pipeline. Batches are up to 500 events, so each scales with batch size and runs on every batch.

1. **Drop dead `event_size_pairs` in the streaming insert path.** The streaming path built a list of `{log_event, size}` tuples, passed it through `stream_batch/2`, and immediately re-extracted the `log_events` — discarding every size. Sizes are read exactly once (computing `batch_size` for telemetry); nothing downstream uses them, and the function's return value is discarded. Now passes the already-built `log_events` directly. Removes two O(n) traversals + ~2N tuple allocations per streaming batch.

2. **Reuse the batch source instead of a per-event cache fetch.** `process_data` called `Sources.Cache.get_by_id/1` for every event, copying the full `Source` struct out of the cache into the process heap once per event. `handle_batch/4` already resolves the identical source once (a BQ pipeline serves a single source), so it's threaded through `fetch_events_from_messages/3` into `process_data/3` and the per-event fetch is dropped.

3. **Collect batch events and metrics in a single pass.** `handle_batch/4` traversed the batch three times (`Enum.map` for log events, `length` for the count, `Enum.sum_by` for byte total). Folded into one `Enum.reduce`. Log events come out reversed, which is fine — each BQ row carries its own `insertId` and order within a batch insert is irrelevant.

4. **Drop the per-event ETS lookup in `ack/3`.** `ack` re-fetched the full `LogEvent` from ETS for every successful event solely to resolve telemetry labels — a second copy of an event `handle_batch/4` had already pulled out of ETS. The per-event ingest telemetry now emits in `handle_batch/4`, where the event is already in hand via `triples`, resolving the label mapping and backend metadata once per batch. `ack/3` now only needs each event's id to finalize queue state and never touches the event body. Adds `Sources.extract_labels/2` (mapping-aware extraction) with `get_labels_from_event/2` delegating to it.

The storage-write path was already clean (takes `log_events` directly).

### Behavior note for #4 (telemetry timing)

`[:logflare, :backends, :ingest]` now fires at batch-insert time rather than at ack. The emitted set is identical (`handle_batch`'s `succeeded` == ack's `successful`) except on the exceptional path where `handle_batch` raises after emitting — consistent with the pipeline already emitting regardless of BigQuery insert errors (insert errors are logged, not propagated to fail messages). Events missing from ETS are not in `triples`, so they emit no ingest telemetry (unchanged from original behavior).

---

## Verification

- `mix test` on `ingest_event_queue_test.exs`, `bigquery/pipeline_test.exs`, `opentelemetry_test.exs`, `buffer_producer_test.exs`, `bigquery_adaptor_test.exs`, `dynamic_pipeline_test.exs`, `sources_test.exs` — all pass (incl. the `@tag :race` concurrent-claim test, the janitor stale-reset/reclaim test, `process_data/3` source-threading tests, and `handle_batch` ingest-telemetry tests).
- `mix lint.diff` clean, `mix test.typings` (Dialyzer) clean.
- Benchmark: `test/profiling/ingest_event_queue_take_pending_ids_bench.exs`.

## Status & open questions

- [ ] **Confirm the original incident behind #3609** — the `update_counter` CAS preserves multi-consumer safety regardless, but understanding what was observed would validate the regression test actually reproduces it.
- [ ] Run the **full** `test/logflare/backends/` tree + a real BQ ingest path (only the directly-affected files were run here).
- [ ] Consider a module-level doc / constructor for the 5-tuple shape to prevent future drift.
- [ ] `test/profiling/ingest_event_queue_trunc.exs` already used a stale 3-tuple spec (pre-existing, CI-excluded) — now two arities behind.
- [ ] Sanity-check the Part 2 #4 telemetry-timing change against any downstream consumer that treats `[:logflare, :backends, :ingest]` as a strictly-acked signal (e.g. billing/accounting).

### DIFF
--- a/lib/logflare/backends/ingest_event_queue.ex
+++ b/lib/logflare/backends/ingest_event_queue.ex
@@ -296,7 +296,10 @@ defmodule Logflare.Backends.IngestEventQueue do
   def add_to_table({sid_bid_pid, tid}, batch, _opts) when is_tuple(sid_bid_pid) do
     objects =
       for %{id: id} = event <- batch do
-        {id, :pending, event, :erlang.external_size(event.body)}
+        # 5th field is the claim counter: 0 = unclaimed. take_pending_ids/2 claims via an
+        # atomic update_counter (0 -> 1 winner), so re-inserting here resets it to 0 and
+        # makes a requeued event claimable again.
+        {id, :pending, event, :erlang.external_size(event.body), 0}
       end
 
     :ets.insert(tid, objects)
@@ -436,7 +439,7 @@ defmodule Logflare.Backends.IngestEventQueue do
   def total_pending({_sid, _bid, _pid} = sid_bid_pid) do
     ms =
       Ex2ms.fun do
-        {_event_id, :pending, _event, _size} -> true
+        {_event_id, :pending, _event, _size, _claim} -> true
       end
 
     with tid when tid != nil <- get_tid(sid_bid_pid),
@@ -454,7 +457,7 @@ defmodule Logflare.Backends.IngestEventQueue do
   def total_by_status({_, _} = sid_bid, status) do
     ms =
       Ex2ms.fun do
-        {_event_id, event_status, _event, _size} when event_status == ^status -> true
+        {_event_id, event_status, _event, _size, _claim} when event_status == ^status -> true
       end
 
     traverse_queues(
@@ -478,7 +481,7 @@ defmodule Logflare.Backends.IngestEventQueue do
   def take_pending(sid_bid_pid, n) when is_integer(n) do
     ms =
       Ex2ms.fun do
-        {_event_id, :pending, event, _size} -> event
+        {_event_id, :pending, event, _size, _claim} -> event
       end
 
     with tid when tid != nil <- get_tid(sid_bid_pid),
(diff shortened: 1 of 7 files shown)


djwhitt:
Superseded by the 3-way split of this work: #3617 (claim counter) and #3618 (handle_batch/streaming allocation cleanups) are merged; the remaining ack/telemetry change is tracked in #3619. Closing this combined draft in favor of those.

PR #3619, opened 2026-06-21
perf(bq pipeline): emit ingest telemetry in handle_batch, drop ack ETS lookup

djwhitt:
## Summary

Final PR of the split from #3616. The first two parts — the atomic claim counter (#3617) and the `handle_batch`/streaming allocation cleanups (#3618) — have merged to `main`, so this PR now stands alone on top of them. It carries the one behavior change in the split (telemetry timing), isolated for independent review.

## Change

`ack` re-fetched the full `LogEvent` from ETS for every successful event solely to resolve telemetry labels — a second copy of an event `handle_batch/4` had already pulled out of ETS. The per-event ingest telemetry now emits in `handle_batch/4`, where the event is already in hand via `triples`, resolving the label mapping and backend metadata once per batch. `ack/3` now only needs each event's id to finalize queue state and never touches the event body.

Adds `Sources.extract_labels/2` (mapping-aware extraction) with `get_labels_from_event/2` delegating to it, so the label mapping can be resolved once per batch and reused across events.

## Behavior note — telemetry timing

`[:logflare, :backends, :ingest]` now fires at batch-insert time rather than at ack. Two consequences worth noting:

- The emitted set is no longer gated on Broadway's per-message `successful` classification — it now covers every event found in ETS for the batch (`triples`). Events missing from ETS are not in `triples`, so they emit no ingest telemetry, unchanged from before.
- The emit happens *after* both insert paths and before any code that can raise, so a crash in `handle_batch` (serialization failure, NIF error, schema-update `MatchError`) means telemetry was never emitted — matching the old ack path, where a `handle_batch` crash also produced no telemetry. Normal BigQuery insert errors do not raise: both the streaming and Storage Write API paths handle them in a `case` and return normally, so telemetry fires for those just as ack did before.

## Performance

The old ack path ran, per successful event, one `:ets.lookup` that deep-copies the full `LogEvent` onto the ack process heap, plus a re-parse of `source.labels`. The new path reuses the events `handle_batch` already fetched and resolves the label mapping once per batch, so it adds no lookup. `test/profiling/bq_pipeline_ack_telemetry_bench.exs` measures the difference against a live ETS table at batch=500:

| body | source | old (ack) | new (handle_batch) | speedup | memory (old → new) |
|---|---|---|---|---|---|
| medium (~25 fields) | no labels | 239.8 µs | 9.95 µs | 24x | 1.96 MB → 50 KB |
| small (~3 fields) | no labels | 115.3 µs | 8.76 µs | 13x | 891 KB → 51 KB |
| medium (~25 fields) | with labels | 438.6 µs | 65.5 µs | 6.7x | 2.15 MB → 107 KB |
| small (~3 fields) | with labels | 243.8 µs | 30.6 µs | 8x | 1.06 MB → 107 KB |

The headline win is allocation: ~10–40x less transient garbage per batch, since the per-event full-event heap copy in ack is gone. Wall-time speedup is 7–24x on the isolated path; in the live pipeline the BigQuery insert dominates, so this shows up as reduced CPU/GC headroom rather than a higher throughput ceiling. (The bench also varies body size with a large refc-binary blob to confirm the copy cost tracks heap footprint, not `external_size` — large binaries are shared by refcount on read.)


Baishan (approved):
lgtm

PR #3625, opened 2026-06-22
perf(bq pipeline): delete non-default acked rows immediately

djwhitt:
For non-default BigQuery queues (`{sid, bid}` with `bid != nil`), the `:ingested` status is effectively write-only: recent logs only ever reads the default queue (`list_recent_logs_local` reads `{sid, nil}`), and the `QueueJanitor` already truncates *all* `:ingested` rows for non-default backends every cycle (`truncate_all? = metrics.avg > 100 or bid != nil`).

This changes the BQ pipeline ack to delete those rows immediately instead of marking them `:ingested` and waiting for the next janitor pass. End-state is identical; it just skips the intermediate `:ingested` dwell (up to one scaled janitor interval) and removes the janitor truncation work, reducing ETS residency for backend queues.

The default queue (`{sid, nil}`) is unchanged — it keeps the `:ingested` remainder that backs recent local logs.

## Change

`if metrics.avg > 100` → `if bid != nil or metrics.avg > 100` in the ack delete/status branch.

## Notes

- The janitor’s `or bid != nil` clause is retained as a safety net for any `:ingested` rows that still appear (lookup-miss path, claim races, other writers).
- Telemetry is unaffected — the lookup that emits per-event telemetry runs before the delete branch.

### DIFF
--- a/lib/logflare/sources/source/bigquery/pipeline.ex
+++ b/lib/logflare/sources/source/bigquery/pipeline.ex
@@ -152,7 +152,12 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
               :ok
           end
 
-          if metrics.avg > 100 do
+          # Non-default queues ({sid, bid} with bid != nil) never serve recent logs
+          # (list_recent_logs_local always reads {sid, nil}) and the janitor truncates
+          # all their :ingested rows every cycle anyway, so delete at ack instead of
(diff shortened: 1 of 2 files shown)


djwhitt:
Closing. I don't think there's enough of these to matter and it just adds another thing to think about when looking at the code.

PR #3630, opened 2026-06-23
perf(clickhouse): reduce allocations in RowBinary row encoding

djwhitt:
## What

Allocation-reduction optimizations to the ClickHouse RowBinary encoder hot path (`ingester.ex`, `row_binary_encoder.ex`). All output is byte-identical and map iteration order is unchanged — these are constant-factor wins on transient allocation, not a payload or behavior change.

- **Hoist batch-constant `mapping_config_id`** — it is the same UUID for every row in a batch, so `encode_batch/2` now encodes it once and reuses the bytes instead of re-parsing it per row. (Only this field is batch-constant; `source_uuid`/`source_name` vary per row because a single pipeline batches events from multiple sources, so they are deliberately not hoisted.)
- **`uuid/1` fast path** — a binary-pattern-match clause for canonical hyphenated UUIDs avoids the `String.replace("-", "")` intermediate; non-canonical input falls through to the original clause.
- **`:maps.fold` map encoder** — the generic `map/3` replaces `Map.to_list` + `length/1` + `Enum.flat_map`, dropping a list materialization, an O(n) length walk, and an unnecessary flatten.
- **Specialized `map_string_string`** — the hottest map encoder, used for every resource/scope/log/span attribute map plus the maps nested in metric exemplars and trace events/links. It now encodes directly with `:maps.fold`, inlining the length-prefix for binary keys/values so each pair emits one flat iolist fragment instead of two nested ones; a guarded fallback clause preserves behavior for non-binary (iodata) values.

## Results

Apple M5, measured with Benchee. End-to-end encode stage — the real `Ingester.encode_batch/2` over realistic 500-event log/metric/trace batches (`bench/clickhouse_encode_e2e.exs`), combined effect of all commits vs the PR base:

| batch | allocation | throughput |
| --- | --- | --- |
| log | −28% | 1.32× |
| metric | −21% | ~1.08× |
| trace | −25% | ~1.12× |

Allocation reduction is the reliable signal here (Benchee reports it deterministically across runs); throughput varies more run-to-run, with log the clearest gain.

Per-optimization micro-benchmarks (`bench/clickhouse_row_encoding.exs`, `bench/clickhouse_map_encoding.exs`, `bench/clickhouse_encoder_primitives.exs`):

| change | speedup | allocation |
| --- | --- | --- |
| batch UUID hoist (500-row log batch) | 1.14× | −13% |
| `uuid/1` fast path (per call) | 1.46× | −27% |
| generic `map/3`, 15 keys | 1.54× | −8% |
| `map_string_string`, 15 keys | 1.41× | −19% |
| `map_string_string`, 100 keys | 3.3× | −17% |

Output is byte-identical (verified by each benchmark's sanity check). This measures the encode stage only — downstream gzip and the HTTP insert are unchanged — so the win is reduced GC pressure on the ingest path rather than a 1:1 ingest-throughput change.

## Benchmarks

`bench/run_encode_e2e.sh [baseline_sha]` runs the e2e encode bench against a baseline SHA and the current HEAD, then prints a before/after comparison. Snapshots are written to a temp dir, never the repo tree (consistent with the other benches here, which print to stdout and persist nothing).

### DIFF
--- a/bench/clickhouse_encode_e2e.exs
+++ b/bench/clickhouse_encode_e2e.exs
@@ -1,0 +1,212 @@
+# Usage: TAG=before mix run --no-start bench/clickhouse_encode_e2e.exs
+#
+# End-to-end encode-stage benchmark for the ClickHouse ingester. Exercises the
+# real production function `Ingester.encode_batch/2` (NOT hand-replicated
+# baselines) over realistic log / metric / trace batches, then forces the
+# resulting iodata to a binary so the full encode cost is measured.
+#
+# Run once per checkout via bench/run_encode_e2e.sh, which tags each run
+# (before/after) and saves Benchee results for cross-SHA comparison.
+
+alias Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester
+alias Logflare.LogEvent
+
+tag = System.get_env("TAG", "untagged")
+save_dir = System.get_env("BENCH_SAVE_DIR", "/tmp")
+File.mkdir_p!(save_dir)
+
+batch_size = String.to_integer(System.get_env("BATCH_SIZE", "500"))
+
+mapping_config_id = "00000000-0000-0000-0001-000000000003"
+source_uuid = String.to_atom("550e8400-e29b-41d4-a716-446655440000")
+
+# Realistic, *varied* resource attributes: ~12 keys spanning the typical OTEL
+# service/host/k8s/cloud namespaces. Key set and value lengths vary by event so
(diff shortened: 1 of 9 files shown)


amokan (approved):
Great work @djwhitt 

> Hoist batch-constant mapping_config_id

This is the only thing I want to call out for the future - if we keep the mapper around and eventually move it to be DB-driven per source (_or whatever_), this particular perf improvement around the mapping config id will need to be reverted. But at this time, the approach here is a sensible solution.

PR #3631, opened 2026-06-23
perf(clickhouse): avoid retaining duplicate event bodies in handle_batch

djwhitt:
## What

`handle_batch` in the ClickHouse pipeline mapped each event's body and built a **new** list of `LogEvent` structs, while the original Broadway `messages` (holding the original, unmapped bodies) stayed alive to be returned for acking. Both body sets were therefore retained simultaneously through the encode + insert window.

This rewrites the messages in place (`%{message | data: %{event | body: mapped_body}}`) so the mapped events flow through to acking and the insert reuses those same structs via a cheap pointer list. The original bodies become unreachable right after the mapping loop — before the slow insert — rather than being held until ack.

## Why it is safe

- Mapping stays in the pipeline, so both insert protocols (HTTP `Ingester` and native `NativeIngester`) are unaffected.
- `@max_retries` is `0`, so `requeue_retriable_messages` is always a no-op — there is no re-mapping risk — and the drop path only uses `event.id`.

## Results

Retained term size at the insert point (`:erts_debug.size/1` of `{messages, events}`, `bench/clickhouse_handle_batch_retention.exs`):

| batch size | before | after | reduction |
| --- | --- | --- | --- |
| 1,000 | 0.97 MB | 0.73 MB | −25.2% |
| 10,000 | 9.69 MB | 7.25 MB | −25.2% |

The reduction is constant across batch sizes.

## Testing

- Dialyzer, Credo, and formatter clean.
- ⚠️ `pipeline_test.exs` could not be run in my environment (no ClickHouse reachable; all 22 tests fail in `setup` at `provision_ingest_tables`, before the changed code runs). These integration tests should be run against a live ClickHouse before merge — which is why this PR is a draft.

### DIFF
--- a/bench/clickhouse_handle_batch_retention.exs
+++ b/bench/clickhouse_handle_batch_retention.exs
@@ -1,0 +1,95 @@
+# Usage: mix run bench/clickhouse_handle_batch_retention.exs
+#
+# Demonstrates the retained-memory difference in handle_batch between:
+#   OLD: build a separate `events` list of mapped LogEvents while the original
+#        `messages` (original bodies) stay alive for acking -> ~2x bodies held
+#        through the insert window.
+#   NEW: rewrite messages in place so `events` and the returned `messages`
+#        share the same mapped structs; original bodies become garbage before
+#        the insert -> ~1x bodies held.
+#
+# Retained size is measured with :erts_debug.size/1, which counts shared
+# sub-terms within a term once -- so it reflects what is actually live at the
+# point insert_log_events runs.
+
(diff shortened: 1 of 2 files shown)


djwhitt:
Slapped a "do not merge" on this. Doing a bit of experimentation, but I'm not sure this is what we want. More systemic changes (e.g. ID passing with bodies in ETS) are likely better.

djwhitt:
Closing this. We're going with ID passing in #3667 instead.

PR #3632, opened 2026-06-24
fix: backend_select supports default backend

msmithstubbs:
Remove hard coded BigQuery default backend from backend select. Fetches default backend instead.

CleanShot 2026-06-24 at 11 21 50@2x

Follow up to #3093


lib/logflare/backends.ex
msmithstubbs, line 1: Can keep lowercase if necessary.

Ziinc (approved):

PR #3633, opened 2026-06-24
feat: increase bigquery timeout to 60 seconds

msmithstubbs:
Removes distinction between BigQuery reservation and non-reservation searches: all now use a 60 second timeout.

Fixes O11Y-2041


msmithstubbs:
Do we want to show a timeout error message to the user? Or leave it as the generic *Backend error! Retry your query...* ?

Ziinc:
@msmithstubbs a timeout error would be much better 😄

msmithstubbs:
Added a query error message when BigQuery search times out.

Ziinc (approved):

PR #3638, opened 2026-06-24
feat: clickhouse bytes processed

Ziinc:
Adds in clickhouse bytes processed via x-clickhouse-summary header from http read response
CleanShot 2026-06-25 at 04 01 12@2x

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
@@ -98,7 +98,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
   def execute_query(%Backend{} = backend, {query_string, params}, _opts)
       when is_non_empty_binary(query_string) and is_list(params) do
     case execute_ch_query(backend, query_string, params) do
-      {:ok, result} -> {:ok, QueryResult.new(result)}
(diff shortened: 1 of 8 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -345,7 +345,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
           statement :: iodata(),
           params :: map | [term] | [row :: [term]] | iodata | Enumerable.t(),
           [Ch.query_option()]
-        ) :: {:ok, Ch.Result.t()} | {:error, QueryError.t()}
+        ) :: {:ok, {[map()], non_neg_integer() | :not_supported}} | {:error, QueryError.t()}
```
djwhitt, line 348: Nit: I think `:not_supported` can flow through to telemetry

djwhitt (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
Ziinc: hmm let me double check...

Ziinc: fixed:
```
    measurements = %{
      total_bytes_processed:
        if(is_integer(data.total_bytes_processed), do: data.total_bytes_processed, else: 0)
    }
```

PR #3641, opened 2026-06-25
feat: add copy schema to clipboard button

msmithstubbs:
* Adds a button to copy the source schema as LLM-friendly markdown. 
* Adds `clipboard_button/1` component as a reusable copy-to-clipboard button.

Text copied to clipboard is in the following format. Cursory tests showed better agent responses when a link to the LQL docs is included.

```
# Logflare source schema

Use this schema when writing Logflare LQL (https://docs.logflare.app/concepts/lql/)

- `event_message` STRING Human-readable event message.
- `id` STRING Event UUID.
- `level` STRING
- `metadata` ARRAY<RECORD>
  - `client_capabilities_str` STRING
  - `client_name` STRING
  - `client_version` STRING
  - `duration_ms` INTEGER
  - `jsonrpc_method` STRING
  - `request_path` STRING
  - `server_version` STRING
  - `session_id` STRING
  - `user_agent` STRING
  - `user_id` STRING
- `namespace` STRING
- `org_id` INTEGER
- `request_id` STRING
- `service` STRING
- `timestamp` TIMESTAMP Ingest timestamp.
```

CleanShot 2026-06-25 at 12 03 34@2x

## Demo

https://github.com/user-attachments/assets/c16b5578-bb7b-4d0b-9847-df5c39eaeeb0

Fixes O11Y-2083

### DIFF
--- a/lib/logflare_web/live/sources/bq_schema_component.ex
+++ b/lib/logflare_web/live/sources/bq_schema_component.ex
@@ -6,7 +6,7 @@ defmodule LogflareWeb.SourceBqSchemaComponent do
   alias Logflare.Sources.Source.BigQuery.SchemaBuilder
 
   @impl true
-  def render(%{source: source}) do
+  def render(%{source: source} = assigns) do
     bq_schema =
       SourceSchemas.Cache.get_source_schema_by(source_id: source.id)
       |> case do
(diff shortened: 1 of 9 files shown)


lib/logflare_web/live/sources/bq_schema_component.ex
```diff
@@ -6,14 +6,20 @@ defmodule LogflareWeb.SourceBqSchemaComponent do
...
+  def render(%{source: source} = assigns) do
     bq_schema =
       SourceSchemas.Cache.get_source_schema_by(source_id: source.id)
       |> case do
         nil -> SchemaBuilder.initial_table_schema()
         %_{bigquery_schema: schema} -> schema
       end
 
-    BqSchema.format_bq_schema(bq_schema)
+    assigns = assign(assigns, :bq_schema, BqSchema.format_bq_schema(bq_schema))
```
Ziinc, line 17: hmm can we use the flatmap schema instead of the BqSchema? would be less tech debt to decouple bq schema from this

Ziinc:
merging in first since this works as intended, but a flatmap based version is preferred.

Ziinc (approved):

PR #3648, opened 2026-06-29
feat: endpoint API supports lf-endpoint-version header

msmithstubbs:
Adds version selection to the endpoint API with the `lf-endpoint-version` header

* Adds endpoint query caching based on endpoint version
* Endpoint queries will include `endpoint_version` label
* If no `lf-endpoint-version` header is set the current version of the endpoint query is used.

Fixes O11Y-1064


lib/logflare/endpoints/results_cache.ex
```diff
@@ -188,28 +192,62 @@ defmodule Logflare.Endpoints.ResultsCache do
...
   end
 
-  def name(query_id, params) do
-    partition = endpoints_part(query_id, params)
+  def name(%EndpointQuery{id: query_id, version_number: version_number}, params) do
     param_hash = :erlang.phash2(params)
-    {:via, :syn, {partition, {query_id, param_hash}}}
+
+    {key, partition_key} =
+      case version_number do
```
msmithstubbs: Caching on version number does mean results for the latest endpoint version can be cached twice: once with the version number, and once without.

lib/logflare/endpoints.ex
```diff
@@ -303,11 +329,30 @@ defmodule Logflare.Endpoints do
...
-          fragment("(?->>'version_number')::integer = ?", version.meta, ^version_number)
+          fragment("?->>'version_number' = ?", version.meta, ^Integer.to_string(version_number))
     )
     |> Repo.one()
   end
 
+  @spec endpoint_from_version(EndpointQuery.t(), Version.t(), pos_integer()) :: EndpointQuery.t()
+  defp endpoint_from_version(%EndpointQuery{} = endpoint, %Version{meta: meta}, version_number)
+       when is_map(meta) do
+    endpoint = Repo.preload(endpoint, :user)
```
Ziinc: Using Users.Cache would help to remove the db query here, especially since this is a hot path

Ziinc: Alternatively we can load the user directly with in the initial EndpointQuery fetch

lib/logflare/endpoints/results_cache.ex
Ziinc: I think to avoid ambiguity, we should always set the version number. If no version created yet for existing endpoints then 0 or nil will work.

Ziinc: It also means that we essentially can remove cache invalidation on updates as the version number will always change so we don't need to be worried of config drift.

test/logflare/endpoints/cache_test.exs
```diff
@@ -49,6 +49,117 @@ defmodule Logflare.Endpoints.CacheTest do
...
+    test "endpoint updates invalidate versioned caches", %{endpoint: endpoint, user: user} do
+      insert(:endpoint_version,
+        endpoint: endpoint,
+        version_number: 1,
+        snapshot_overrides: %{"query" => "select 'historical' as testing"}
+      )
+
+      assert {:ok, versioned_endpoint} = Endpoints.get_endpoint_query_at_version(endpoint, 1)
+
+      GoogleApi.BigQuery.V2.Api.Jobs
```
Ziinc: Moving forward let's write tests again the Clickhouse backend instead of relying on BQ stubs

Ziinc (changes requested):

lib/logflare/endpoints/results_cache.ex
msmithstubbs: We would still have to invalidate the cache on update, right? 

If I'm requesting an endpoint without specifying a version I'll get cached version 0. If the endpoint is updated the default endpoint served will still be version 0 without the updated config.

msmithstubbs: Updated to use `:latest` as the version number when no version is specified. Think this is clearer. That work?

lib/logflare/endpoints.ex
msmithstubbs: `map_query_sources/1` actually just need `user_id` so removed the preload.

test/logflare/endpoints/cache_test.exs
msmithstubbs: Noted. Switched relevant tests to use Clickhouse.

lib/logflare_web/controllers/endpoints_controller.ex
```diff
@@ -98,13 +110,44 @@ defmodule LogflareWeb.EndpointsController do
...
   end
 
-  # only parse body for get when `?sql=` and `?lql=` are empty and it is sandboxable
+  @spec load_endpoint_query(Plug.Conn.t(), any()) :: Plug.Conn.t()
+  defp load_endpoint_query(%{assigns: %{endpoint: endpoint}} = conn, _opts) do
+    with [version_str | _] <- get_req_header(conn, "lf-endpoint-version"),
+         {:ok, version_number} <- parse_endpoint_version(version_str),
+         {:ok, endpoint_query} <-
+           Endpoints.get_endpoint_query_at_version(endpoint, version_number) do
+      assign(conn, :endpoint_query, endpoint_query)
```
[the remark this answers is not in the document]
msmithstubbs: This is true, but I'm not sure if it falls within the intent of this feature. Please advise, @Ziinc

PR #3650, opened 2026-06-29
feat: re-enable pipeline retries for ClickHouse backend

amokan:
This re-enables pipeline-level insert retries for the ClickHouse backend, which had been disabled for a while now because of compounding issues when we'd run into "too many parts" errors and things like that.

This PR introduces a simple ETS-based circuit breaker (_supervised and scoped per ClickHouse backend_) that allows _up to_ 2 retries depending on the state of the circuit breaker. The initial insert does not touch the circuit breaker by design.

The goal of this is to allow for additional retries but to avoid a constant blast of them when things are not in an ideal state on the ClickHouse side.


chasers (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -133,6 +135,24 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
         %{backend_id: backend_id}
       )
       when batcher in [:ch_fresh, :ch_stale] and is_event_type(event_type) do
+    emit_batch_telemetry(batch_info, backend_id, event_type, batcher, day_bucket)
+
+    backend = Backends.Cache.get_backend(backend_id)
+
+    case insert_batch(backend, messages, event_type, batcher, batch_info, day_bucket) do
+      :ok -> messages
+      {:error, reason} -> Enum.map(messages, &Message.failed(&1, reason))
```
Ziinc, line 144: Just a thought... Probably can upstream something to mark the entire batch as failed internally in Broadway so that we don't need to iterate over the entire batch just to mark it...

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -202,26 +236,47 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
-      drop_exhausted_messages(exhausted, backend_id)
-      requeue_retriable_messages(retriable, backend_id)
+      ack_backend_failures(backend_id, messages)
     end)
   end
 
-  @spec drop_exhausted_messages(exhausted :: [Message.t()], backend_id :: pos_integer()) :: :ok
-  defp drop_exhausted_messages([], _backend_id), do: :ok
+  @spec ack_backend_failures(backend_id :: pos_integer(), messages :: [Message.t()]) :: :ok
+  defp ack_backend_failures(backend_id, messages) do
```
Ziinc, line 244: We could move requeue logic to handle_failed/2 to keep ack logic more minimal

PR #3653, opened 2026-07-01
feat: restore endpoint version

msmithstubbs:
Adds the ability for users to restore an endpoint to a previous snapshot. 

* Endpoint versions include an action dropdown with a Restore button
  * To provide user feedback the 'Restoring' banner is optimistically displayed and hidden after the version is restored.
* Removes token as an endpoint snapshot field: this is autogenerated on database write so can't be capture in first snapshot, and then doesn't change anyway.

CleanShot 2026-07-01 at 22 01 29@2x

## Demo
https://github.com/user-attachments/assets/1092242a-f09c-4bce-9665-654c3aa11893

### DIFF
--- a/lib/logflare/endpoints.ex
+++ b/lib/logflare/endpoints.ex
@@ -169,6 +169,35 @@ defmodule Logflare.Endpoints do
     end
   end
 
+  @spec restore_query_version(EndpointQuery.t(), integer(), origin()) ::
+          {:ok, EndpointQuery.t(), integer()}
+          | {:error, Ecto.Changeset.t()}
+  def restore_query_version(%EndpointQuery{id: endpoint_id} = query, version_number, origin)
+      when is_integer(version_number) do
+    with {:version, %Version{id: version_id, meta: meta}} <-
+           {:version, get_endpoint_query_version(endpoint_id, version_number)},
+         {:restorable, true} <-
+           {:restorable, version_id != current_endpoint_version_id(endpoint_id)},
+         {:snapshot, snapshot} when is_map(snapshot) and map_size(snapshot) > 0 <-
+           {:snapshot, Map.get(meta, "endpoint_snapshot")},
+         {:ok, endpoint} <-
+           update_query(query, EndpointQuery.version_snapshot_attrs(snapshot), origin) do
+      {:ok, endpoint, version_number}
+    else
+      {:error, %Ecto.Changeset{}} = error ->
+        error
+
+      {:restorable, false} ->
+        {:error, invalid_restore_changeset(query, "Version is already current")}
+
+      {:snapshot, _} ->
+        {:error, invalid_restore_changeset(query, "Version snapshot is missing")}
+
+      {:version, nil} ->
+        {:error, invalid_restore_changeset(query, "Version not found")}
+    end
+  end
+
   @doc """
   Derives the SQL language from a backend ID.
 
@@ -253,6 +282,13 @@ defmodule Logflare.Endpoints do
 
   defp version_origin(%OauthAccessToken{id: id}), do: "API: id #{id}"
 
+  @spec invalid_restore_changeset(EndpointQuery.t(), String.t()) :: Ecto.Changeset.t()
+  defp invalid_restore_changeset(%EndpointQuery{} = query, message) do
+    query
+    |> change_query(%{})
+    |> Ecto.Changeset.add_error(:base, message)
+  end
+
   @spec lock_endpoint_query(EndpointQuery.t()) :: EndpointQuery.t()
   defp lock_endpoint_query(%EndpointQuery{id: query_id}) do
     from(query in EndpointQuery,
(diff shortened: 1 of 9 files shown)


lib/logflare/endpoints.ex
```diff
@@ -169,6 +169,35 @@ defmodule Logflare.Endpoints do
...
+          {:ok, EndpointQuery.t(), integer()}
+          | {:error, Ecto.Changeset.t()}
+  def restore_query_version(%EndpointQuery{id: endpoint_id} = query, version_number, origin)
+      when is_integer(version_number) do
+    with {:version, %Version{id: version_id, meta: meta}} <-
+           {:version, get_endpoint_query_version(endpoint_id, version_number)},
+         {:restorable, true} <-
+           {:restorable, version_id != current_endpoint_version_id(endpoint_id)},
+         {:snapshot, snapshot} when is_map(snapshot) and map_size(snapshot) > 0 <-
+           {:snapshot, Map.get(meta, "endpoint_snapshot")},
```
amokan: Nit - If for some weird reason `meta` were to be nil, this would raise rather than flow down to the "Version snapshot is missing" branch.

You could introduce a couple private functions (_see below_) to pattern match on that and then  replace 182 with `{:snapshot, endpoint_snapshot(meta)},` 🤷‍♂️ 

```elixir
defp endpoint_snapshot(%{"endpoint_snapshot" => snapshot}), do: snapshot
defp endpoint_snapshot(_meta), do: nil
```

amokan (approved):
Minor nit - but other than that it looks good!

lib/logflare/endpoints.ex
amokan: @msmithstubbs check those depthfirst 🤖 reviews up above, which I didn't notice until after my review. Seem plausible, but feel free to call out if not an issue.

lib/logflare/endpoints.ex
```diff
@@ -169,6 +169,35 @@ defmodule Logflare.Endpoints do
...
 
+  @spec restore_query_version(EndpointQuery.t(), integer(), origin()) ::
+          {:ok, EndpointQuery.t(), integer()}
+          | {:error, Ecto.Changeset.t()}
+  def restore_query_version(%EndpointQuery{id: endpoint_id} = query, version_number, origin)
+      when is_integer(version_number) do
+    with {:version, %Version{id: version_id, meta: meta}} <-
+           {:version, get_endpoint_query_version(endpoint_id, version_number)},
+         {:restorable, true} <-
+           {:restorable, version_id != current_endpoint_version_id(endpoint_id)},
```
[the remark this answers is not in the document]
msmithstubbs: I'm not sure this is a useful improvement. Restoring and endpoint creates a new version so the history will show what has happened.

lib/logflare/endpoints.ex
msmithstubbs: Thanks, implemented as you suggested and addressed high and medium depthfirst issues.

I'm not sure https://github.com/Logflare/logflare/pull/3653#discussion_r3653661856 is worth addressing but let me know if you feel otherwise.

lib/logflare_web/live/endpoints/endpoints_versions_live.ex
```diff
@@ -215,7 +262,51 @@ defmodule LogflareWeb.EndpointsVersionsLive do
...
 
+  def handle_event("restore-version", %{"version-number" => version_number}, socket) do
+    endpoint = socket.assigns.endpoint
+    user = socket.assigns.team_user || socket.assigns.user
+
+    case Integer.parse(version_number) do
+      {version_number, ""} ->
+        socket =
+          socket
+          |> start_async(:version_restored, fn ->
```
Ziinc, line 273: If this is done async then why use an event listener as opposed do sending a flash here?

lib/logflare_web/live/endpoints/endpoints_versions_live.ex
```diff
@@ -27,14 +27,29 @@ defmodule LogflareWeb.EndpointsVersionsLive do
         </.subheader_path_link>/versions
       </:path>
     </.subheader>
-
+    <script>
+      window.addEventListener("phx:hide-restoring-indicator", () => {
```
Ziinc, line 31: I didn't catch this indicator in the demo 😂 I think a flash to follow the rest of the ui would be more obvious

Ziinc (commented):
I think the inline restoring causes the UI to jump too much and be too janky, probably better to use a flash for feedback and add a transition for the new listitem to slide in (if possible)

lib/logflare_web/live/endpoints/endpoints_versions_live.ex
msmithstubbs: Flash is already in place, so I'll follow up with a PR that removes the loading banner and uses a slide in animation.

PR #3655, opened 2026-07-01
fix: address credo performance offenses

nelsonmestevao:


test/logflare/opentelemetry_test.exs
```diff
@@ -51,7 +51,7 @@ defmodule Logflare.OpenTelemetryTest do
       Pipeline.handle_batch(:bq, messages, batch_info, context)
 
       spans = collect_spans()
-      assert length(spans) > 0
+      assert is_list(spans) and spans != []
```
amokan: Minor Nit...

May be worth doing something like an `assert [_|_] = spans` for a bunch of these `is_list(foo) and foo != []` assertions.

If you feel better about a guard-based approach, I have no issue with you adding an `is_non_empty_list` guard to `Logflare.Utils.Guards`.

amokan (approved):
Looks good. Only a minor nit around some of the list-based assertions.

test/logflare/opentelemetry_test.exs
nelsonmestevao: I like your suggestion of a `is_non_empty_list/1` guard 🙏 I will follow that one 💪

nelsonmestevao: Done in last push 👍

PR #3662, opened 2026-07-03
Feat/s3 producer pipeline

Baishan:
- Add a spool-based log ingestion pipeline: producer side buffers/uploads batches to S3/GCS, consumer side downloads and replays them back through the normal ingest path
- Add pluggable queue (SQS, GCP Pub/Sub) and storage (S3, GCS) adaptors behind `Queue`/`Storage` behaviours for the spool pipeline
- Add `enable_spooling` flag on sources (migration + schema + edit-source UI toggle) to opt sources into the spool path
- Add `MemoryMonitor` GenServer for spool backpressure: tracks system memory pressure and per-source destination-queue backlog, throttling the consumer instead of dropping events
- Extend `IngestEventQueue`/`BufferProducer`/`QueueJanitor` with stale `:processing` cleanup and startup-queue migration for spool producers
- Add telemetry throughout the spool pipeline (throttling, queue depth, stale-event handling)
- Add a dev-only LiveView dashboard (`DevDashboard`) for inspecting spool/ingest state locally
- Add GCS/staging provisioning (Cloudbuild config, lifecycle policy) and local docker-compose setups for MinIO/GCP emulation
- Add `scripts/replay_spool_files.exs` for manually replaying spooled files

closes O11Y-2139 and O11Y-2140


lib/logflare/backends/spool/producer_pipeline.ex
```diff
@@ -0,0 +1,398 @@
...
+  def transform(event, _opts) do
+    %Message{
+      data: event,
+      acknowledger: {__MODULE__, :no_ack_ref, :ack_data}
+    }
+  end
+
+  @impl Broadway.Acknowledger
+  def ack(_ref, successful, failed) do
+    for %{data: {id, tid, _size}} <- successful ++ failed do
```
djwhitt: This deletes failed messages as well as successful ones. In this producer, a storage `put` error turns the batch messages into `Message.failed/2`, so Broadway will still call this ack callback and the ETS rows get deleted even though no S3/GCS object was written. There is a similar loss/stranding case when the upload succeeds but `notify_queue/4` fails: the batch is returned as successful, so these rows are deleted even though no SQS/PubSub message points the consumer at the object. Failed uploads should be left retryable/requeued, and queue publish failures should not ack/delete the events as successfully spooled.

lib/logflare/backends/spool/producer_pipeline.ex
```diff
@@ -0,0 +1,398 @@
...
+
+    emit_storage_put_telemetry(:ndjson_gz, byte_size(body), result)
+    result
+  end
+
+  defp upload_plain(messages, bucket, file_key, storage_mod) do
+    body =
+      Enum.flat_map(messages, fn %{data: {id, tid, _size}} ->
+        case :ets.lookup(tid, id) do
+          [{^id, _status, log_event, _byte_size}] -> [encode_line(log_event), "\n"]
```
djwhitt: This pattern still matches the old 4-tuple ETS row shape, but `IngestEventQueue.add_to_table/3` now inserts 6-tuples (`{id, status, event, size, claim, claimed_at}`) after the stale-processing/claim changes from main. With real queued spool events this lookup will return a 6-tuple and fall through to a `CaseClauseError`, so the producer cannot serialize batches. The current producer test misses this because it manually inserts `{id, :pending, log_event, byte_size}` into ETS instead of using `IngestEventQueue.add_to_table/2`; updating the fixture to use the real queue path should catch this.

lib/logflare/backends/spool/producer_pipeline.ex
djwhitt: Is this best-effort behavior intended only while the spool path is being tested with non-critical sources?

lib/logflare/backends.ex
```diff
@@ -589,11 +589,39 @@ defmodule Logflare.Backends do
...
+  @doc """
+  Dispatches events from the spool consumer directly to backends, bypassing the spool producer path.
+  Use this in the consumer pipeline to avoid re-routing events back to the spool in `:both` mode.
+  """
+  @spec dispatch_from_spool([map()], Source.t()) :: {:ok, non_neg_integer()}
+  def dispatch_from_spool(spool_records, source) do
+    ensure_source_sup_started(source)
+    log_events = Enum.map(spool_records, &LogEvent.make_from_spool(&1, source))
+    count = length(log_events)
+    increment_counters(source, count)
```
djwhitt: Do we want to increment the normal source counters again when replaying from spool? `ingest_logs/3` already calls `increment_counters/2` before sending spooled sources to the producer, so this appears to double-count every spooled event in source rates/total logs. If the intent is to track replay volume separately, could this be separate spool-consumer telemetry instead of the normal ingest counters?

lib/logflare/backends.ex
```diff
@@ -589,11 +589,39 @@ defmodule Logflare.Backends do
     {log_events, errors} = split_valid_events(source, event_params)
     count = Enum.count(log_events)
     increment_counters(source, count)
-    maybe_broadcast_and_route(source, log_events)
-    dispatch_to_backends(source, backend, log_events)
+
+    if spool_producer_mode?() and source.enable_spooling do
```
djwhitt: I think this branch can break explicit-backend/rule routing, especially in `SPOOL_MODE=both`. `SourceRouter` routes backend rules by calling `Backends.ingest_logs([le], source, backend)` after setting `via_rule_id`, but this branch ignores `backend` and writes only source/event fields to the spool. On a node running both producer and consumer, `dispatch_from_spool/2` calls `maybe_broadcast_and_route/2`, the backend-rule path calls back into `ingest_logs/3`, and this branch can re-spool the routed event instead of dispatching it to the target backend; because `via_rule_id`/backend target are not persisted in the spool record, it can be replayed as a fresh source-wide event again. Should explicit-backend/routed ingests bypass spooling, or should the spool record preserve the intended backend/rule metadata?

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -0,0 +1,445 @@
...
+    if state.poll_timer, do: Process.cancel_timer(state.poll_timer)
+    %{state | poll_timer: Process.send_after(self(), :poll, delay)}
+  end
+
+  defp buffered?(%{current: nil}), do: false
+  defp buffered?(%{current: %{lines: []}}), do: false
+  defp buffered?(_), do: true
+
+  defp maybe_ack_exhausted(%{current: %{lines: [], handle: handle}} = state) do
+    ack_and_notify(state.queue_mod, state.queue_url, handle, :buffer_exhausted)
```
djwhitt, line 129: This looks like the queue message is acked once the spool file has been fully emitted into Broadway, rather than after the records have been accepted into the downstream `IngestEventQueue` by `dispatch_from_spool/2`. I can see that being an intentional initial tradeoff given the existing ingest path also only guarantees handoff into local queues, but could we document the intended ack boundary here? It would help clarify whether SQS/PubSub ack is meant to mean "file downloaded/emitted" vs "records accepted into local ingest buffers".

djwhitt: Heh, that comment probably could have used some deslopping. I'm guessing this is intended initially for the MVP, but just wanted to confirm.

lib/logflare/backends/spool/producer_pipeline.ex
Baishan: I'ved added in the logic to requeue these but I want retries to 0 for initial deployment like the other pipelines.

lib/logflare/backends/spool/producer_pipeline.ex
Baishan: Nice catch, I merged main into it and this was uncaught. Updated the tests.

lib/logflare/backends.ex
Baishan: This should be removed. I remember removing it before 🤷

lib/logflare/backends.ex
Baishan: This is a big one that could have resulted in an infinite loop. I've added a param to only allow spooling form the initial http requests and some redundant checks to make sure its not being respooled.

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
Baishan: For first few milestones of this we will ack when delivered to broadway and control memory pressure to prevent ooms never overloading the queues. A complete end to end ack is difficult because a batch is fanned out to many sources and needs planning as its a much bigger change to how the pipelines are structured. I'm thinking of some kind of write ahead log that persists crashes or a reference counting ack system.

lib/logflare/backends/spool/queue/sqs.ex
```diff
@@ -0,0 +1,61 @@
...
+
+        {:ok, normalized}
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+  @impl Logflare.Backends.Spool.Queue
+  def ack(queue_url, handle) do
```
djwhitt, line 33: `ack/2` always returns `:ok`, which also masks real `DeleteMessage` failures. Can we ignore the ElasticMQ parse-error case while still returning `{:error, reason}` for actual delete failures?

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -0,0 +1,445 @@
...
+  end
+
+  defp handle_received_message(handle, body, bucket, queue_url, queue_mod, storage_mod) do
+    case Jason.decode(body) do
+      {:ok, %{"file_key" => file_key}} when is_binary(file_key) ->
+        case download_and_parse(bucket, file_key, storage_mod) do
+          {:ok, lines} ->
+            {:ok, handle, lines}
+
+          {:error, %Tesla.Env{status: 404}} ->
```
djwhitt: This only catches the GCS/Tesla 404 shape. S3/ExAws missing objects come back as `{:http_error, 404, ...}` / `NoSuchKey`, so stale SQS messages may be nacked forever. Could the storage adapters normalize missing objects to something like `{:error, :not_found}` and have the consumer handle that?

djwhitt (approved):

amokan (approved):
Ok. Only points I found any potential concerns are covered by @djwhitt's comments.

But really I have no issue in seeing this get tested. We can follow-up with adjustments or fixes.

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
Baishan: Done

lib/logflare/backends/spool/queue/sqs.ex
Baishan: done

PR #3664, opened 2026-07-06
feat: add Management API endpoint for alerts (Api.AlertController)

djwhitt:
## Summary

Adds Management API CRUD for `Logflare.Alerting.AlertQuery` at `/api/alerts`, enabling Terraform-provider coverage for alerts under O11Y-2143.

The alert resource's list and lookup paths follow existing team-access semantics; backend attachment has the limitation below.

### Backend attachment contract

- Create and update accept `backend_ids`, a complete replacement set of positive integer backend IDs.
- IDs are deduplicated and must belong directly to the authenticated user.
- `backends` is response-only and returns the attached backend objects. Supplying `backends` in a request returns `400`; use `backend_ids` instead.
- Invalid `backend_ids` return `400` rather than raising.

Alerts now preload and serialize attached backends in all response-producing actions, so Terraform can reconcile the association.

### OpenAPI

- Separate create and update request schemas: create requires alert fields; update permits partial updates.
- Documents `backend_ids` only on write schemas and `backends` only on the response schema.
- Correctly documents nullable optional fields and `400`, `404`, and `422` responses.

### Known limitation / follow-up

Alert list, fetch, update, and delete are team-access scoped, but `backend_ids` are resolved only among the token owner's directly owned backends. Consequently, a member can manage an existing team alert but cannot discover or attach that team's owner-owned backends through the Management API. `/api/backends` list and show have the same direct-owner scope.

This does not affect the owner-token Terraform use case. Team-aware backend discovery and attachment should be addressed together in a follow-up: make backend list/show and alert attachment use the existing backend access helpers, then cover team-owned attachment, unrelated-backend rejection, and team backend discovery.

### Stacking

#3663 has merged. This branch should be rebased onto `main` before final review so its already-merged changes are removed from this PR diff.

Implements O11Y-2143.

## Validation

- `../bin/test test/logflare_web/controllers/api/alert_controller_test.exs test/logflare/alerting_test.exs` — 47 tests and 1 doctest passing
- `../bin/x mix compile --warnings-as-errors`
- formatting check
- strict Credo on affected files

### DIFF
--- a/lib/logflare/alerting.ex
+++ b/lib/logflare/alerting.ex
@@ -102,6 +102,23 @@ defmodule Logflare.Alerting do
     |> Repo.one()
   end
 
+  @doc """
+  Fetches a single alert query by arbitrary conditions, scoped to the
+  alert queries the user has team access to.
+  """
+  @spec fetch_alert_query_by_user_access(User.t() | TeamUser.t(), keyword()) ::
+          {:ok, AlertQuery.t()} | {:error, :not_found}
+  def fetch_alert_query_by_user_access(user_or_team_user, kw) when is_list(kw) do
+    AlertQuery
+    |> Teams.filter_by_user_access(user_or_team_user)
+    |> where([alert_query], ^kw)
+    |> Repo.one()
+    |> case do
+      nil -> {:error, :not_found}
+      alert_query -> {:ok, alert_query}
+    end
+  end
+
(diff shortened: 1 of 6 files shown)


djwhitt:
Superseded by #3682. The rebased alert API branch now lives in Logflare/logflare.

PR #3666, opened 2026-07-08
feat: S3 Tables backend

bblaszkow06:
Adds a backend adaptor that writes ingested logs, metrics, and traces to **AWS S3 Tables (Apache Iceberg)**.

## Overview

image

## Implementation

- **Rustler NIF (`native/s3_tables_ex`)** wrapping the `iceberg-rust` library and `iceberg-catalog-s3tables` — handles catalog init, table provisioning, and Iceberg snapshot commits. Rows arrive as NDJSON and are decoded into Arrow `RecordBatch`es, then written as Parquet data files.
- **Consolidated Broadway pipeline** (`s3_tables_adaptor/pipeline.ex`) — one pipeline per backend across all sources. Events are batched by `{event_type, day_bucket}` so each append targets a single type-specific Iceberg table (`otel_logs`/`otel_metrics`/`otel_traces`) and, typically, one day partition. Failed batches are requeued with a bounded retry counter and dropped once exhausted.
- **`CatalogManager`** — a per-backend GenServer that provisions tables on boot and caches the catalog resource in the `Registry` value for lock-free hot-path reads (with column-drift warnings for existing tables).
- **Based on the ClickHouse pipeline** — the batching, mapping, and requeue logic mirror the ClickHouse consolidated adaptor. The shared OTEL-column fixups (trace-duration computation, severity-number resolution) were extracted from the ClickHouse pipeline into `Logflare.Mapper.PostProcess`, now used by both.

### Non-blocking NIFs

The Iceberg operations (catalog load, Parquet upload, snapshot commit with retries) are executed on a shared tokio runtime. However, to avoid blocking any scheduler, since Rustler doesn't support yielding NIFs, each NIF returns `:ok` **immediately**. The result is sent via a message from the tokio runner thread to a caller

## Known limitations

- Degraded per-source query performance - missing write-side sort ([O11Y-2215](https://linear.app/supabase/issue/O11Y-2215/source-level-tenancy-in-s3-tables-backend-write-side-sort-declared))
- While AWS cleans up stale data, the removal after the retention period must be handled by the application [O11Y-2210](https://linear.app/supabase/issue/O11Y-2210/data-retention-for-s3-tables-backend)
- Refactor opportunities: [O11Y-2195](https://linear.app/supabase/issue/O11Y-2195/extract-shared-consolidated-pipeline-ackretrydrop-logic-clickhouse-s3) [O11Y-2171](https://linear.app/supabase/issue/O11Y-2171/generalize-clickhouse-circuitbreaker-for-reuse-across-backend-adaptors)


lib/logflare/backends/adaptor/s3_tables_adaptor.ex
```diff
@@ -0,0 +1,119 @@
+defmodule Logflare.Backends.Adaptor.S3TablesAdaptor do
+  @moduledoc """
+  Backend adaptor that writes batches of logs to AWS S3 Tables (Apache Iceberg).
+
+  Runs consolidated: one adaptor tree per backend across all sources (see
+  `Logflare.Backends.ConsolidatedSup`). Ingestion is not yet implemented — the
+  pipeline drains batches without writing them.
```
Ziinc: this looks outdated

lib/logflare/backends/adaptor/s3_tables_adaptor.ex
```diff
@@ -0,0 +1,119 @@
...
+  alias __MODULE__.Pipeline
+  alias Ecto.Changeset
+  alias Logflare.Backends
+  alias Logflare.Backends.Adaptor
+  alias Logflare.Backends.Backend
+
+  @behaviour Adaptor
+
+  @min_batch_timeout 1_000
+  @max_batch_timeout 5_000
```
Ziinc: probably want a much larger batch timeout for the pipelines, especially for a consolidated ingest

bblaszkow06: These values come from the existing S3 pipeline, ClickHouse uses 5 seconds without configurability.
I don't have an intuition on how many events per second might be expected in production. What's your take on the timeout values? For the time being I've raised the min/max to 5sec/60sec
Should I remove the configurability of the timeout?

lib/logflare/backends/adaptor/s3_tables_adaptor.ex
```diff
@@ -0,0 +1,123 @@
...
+
+  @doc false
+  @impl Adaptor
+  def cast_config(%{} = params, existing_config \\ %{}) do
+    types = %{
+      table_bucket_arn: :string,
+      namespace: :string,
+      access_key_id: :string,
+      secret_access_key: :string,
+      batch_timeout: :integer
```
Ziinc, line 55: we should set a default for the user, like 30s

Ziinc: i would consider `:batch_timeout` as an optional tuning parameter as an advanced option and should not be required for configuring initial backend.

lib/logflare/backends/adaptor/s3_tables_adaptor.ex
```diff
@@ -0,0 +1,123 @@
...
+  """
+  @impl Adaptor
+  @spec test_connection(Backend.t()) :: :ok | {:error, term()}
+  def test_connection(%Backend{} = backend) do
+    config = Adaptor.get_backend_config(backend)
+
+    case Native.init_catalog(config) do
+      {:ok, _catalog} -> :ok
+      {:error, reason} -> {:error, reason}
+    end
```
Ziinc, line 96: nit: can be a with/do

lib/logflare/backends/adaptor/s3_tables_adaptor/native.ex
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  defp wrap_sending_nif(nif_fun, timeout \\ @default_timeout) do
+    ref = make_ref()
+    :ok = nif_fun.(ref)
+
+    receive do
+      {^ref, result} -> result
+    after
+      # TODO: this may leak late messages to a caller, consider spawning a task
```
Ziinc, line 69: this is fine, Task would be more expensive.

lib/logflare/backends/adaptor/s3_tables_adaptor/native.ex
```diff
@@ -0,0 +1,72 @@
+defmodule Logflare.Backends.Adaptor.S3TablesAdaptor.Native do
+  @moduledoc false
+  # Public interface over the `s3_tables_ex` NIF.
+  # The raw NIF stubs live in `Native.Nifs`; keeping them separate allows
+  # mocking with Mimic
+
+  alias Logflare.Backends.Adaptor.S3TablesAdaptor.Native.Nifs
+
+  @default_timeout 5_000
```
Ziinc:
```suggestion
  @default_timeout 10_000
```

lib/logflare/backends/adaptor/s3_tables_adaptor/pipeline.ex
```diff
@@ -0,0 +1,278 @@
...
+  alias Logflare.Mapper.PostProcess
+  alias Logflare.Utils
+
+  @behaviour Broadway
+  @behaviour Broadway.Acknowledger
+
+  @producer_concurrency 1
+  @processor_concurrency 5
+
+  @max_retries 2
```
Ziinc:
```suggestion
  @max_retries 1
```
if value is too high can result in oom on backend non-responsiveness

lib/logflare/backends/pipeline/batch_splitter.ex
```diff
@@ -0,0 +1,50 @@
...
+  @spec max_batch_size() :: pos_integer()
+  def max_batch_size, do: @max_batch_size
+
+  @doc """
+  Returns the `{acc, fun}` tuple for a Broadway batcher's `:batch_size` option.
+  """
+  @spec build() :: {tuple(), (Message.t(), tuple() -> {:emit | :cont, tuple()})}
+  def build do
+    {
+      {@max_batch_size, @max_batch_length},
```
Ziinc: we can probably make this configurable, for s3 larger payloads would likely be preferable.

lib/logflare/backends/adaptor/s3_tables_adaptor/catalog_manager.ex
```diff
@@ -0,0 +1,146 @@
...
+          String.t(),
+          Backend.t()
+        ) :: :ok
+  defp warn_on_schema_drift(catalog, event_type, table_name, backend) do
+    with {:ok, info} <- Native.table_info(catalog, table_name),
+         :ok <- check_version(info, event_type) do
+      :ok
+    else
+      {:error, :version_mismatch, error_meta} ->
+        metadata = [backend_id: backend.id, table_name: table_name] ++ error_meta
```
Ziinc: user_id

Ziinc: backend_uuid

lib/logflare/backends/adaptor/s3_tables_adaptor/catalog_manager.ex
```diff
@@ -0,0 +1,146 @@
...
+        Logger.warning("S3 Tables Iceberg table schema mismatch", metadata)
+
+      {:error, reason} ->
+        Logger.warning("S3 Tables Iceberg table schema check failed",
+          backend_id: backend.id,
```
Ziinc: same needs more meta

lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -347,6 +347,37 @@
...
+            <small class="form-text text-muted">
+              The Iceberg namespace to ingest into.
+            </small>
+          </div>
+          <div class="form-group">
+            {label(f_config, :batch_timeout, "Batch Timeout")}
+            {text_input(f_config, :batch_timeout, class: "form-control", type: "number", value: 5_000)}
+            <small class="form-text text-muted">
+              The batch timeout in milliseconds.
+            </small>
```
Ziinc, line 439: should be optional config with default set. i think 5s for default is too short.

test/logflare/backends/adaptor/s3_tables_adaptor_test.exs
```diff
@@ -0,0 +1,196 @@
...
+          S3TablesAdaptor,
+          Map.put(@valid_config, :batch_timeout, 60_001)
+        )
+
+      assert %Ecto.Changeset{valid?: false} = cs
+
+      assert {_message, [validation: :number, kind: :less_than_or_equal_to, number: 60_000]} =
+               cs.errors[:batch_timeout]
+    end
+  end
```
Ziinc: can consloidate into a for/do

test/logflare/backends/adaptor/s3_tables_adaptor_test.exs
```diff
@@ -0,0 +1,196 @@
...
+    test "invalid credentials" do
+      assert {:error, err} = S3TablesAdaptor.Native.init_catalog(@valid_config)
+      assert err =~ "invalid"
+    end
+
+    setup do
+      table_bucket_arn = System.fetch_env!("LOGFLARE_S3_TABLES_TEST_BUCKET_ARN")
+      namespace = System.fetch_env!("LOGFLARE_S3_TABLES_TEST_NAMESPACE")
+      access_key_id = System.fetch_env!("AWS_ACCESS_KEY_ID")
+      secret_access_key = System.fetch_env!("AWS_SECRET_ACCESS_KEY")
```
Ziinc, line 223: why is env var usage here?

test/logflare/backends/adaptor/s3_tables_adaptor_test.exs
```diff
@@ -0,0 +1,196 @@
...
+        1..2
+        |> Task.async_stream(
+          fn n ->
+            row = %{
+              "id" => Ecto.UUID.generate(),
+              "event_message" => "concurrent append #{n}",
+              "timestamp" => now_ns
+            }
+
+            S3TablesAdaptor.Native.append_batch(catalog, table_name, Jason.encode!(row) <> "\n")
```
Ziinc, line 330: instead of testing the nif directly, one method to not test the implementation detail is to start a mock api bandit server as part of the test supervision and hook into that pipeline

test/logflare/backends/adaptor/s3_tables_adaptor/pipeline_test.exs
```diff
@@ -0,0 +1,295 @@
...
+        acknowledger: {Pipeline, :ack_id, %{backend_id: @backend_id}},
+        status: {:failed, :commit_conflict}
+      }
+
+      test_pid = self()
+
+      Mimic.expect(IngestEventQueue, :delete_batch, fn {:consolidated, @backend_id}, events ->
+        send(test_pid, {:deleted_batch, events})
+        :ok
+      end)
```
Ziinc: not a fan of mocking internal modules. feels like a lot of these tests in this mod should be part of S3TablesAdaptorTest

test/logflare/backends/adaptor/s3_tables_adaptor/iceberg_schema_test.exs
```diff
@@ -0,0 +1,79 @@
...
+        assert field.type in @allowed_types
+      end
+    end
+
+    test "every event type has a required timestamp field typed timestamptz" do
+      for event_type <- IcebergSchema.event_types() do
+        assert %{type: "timestamptz", required: true} =
+                 Enum.find(IcebergSchema.fields(event_type), &(&1.name == "timestamp"))
+      end
+    end
```
Ziinc: can be combined

lib/logflare/backends/adaptor/s3_tables_adaptor/native/nifs.ex
```diff
@@ -0,0 +1,19 @@
...
+  use Rustler, otp_app: :logflare, crate: "s3_tables_ex"
+
+  def init_catalog(_ref, _config), do: :erlang.nif_error(:nif_not_loaded)
+
+  def ensure_table(_ref, _catalog, _table_name, _fields, _properties),
+    do: :erlang.nif_error(:nif_not_loaded)
+
+  def table_info(_ref, _catalog, _table_name), do: :erlang.nif_error(:nif_not_loaded)
+
+  def drop_table(_ref, _catalog, _table_name), do: :erlang.nif_error(:nif_not_loaded)
```
Ziinc, line 13: is this for convenience or iex usage?? i don't see where this is used. if its for actual usage then we should put it at S3TablesAdaptor level

native/s3_tables_ex/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+        match self {
+            AppendError::CommitConflict => atoms::commit_conflict().encode(env),
+            AppendError::Timeout => atoms::timeout().encode(env),
+            AppendError::Other(message) => message.encode(env),
+        }
+    }
+}
+
+/// Decodes newline-delimited JSON rows into Arrow record batches, writes them
+/// as Iceberg parquet data files (fanned out per day partition), and commits a
```
Ziinc: hmm wonder if we need to partition by other fields, or to make the partitioning more configurable...

Ziinc: we can leave it as day for first iteration for now.

native/s3_tables_ex/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+/// fast-append transaction. Commit conflicts are retried by the iceberg crate
+/// itself (bounded by the `commit.retry.*` table properties); exhaustion
+/// surfaces as `{:error, :commit_conflict}`.
+///
+/// Contract: integer values in `timestamptz` columns are interpreted as unix
+/// **nanoseconds** (the unit the mapper emits) and are scaled down to the
+/// microseconds Iceberg stores; RFC3339 strings are accepted in any unit.
```
Ziinc: ideally we should be able to tell Mapper that we can only accept microseconds so that we don't do the decode again just to convert this value

native/s3_tables_ex/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+) -> Result<AppendOk, AppendError> {
+    let table_ident = catalog.table_ident(table_name);
+    let table = catalog.catalog.load_table(&table_ident).await?;
+
+    let iceberg_schema = table.metadata().current_schema().clone();
+    let arrow_schema = Arc::new(schema_to_arrow_schema(&iceberg_schema)?);
+    // arrow-json takes raw JSON integers in a timestamp column as already
+    // being in the column's unit, but callers send mapper-produced
+    // nanoseconds -- so decode against a nanosecond variant of the schema
+    // and cast down to the table's microsecond columns afterwards
```
Ziinc: pretty sure we can do this at the mapper side instead while it is getting built

native/s3_tables_ex/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+    let row_count: u64 = batches.iter().map(|batch| batch.num_rows() as u64).sum();
+
+    if row_count == 0 {
+        return Ok(AppendOk {
+            row_count: 0,
+            data_files: 0,
+        });
+    }
+
+    let splitter = RecordBatchPartitionSplitter::try_new_with_computed_values(
```
Ziinc, line 341: i think we should do the partitioning at the Broadway level to avoid having to work with multiple partitions here, which could result in small files.

I also think that we can avoid dealing with multiple recordbatches if we are able to just directly work with only one batch at a time (i.e  ndjson -> 1 record batch -> 1 parquet file)

native/s3_tables_ex/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+        DataType::List(child) => DataType::List(Arc::new(to_nanosecond_field(child))),
+        DataType::LargeList(child) => DataType::LargeList(Arc::new(to_nanosecond_field(child))),
+        other => other.clone(),
+    }
+}
+
+/// Casts a batch decoded with the nanosecond schema back to the table's
+/// arrow schema, scaling timestamp columns down to microseconds and
+/// restoring the Iceberg field-id metadata the writers rely on.
+fn cast_batch(
```
Ziinc: can be avoided by directly adjusting Mapper behaviour

Ziinc (commented):
we can do the Mapper adjustments in a followup since this PR is quite massive.

Ziinc (changes requested):
Lets leave out the Mapper changes for now, we should be good to merge once the pipeline adjustments are in. Preferably we can do the partitioning on elixir side.

lib/logflare/backends/adaptor/s3_tables_adaptor.ex
Ziinc: We should remove the timeout configuration, should be handled by system only

Ziinc: For consolidated pipelines we generally want very large batches, like 10k-50k events. Having a max of 25k let's us optimize for larger file sizes while keeping a ceiling on the payload size. We csn tune it further later

Ziinc: 20k events for 3s batch window is fine

bblaszkow06:
Closed in favour of https://github.com/bblaszkow06/logflare/pull/26, with mapper fixes in https://github.com/Logflare/logflare/pull/3882
and mock S3 Tables server in https://github.com/bblaszkow06/logflare/pull/27

PR #3667, opened 2026-07-08
perf: reduce process heap allocations for clickhouse pipeline

Baishan:
Uses id passing in the clickhouse pipeline.
Perform incremental compression so the entire uncompressed batch never needs to be held in the process heap during compression.

Notable reduction of long_gc events and a drop in memory. Slight increase in cpu is expected.


lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -741,6 +742,47 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+  """
+  @spec reset_processing_to_pending(table_key() | consolidated_table_key()) :: :ok
+  def reset_processing_to_pending(sid_bid_pid) do
+    ms =
+      Ex2ms.fun do
+        {id, :processing, _event, _size, _claim, _claimed_at} -> id
+      end
+
+    with tid when tid != nil <- get_tid(sid_bid_pid),
+         {ids, _cont} <- :ets.select(tid, ms, 10_000) do
```
amokan: I'm struggling to understand this `10_000`. What happens to the rest of the items in this scenario?

amokan (approved):
I'm down with giving this a go - but the 10k cap in the `reset_processing_to_pending` func is a head-scratcher.

The reality is that we just let these fail now, so it is better - but something we'd want to follow-up another PR for to fix long-term if this does stick.

Lets get the CI situation ironed out before merging though 🙏

lib/logflare/backends/ingest_event_queue.ex
Baishan: I've removed this. I think it was added during debugging and I didn't catch it here.

PR #3668, opened 2026-07-09
fix: eliminate ClickHouse pipeline CPU regression

djwhitt:
## Summary

Stacked on #3667. This keeps the ClickHouse ID-passing + streaming compression memory savings while recovering CPU:

- hoists `mapping_config_id` UUID encoding once per batch for streamed RowBinary rows
- has the consolidated producer emit ID + small routing metadata so ClickHouse avoids the extra `handle_message/3` ETS lookup
- keeps full `LogEvent` payloads in ETS until batch compression/ack
- adds targeted benchmark/profiling scripts for CPU and memory investigation

## CPU results

Fixed-work CPU runner, `BATCH_SIZE=1000 BATCHES=20`:

| scenario | old_queue runtime µs/row | optimized runtime µs/row |
| --- | ---: | ---: |
| log/small | 31.95 | 30.40 |
| log/realistic | 51.75 | 31.55 |
| metric/small | 72.85 | 66.25 |
| metric/realistic | 99.05 | 81.55 |
| trace/small | 57.85 | 49.30 |
| trace/realistic | 70.50 | 64.25 |

Larger 50k-row log realistic run:

- `old_queue`: 61.50 runtime µs/row
- optimized `id_queue_metadata`: 46.42 runtime µs/row

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
@@ -125,8 +125,8 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
     }
   end
 
-  def transform({id, tid, _size, _event_type, _day_bucket, _freshness}, opts) do
-    transform({id, tid, 0}, opts)
+  def transform({id, tid, size, _event_type, _day_bucket, _freshness}, opts) do
+    transform({id, tid, size}, opts)
   end
 
   def transform({id, tid, _size}, opts) do
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
@@ -11,8 +11,8 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
   `{event_type, day_bucket}` so each insert targets a single ClickHouse
   partition. Multiple sources are processed together in a single pipeline.
 
-  Uses ID-passing: the producer emits `{id, tid, size}` tuples and events
-  remain in ETS as `:processing` throughout. Each batcher fetches events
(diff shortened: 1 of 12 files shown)


lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -528,16 +528,55 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+          | {:error, :not_initialized}
+  def take_pending_ids_with_metadata(_, 0), do: {:ok, [], nil}
+
+  def take_pending_ids_with_metadata(sid_bid_pid, n) when is_integer(n) do
+    ms = [
+      {
+        {:"$1", :pending,
+         %{
+           __struct__: LogEvent,
+           event_type: :"$3",
```
Baishan, line 623: I didnt know you could do this with ets

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -103,8 +110,27 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
+          opts :: keyword()
+        ) :: Message.t()
+  def transform({id, tid, _size, event_type, day_bucket, freshness}, opts)
+      when is_event_type(event_type) and freshness in [:fresh, :stale] do
+    %Message{
+      data: {id, tid, event_type, day_bucket},
+      acknowledger: {__MODULE__, :ack_id, %{backend_id: opts[:backend_id]}}
+    }
+    |> Message.put_batcher(ingest_freshness_to_batcher(freshness))
+    |> Message.put_batch_key({event_type, day_bucket})
```
Baishan: this seems to be done twice now, once in transform and handle_message. It probably makes sense to keep it in handle_message as there are more processors than producers to spread the load.

Baishan: Oh i see the new handle message that will fire before the other one. Probably still makes sense to move this into handle_message and kill the old one.

PR #3672, opened 2026-07-10
fix(clickhouse): trip circuit breaker immediately on TOO_MANY_PARTS

djwhitt:
## Context

Incident analysis of a ClickHouse parts explosion (parts climbing through `parts_to_throw_insert` → error 252 `TOO_MANY_PARTS` → a large insert-retry storm). This PR addresses the part of that failure mode that is **correct to fix regardless of the still-unconfirmed root cause**.

## Changes

**Trip the circuit breaker immediately on `TOO_MANY_PARTS` (252)**

Once a partition is over `parts_to_throw_insert`, retrying inserts creates *more* parts and makes it worse. Previously a 252 was treated like any other failure and had to accumulate toward the breaker's 20-failures/30s threshold, while requeued retries kept hammering an already-throttled cluster.

- New synchronous `CircuitBreaker.trip/1` opens the breaker at once, bypassing the failure-count threshold, and does not return until `check/1` can see it open. `maybe_trip` is refactored into a shared `open_breaker/4` that tags the open reason (`:threshold` | `:forced`) in the log line and telemetry metadata.
- Insert failures now route through `record_insert_failure/2`: a `Code: 252` marker in the HTTP error body → `trip/1`; every other error keeps the existing `record_failure/1` counting behavior.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/circuit_breaker.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/circuit_breaker.ex
@@ -81,6 +81,33 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.CircuitBreaker do
   end
 
   @doc """
+  Immediately opens the breaker, bypassing the failure-count threshold.
+
+  Intended for failures that are self-amplifying under retry — notably a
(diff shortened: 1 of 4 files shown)


djwhitt:
Superseded by #3711, opened from the upstream Logflare/logflare repository.

PR #3674, opened 2026-07-10
feat(telemetry): expose batch flush trigger and per-backend missing IDs

djwhitt:
## Why

While investigating ClickHouse `TOO_MANY_PARTS` (error 252), we want to confirm two suspected drivers of small / under-filled inserts before changing insert behavior:

1. **How often do batches flush on `:timeout` vs `:size`?** A high `:timeout` share means batches ship before reaching `batch_size` → smaller parts.
2. **How many event IDs go missing from ETS between claim and batch-insert?** With the ID-passing pipeline, events stay in ETS as `:processing` and are only fetched at insert time; any ID gone by then is dropped, so the rows actually written can be far fewer than the batch Broadway sized.

Both signals were emitted in telemetry events but **not registered as metrics**, so nothing aggregated them. This PR wires them up. No behavior changes to the pipelines — telemetry only.

## Changes

- **`batch_trigger` → metadata.** It was in the `handle_batch` *measurements* map, where `Telemetry.Metrics` tags can't read it. Moved to metadata across all five emitters (clickhouse, bigquery, webhook, http_based, spool). `batch_size` stays in measurements; the dev dashboard (only measurements consumer) reads `batch_size` only and is unaffected.
- **New counter** `logflare.backends.pipeline.handle_batch.batch_trigger`, tagged `backend_type` + `batch_trigger` (`:size` | `:timeout` | `:flush`).
- **`missing_ids` now sliceable.** Added `backend_id`/`event_type` tags to the existing `sum` and added a `distribution` (per-batch missing counts). A `tag_values` fallback normalizes heterogeneous emitter metadata (ClickHouse carries `backend_id`+`event_type`; BigQuery carries `source_id`).

## Testing

- New regression test locks in the contract that `batch_trigger` is emitted in metadata (not measurements) and `batch_size` stays in measurements.
- `test/logflare/bigquery/pipeline_test.exs` passes (22/22).
- ClickHouse pipeline tests require a live ClickHouse and were not run in this environment.

## Follow-ups (not in this PR)

Depending on what the metrics show: server-side `async_insert` tuning / extending async to the fresh lane, excluding `:processing` rows from the overflow purge, or raising the fresh batch timeout.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
@@ -228,13 +228,14 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
   defp emit_batch_telemetry(batch_info, backend_id, event_type, batcher, day_bucket) do
     :telemetry.execute(
       [:logflare, :backends, :pipeline, :handle_batch],
-      %{batch_size: batch_info.size, batch_trigger: batch_info.trigger},
+      %{batch_size: batch_info.size},
       %{
         backend_type: :clickhouse,
         backend_id: backend_id,
         event_type: event_type,
         batcher: batcher,
-        day_bucket: day_bucket
+        day_bucket: day_bucket,
+        batch_trigger: batch_info.trigger
       }
     )
   end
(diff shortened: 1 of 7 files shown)


djwhitt:
Closing. You can infer this from batch size.

PR #3676, opened 2026-07-10
chore: fix time-of-day flaky test

pmbauer:
## Background

[example flaky run](<https://github.com/Logflare/logflare/actions/runs/29106216571/job/86407270550?pr=3675>)

`SearchLVTest` failed intermittently in CI.

* The `t:yesterday` case ingested an event fixed at yesterday noon.
* Ingestion [now drops events older than 24h](<https://github.com/Logflare/logflare/commit/f3395503dbcb69b45082018686d09a10df56786b>), so whenever the test ran after \~12:00 UTC, yesterday-noon was >24h old and silently dropped with only 1 of 2 expected events landed.
* It passed in the morning, failed in the afternoon (UTC)

## Change

* Compute the `t:yesterday` fixture timestamp as the midpoint between the staleness boundary (`now - max_event_age_us`) and yesterday's end, so it's always both within "yesterday" and inside the 24h ingest ingestion window
* if we roll back to a 72h cutoff, we clamp the lower bound for the midpoint comp to yesterday's start, so we always land in yesterday IFF the cutoff is >= 24H.
* Expose `Backends.max_event_age_us/0` and `Backends.max_future_event_us/0` to avoid a hard-coded 24h boundary

> \[!note\]
> This test could break again if we drop the ingestion cutoff below 24 hours, so I'm not crazy about this fix, but it will fix the flake for now.  If we ever do drop below 24 hours, the premise of this test needs revisited anyway.

### DIFF
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -46,6 +46,21 @@ defmodule Logflare.Backends do
   def max_buffer_queue_len, do: @max_pending_buffer_len_per_queue
 
   @doc """
+  Returns the maximum age, in microseconds, of an event that will be accepted at
+  ingestion. Events older than this are dropped as stale.
+  """
+  @spec max_event_age_us() :: non_neg_integer()
+  def max_event_age_us, do: @max_event_age_us
+
+  @doc """
+  Returns the maximum time, in microseconds, that an event's timestamp may be in
+  the future and still be accepted at ingestion. Events further ahead than this
+  are dropped.
+  """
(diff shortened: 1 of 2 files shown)


amokan (approved):
Only nit would be that `Backends.max_future_event_us/0` is not used anywhere - but I think I prefer the symmetry of having both public functions in place.

Other than that, totally agree that this change helps that test be a bit more resilient and flexible no matter which way we go with bounds on the incoming events.

Longer-term, it would make sense to not have that LV search test even be driven by running data through the ingest layer though.

PR #3677, opened 2026-07-10
Fix/loadbalance ets queus

Baishan:
Queues where always round robined in the same order once they scaled up. This eventually results in uneven queue distribution and potentially some queues never getting load. 

The fix is to order by queue size before sending to the round robin function.

Also increase max demand for clickhouse processors so they dont need to make as many requests to ets.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
@@ -85,7 +85,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
         concurrency: @producer_concurrency
       ],
       processors: [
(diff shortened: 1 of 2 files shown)


djwhitt (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -85,7 +85,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
         concurrency: @producer_concurrency
       ],
       processors: [
-        default: [concurrency: @processor_concurrency, min_demand: 1, max_demand: 100]
+        default: [concurrency: @processor_concurrency, min_demand: 100, max_demand: 1000]
```
amokan, line 88: Let's move these `min_demand` and `max_demand` values to module attributes 🙏

PR #3680, opened 2026-07-10
chore: use `:broadway` v1.3 over forked copy

amokan:
_This is one we'll want to monitor on staging/canary, but this seems like a solid and low-risk move._

We've been using our [own fork of Broadway](https://github.com/Logflare/broadway/tree/fix/batcher-fullsweep-after) for some time, but I noticed that the pinned branch we were on is actually in the canonical Broadway repo and is included in the latest v1.3 release.

Delta between our forked/pinned branch and Broadway v1.3 can be found here - https://github.com/Logflare/broadway/compare/fix/batcher-fullsweep-after...elixir-broadway:broadway:v1.3.0

[Here](https://github.com/elixir-broadway/broadway/pull/350) is Ziinc's merged PR into Broadway from the pinned version we were on.

### Why Switch To v1.3?

A couple standout items that could be beneficial to us in one way or another are:

- https://github.com/elixir-broadway/broadway/pull/361
- https://github.com/elixir-broadway/broadway/pull/370
- https://github.com/elixir-broadway/broadway/pull/363


djwhitt (approved):
Reviewed the changelog. Looks low risk, and I really want that process labeling capability.

PR #3682, opened 2026-07-13
feat: add Management API endpoint for alerts (Api.AlertController)

djwhitt:
## Summary

Adds Management API CRUD for `Logflare.Alerting.AlertQuery` at `/api/alerts`, enabling Terraform-provider coverage for alerts under O11Y-2143.

The alert resource's list and lookup paths follow existing team-access semantics; backend attachment has the limitation below.

### Backend attachment contract

- Create requires `name`, `query`, and `cron`; `language` defaults to `bq_sql` when omitted.
- Create and update accept `backend_ids`, a complete replacement set of positive integer backend IDs.
- IDs are deduplicated and must belong directly to the authenticated user.
- `backends` is response-only and returns the attached backend objects. Supplying `backends` in a request returns `400`; use `backend_ids` instead.
- Invalid `backend_ids` return `400` rather than raising.

Alerts now preload and serialize attached backends in all response-producing actions, so Terraform can reconcile the association.

### OpenAPI

- Separate create and update request schemas: create requires alert fields; update permits partial updates.
- Documents `backend_ids` only on write schemas and `backends` only on the response schema.
- Documents `backend_ids` as positive integers, nullable backend metadata, and the actual naïve timestamp serialization.
- Correctly documents nullable optional fields and `400`, `404`, and `422` responses.

### Known limitation / follow-up

Alert list, fetch, update, and delete are team-access scoped, but `backend_ids` are resolved only among the token owner's directly owned backends. Consequently, a member can manage an existing team alert but cannot discover or attach that team's owner-owned backends through the Management API. `/api/backends` list and show have the same direct-owner scope.

This does not affect the owner-token Terraform use case. Team-aware backend discovery and attachment should be addressed together in a follow-up: make backend list/show and alert attachment use the existing backend access helpers, then cover team-owned attachment, unrelated-backend rejection, and team backend discovery.

Implements O11Y-2143.

### DIFF
--- a/lib/logflare/alerting.ex
+++ b/lib/logflare/alerting.ex
@@ -140,10 +140,13 @@ defmodule Logflare.Alerting do
 
   """
   def create_alert_query(%User{} = user, attrs \\ %{}) do
+    backends = Map.get(attrs, :backends) || Map.get(attrs, "backends")
+
     user
     |> Ecto.build_assoc(:alert_queries)
     |> Repo.preload(:user)
     |> AlertQuery.changeset(attrs)
+    |> put_backends(backends)
     |> Repo.insert()
   end
 
@@ -165,15 +168,15 @@ defmodule Logflare.Alerting do
     alert_query
     |> preload_alert_query()
     |> AlertQuery.changeset(attrs)
-    |> then(fn
-      changeset when not is_nil(backends) ->
-        Ecto.Changeset.put_assoc(changeset, :backends, backends)
-
-      changeset ->
-        changeset
-    end)
+    |> put_backends(backends)
     |> Repo.update()
     |> handle_schedule_change(alert_query)
+  end
+
+  defp put_backends(changeset, nil), do: changeset
+
+  defp put_backends(changeset, backends) do
+    Ecto.Changeset.put_assoc(changeset, :backends, backends)
   end
 
   defp handle_schedule_change({:ok, %AlertQuery{} = updated} = result, previous) do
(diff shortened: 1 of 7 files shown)


djwhitt:
## Manual curl QA

Ran against the rebased branch in the local single-tenant Postgres service at `http://localhost:4000`, using the configured private Management API token. The manual backend and alert were deleted after verification.

| Operation | Result |
| --- | --- |
| `GET /health/` | `200` healthy |
| `GET /api/alerts` | `200` |
| `POST /api/backends` | `201`; created an owned BigQuery backend for attachment |
| `POST /api/alerts` with `backend_ids` and no `language` | `201`; defaulted `language` to `bq_sql` and returned the attached backend |
| `GET /api/alerts/:token` | `200`; returned the serialized backend association |
| `PATCH /api/alerts/:token` | `204`, empty body, `content-type: text/plain; charset=utf-8` |
| `PUT /api/alerts/:token` with `{"backend_ids": []}` | `200`; cleared all backends |
| `PUT /api/alerts/:token` with response-only `backends` | `400`, `backends is read-only; use backend_ids` |
| `DELETE /api/alerts/:token` | `204`, empty body, `content-type: text/plain; charset=utf-8` |
| `GET /api/alerts/:token` after delete | `404`, `{"error":"Not Found"}` |

The single-tenant environment has one authenticated account, so cross-account and team authorization remain covered by the automated controller tests.

amokan (approved):
Looks good!

PR #2851, opened 2025-10-20
feat: Switch team based on query param

msmithstubbs:
This PR supports switching the current team context by setting a `t` query param to the URL. For example, `/dashboard?t=50` will load the dashboard for the team where id = 50.

https://github.com/user-attachments/assets/c516a6a2-b569-42b5-9600-c86e88e82aec

There is no requirement to switch team users, and the current team is not recorded in the session. The current signed in user is identified in the session with `current_email`. This allows users to have multiple browser windows signed in to different teams simultaneously.

TeamContext module takes an email address and an optional team_id param, and returns a user, team, and, optionally, team_user.  If no team_id is provided it defaults to a user record matching that email, and the user's home team.

### Dynamically resolving current team

If a user is signed and requests a resource without the `t=` query param we will set the team context based on the requested resource. 

If they are requesting a page that doesn't specify a resource (for example, `/sources`) the user or team user home team will be set.

### Invalid team_id

https://github.com/user-attachments/assets/31828fbb-5e59-4302-9fa9-60186bf2c2a9

If a user requests a page with an invalid `t=` param the SetTeamContext plug will catch it and retry the request without the invalid id. 

### 403 error page

If a user requests a resource the don't have access to they are shown the 403 forbidden page. 
We no longer show the current teams list and suggest they switch because switching is no longer required.  

CleanShot 2025-11-19 at 09 28 57@2x

### Other changes

* Add new `/login/single_tenant` path
 * Last accessed team is no longer tracked. When logging in you are signed in to your home team unless a team_id param is present.

Apologies for the size of this PR.

PR #3687, opened 2026-07-15
fix(api): make backend and endpoint access team-aware

djwhitt:
## Summary
Make Management API backend/endpoint access consistent with the team-access model introduced for browser/LiveView flows, so team members can discover and use resources owned elsewhere on their team via the API.

- **Backends:** `GET /api/backends` (list) and `GET /api/backends/:token` (show) now respect team access. Alert backend attachment (`PUT /api/alerts/:token`) can now reference any team-accessible backend.
- **Endpoints:** list, show, update, and delete now respect team access.
- **Intentional asymmetry:** backend *mutations* (update, delete, test_connection) and backend/endpoint *creation* remain owner-scoped. Team members get 404 for those, matching current behavior. Only discovery/attachment (backends) and full read+write (endpoints) become team-aware here; broadening backend mutation is a deliberate follow-up rather than part of this change.

## Behavior change
Existing API tokens calling `GET /api/backends` and `GET /api/endpoints` will now receive team-accessible resources in addition to owner-owned ones. Inaccessible resources return 404 (Not Found) consistently.

## Implementation notes
- Controllers now route through the existing `*_by_user_access` helpers instead of owner-only `fetch_backend_by`/`get_by(user_id:)`.
- `get_backend_by_user_access/2` and `get_endpoint_query_by_user_access/2` gain a keyword-filter form so callers can look up by `token:` as well as `id:`.
- `list_backends_by_user_access/2` gains a `filters` argument and now shares a private `filter_backends/2` with `list_backends/1`.
- Because `Teams.filter_by_user_access/2` injects team joins plus `distinct: true`, `has_sources_or_rules` was updated to bind trailing joins via `[b, ..., sb, r]` and a `distinct_backends/1` guard avoids re-adding a conflicting `distinct`. **← main logic to review.**

## Context
[#2851](https://github.com/Logflare/logflare/pull/2851) introduced query-param team switching and the `*_by_user_access` helpers for browser/LiveView flows; the Management API only authenticated the token owner. [#3682](https://github.com/Logflare/logflare/pull/3682) added the Alerts Management API and documented this follow-up: alerts were team-accessible while backend discovery/attachment stayed owner-scoped. This PR closes that gap.

## Stacking
Depends on #3682. Rebase onto `main` and retarget after #3682 merges.

## Non-goals
- Backend create/update/delete/test_connection stay owner-only.
- No changes to browser/LiveView flows or single-tenant behavior.

## Validation
Local: `../bin/test test/logflare_web/controllers/api/alert_controller_test.exs test/logflare_web/controllers/api/backend_controller_test.exs test/logflare_web/controllers/api/endpoint_controller_test.exs test/logflare/backends_test.exs`

New coverage: team read of backend list/show; team rejection of backend update/delete/test (404); team endpoint list/show/update/delete + backend attachment; alert attach allowed + unrelated backend rejected; `has_sources_or_rules` filter under team access. CI green (Elixir CI, typings, format, coverage, E2E).

Linear: O11Y-2154

### DIFF
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -65,8 +65,14 @@ defmodule Logflare.Backends do
   """
   @spec list_backends(keyword()) :: [Backend.t()]
   def list_backends(filters) when is_list(filters) do
-    filters
-    |> Enum.reduce(from(b in Backend), fn
+    Backend
+    |> filter_backends(filters)
+    |> Repo.all()
+    |> Enum.map(fn sb -> typecast_config_string_map_to_atom_map(sb) end)
+  end
+
+  defp filter_backends(query, filters) do
+    Enum.reduce(filters, query, fn
       {:types, types}, q when is_list(types) ->
         where(q, [b], b.type in ^types)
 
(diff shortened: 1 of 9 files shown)


amokan (approved):
Approved.

I **do not** think we should take the Copilot advice around a guard for keywords.

PR #3688, opened 2026-07-15
feat: pagination for search

msmithstubbs:
This PR adds pagination support to the search results live view.

* load more buttons are added to the top and bottom of the search results list
* the top button loads earlier events. It will be hidden if there are no more previous events to load
* the bottom button loads later events. 
* the pagination buttons are hidden when tailing
* if the search query include a timestamp `t:` filter and the loaded page includes results outside the queried timestamp range the results are displayed. The `t:` filter is updated to reflect the fetched timestamp range.

Follows #3720 

- [x]  events are now rendered with LiveView stream
- [x] ~~scrolling to the top of the event list will fetch the next 100 previous events~~
- [x] add ‘load more’ buttons to extend search range earlier or later

CleanShot 2026-07-24 at 15 04 24@2x

#### Load earlier events

https://github.com/user-attachments/assets/e2e11b1f-7650-4acb-80de-e332e264cd71

#### Load next page, timestamp filter LQL updated
https://github.com/user-attachments/assets/ba83db80-f1ea-4388-a9c4-0154aea431f4

Fixes O11Y-286

### DIFF
--- a/lib/logflare/logs/event_page.ex
+++ b/lib/logflare/logs/event_page.ex
@@ -1,0 +1,49 @@
+defmodule Logflare.Logs.EventPage do
+  @moduledoc false
+
+  use TypedStruct
+
+  alias Logflare.LogEvent
+
+  @type boundary :: integer() | DateTime.t() | NaiveDateTime.t()
+  @type cursor :: %{timestamp: integer(), id: String.t()}
+  @type direction :: :previous | :next
+  @type intent :: :initial | :previous | :next
+  @type request :: %{
+          intent: intent(),
+          boundary: boundary() | nil,
+          cursor: cursor() | nil
+        }
+
+  typedstruct do
+    field :rows, [LogEvent.t()], enforce: true
+    field :request, request(), enforce: true
+    field :cursor, cursor() | nil, enforce: true
+    field :next_cursor, cursor() | nil, default: nil
+    field :has_more?, boolean(), enforce: true
+    field :events, :any, default: nil
+  end
+
+  @spec direction(intent()) :: direction()
+  def direction(:next), do: :next
+  def direction(_intent), do: :previous
+
+  @spec boundary(intent(), %{min: boundary(), max: boundary()} | nil) :: boundary() | nil
+  def boundary(:next, %{max: max}), do: max
+  def boundary(_intent, _bounds), do: nil
+
+  @spec valid_request?(term(), term(), boundary() | nil) :: boolean()
+  def valid_request?(:previous, cursor, _boundary), do: valid_cursor?(cursor)
+
+  def valid_request?(:next, cursor, boundary) do
+    not is_nil(boundary) or valid_cursor?(cursor)
+  end
+
+  def valid_request?(_intent, _cursor, _boundary), do: false
+
+  @spec valid_cursor?(term()) :: boolean()
+  def valid_cursor?(%{timestamp: timestamp, id: id}) when is_integer(timestamp) and is_binary(id),
+    do: true
+
+  def valid_cursor?(_cursor), do: false
+end
--- a/lib/logflare/logs/event_page.ex
+++ b/lib/logflare/logs/event_page.ex
@@ -27,11 +27,25 @@ defmodule Logflare.Logs.EventPage do
   def direction(:previous), do: :previous
   def direction(:initial), do: :previous
 
+  @doc """
+  Returns whether an event-page request has a valid intent and cursor.
 
+  An intent can be one of:
+  * `:next` - fetching the next page of newer events after the curosr
+  * `:previous` - fetch a page of older events before the cursor
+  * `:initial` - no cursor, used for the first search by Search LiveView.
+  """
+  @spec valid_request?(term(), term()) :: boolean()
+  def valid_request?(intent, cursor) when intent in [:previous, :next], do: valid_cursor?(cursor)
 
   def valid_request?(_intent, _cursor), do: false
 
+  @doc """
+  Validate the cursor for the page query has a timestamp and id.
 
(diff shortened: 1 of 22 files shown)


lib/logflare_web/live/search_live/logs_search_lv.ex
```diff
@@ -341,6 +344,9 @@ defmodule LogflareWeb.Source.SearchLV do
     socket =
       socket
       |> assign(:search_timezone, tz)
+      |> assign(:search_op_log_events, nil)
+      |> assign(:search_op_log_aggregates, nil)
+      |> stream(:log_events, [], reset: true)
```
msmithstubbs: Changing  the timezone won't update the rendered timestamps in a stream, so clear it immediately.

lib/logflare/logs/event_page.ex
```diff
@@ -0,0 +1,49 @@
...
+  @spec valid_request?(term(), term(), boundary() | nil) :: boolean()
+  def valid_request?(:previous, cursor, _boundary), do: valid_cursor?(cursor)
+
+  def valid_request?(:next, cursor, boundary) do
+    not is_nil(boundary) or valid_cursor?(cursor)
+  end
+
+  def valid_request?(_intent, _cursor, _boundary), do: false
+
+  @spec valid_cursor?(term()) :: boolean()
```
Ziinc, line 49: Docstrings for functions in this mod

lib/logflare/logs/search_operations.ex
```diff
@@ -47,6 +48,44 @@ defmodule Logflare.Logs.SearchOperations do
   @spec max_chart_ticks :: integer()
   def max_chart_ticks, do: @default_max_n_chart_ticks
 
+  @spec default_limit :: pos_integer()
```
Ziinc, line 54: Docstrings

lib/logflare/logs/search_operations.ex
```diff
@@ -87,12 +126,21 @@ defmodule Logflare.Logs.SearchOperations do
...
   end
 
+  @spec query_for_execution(SO.t()) :: Ecto.Query.t()
+  defp query_for_execution(%SO{
+         type: :events,
+         query: %Ecto.Query{limit: %Ecto.Query.LimitExpr{}} = query
+       }),
+       do: limit(query, ^fetch_limit())
+
+  defp query_for_execution(%SO{query: query}), do: query
```
Ziinc, line 145: Don't we always expect a limit to be set? If so then this clause never runs

lib/logflare/logs/search_operations.ex
```diff
@@ -122,15 +177,67 @@ defmodule Logflare.Logs.SearchOperations do
...
+  defp order_events(query, :previous),
+    do: order_by(query, [t], desc: t.timestamp, desc: t.id)
+
+  defp order_events(query, :next), do: order_by(query, [t], asc: t.timestamp, asc: t.id)
+
+  @spec apply_cursor(SO.t()) :: SO.t()
+  def apply_cursor(
+        %SO{event_page_request: %{intent: intent, cursor: %{timestamp: timestamp, id: id}}} = so
+      ) do
+    timestamp = normalize_event_timestamp(so.backend_type, timestamp)
```
Ziinc, line 207: Are we considering different granularity? For e.g ts shift for daily is different from Secondly etc.

Ziinc: Not seeing an offset amount, unless we're querying across the day partition and shifting the range within?

lib/logflare/logs/search_query_executor.ex
```diff
@@ -188,9 +184,50 @@ defmodule Logflare.Logs.SearchQueryExecutor do
...
+  defp handle_event_result(lv_pid, events_so, state) do
+    Logger.debug(
+      "SearchQueryExecutor: Getting search events for #{pid_to_string(lv_pid)} / #{state.source_id} source..."
+    )
+
+    page_size = SearchOperations.default_limit()
+    raw_rows = events_so.rows
+    has_sentinel_row? = length(raw_rows) >= SearchOperations.fetch_limit()
+
+    {page_rows, sentinel_row} =
```
Ziinc, line 198: Would be good to add an inline comment on what is defined as a sentinal row

lib/logflare/lql/rules.ex
```diff
@@ -299,6 +360,32 @@ defmodule Logflare.Lql.Rules do
     FilterRule.shorthand_timestamp?(filter_rule)
   end
 
+  defp timestamp_bounds(%FilterRule{operator: :range, values: [min, max]})
+       when not is_nil(min) and not is_nil(max),
```
Ziinc, line 364: Ah it's here. Should consider granularity though to scanned resultset

Ziinc:
For next minor release

lib/logflare/logs/search_operations.ex
msmithstubbs: This function is called by the chart's aggregate query as well which doesn't have a limit. 

The distinction here is that the query we want to execute isn't exactly the same as the query we build. The one we build can be stored elsewhere, such as an endpoint or alert. The constructed query just has the default limit set. 

`query_for_execution` applies the larger fetch limit at the last moment before execution rather than building it into the query.

lib/logflare/lql/rules.ex
msmithstubbs: Can you elaborate on this one?

Ziinc: ah i was assuming that the cursor shift was fixed, but now i see that it is dependent on the log event list directly 👍

Ziinc (approved):

PR #3690, opened 2026-07-16
feat: generational queue cleanup

Baishan:
:pending and :processing events were  stored on the same ets queue requiring table scans to fetch pending events. This became problematic with clickhouse as the number of :processing events can be massive. It also introduced problems with queuejanitor cleanups needing to scan for stale events.

This pr introduces a new generational approach to queueing events. Events are placed into ets queues which are rotated every minute by the generation janitor. Stale events can be dropped by dropping the entire stale generation without needing to scan through all events. Each pipeline also has an ets table storing pointers to the events in the generation queue. This allows the pointer table to have stricly only pending events improving times of taking pending and keeping that table relatively small. 

This also addresses a problem with needing a massive @consoildated_max_queue_size as it no longer needs to account for :processing events and should prevent spikes getting dumped onto the one table.

The :ingested status was primarly used to show cached events for slow moving queues. This is replicated using a recent events table that is  truncated by queuejanitor without needing to scan through pending.

The pr also addresses load balancing and scaling up of clickhouse pipelines. Previously one clogged queue would trigger an indefinite scale up event that kept reducing the avg batchsize. I get around this by scaling up only if the avg count of all queues is above the threshold and by normalizing inserts accross pipelines so empty pipelines are fed more than backlogged ones.

closes O11Y-2185 and O11Y-2184


lib/logflare/backends/ingest_event_queue/generation_janitor.ex
```diff
@@ -0,0 +1,108 @@
...
+  def handle_info(:rotate, state) do
+    do_rotate(state)
+    schedule(state)
+    {:noreply, state}
+  end
+
+  @doc false
+  # Exposed for testing/benchmarking, same convention as QueueJanitor.do_drop/2.
+  @spec do_rotate(map()) :: :ok
+  def do_rotate(state) do
```
djwhitt, line 92: `do_rotate/1` creates a new generation for every key returned by `list_generation_queues_keys/0`, but that list is derived from the generations table itself. Since each pass creates the replacement before removing old rows, an idle key can never disappear from the list.

As a result, every source/backend that has ever initialized a generation (including ones that never ingest, since initialization is eager) retains ~2 ETS tables, a generations-table entry, and a `:persistent_term` entry forever. Deleted source/backend supervisors do not own these tables, so stopping them does not reclaim this state either.

Could we prune empty/inactive queue keys—dropping their remaining generations and erasing the persistent-term entry—or make generation creation fully lazy? Otherwise ETS table count and the global rotation sweep grow with every source/backend ever seen.

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -348,14 +635,39 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+  end
+
+  # Writes each event's full body into its current generation's table first, then the
+  # lightweight pointer row into `queue_tid` (the target pipeline's own table) — in that
+  # order, so a crash mid-batch can only leave an inert orphaned event (cleaned up by
+  # generation rotation), never a pointer with no backing data. Pointer row shape:
+  # {event_id, generation_tid, size, retries, event_type, day_bucket, ingest_freshness}.
+  defp insert_pointer_batch(sid_bid_pid, queue_tid, batch) do
+    queues_key = pointer_queues_key(sid_bid_pid)
+    gen_tid = current_generation_tid(queues_key)
```
djwhitt, line 685: There is a race here with `GenerationJanitor`: this reads the current TID from `:persistent_term`, but the janitor can publish a replacement and delete this table before the `:ets.insert(gen_tid, ...)` below runs. That insert is outside the rescue block, so it raises `ArgumentError` to the ingestion caller; mid-batch, earlier event bodies may already have been written without any pointer rows.

Retrying only on `ArgumentError` is not quite sufficient because ETS table references can be recycled (the new generation test already observes this). Can we give a selected generation a lease/reference count, or otherwise synchronize rotation with the event-store write, so a writer cannot target a retiring table?

lib/logflare/backends/ingest_event_queue/generation_janitor.ex
Baishan: I'll look into removing these although afaik we never actually kill sources that are dead. I'm going to move away form persistent term as is triggers a global gc event on write and instead put the keys in ets.

lib/logflare/backends/ingest_event_queue.ex
Baishan: The janitor rotates every 60s but only deletes ets tables that are more than 2 generations old. So there is a 60 second window where events can safely be written to the old table to manage the race conditions.

djwhitt: Yeah, fair enough. I guess that's still timing dependent, but it seems unlikely to be an issue in practice.

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -709,46 +967,50 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+    :ets.delete(tid, id)
 
-    with tid when tid != nil <- get_tid(sid_bid_pid),
-         size when is_integer(size) <- :ets.info(tid, :size),
-         {selected, _cont} <- :ets.select(tid, ms, min(n, max(size, 1))) do
-      {:ok, selected}
-    else
-      nil -> {:error, :not_initialized}
-      :"$end_of_table" -> {:ok, []}
+    case :ets.take(gen_tid, id) do
```
djwhitt, line 1020: Since generation expiry intentionally leaves pending pointers behind, this path can receive a `gen_tid` whose table has already been deleted. `:ets.take/2` raises `ArgumentError` for a deleted table rather than returning `[]`; I reproduced this by dropping an event's generation before calling `pop_pending/2`.

This affects the non-ID-passing `BufferProducer` path. The pointer is deleted before the exception, and terminating the producer can reclaim its entire per-pipeline pointer table, including pointers to younger generations.

Could `resolve_and_delete_pending/2` rescue `ArgumentError` per pointer and return `nil`, matching `lookup_event/2`'s stale-generation behavior? `pop_pending/2` already rejects `nil`, so expired pointers would be skipped without crashing the producer. A regression test could leave a pointer behind, drop its generation, and assert that `pop_pending/2` returns `{:ok, []}`.

lib/logflare/backends/ingest_event_queue.ex
Baishan: I will catch argumenterrors wherever they can be thrown anyway.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -447,53 +389,44 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
+      "Requeuing #{length(retriable)} ClickHouse events for retry",
       backend_id: backend_id
     )
 
-    IngestEventQueue.delete_batch(key, events_to_requeue)
-    IngestEventQueue.add_to_table(key, events_to_requeue)
+    events =
+      for pointer <- retriable,
+          event = IngestEventQueue.lookup_event(pointer.tid, pointer.id),
+          not is_nil(event) do
```
amokan, line 406: Given the drops that _could_ happen here silently due to a rotation - what if we did something like a `length(retriable) - length(events)` after the comprehension to determine the delta and log a message if it is greater than zero or something.

lib/logflare/backends/ingest_event_queue/generation_janitor.ex
```diff
@@ -0,0 +1,108 @@
...
+    for queues_key <- IngestEventQueue.list_generation_queues_keys() do
+      rotate_queue(queues_key, state)
+    end
+
+    IngestEventQueue.sweep_recent_events(state.max_age_ms)
```
djwhitt: Is the approximately 1–2 minute retention window for recent logs intentional? The PR describes this cache as replacing retained `:ingested` rows for slow-moving sources, but recent pointers reference the ordinary generation store. Once that generation is dropped, the event becomes unresolvable even if its recent pointer remains; this sweep then applies the same two-minute bound to the pointer itself.

Previously the latest 100 `:ingested` events were retained without an age limit. There is also an existing five-minute dependency in `Sources.source_idle?/1`, which calls `has_recent_logs_within?(source, 5)`. With this change, a recent event can disappear after roughly one minute and allow the source supervisor to stop before that five-minute window expires.

If the shorter recent-log window is intentional, this is fine. Otherwise, recent events would need independent storage or to keep their backing generation alive; increasing only the pointer sweep interval would not preserve them.

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -57,9 +79,300 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+  `:persistent_term` from inside the serialized `handle_call` before creating a table, so
+  only the first caller actually creates one and the rest just read it back.
+  """
+  @spec current_generation_tid(
+          queues_key()
+          | consolidated_queues_key()
+          | spool_producer_queues_key()
+        ) :: :ets.tid()
+  def current_generation_tid(queues_key) do
+    case :persistent_term.get(generation_key(queues_key), nil) do
```
djwhitt: One additional failure mode to keep in mind during the planned move to ETS: the `IngestEventQueue` GenServer owns the generation tables, so they disappear if that process restarts, while these `:persistent_term` entries survive. I reproduced killing/restarting the GenServer and found that `current_generation_tid/1` returned the same now-deleted TID; subsequent event insertion then raises `ArgumentError`.

Moving the current-generation mapping into an ETS table owned and recreated by `IngestEventQueue` should address this, provided stale mappings cannot survive the restart. Could we add a regression test that restarts `IngestEventQueue` and verifies that the next lookup creates and returns a live generation?

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -664,15 +674,27 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
+    # one outlier just adds pipelines the fleet doesn't actually need — spreading the
+    # same total inflow thinner and shrinking every pipeline's average ClickHouse batch
+    # size, which is the whole thing this function is meant to avoid.
+    avg_len =
+      case lens_no_startup_values do
+        [] -> 0
+        values -> Enum.sum(values) / length(values)
+      end
+
+    fleet_above_threshold? = avg_len >= @scaling_threshold
```
djwhitt: The linked issue says scale-up should happen only when all existing pipeline queues are above the threshold, but using the average still allows a sufficiently large outlier to trigger scaling while other pipelines are empty. For example, `[30_000, 0]` and `[60_000, 0, 0, 0]` both average to the 15,000 threshold and scale up.

This is bounded compared with the previous `Enum.any?/2` behavior, but it can still add an unnecessary pipeline and reduce batch sizes. Since the new weighted routing preferentially fills emptier queues, could we gate scale-up on the least-loaded queue instead?

```elixir
fleet_above_threshold? =
  lens_no_startup_values != [] and
    Enum.all?(lens_no_startup_values, &(&1 >= @scaling_threshold))
```

That would directly implement the ticket's “all pipelines above threshold” requirement. Is the average-based behavior an intentional deviation?

lib/logflare/backends/ingest_event_queue/queue_janitor.ex
```diff
@@ -106,30 +81,31 @@ defmodule Logflare.Backends.IngestEventQueue.QueueJanitor do
...
-      IngestEventQueue.truncate_table(table_key, :ingested, 0)
+      IngestEventQueue.truncate_recent(queues_key, 0)
     else
-      IngestEventQueue.truncate_table(table_key, :ingested, state.remainder)
+      IngestEventQueue.truncate_recent(queues_key, state.remainder)
     end
 
     size = IngestEventQueue.get_table_size(table_key)
 
     if is_integer(size) and size > state.max and pid != nil do
```
amokan, line 106: This could be something that is acceptable short-term (_it's already better than what's on prod_) - but the `pid != nil` guard means the startup queue will never drop pending.

Given that pointer rows only clear at drain rate - this _could_ balloon up a bit under a heavy sustained load.

Probably not a hill to die on now, but felt worthy to add a comment to follow-up on later.

lib/logflare/backends/ingest_event_queue/generation_janitor.ex
Baishan: Nice catch, I didnt see the check for idle here was 5 minutes. We could copy the whole event into recent and have that persist longer. I opted not to do it to simplify fetching them from one place only. I would argue caching these on most sources indefiniltly is not a great idea but we could retain that behaviour.

djwhitt: Yeah, not sure that's desired either, but figured we should be explicit about what's happening here.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
Baishan: Load should be roughly balanced between them due to the weight_by_load so they should normalize quickly enough. But I think explicitly making it all makes sense either way.

djwhitt:
### Queue data flow (head `f23baaac`)

This is my reading of the new pointer + generation architecture, as a review aid.

<details open>
<summary>Show diagram</summary>

```mermaid
flowchart TB
    Ingest["Backends dispatch<br/>batch of LogEvents"] --> Add["IngestEventQueue.add_to_table/3"]
    Add --> Select{"Select target pointer queue"}
    Select -->|"eligible pipeline"| Write["insert_pointer_batch/3"]
    Select -->|"none eligible"| Startup["Startup pointer queue"]
    Startup --> Write

    Write --> Gen[("Shared generation ETS<br/>per queues_key<br/>{id, LogEvent}")]
    Write --> Ptr[("Pointer ETS<br/>per pipeline or startup<br/>{id, gen_tid, size, retries, routing metadata}")]

    Ptr --> Producer["BufferProducer"]
    Producer --> Mode{"id_passing?"}

    Mode -->|"yes"| Claim["pop_pending_pointers/2<br/>remove pointer"]
    Claim --> IDAdapters["BigQuery / ClickHouse / Spool"]
    IDAdapters --> Lookup["lookup_event(gen_tid, id)"]
    Lookup --> Ack{"Ack result"}
    Ack -->|"success"| Delete["delete_id/2"]
    Ack -->|"BQ low-rate default"| Recent[("Recent-pointer ETS<br/>{id, gen_tid}")]
    Ack -->|"retry"| Retry{"Retry mode"}
    Retry -->|"BigQuery / ClickHouse"| Add
    Retry -->|"Spool"| Reinsert["reinsert_pointer/1"]
    Reinsert --> Ptr

    Mode -->|"no"| Pop["pop_pending/2<br/>remove pointer and event body"]
    Pop --> FullAdapters["Webhook / Syslog / HTTP / Postgres / S3"]

    Delete -->|"delete event body"| Gen
    Recent -. "references event body" .-> Gen
```

</details>

djwhitt:
### Generation rotation and pointer lifetime (head `f23baaac`)

The ID-passing path is shown explicitly; non-ID consumers resolve and remove the event body inside `pop_pending/2`.

<details open>
<summary>Show diagram</summary>

```mermaid
sequenceDiagram
    autonumber
    participant I as Ingest caller
    participant Q as IngestEventQueue
    participant C as Current-generation mapping
    participant G0 as Generation G0 ETS
    participant G1 as Generation G1 ETS
    participant P as Pointer ETS
    participant A as Adapter pipeline
    participant R as Recent-pointer ETS
    participant J as GenerationJanitor

    I->>Q: add_to_table(batch)
    Q->>C: current_generation_tid(key)
    C-->>Q: G0 TID
    loop each event
        Q->>G0: insert {id, LogEvent}
        Q->>P: insert {id, G0 TID, metadata}
    end

    J->>Q: new_generation(key)
    Q->>G1: create table
    Q->>C: publish G1 as current
    Note over G0,G1: G0 remains readable during its retention window

    I->>Q: add_to_table(new batch)
    Q->>G1: insert event bodies
    Q->>P: insert pointers to G1

    alt G0 event is processed before expiry
        P->>Q: pop_pending_pointers/2
        Q-->>A: LogEventPointer containing G0 TID
        A->>Q: lookup_event(G0, id)
        Q->>G0: lookup id
        G0-->>Q: LogEvent
        Q-->>A: LogEvent
        opt Keep successful BigQuery event visible as recent
            A->>R: record {id, G0 TID}
        end
    else G0 event remains pending
        Note over P,G0: Pointer continues to reference G0
    end

    J->>Q: drop_generation(key, G0)
    Q->>G0: delete whole ETS table
    Note over P,R: Pending or recent pointers may still contain G0 TID

    P->>Q: claim an expired pointer
    Q-->>A: pointer containing G0 TID
    A->>Q: lookup_event(G0, id)
    Q-->>A: nil plus stale-table telemetry

    J->>Q: sweep_recent_events(max_age_ms)
    Q->>R: remove expired recent pointers
```

</details>

lib/logflare/backends/ingest_event_queue/queue_janitor.ex
Baishan: I was experiementing with having them prioritize draining startup. The thinking was it would encourage scale up if there are events getting dumped due to the cut off. But then I decided it could get into a state where its only consuming startup. For clickhouse it might not actually be the worst thing to only consume from a shared startup here but thats for another day.

My assumptions for how this works now is that the pipes scale up above 15000 pending and only dump into startup abote 60000. So there will be relatively empty queues starting up before the cut off that can consume these as they keep up. The problem is what happens during scale down. They can be dumped into startup and are relying to queues that have extra load now consuming them. I think more work needs to be done to properly plan what should happen with this.

lib/logflare/backends/ingest_event_queue.ex
Baishan: Added the rescue

lib/logflare/backends/ingest_event_queue.ex
Baishan: changed to use ets @ets_current_generation. These should be tied to ingesteventqueue so if that dies all the tables will die with it

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
Baishan: Made this explicit like above^

lib/logflare/backends/ingest_event_queue.ex
Baishan: Catching these above ^

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
Baishan: Done

lib/logflare/backends/ingest_event_queue/generation_janitor.ex
Baishan: I've updated it to store the full events in recent. These will be truncated to 100 buy queue janitor and purged after 10 minutes. I think 10 minutes is reasonable but we can make it much longer. I just didn't want it to affect the source_idle behaviour.  We can measure the size of this table in canary and increase the timeout to much longer if it makes sense.

lib/logflare/backends/ingest_event_queue/generation_janitor.ex
Baishan: Ok these should be deleted soon after MapperJanitor cleans up dead queues now.

lib/logflare/backends/buffer_producer.ex
```diff
@@ -310,118 +303,84 @@ defmodule Logflare.Backends.BufferProducer do
...
+          startup_key :: table_key(),
+          n :: non_neg_integer(),
+          fetch_fn :: (table_key(), non_neg_integer() -> [term()])
+        ) :: [term()]
+  defp fetch_own_first(own_key, startup_key, n, fetch_fn) do
+    own_events = fetch_fn.(own_key, n)
+    remaining = n - length(own_events)
+
+    if remaining > 0 do
+      own_events ++ fetch_fn.(startup_key, remaining)
```
djwhitt, line 363: Documenting one low-probability consequence of letting live producers read from the startup queue: producer initialization still moves startup into the new producer queue via `IngestEventQueue.move/2`, which copies each row and then deletes it. If `move/2` has read a pointer and an existing producer takes that pointer before the move inserts it into the target, the move recreates the already-claimed pointer in the new queue. For the ID-passing paths, both pipelines can then resolve it before either ack deletes the event body, resulting in a duplicate event.

I reproduced the race under deliberately heavy contention: 4 IDs out of 200,000 were both returned as claimed and left pending in the move target. Real incidence should be substantially lower given the narrow timing and routing behavior, so I do not think this needs to block the change; recording it here as an accepted occasional-duplicate risk around pipeline scale-up.

lib/logflare/backends/ingest_event_queue/generation_janitor.ex
```diff
@@ -0,0 +1,183 @@
...
+  defp table_size(tid) do
+    case :ets.info(tid, :size) do
+      n when is_integer(n) -> n
+      _ -> 0
+    end
+  end
+
+  defp emit_drop_telemetry(queues_key, dropped_count, dropped_size) do
+    :telemetry.execute(
+      [:logflare, :ingest_event_queue, :generation_janitor, :drop],
```
djwhitt, line 180: These `:drop`/`:prune` events are not currently included in `Logflare.Telemetry.metrics/0`. `OtelMetricExporter` only attaches handlers for explicitly defined metric event names, so these events are otherwise discarded.

The existing `missing_ids` and `stale_table` metrics provide partial, delayed coverage when a surviving pointer is eventually read, but they cannot observe an abandoned claim whose pointer was already removed. This event is the only place that directly knows how many event bodies generation cleanup evicted.

Not blocking, but it would be useful for staging to export sums for `drop.events`, `drop.generations`, and `prune.generations`. I would avoid tagging by `queues_key` because of cardinality.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -447,53 +389,66 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
+  defp emit_requeue_lookup_miss_telemetry(_backend_id, 0), do: :ok
+
+  defp emit_requeue_lookup_miss_telemetry(backend_id, missing_count) do
+    Logger.warning(
+      "Dropped #{missing_count} ClickHouse event(s) during retry requeue: pointer's generation was already gone by lookup time",
+      backend_id: backend_id
+    )
+
+    :telemetry.execute(
+      [:logflare, :ingest_event_queue, :requeue_lookup_miss],
```
djwhitt, line 432: This telemetry event is also absent from `Logflare.Telemetry.metrics/0`, so only the warning log is currently exported. Could we add an aggregate `requeue_lookup_miss.count` sum?

The PR also removes the only emitter for the existing `stale_processing.reset` and `stale_processing.dropped` metrics, so those definitions can be removed or replaced with the new generation/requeue metrics. This is non-blocking, but would make the rollout signals match the new queue lifecycle.

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -512,139 +853,81 @@ defmodule Logflare.Backends.IngestEventQueue do
...
-      # One monotonic read per batch (not per event); every row claimed in this batch shares
-      # the same claimed_at stamp, which is all the stale-recovery janitor needs.
-      claimed_at = System.monotonic_time(:millisecond)
-
-      # The claim counter's 0 -> 1 winner takes the event; a 2+ result (another consumer,
-      # or a write_concurrency duplicate row within this select) loses, so no dedup needed.
-      confirmed = Enum.filter(selected, &claim_pending(tid, id_fun.(&1), claimed_at))
+      confirmed =
+        selected
+        |> Enum.filter(fn {id, _, _, _, _, _, _} -> :ets.take(tid, id) != [] end)
```
djwhitt: `:ets.take/2` returns the row that was actually claimed, but this discards it and builds the pointer from the earlier `select/3` result.

In the narrow case where the same ID is taken and reinserted between the select and take, this removes the new row while returning its old generation/retry metadata. That could roll back the retry count or leave the reinserted event body orphaned if it moved generations.

Could we build the pointer from the row returned by `take` instead?

```elixir
selected
|> Enum.reduce([], fn {id, _, _, _, _, _, _}, acc ->
  case :ets.take(tid, id) do
    [actual_row] -> [to_pointer(actual_row, tid) | acc]
    [] -> acc
  end
end)
|> Enum.reverse()
```

This should also avoid the intermediate filtered list, so it is likely slightly more efficient rather than introducing a performance tradeoff. Non-blocking, but a cheap correctness improvement.

lib/logflare/backends/ingest_event_queue/generation_janitor.ex
```diff
@@ -0,0 +1,183 @@
...
+
+  defp has_live_queue?(queues_key), do: IngestEventQueue.list_queues(queues_key) != []
+
+  defp drop_aged_generations(queues_key, state) do
+    cutoff = System.monotonic_time(:millisecond) - state.max_age_ms
+
+    {dropped_count, dropped_size} =
+      queues_key
+      |> IngestEventQueue.list_generations()
+      |> Enum.filter(fn {_tid, created_at} -> created_at <= cutoff end)
```
djwhitt, line 149: One retention detail to confirm: `max_age_ms` is applied to the generation's creation time, not the event's insertion time.

With a 60-second rotation interval and `max_age_ms` of two minutes, a generation created at `t=0` can receive events until approximately `t=60`, then be dropped at approximately `t=120`. An event inserted near `t=60` therefore survives only about one minute. Normal effective event retention is roughly **60–120 seconds**, rather than the two-to-four-minute bound described in the moduledoc.

This affects pending events during a backend slowdown or backlog, independently of the recent-events cache fixed in the earlier thread. Is the approximately one-minute minimum retention intentional? Non-blocking if so, but the documentation should reflect it; if two minutes is intended as the minimum, the generation age would need another rotation interval of headroom.

djwhitt (approved):

lib/logflare/backends/ingest_event_queue.ex
Baishan: Done

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
Baishan: Done

lib/logflare/backends/ingest_event_queue/generation_janitor.ex
Baishan: Added

lib/logflare/backends/ingest_event_queue/generation_janitor.ex
Baishan: Added interval when dropping

lib/logflare/backends/buffer_producer.ex
Baishan: Modified move to use :ets.take to prevent this.

PR #3692, opened 2026-07-16
fix(api): enforce and correct Management API OpenAPI contracts

djwhitt:
## Summary
- add generated-spec smoke and authenticated management-route coverage checks
- validate each dispatched, documented API response against its matched OpenAPI operation and status in `ConnCase`
- document the account endpoint and shared management-auth 401 responses
- correct response schema/content-type drift exposed by the new checks
- split source, backend, endpoint, rule, and team request schemas from response schemas
- document rule validation 422 responses and cover those response paths

## Carve-out
Route coverage is limited to routes behind `:require_mgmt_api_auth`; ingestion, partner, and spec-serving routes remain outside this management-API guardrail.

### DIFF
--- a/lib/logflare_web/controllers/api/rule_controller.ex
+++ b/lib/logflare_web/controllers/api/rule_controller.ex
@@ -3,16 +3,16 @@ defmodule LogflareWeb.Api.RuleController do
   use OpenApiSpex.ControllerSpecs
 
   alias Logflare.Backends
-  alias Logflare.Sources
   alias Logflare.Rules
+  alias Logflare.Sources
   alias LogflareWeb.OpenApi.Accepted
   alias LogflareWeb.OpenApi.List
   alias LogflareWeb.OpenApi.NotFound
   alias LogflareWeb.OpenApi.UnprocessableEntity
   alias LogflareWeb.OpenApiSchemas.RuleApiSchema
-  alias LogflareWeb.OpenApiSchemas.RuleParams
   alias LogflareWeb.OpenApiSchemas.RuleBatchResponse
   alias LogflareWeb.OpenApiSchemas.RuleCreateResponse
+  alias LogflareWeb.OpenApiSchemas.RuleParams
 
   action_fallback(LogflareWeb.Api.FallbackController)
 
--- a/lib/logflare_web/controllers/api/rule_controller.ex
+++ b/lib/logflare_web/controllers/api/rule_controller.ex
@@ -6,10 +6,13 @@ defmodule LogflareWeb.Api.RuleController do
   alias Logflare.Sources
   alias Logflare.Rules
   alias LogflareWeb.OpenApi.Accepted
-  alias LogflareWeb.OpenApi.Created
   alias LogflareWeb.OpenApi.List
   alias LogflareWeb.OpenApi.NotFound
+  alias LogflareWeb.OpenApi.UnprocessableEntity
   alias LogflareWeb.OpenApiSchemas.RuleApiSchema
+  alias LogflareWeb.OpenApiSchemas.RuleParams
+  alias LogflareWeb.OpenApiSchemas.RuleBatchResponse
+  alias LogflareWeb.OpenApiSchemas.RuleCreateResponse
 
   action_fallback(LogflareWeb.Api.FallbackController)
 
@@ -51,9 +54,11 @@ defmodule LogflareWeb.Api.RuleController do
 
   operation(:create,
     summary: "Create rule. Allows batch creation if as a list.",
-    request_body: RuleApiSchema.params(),
+    request_body: RuleParams.params(),
     responses: %{
-      201 => Created.response(RuleApiSchema),
+      201 => RuleCreateResponse.response(),
+      400 => RuleBatchResponse.response(),
+      422 => UnprocessableEntity.response(),
       404 => NotFound.response()
     }
   )
@@ -124,11 +129,12 @@ defmodule LogflareWeb.Api.RuleController do
   operation(:update,
     summary: "Update rule",
     parameters: [token: [in: :path, description: "Rule UUID", type: :string]],
-    request_body: RuleApiSchema.params(),
+    request_body: RuleParams.params(),
     responses: %{
       204 => Accepted.response(),
       200 => Accepted.response(RuleApiSchema),
-      404 => NotFound.response()
+      404 => NotFound.response(),
+      422 => UnprocessableEntity.response()
     }
   )
 
--- a/lib/logflare_web/controllers/api/source_controller.ex
+++ b/lib/logflare_web/controllers/api/source_controller.ex
@@ -9,9 +9,11 @@ defmodule LogflareWeb.Api.SourceController do
   alias LogflareWeb.OpenApi.Created
   alias LogflareWeb.OpenApi.List
   alias LogflareWeb.OpenApi.NotFound
+  alias LogflareWeb.OpenApi.UnprocessableEntity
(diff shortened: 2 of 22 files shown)


lib/logflare_web/controllers/api/rule_controller.ex
```diff
@@ -6,10 +6,13 @@ defmodule LogflareWeb.Api.RuleController do
   alias Logflare.Sources
   alias Logflare.Rules
   alias LogflareWeb.OpenApi.Accepted
-  alias LogflareWeb.OpenApi.Created
   alias LogflareWeb.OpenApi.List
   alias LogflareWeb.OpenApi.NotFound
+  alias LogflareWeb.OpenApi.UnprocessableEntity
   alias LogflareWeb.OpenApiSchemas.RuleApiSchema
+  alias LogflareWeb.OpenApiSchemas.RuleParams
+  alias LogflareWeb.OpenApiSchemas.RuleBatchResponse
```
amokan, line 13: super minor nit because I'm a weirdo - would be nice to alphabetize these aliases

lib/logflare_web/controllers/api/source_controller.ex
```diff
@@ -46,10 +48,11 @@ defmodule LogflareWeb.Api.SourceController do
 
   operation(:create,
     summary: "Create source",
-    request_body: Source.params(),
+    request_body: {"Source Parameters", "application/json", SourceParams.schema()},
```
amokan, line 51: Curious why we lost the `Source.params()` here and on line 72?

lib/logflare_web/open_api_schemas.ex
```diff
@@ -71,13 +71,33 @@ defmodule LogflareWeb.OpenApiSchemas do
...
   defmodule AccessToken do
     @properties %{
       id: %Schema{type: :integer},
-      token: %Schema{type: :string},
-      description: %Schema{type: :string},
+      token: %Schema{type: :string, nullable: true},
+      description: %Schema{type: :string, nullable: true},
       scopes: %Schema{type: :string},
-      inserted_at: %Schema{type: :string, format: :"date-time"}
+      inserted_at: %Schema{type: :string}
```
amokan: Was the format no longer needed on this? Originally it had `date-time` which [translates to RFC3339](https://spec.openapis.org/registry/format/date-time).

test/support/conn_case.ex
```diff
@@ -120,8 +128,79 @@ defmodule LogflareWeb.ConnCase do
     Plug.Conn.put_req_header(conn, "authorization", "Bearer #{access_token.token}")
   end
 
-  def assert_schema(data, schema_name) do
-    OpenApiSpex.TestAssertions.assert_schema(data, schema_name, LogflareWeb.ApiSpec.spec())
+  for method <- @http_methods do
```
amokan, line 138: While I'm not a big stickler on macro cleverness - we should probably add a comment in here indicating what we're doing so future us (_or an AI agent_) don't trip over this given that any sort of future `get(conn, "foo")` stuff will get spec validation and potentially catch someone off guard.

Probably deserves a comment up in the moduledoc and a super small inline comment 🤷

amokan: Also think we want to ensure the escape hatch in `dispatch/5` is called out in the module doc.

amokan (changes requested):
Couple minor notes

lib/logflare_web/controllers/api/rule_controller.ex
djwhitt: Good catch—agree this should be alphabetical. I’ll clean up the aliases.

lib/logflare_web/controllers/api/source_controller.ex
djwhitt: This was intentional. `Source.params()` reuses the response schema, which includes output-only fields like `id`, `token`, `api_quota`, and `backends`. `SourceParams` is a dedicated write schema for create/update, and the API spec test now verifies that separation.

lib/logflare_web/open_api_schemas.ex
djwhitt: The API currently serializes this Ecto `NaiveDateTime` without a timezone, so it isn’t strictly RFC3339. I removed `format: :"date-time"` to match the actual wire response. Restoring it would require changing timestamp serialization separately.

test/support/conn_case.ex
djwhitt: Addressed in c7b1476a75423a080856acddcf93ac3b7114c6d6

lib/logflare_web/open_api_schemas.ex
djwhitt: Thinking about this more, perhaps we should just make the response proper RFC3339.

lib/logflare_web/controllers/api/rule_controller.ex
djwhitt: Addressed in c7b1476a75423a080856acddcf93ac3b7114c6d6

lib/logflare_web/open_api_schemas.ex
djwhitt: Implemented in e5f1bfd1. Access-token responses now convert stored UTC `NaiveDateTime` values to RFC3339 `DateTime`s (with `Z`), and the OpenAPI `date-time` format has been restored. Added response and schema contract coverage.

lib/logflare_web/controllers/user_controller.ex
```diff
@@ -26,6 +29,19 @@ defmodule LogflareWeb.UserController do
     bigquery_additional_projects
   )a
 
+  tags(["management"])
+
+  operation(:api_show,
+    summary: "Fetch account",
+    responses: %{200 => One.response(UserSchema)}
+  )
```
Ziinc, line 37: just a note that this should move to its own api controller in the future to keep this an only html controller

test/support/conn_case.ex
```diff
@@ -120,8 +134,80 @@ defmodule LogflareWeb.ConnCase do
...
+        assert_documented_response!(operation, conn)
+        OpenApiSpex.TestAssertions.assert_operation_response(conn, operation.operationId)
+        conn
+    end
+  end
+
+  defp open_api_operation(conn) do
+    with %{route: route} <-
+           Phoenix.Router.route_info(Router, conn.method, conn.request_path, conn.host),
+         true <- String.starts_with?(route, "/api"),
```
Ziinc: should move up this filter so that it isn't 3 levels deep, for example having it in `dispatch_and_assert_open_api_response/5` by just checking on the conn.path

Ziinc: this chain of logic feels like it could be simpler if done within  a single function with a with/do or pipe

Ziinc (approved):
can be split out in future, feels like multiple changes in 1 PR.

lib/logflare_web/controllers/user_controller.ex
djwhitt: Worth doing in this PR you think? (I'm fine either way)

test/support/conn_case.ex
djwhitt: Addressed in 29953b37. The `/api` filter now runs immediately after dispatch, and the route/spec/operation lookup plus assertions are flattened into a single `with` flow.

lib/logflare_web/controllers/user_controller.ex
djwhitt: Going with no for now, but may try to get it when doing future API work.

amokan (approved):
Looks good!

PR #3693, opened 2026-07-17
feat: add support for multiple labeled clickhouse read clusters

amokan:
## Why?

Every ClickHouse query currently lands on a single cluster, so a heavy workload such as dashboard log scans or a burst of MCP/agent queries can potentially starve resources away from competing queries.

This change allows multiple read-only clusters to be configured for a backend and queries to be routed using an optional `LF-ENDPOINT-CLICKHOUSE-READ-CLUSTER-LABEL` header provided to an endpoint, similar to the existing `LF-ENDPOINT-BIGQUERY-RESERVATION` header. Ultimately allowing us to isolate certain classes of queries to their own backing read clusters.

## Key Changes

- ClickHouse backends can now define multiple labeled read clusters via a `read_only_urls` config option as well as a `default_read_cluster` label to fall back to.
- The adaptor reads the header into `opts[:read_cluster]` and selects a per-label read pool, running one supervised `ConnectionManager` and pool per `{backend, label}`. This follows the existing BigQuery reservation header → opts → adaptor path.
- Resolution degrades gracefully so an unknown or absent label falls back to the default cluster (_or a legacy read URL when nothing is configured_). An unhealthy labeled pool retries once on the default. With no header and no `read_only_urls` config, behavior is identical to today, so this is a no-op until a backend has opted in to this feature.
- Updates the backend-form LV to handle the multiple read only cluster/label situation.

## Additional Notes

- The existing `read_only_url` config option will be removed in a later PR


lib/logflare_web/live/backends/read_cluster_urls_component.ex
```diff
@@ -0,0 +1,132 @@
...
+              value: row_label,
+              placeholder: "caller label",
+              class: "form-control"
+            )}
+            {text_input(@form, "read_cluster_url_#{i}",
+              value: row_url,
+              placeholder: "https://read-cluster:8443",
+              class: "form-control"
+            )}
+            <button type="button" class="btn btn-outline-danger" phx-click="remove_row" phx-target={@myself}>
```
djwhitt: This button doesn’t identify which row was clicked. `remove_row` only decrements `row_count`, while rendering keeps the first N entries, so removing a first/middle row actually drops the final cluster on save. Could we send the row index and remove that specific entry?

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -139,6 +142,8 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
        port: :integer,
        pool_size: :integer,
        read_only_url: :string,
+       read_only_urls: {:map, :string},
```
djwhitt, line 229: `test_connection/1` only checks `read_only_url`; it skips backends configured exclusively with the new `read_only_urls`. That lets the connection-test endpoint return success when a labeled/default read cluster is unreachable or missing `SELECT`. Can we validate each configured labeled URL as part of the connection test?

lib/logflare_web/live/backends/read_cluster_urls_component.ex
amokan: good catch

lib/logflare_web/live/backends/read_cluster_urls_component.ex
```diff
@@ -0,0 +1,138 @@
...
+  @impl true
+  def render(assigns) do
+    ~H"""
+    <div>
+      <div class="form-group">
+        <label>Per-Caller Read Cluster URLs (Optional)</label>
+        <small class="form-text text-muted">
+          Route ClickHouse reads to a dedicated cluster per caller. Each row maps a free-form
+          caller label to a read cluster URL. The query source sends the label in the <code>LF-ENDPOINT-READ-CLUSTER</code> header and Logflare routes the read to the matching cluster. Leave empty to use the
+          Read-Only URL above.
```
Ziinc: feels weird that it is an either/or. `ready_only_url` could function as the "default" label and `read_only_urls` could function as the allowlist.
Also, renaming to `LF-ENDPOINT-CLICKHOUSE-READ-CLUSTER-LABEL` would be more self-documenting and indicate that this is specific to clickhouse (and follows LF-ENDPOINT-BIGQUERY-RESERVATION pattern

test/logflare_web/live/backends/backends_live_test.exs
```diff
@@ -404,6 +431,133 @@ defmodule LogflareWeb.BackendsLiveTest do
...
+            type: "clickhouse",
+            config: %{
+              url: "http://localhost",
+              database: "test_db",
+              port: 8123,
+              username: "user",
+              password: "pass",
+              read_cluster_label_0: "reporting",
+              read_cluster_url_0: "http://reporting-read.local:8123",
+              default_read_cluster: "reporting"
```
Ziinc, line 485: instead of a dedicated new field for the default, `read_cluster_url` could function as accepting a url or an label (that maps to one of the allowlist read_cluster_urls.label values, that would make it more cohesive.
Or just remove the read_cluster_url and migrate it into read_cluster_urls as a `default` label and set the default_read_cluster to default to the `default` label

Ziinc (changes requested):
the config handling of `read_cluster_url` feels like a code smell, doesn't integrate very well with the label mapping system (which i'm fine with)

lib/logflare_web/live/backends/read_cluster_urls_component.ex
amokan: > feels weird that it is an either/or.

Not really - `read_only_url` is going away after this lands. No reason for it to stay.

amokan: > Also, renaming to LF-ENDPOINT-CLICKHOUSE-READ-CLUSTER-LABEL would be more self-documenting and indicate that this is specific to clickhouse (and follows LF-ENDPOINT-BIGQUERY-RESERVATION pattern

I'm down with that header name. Will make changes to the other PRs around this effort for that.

Ziinc (approved):
still needs the header tweak

test/logflare_web/live/backends/backends_live_test.exs
Ziinc: ignore me, missed a crucial description note.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -356,21 +428,34 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
       when is_list_or_map(params) and is_list(opts) do
-    with :ok <- ensure_query_connection_manager_started(backend) do
-      pool_via = connection_pool_via(backend)
+    requested = Keyword.get(opts, :read_cluster)
+    label = resolve_read_cluster_label(backend, requested)
+
+    warn_on_unconfigured_read_cluster(backend, requested, label)
+
+    case do_ch_query_on_label(backend, statement, params, label) do
+      {:error, %QueryError{kind: :connection_error}} = error ->
```
Ziinc, line 549: can add a user monitoring warning log here (metadata set with :user_id)

lib/logflare/backends/adaptor/clickhouse_adaptor/connection_manager.ex
```diff
@@ -131,63 +141,83 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.ConnectionManager do
...
+  @doc """
+  Resolves the host for a given `label`.
+
+  Falls back to the `default_read_cluster` entry, then to the deprecated
+  `read_only_url`, and finally to the primary `url`.
+  """
+  @spec read_host(Backend.t() | nil, String.t() | nil) :: String.t() | nil
+  def read_host(%Backend{config: config}, label) do
+    case extract_url_components(read_url(config, label)) do
+      {:ok, {_scheme, hostname, _port}} -> hostname
```
Ziinc, line 207: might have user/pass in hostname for basic auth, would be good to redact in case.
https://clickhouse.com/docs/concepts/features/interfaces/http#authentication

lib/logflare/backends/adaptor/clickhouse_adaptor/query_connection_sup.ex
```diff
@@ -228,12 +230,24 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.QueryConnectionSup do
...
+
+    :ok
   end
 
+  @spec lookup_managers(pos_integer()) :: [pid()]
+  defp lookup_managers(backend_id) when is_pos_integer(backend_id) do
+    Registry.select(BackendRegistry, [
+      {{{ConnectionManager, backend_id}, :"$1", :_}, [], [:"$1"]},
+      {{{ConnectionManager, backend_id, :_}, :"$1", :_}, [], [:"$1"]}
+    ])
```
Ziinc: Ex2ms for readability

lib/logflare_web/live/backends/backends_live.ex
```diff
@@ -70,51 +71,55 @@ defmodule LogflareWeb.BackendsLive do
...
-    socket =
-      case Backends.update_backend(socket.assigns.backend, params) do
-        {:ok, backend} ->
-          socket
-          |> assign(:show_rule_form?, false)
-          |> refresh_backend(backend.id)
-          |> refresh_backends()
-          |> put_flash(:info, "Successfully updated backend")
-          |> push_patch(to: with_team_param(~p"/backends/#{backend.id}", socket.assigns.team))
+    with {:ok, params} <- assemble_read_clusters(transform_params(params)) do
```
Ziinc: feels like it should be part of the transform_params function? or can rename to transform_params_read_cluster to make it clear that it accepts params and is part of the param transformation logic

nit

test/logflare_web/controllers/endpoints_controller_test.exs
```diff
@@ -992,4 +993,56 @@ defmodule LogflareWeb.EndpointsControllerTest do
...
+      user: user
+    } do
+      pid = self()
+      endpoint = insert(:endpoint, user: user, enable_auth: false)
+
+      Endpoints
+      |> expect(:run_cached_query, fn _query, _params, opts ->
+        send(pid, {:read_cluster, Keyword.get(opts, :read_cluster)})
+        {:ok, %{rows: []}}
+      end)
```
Ziinc: would be better to expect on Ch to verify the read cluster url instead

Ziinc: stubbing the context means that we don't get a full integration test of controller+context

test/logflare_web/controllers/endpoints_controller_test.exs
```diff
@@ -992,4 +993,56 @@ defmodule LogflareWeb.EndpointsControllerTest do
...
+      refute conn.halted
+
+      assert_received {:read_cluster, "api"}
+    end
+
+    test "sets :read_cluster to nil when the header is absent", %{conn: init_conn, user: user} do
+      pid = self()
+      endpoint = insert(:endpoint, user: user, enable_auth: false)
+
+      Endpoints
```
Ziinc: would expect on Ch instead.

Ziinc (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor/connection_manager.ex
amokan: I'm just going to ensure all aspects of basic auth are stripped out of URLs as this would not even be a valid option for the type of read-read warehouse we have to use as those credentials are synced from the primary. Also worth noting that OSS ClickHouse is not really a use-case here either as it does not support read-only compute isolation.

Ziinc (approved):

PR #3701, opened 2026-07-17
fix: decouple event id in gen table from actual log event id

Baishan:
Decouple gen event id from the real event id to allow duplicates in loadtest

### DIFF
--- a/lib/logflare/backends/ingest_event_queue.ex
+++ b/lib/logflare/backends/ingest_event_queue.ex
@@ -675,24 +675,32 @@ defmodule Logflare.Backends.IngestEventQueue do
     end
   end
 
-  # Writes each event's full body into its current generation's table first, then the
-  # lightweight pointer row into `queue_tid` (the target pipeline's own table) — in that
-  # order, so a crash mid-batch can only leave an inert orphaned event (cleaned up by
-  # generation rotation), never a pointer with no backing data. Pointer row shape:
-  # {event_id, generation_tid, size, retries, event_type, day_bucket, ingest_freshness}.
+  # Claims the pointer row first — insert_new/2 into `queue_tid`, keyed by the event's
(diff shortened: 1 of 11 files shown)


djwhitt (approved):

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -675,24 +675,32 @@ defmodule Logflare.Backends.IngestEventQueue do
...
 
-        {id, gen_tid, :erlang.external_size(event.body), event.retries || 0, event.event_type,
-         event.day_bucket, event.ingest_freshness}
+      row =
+        {id, gen_tid, gen_event_id, :erlang.external_size(event.body), event.retries || 0,
+         event.event_type, event.day_bucket, event.ingest_freshness}
+
+      if :ets.insert_new(queue_tid, row) do
+        :ets.insert(gen_tid, {gen_event_id, event})
       end
```
djwhitt, line 701: Could we keep the event-first write order here? Publishing the pointer before inserting its backing generation row creates a race with `pop_pending/2`: a consumer can remove the pointer, fail to resolve the event, and permanently drop it before the backing insert runs. I reproduced this with concurrent writers and one reader: 8 of 1,000,000 unique events were lost, leaving 8 orphaned generation rows.

Since `gen_event_id` is unique, we can publish the backing row first and remove only that new row when the pointer claim is rejected:

```suggestion
      :ets.insert(gen_tid, {gen_event_id, event})

      unless :ets.insert_new(queue_tid, row) do
        :ets.delete(gen_tid, gen_event_id)
      end
```

Baishan: I'll need another look later. I don't want to do this^ because there are a ton of duplicates that weren't inserted at all due to :pending before. This would insert and delete the event for each of them. It might warrant a lookup first

PR #3702, opened 2026-07-17
fix(ingest): publish events before queue pointers

djwhitt:
## Summary

- write each generation-store event before publishing its queue pointer
- delete only the new generation row when a duplicate pointer is rejected
- cover duplicate-pointer cleanup without changing the existing first-pointer-wins behavior

## Context

Follow-up to #3701 and [its review discussion](https://github.com/Logflare/logflare/pull/3701#discussion_r3606877665). Publishing the pointer first allows `pop_pending/2` to claim it before the backing event exists, permanently dropping the event and leaving the later generation insert orphaned.

A focused stress reproduction on the merged ordering lost 8 of 1,000,000 unique events with 24 writers and one reader. With this change, all 1,000,000 events were processed and no generation rows remained.

### DIFF
--- a/lib/logflare/backends/ingest_event_queue.ex
+++ b/lib/logflare/backends/ingest_event_queue.ex
@@ -675,15 +675,11 @@ defmodule Logflare.Backends.IngestEventQueue do
     end
   end
 
-  # Claims the pointer row first — insert_new/2 into `queue_tid`, keyed by the event's
-  # own id, is atomic per key, so a duplicate id already pending in this same queue is
-  # rejected outright rather than overwriting (and thereby orphaning) whatever
-  # generation-store row the existing pointer references. Only once the claim succeeds
-  # is the event's full body written into the current generation's table (`gen_tid`),
-  # under a freshly generated `gen_event_id` rather than the event's own id — this
-  # decouples the shared store's key from the (possibly duplicated) event id, so
-  # independent producers that each receive their own copy of the same id never collide
-  # on the same generation-store row. Pointer row shape: {event_id, generation_tid,
+  # Writes the event body into the generation table before publishing its pointer, so a
+  # consumer can never claim a pointer whose backing row is not yet visible. The fresh
+  # `gen_event_id` makes this row independent from any duplicate event id. If insert_new/2
+  # rejects the pointer or the queue disappears during publication, only this insertion's
+  # generation row is removed. Pointer row shape: {event_id, generation_tid,
   # gen_event_id, size, retries, event_type, day_bucket, ingest_freshness}.
   defp insert_pointer_batch(sid_bid_pid, queue_tid, batch) do
     queues_key = pointer_queues_key(sid_bid_pid)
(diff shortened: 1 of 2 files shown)


lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -696,8 +692,10 @@ defmodule Logflare.Backends.IngestEventQueue do
         {id, gen_tid, gen_event_id, :erlang.external_size(event.body), event.retries || 0,
          event.event_type, event.day_bucket, event.ingest_freshness}
 
-      if :ets.insert_new(queue_tid, row) do
-        :ets.insert(gen_tid, {gen_event_id, event})
+      :ets.insert(gen_tid, {gen_event_id, event})
+
+      unless :ets.insert_new(queue_tid, row) do
```
amokan:
```suggestion
      if not :ets.insert_new(queue_tid, row) do
```

`unless` was [depreciated in Elixir 1.18](https://elixir.hexdocs.pm/1.18.4/Kernel.html#unless/2)... should replace with `if not`

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -696,8 +692,10 @@ defmodule Logflare.Backends.IngestEventQueue do
         {id, gen_tid, gen_event_id, :erlang.external_size(event.body), event.retries || 0,
          event.event_type, event.day_bucket, event.ingest_freshness}
 
-      if :ets.insert_new(queue_tid, row) do
-        :ets.insert(gen_tid, {gen_event_id, event})
+      :ets.insert(gen_tid, {gen_event_id, event})
```
amokan, line 695: 🤖 

> In the rescue ArgumentError fallback (queue owner died mid-batch, table gone), the new ordering means :ets.insert(gen_tid, ...) for the in-flight item can succeed before :ets.insert_new(queue_tid, row) raises. That raise skips the unless cleanup branch entirely (it never returns false, it throws), so the loop aborts and the whole batch is retried under the fallback identity (add_to_table(put_elem(sid_bid_pid, 2, nil), batch)), writing a second, fresh generation row for the same event. The first row is now orphaned until GenerationJanitor rotates it out.

The truth is this ^^^ is not really a _new_ thing and not worth blocking on - but just wanted to call it out

amokan (approved):
Looks good. I'd suggest swapping that `unless` statement out though.

Probably should let @Baishan merge on Monday if he's good with it.

lib/logflare/backends/ingest_event_queue.ex
djwhitt: Yeah, lot's of small latent bugs in there like this I think. 🤖s are good at finding them.

Ziinc (approved):

lib/logflare/backends/ingest_event_queue.ex
djwhitt: Updated this to `if not`.

lib/logflare/backends/ingest_event_queue.ex
djwhitt: Addressed this as well: if pointer publication raises because the queue disappeared, the just-written generation row is now removed before the existing startup-queue fallback runs. Added regression coverage asserting only the fallback row remains.

PR #3711, opened 2026-07-21
fix(clickhouse): trip circuit breaker immediately on TOO_MANY_PARTS

djwhitt:
## Context

Incident analysis of a ClickHouse parts explosion showed parts climbing through `parts_to_throw_insert`, followed by error 252 (`TOO_MANY_PARTS`) and a large insert-retry storm. Once a partition is over the throw threshold, retrying inserts only adds load while background merges need time to catch up.

## Changes

- Do not retry `TOO_MANY_PARTS` responses in the Tesla HTTP retry middleware.
- Add synchronous `CircuitBreaker.trip/1` to open the breaker immediately, bypassing the normal failure-count threshold.
- Share ClickHouse error classification between the HTTP ingester and pipeline so code 252 trips immediately while other errors retain existing failure counting.
- Record the open reason (`:threshold` or `:forced`) in logs and telemetry.
- Cover the current generation-pointer pipeline and verify error 252 results in only one HTTP request.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
@@ -17,6 +17,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
   @max_delay 5_000
   @pool_timeout 8_000
   @receive_timeout 20_000
+  @too_many_parts_marker "Code: 252."
 
   @doc """
   Inserts a list of `LogEvent` structs into ClickHouse.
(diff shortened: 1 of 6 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -17,6 +17,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
   @max_delay 5_000
   @pool_timeout 8_000
   @receive_timeout 20_000
+  @too_many_parts_marker "Code: 252"
```
amokan:
```suggestion
  @too_many_parts_marker "Code: 252."
```

Minor nit, but this would protect against a future/hypothetical code such as `2529` since ClickHouse emits a trailing period after the code.

amokan (approved):
Looks good! I have one note around the trailing period for the error code string, but not required.

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
djwhitt: Will look into this. I think there may be multiple 252 messages we need to match.

djwhitt: Confirmed ClickHouse emits the trailing period. Updated the marker to `Code: 252.` and added coverage that excludes a hypothetical prefixed code such as `2529`.

PR #3720, opened 2026-07-22
refactor: use assigns for log event modal state

msmithstubbs:
Refactors SearchLV modals to use assigns rather than params from HTML attributes. This prepares SearchLV for rendering events with a LiveView stream where HTML attributes, such as `tailing?`, aren't updated after an event is rendered

## Demo/QA
Functionality is the same but this video demonstrates event context modal pausing and resuming tailing. 

https://github.com/user-attachments/assets/814dc742-c531-4974-9b8c-9840c65af6bf


Ziinc (approved):

Ziinc:
It looks like opening up the context modal will pause the reloading, but opening up the event does not. Looks like a small discrepancy, can be a follow-up.

PR #3722, opened 2026-07-23
refactor(cloudbuild): replace GCE container declarations with startup scripts

djwhitt:
## Summary

- add an idempotent Container-Optimized OS startup script that authenticates to GCR and starts Logflare with the existing privileged, host-networked, always-restarting container behavior
- preserve Konlet's host firewall rules and rotated `json-file` logging so the existing COS logging agent continues exporting container logs
- create plain GCE instance templates with startup, shutdown, image, and per-cluster environment metadata instead of `create-with-container`
- migrate the manual instance-template helper and reject stale existing templates that still use the deprecated container declaration

## Validation

- `shellcheck cloudbuild/gce-startup.sh cloudbuild/create-instance-template.sh scripts/create-instance-templates.sh test/cloudbuild/gce_scripts_test.sh`
- `bash -n` for all changed shell scripts
- `test/cloudbuild/gce_scripts_test.sh`
- parsed both changed Cloud Build configs with PyYAML and executed their environment-generation steps with representative substitutions

## Staging validation

Exercised the changed staging Cloud Build config in two stages, using the existing staging build/deploy tooling. No serving traffic was affected at any point.

**Isolated instance test:** built a fresh instance template from this branch and booted a single, disposable instance in its own instance group (not the real serving group), then tore the disposable resources down afterward. Confirmed:

- the generated instance template carries `startup-script`/`shutdown-script`/image/env metadata and no `gce-container-declaration`
- the startup script completed successfully on first boot: GCR auth succeeded, the image pulled, and the container started
- the running container matches today's behavior: privileged, host networking, always-restart, and the same rotated `json-file` logging configuration
- the host firewall rules the startup script installs are present and idempotent on rerun
- the instance joined the real staging Erlang cluster and reported the expected cluster peers, cross-checked two independent ways: the app's own health endpoint, and a separate live RPC call into the running node
- container logs continue to land in Cloud Logging under the same log stream, with the expected container and application labels intact

**Canary on the real serving group:** rolled the same tested template out as a single-instance canary on the real staging serving group, then rolled it back. Confirmed:

- the canary instance received real load-balancer-routed traffic and joined the live staging cluster alongside the existing fleet, with no impact to the rest of the group
- the shutdown script drains the instance gracefully during the rolling replacement back to the previous template

### DIFF
--- a/test/cloudbuild/gce_scripts_test.sh
+++ b/test/cloudbuild/gce_scripts_test.sh
@@ -2,10 +2,13 @@
 
 set -euo pipefail
 
-readonly REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+readonly REPO_ROOT
 readonly STARTUP_SCRIPT="${REPO_ROOT}/cloudbuild/gce-startup.sh"
 readonly CREATE_SCRIPT="${REPO_ROOT}/cloudbuild/create-instance-template.sh"
-readonly TEST_DIR="$(mktemp -d)"
+readonly TEMPLATE_SCRIPT="${REPO_ROOT}/scripts/create-instance-templates.sh"
+TEST_DIR="$(mktemp -d)"
+readonly TEST_DIR
 
 trap 'rm -rf "${TEST_DIR}"' EXIT
 
@@ -162,6 +165,16 @@ if [[ "$3" == "describe" ]]; then
       printf '%s\n' '{"properties":{"metadata":{"items":[{"key":"startup-script"}]}}}'
       ;;
   esac
+elif [[ "$3" == "create" && -n "${ENV_CAPTURE_DIR:-}" ]]; then
+  template="$4"
+  for arg in "$@"; do
+    case "${arg}" in
+      --metadata-from-file=*logflare-container-env=*)
+        env_file="${arg##*logflare-container-env=}"
+        cp "${env_file}" "${ENV_CAPTURE_DIR}/${template}.env"
+        ;;
+    esac
+  done
 fi
 STUB
   chmod +x "${path}"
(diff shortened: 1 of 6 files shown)


test/cloudbuild/gce_scripts_test.sh
```diff
@@ -0,0 +1,256 @@
...
+  : >"${call_log}"
+  : >"${iptables_state}"
+  install_startup_stubs "${TEST_DIR}/startup-bin"
+
+  run_startup_script "${call_log}" "${iptables_state}" "${docker_home}" "${env_file}"
+
+  [[ "$(stat -c '%a' "${env_file}")" == "600" ]] || fail "container env file is not mode 600"
+  assert_contains "${env_file}" "RELEASE_COOKIE=test-cookie"
+  assert_contains "${call_log}" "credential HOME=${docker_home} configure-docker --registries=gcr.io"
+  assert_contains "${call_log}" "docker run --name=logflare --privileged --restart=always --network=host --detach --log-driver=json-file --log-opt max-size=500m --log-opt max-file=3 --env-file ${env_file} gcr.io/test-project/logflare_app:test-tag"
```
Ziinc: Assertions are overly comprehensive, we just want to assert the the script (and important commsnds/options) are created correctly, we're testing implementation of config values at this point

test/cloudbuild/gce_scripts_test.sh
```diff
@@ -0,0 +1,256 @@
...
+  assert_contains "${output}" "Created        : 8"
+  assert_contains "${capture_dir}/logflare-prod-1-2-3-prod-a.env" "LOGFLARE_ALERTS_ENABLED=true"
+  assert_contains "${capture_dir}/logflare-prod-1-2-3-prod-b.env" "LOGFLARE_ALERTS_ENABLED=false"
+}
+
+test_startup_script
+test_create_script
+test_manual_template_script
+
+echo "gce script tests passed"
```
Ziinc: This test harness is not the easiest for a non-clanker to parse and scan through, would advise simplifying or using node+jest where the utilities are already baked into the test harness.
Fine to keep using bash for the test script, just needs to be more readable

Ziinc (approved):

test/cloudbuild/gce_scripts_test.sh
djwhitt: Simplified the Bash test from 256 to 83 lines and organized it into two focused configuration checks, removing the custom command-stub harness.

test/cloudbuild/gce_scripts_test.sh
djwhitt: Simplified in 6ee283bb. The test now checks only the important startup and instance-template configuration, without emulating Docker, iptables, metadata, reruns, or failure behavior.

djwhitt: Followed up in f2f830b7 so the assertions target the semantic contract—required commands, options, metadata keys, and absence of the legacy command—without pinning exact shell/YAML formatting or variable expansions.

PR #3723, opened 2026-07-23
chore: remove stale batcher lane from ClickHouse pipeline

amokan:
Now that we are not allowing data older than 24h in, the stale lane in the ClickHouse pipeline is just bloat.

Note that this also removes the `freshness` tag from the telemetry so any grafana charts will need to be updated when this rolls.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
@@ -6,10 +6,8 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
   consolidated queue and inserting them into the backend's type-specific
   ingest tables (otel_logs, otel_metrics, otel_traces).
 
-  Events are routed to one of two batchers (`:ch_fresh` or `:ch_stale`)
-  based on `LogEvent.ingest_freshness`, and batched by a composite key of
-  `{event_type, day_bucket}` so each insert targets a single ClickHouse
-  partition. Multiple sources are processed together in a single pipeline.
+  Events are batched by a composite key of `{event_type, day_bucket}` so each
+  insert targets a single ClickHouse partition.
 
   Uses ID-passing: the producer emits `LogEventPointer`s (id + routing metadata)
   while the full events live in a separate generation store (see
@@ -44,18 +42,14 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
   @processor_concurrency 6
   @processor_min_demand 100
   @processor_max_demand 1_000
-  @fresh_batch_size 60_000
-  @fresh_batch_timeout 5_000
-  @fresh_batcher_concurrency 32
-  @stale_batch_size 60_000
-  @stale_batch_timeout 12_000
-  @stale_batcher_concurrency 8
+  @batch_size 60_000
+  @batch_timeout 5_000
+  @batcher_concurrency 32
   @max_retries 1
-  # One full batch per fresh/stale batcher lane, used as a generous safety valve rather
+  # Two full batches across every batcher, used as a generous safety valve rather
   # than a fine-grained flow-control knob — see BufferProducer.capped_fetch_amount/2.
   # It should only cap genuinely runaway backlog during healthy operation.
-  @max_in_flight 2 * @fresh_batch_size * @fresh_batcher_concurrency +
-                   @stale_batch_size * @stale_batcher_concurrency
+  @max_in_flight 2 * @batch_size * @batcher_concurrency
 
   @doc false
   @spec max_retries() :: non_neg_integer()
@@ -63,7 +57,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
 
   @doc false
   @spec max_batch_size() :: pos_integer()
-  def max_batch_size, do: @fresh_batch_size
+  def max_batch_size, do: @batch_size
 
   @doc false
   @spec max_in_flight() :: pos_integer()
@@ -97,7 +91,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
              consolidated: true,
              id_passing: true,
              max_in_flight: @max_in_flight,
-             seed_batch_size: @fresh_batch_size
+             seed_batch_size: @batch_size
            ]},
         transformer: {__MODULE__, :transform, [backend_id: backend.id]},
         concurrency: @producer_concurrency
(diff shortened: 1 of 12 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -44,26 +42,22 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
   @processor_min_demand 100
   @processor_max_demand 1_000
-  @fresh_batch_size 60_000
-  @fresh_batch_timeout 5_000
-  @fresh_batcher_concurrency 32
-  @stale_batch_size 60_000
-  @stale_batch_timeout 12_000
-  @stale_batcher_concurrency 8
+  @batch_size 60_000
+  @batch_timeout 5_000
```
Ziinc, line 46: lets tweak up the batch timeout as well, 6500 maybe?
```suggestion
  @batch_timeout 6_500
```

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -44,26 +42,22 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
+  @batcher_concurrency 32
```
Ziinc, line 47: should by dynamic based on scheduiler online with a min of somthing like 4. for self-hosters. this will increase ram usage on small instances in future once we support clickhouse for self hosting.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -44,26 +42,22 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
   @max_retries 1
-  # One full batch per fresh/stale batcher lane, used as a generous safety valve rather
+  # Two full batches across every batcher, used as a generous safety valve rather
   # than a fine-grained flow-control knob — see BufferProducer.capped_fetch_amount/2.
   # It should only cap genuinely runaway backlog during healthy operation.
-  @max_in_flight 2 * @fresh_batch_size * @fresh_batcher_concurrency +
-                   @stale_batch_size * @stale_batcher_concurrency
+  @max_in_flight 2 * @batch_size * @batcher_concurrency
```
Ziinc, line 52: would it be better to track requests in flight as opposed to volume in flight? 
small volume with large number of requests will have adverse effect as well, 
while large volume with small number of requests is actually ideal.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -151,32 +140,19 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
+        %Message{data: %LogEventPointer{event_type: event_type, day_bucket: day_bucket}} = message,
         _context
       )
-      when is_event_type(event_type) and freshness in [:fresh, :stale] do
+      when is_event_type(event_type) do
     message
-    |> Message.put_batcher(ingest_freshness_to_batcher(freshness))
-    |> Message.put_batch_key({event_type, message.data.day_bucket})
+    |> Message.put_batcher(:ch)
+    |> Message.put_batch_key({event_type, day_bucket})
```
Ziinc, line 149: i'm thinking that having a different batch timeout for different less-important event_types like traces/logs could allow us to also increase the batch size for these tables, especially since there is no latency requirement for these tables. 
we can customize the batch_size function to achieve this

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -31,10 +31,9 @@ defmodule Logflare.Backends.IngestEventQueue do
   @ets_recent_events :ingest_event_queue_recent_events
   @max_queue_size 30_000
   @consolidated_max_queue_size 120_000
-  @pointer_batch_key_match_spec (for freshness <- [:fresh, :stale],
-                                     event_type <- [:log, :metric, :trace] do
-                                   {{:_, :_, :_, :_, :_, event_type, :"$1", freshness},
-                                    [{:is_integer, :"$1"}], [{{freshness, event_type, :"$1"}}]}
+  @pointer_batch_key_match_spec (for event_type <- [:log, :metric, :trace] do
```
Ziinc, line 34: just a note that using :ex2ms will make this a lot more readable

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -961,15 +960,14 @@ defmodule Logflare.Backends.IngestEventQueue do
...
         n
       )
-      when freshness in [:fresh, :stale] and
-             event_type in [:log, :metric, :trace] and is_integer(day_bucket) and
+      when event_type in [:log, :metric, :trace] and is_integer(day_bucket) and
              is_integer(n) and n > 0 do
     ms = [
-      {{:"$1", :"$2", :"$3", :"$4", :"$5", event_type, day_bucket, freshness}, [],
-       [{{:"$1", :"$2", :"$3", :"$4", :"$5", event_type, day_bucket, freshness}}]}
+      {{:"$1", :"$2", :"$3", :"$4", :"$5", event_type, day_bucket}, [],
```
Ziinc, line 969: as above, using ex2ms would make this more readable

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -1044,7 +1041,7 @@ defmodule Logflare.Backends.IngestEventQueue do
   def pop_pending(_, 0), do: {:ok, []}
 
   def pop_pending(sid_bid_pid, n) when is_integer(n) do
-    ms = [{{:"$1", :"$2", :"$3", :_, :_, :_, :_, :_}, [], [{{:"$1", :"$2", :"$3"}}]}]
+    ms = [{{:"$1", :"$2", :"$3", :_, :_, :_, :_}, [], [{{:"$1", :"$2", :"$3"}}]}]
```
Ziinc, line 1044: ex2ms

Ziinc (approved):

lib/logflare/backends/ingest_event_queue.ex
djwhitt: TIL

Baishan (approved):

djwhitt (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
Baishan: The batch scheduler will keep making batches even if handle_batch blocks. Max  in_flight limits how many batches can be queued up throttling the producer if the pipeline is backlogged. Its only used as load shedding mechanism and doesn't affect number of requests. Requests scale up the faster batches are filled up to batcher_concurrency so they are limited by batch_timeout.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
amokan: I was going to do the same, but want to avoid making too many changes in this PR given this is part of some larger adjustments - but will for sure take that into account.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
amokan: Yep. Again, I think this is temporary - want to eliminate any fogginess in the very short-term. I'll make a note to adjust this in a week.

PR #3726, opened 2026-07-23
chore: re-introduce some ClickHouse async config options

amokan:
Brings back the option for ClickHouse async inserts, this time based on a configurable batch size boundary and support for a dedicated async ingest cluster. If enabled, this allows us to route batches under a row count threshold as async inserts. When a dedicated async cluster is configured, this allows for larger batch building on the ClickHouse side if you have a primary cluster that scaled wide horizontally.

New optional configs for the ClickHouse backend:
- `use_async_inserts_for_small_batches` (boolean, defaults to false)
- `async_insert_max_rows` (integer, defaults to 1000)
- `async_insert_cluster_url` (string)


lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -356,22 +362,34 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
...
           table :: String.t(),
           TypeDetection.event_type(),
-          opts :: keyword()
+          opts :: keyword(),
+          async? :: boolean()
         ) :: String.t()
-  defp build_request_url(connection_opts, table, event_type, opts) do
-    base_url = Keyword.get(connection_opts, :url)
+  defp build_request_url(connection_opts, table, event_type, opts, async?) do
+    base_url = insert_base_url(connection_opts, async?)
```
djwhitt: Could we derive the complete origin, including the port, from the selected async URL? `build_request_url/5` currently takes the async URL’s scheme and host but still uses the primary backend’s `:port`. I reproduced this with primary port `8123` and `async_insert_cluster_url: "https://async-cluster.local:9444"`—the Finch request used port `8123`.

This will break dedicated endpoints whenever their ports differ. Please preserve `uri.port` for the dedicated URL, use its scheme’s default when absent, and add an assertion for `request.port`.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -141,7 +141,10 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
        read_only_url: :string,
        insert_protocol: :string,
        native_port: :integer,
-       native_pool_size: :integer
+       native_pool_size: :integer,
+       use_async_inserts_for_small_batches: :boolean,
+       async_insert_cluster_url: :string,
```
djwhitt, line 147: Should `test_connection/1` also validate this endpoint when async routing is enabled? At present, a backend with `use_async_inserts_for_small_batches: true` and an unreachable `async_insert_cluster_url` still returns `:ok`, because only the primary and read endpoints are checked.

The provisioner relies on this check, so it can start successfully and then route qualifying batches to an unusable endpoint; those events are eventually dropped after retry exhaustion. Please add a conditional connectivity/grant check and regression coverage for the async endpoint.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -359,21 +361,25 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
-  defp finalize_insert(backend, event_type, compressed, good, bad) do
+  defp finalize_insert(backend, event_type, compressed, good_count, good, bad) do
+    insert_opts = [async: async_insert?(backend, good_count)]
+
     case ClickHouseAdaptor.insert_log_events_compressed(
            backend,
            event_type,
-           compressed
+           compressed,
+           insert_opts
```
djwhitt, line 382: Suggestion: now that this call can route to a dedicated async endpoint, the error log in `insert_log_events_compressed/4` still records `config_host(backend)`, which is always the primary host. Could we make that metadata route-aware so async failures identify the endpoint that actually failed?

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
amokan: Fixed

lib/logflare/backends/adaptor/clickhouse_adaptor/endpoint_utils.ex
```diff
@@ -0,0 +1,124 @@
...
+
+      iex> resolve_port(URI.parse("https://host"), 8443)
+      8443
+
+      iex> resolve_port(%URI{port: nil}, 8123)
+      8123
+  """
+  @spec resolve_port(URI.t(), pos_integer() | nil) :: pos_integer() | nil
+  def resolve_port(%URI{port: port}, fallback)
+      when port in [nil, 80, 443] and is_pos_integer(fallback),
```
djwhitt: `URI.parse/1` normalizes both an omitted port and an explicitly supplied standard port to `80`/`443`, so this guard also overwrites valid URLs such as `http://host:80` and `https://host:443` with the primary backend port. I reproduced an async URL ending in `:80` producing a Finch request to the primary port `8123`.

Could we detect explicit-port presence from the original URL using `:uri_string.parse/1`, which preserves that distinction?

```elixir
explicit_port =
  case :uri_string.parse(url) do
    %{port: port} -> port
    _ -> nil
  end

port = explicit_port || fallback_port || default_port(scheme)
```

This retains the current fallback behavior when the port is omitted while preserving every explicitly configured port. Please also add a request-port regression case for explicit `:80` or `:443`.

djwhitt (approved):

PR #3727, opened 2026-07-23
feat(config): make credential and cert paths configurable

djwhitt:
Part of [O11Y-2137](https://linear.app/supabase/issue/O11Y-2137).

## What

Makes Logflare's file-based credentials and certificates locatable via environment variables, and lets the BigQuery service account key be supplied inline as JSON.

New environment variables (all with backward-compatible defaults):

| Variable | Default | Purpose |
| --- | --- | --- |
| `GOOGLE_APPLICATION_CREDENTIALS` | `gcloud.json` | Path to the BigQuery service account key file |
| `GOOGLE_APPLICATION_CREDENTIALS_JSON` | _(unset)_ | Service account key JSON provided inline; takes precedence over the file when set |
| `LOGFLARE_TLS_CERT_PATH` | `cert.pem` | gRPC TLS certificate path |
| `LOGFLARE_TLS_KEY_PATH` | `cert.key` | gRPC TLS key path |
| `DB_SSL_CA_CERT_PATH` | `db-server-ca.pem` | Internal DB SSL server CA cert path |
| `DB_SSL_CLIENT_CERT_PATH` | `db-client-cert.pem` | Internal DB SSL client cert path |
| `DB_SSL_CLIENT_KEY_PATH` | `db-client-key.pem` | Internal DB SSL client key path |

## Why

Today `config/runtime.exs` reads these secrets as fixed filenames in the working directory (`gcloud.json`, `cert.pem`, `db-*.pem`). That assumes secrets are baked into the image at a known path, which is how the current GCE deployment works but does not fit environments that inject secrets as environment variables or mounted secret volumes (e.g. Kubernetes / Helm).

These changes decouple credential loading from a fixed working directory:

- **Configurable paths** let a deployment mount a secret volume anywhere and point Logflare at it.
- **Inline JSON credentials** let the BigQuery key ride the same env-var/secret injection path as every other secret, with no file mount at all.

All defaults preserve the existing filenames, so current deployments are unaffected.

## Changes

- `config/runtime.exs`: resolve credential/cert locations from env vars with the existing filenames as defaults; support inline `GOOGLE_APPLICATION_CREDENTIALS_JSON`.
- `docs/docs.logflare.com/docs/self-hosting/index.md`: document the new variables under the Database SSL and BigQuery setup sections.

## Notes

Backward compatible — no behavior change when the new variables are unset.

## Related PRs

Part of the Logflare-on-Kubernetes secrets migration:

- #3719 — initial Helm chart (base for the chart work)
[...]

PR #3728, opened 2026-07-23
feat(helm): wire in secret file mounts, reloader, and extra config

djwhitt:
Part of [O11Y-2137](https://linear.app/supabase/issue/O11Y-2137).

> Stacked on top of #3719 (`feat/helm`). Merge target is `feat/helm` (or `main` once #3719 lands). Still iterating, so the chart version is intentionally **not** bumped yet — the version-gate check will be red until it is.

## What

Wires the chart to carry the full secret/config surface a real deployment needs, so it can consume the ExternalSecret-materialised Secrets (`logflare-secrets`, `logflare-cert-files`) from the k8s-observability side.

- **`certFilesSecret` / `certFilesMountPath`** — mounts a Secret whose keys are cert filenames (`db-server-ca.pem`, `db-client-cert.pem`, `db-client-key.pem`, `cert.pem`, `cert.key`) as files, and sets `DB_SSL_CA_CERT_PATH` / `DB_SSL_CLIENT_CERT_PATH` / `DB_SSL_CLIENT_KEY_PATH` / `LOGFLARE_TLS_CERT_PATH` / `LOGFLARE_TLS_KEY_PATH` to point at the mount. Logflare reads these from disk, so `envFrom` can't deliver them.
- **`reloader`** — optional Stakater Reloader annotation so the Deployment rolls when its ConfigMap or referenced Secrets change (the chart can't checksum Secrets it doesn't own).
- **`extraConfig`** — free-form map of non-secret env vars rendered verbatim into the ConfigMap, for the long tail of config not modeled as first-class values (e.g. `SPOOL_*`, `LIBCLUSTER_TOPOLOGY`, `GOOGLE_*_SA`, per-env cluster sizing).

## Notes

- All new behavior is **default-off**; existing rendering is unchanged (existing unit tests stay green). Added helm-unittest coverage for each new path and documented them in the README.
- The BigQuery service-account key is handled as an env-var secret (`GOOGLE_APPLICATION_CREDENTIALS_JSON`), not a mounted file.
- **Depends on** the configurable-path / inline-JSON support in `config/runtime.exs` (logflare #3727) for the `*_PATH` vars and `GOOGLE_APPLICATION_CREDENTIALS_JSON` to take effect at runtime.
- Secret var names are deployment-mode dependent (e.g. SaaS uses `SUPER_USER_AUTHENTICATION_TOKEN`; single-tenant uses `LOGFLARE_PUBLIC/PRIVATE_ACCESS_TOKEN`). The chart stays agnostic — `envFrom` injects whatever keys the referenced Secret has — so this isn't hardcoded.

## Related PRs

Part of the Logflare-on-Kubernetes secrets migration:

- #3719 — initial Helm chart (this PR stacks on it)
- #3727 — app: configurable credential/cert paths + inline JSON (runtime dependency)
- supabase/k8s-observability#574 — encrypted secrets synced to SSM Parameter Store
- supabase/k8s-observability#573 — ExternalSecrets that materialize the Secrets this chart consumes

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -47,6 +47,25 @@ logflare:
   secretRefs: []
   # - logflare-secrets
 
+  # Cert/key files (internal DB SSL and gRPC TLS) come from a mounted Secret
+  # rather than envFrom, because Logflare reads them from disk. Set to the name
+  # of a Secret you manage yourself whose keys are the cert filenames
+  # (db-server-ca.pem, db-client-cert.pem, db-client-key.pem, cert.pem,
+  # cert.key); empty disables the mount. The chart points the matching
+  # DB_SSL_*_PATH / LOGFLARE_TLS_*_PATH env vars at the mount for you.
+  certFilesSecret: ""
+  certFilesMountPath: /etc/logflare/certs
+
+  # Roll the Deployment automatically when its ConfigMap or referenced Secrets
+  # change. Requires the Stakater Reloader controller to be installed.
+  reloader: false
+
+  # Free-form non-secret environment variables rendered verbatim into the
+  # ConfigMap. Use this for configuration not modeled as first-class values
+  # above (e.g. LOGFLARE_LOG_LEVEL, LIBCLUSTER_TOPOLOGY, SPOOL_*, GOOGLE_*_SA).
+  extraConfig: {}
+    # LOGFLARE_LOG_LEVEL: info
+
   # Logflare's own metadata Postgres database (distinct from the optional
   # "postgres" event-storage backend below).
   db:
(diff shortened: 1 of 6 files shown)


stigglor (approved):

helm/values.yaml
```diff
@@ -47,6 +47,25 @@ logflare:
...
+  # rather than envFrom, because Logflare reads them from disk. Set to the name
+  # of a Secret you manage yourself whose keys are the cert filenames
+  # (db-server-ca.pem, db-client-cert.pem, db-client-key.pem, cert.pem,
+  # cert.key); empty disables the mount. The chart points the matching
+  # DB_SSL_*_PATH / LOGFLARE_TLS_*_PATH env vars at the mount for you.
+  certFilesSecret: ""
+  certFilesMountPath: /etc/logflare/certs
+
+  # Roll the Deployment automatically when its ConfigMap or referenced Secrets
+  # change. Requires the Stakater Reloader controller to be installed.
```
Ziinc, line 60: Feels like reloading should be done outside of this chart, would defer to @stigglor opinion here

djwhitt: Would adding a `deploymentAnnotations` map to make it open ended be reasonable?

PR #3734, opened 2026-07-24
feat(helm): add secret/config consumption wiring (cert mounts, reloader, extraConfig)

djwhitt:
Stacks the secret/config **consumption** wiring from #3728 onto the `feat/helm-external-secrets` branch (#3729), so the chart that #575 deploys can actually use the secrets ESO materializes. #3729 renders the ExternalSecrets; this adds the half that reads them.

Targets `feat/helm-external-secrets` so everything lands on the branch supabase/k8s-observability#575 points at. Supersedes #3728 (which was a parallel branch off `feat/helm`) — close that one once this merges.

## What this adds (cherry-picked from the approved #3728)

- **Cert-file mount:** `logflare.certFilesSecret` / `certFilesMountPath` mount the `logflare-cert-files` Secret as files and set `DB_SSL_CA_CERT_PATH` / `DB_SSL_CLIENT_CERT_PATH` / `DB_SSL_CLIENT_KEY_PATH` / `LOGFLARE_TLS_CERT_PATH` / `LOGFLARE_TLS_KEY_PATH`. Without this, ESO creates `logflare-cert-files` but the app never sees the certs (it reads them from disk).
- **`logflare.extraConfig`:** free-form non-secret env vars rendered into the ConfigMap — for the long tail of per-env config (e.g. `SPOOL_*`, `LIBCLUSTER_TOPOLOGY`, `LOGFLARE_LOG_LEVEL`).
- **`logflare.reloader`:** optional Stakater Reloader annotation so the Deployment rolls on ConfigMap/Secret changes.

All default-off; existing rendering is unchanged. Includes the helm-unittest coverage from #3728. Clean cherry-pick — no version double-bump (Chart.yaml stays at #3729's `0.2.0`).

## Follow-ups (not blocking)

- README now documents cert files two ways — the dedicated `certFilesSecret` (this PR) and the generic `volumes`/`volumeMounts` example (#3729). Worth consolidating on `certFilesSecret` since it also wires the `*_PATH` env vars.
- supabase/k8s-observability#575's `valuesObject` should set `certFilesSecret: logflare-cert-files` (rather than manual volumes) to get the mount + path env vars.

Depends on / stacks with #3719 and #3729.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -219,8 +219,27 @@ Per-entry `refreshInterval` and `creationPolicy` overrides are also supported; s
 | `logflare.backend.bigquery.datasetLocation` | `GOOGLE_DATASET_LOCATION` | bigquery only |
 | `logflare.backend.postgres.schema` | `POSTGRES_BACKEND_SCHEMA` | postgres only |
 | `logflare.secretRefs` | see [Loading secrets](#loading-secrets) | |
+| `logflare.certFilesSecret` | — | Name of a Secret whose keys are cert filenames; mounted as files. See [Certificate files](#certificate-files) |
+| `logflare.certFilesMountPath` | `DB_SSL_*_PATH`, `LOGFLARE_TLS_*_PATH` | Mount path for `certFilesSecret`; the chart points the cert path env vars here |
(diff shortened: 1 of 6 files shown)


djwhitt:
Closing in favor of #3728 — the separate stacked PR adds more confusion than it's worth.

PR #3737, opened 2026-07-24
chore: remove ClickHouse native protocol implementation

amokan:
The ClickHouse native protocol implementation for inserts was not being used so it makes sense to remove this code. Can always be implemented again later from git history if we find the urge to do so.


djwhitt (approved):
Looks great! Neither I nor my 🤖 could find any issues.

PR #3741, opened 2026-07-26
perf(logs): optimize LogEvent ingest transformations

djwhitt:
## Optimization summary

1. **Construct `LogEvent` directly** — removes the Ecto changeset, intermediate map merge, and impossible pre-validation branches from `LogEvent.make/2` after the mapper has already guaranteed a valid body map.
2. **Build deep key maps with `:maps.fold/3`** — avoids the temporary tuple list created by `Enum.map/2` plus `Map.new/1` during recursive key updates.
3. **Build cleaned metadata maps with `:maps.fold/3`** — filters nested empty values directly into the destination map instead of going through the `Enumerable` protocol.
4. **Fast-path BigQuery-safe field names** — returns already-valid keys unchanged and reserves the full rewrite for keys that actually need normalization.
5. **Fuse ingest cleaning and key normalization** — combines two full recursive body traversals into one map/list walk, with explicit database and ingestion mapping entry points selecting the appropriate cleaner.
6. **Scan safe keys in four-byte chunks** — reduces recursive calls while checking the common ASCII-safe key path.

### 1. Construct `LogEvent` directly

`LogEvent.make/2` previously cast the mapper result through an Ecto changeset, validated that `body` was present, assembled another map with `Map.merge/2`, and finally converted that map into a struct.

The mapper always returns a body map with the generated ID and normalized timestamp merged into it, so the changeset could not add useful validation on this path. The new code constructs `%LogEvent{}` directly and leaves the existing transformation and BigQuery schema validators in place. It also removes the now-unreachable invalid-event transform and pre-validation clauses.

### 2. Build deep key maps with `:maps.fold/3`

`EnumDeepUpdate.update_all_keys_deep/2` previously converted every map level into a list of transformed key/value tuples and then rebuilt the map with `Map.new/1`.

The replacement folds the source map directly into a destination map with `:maps.fold/3` and `:maps.put/3`. Recursive map and list behavior is unchanged, but the temporary list and associated enumeration work are removed.

### 3. Build cleaned metadata maps with `:maps.fold/3`

`MetadataCleaner.deep_reject_nil_and_empty/1` now uses the native map fold for map nodes. Empty values are skipped, nested containers are recursively cleaned, and retained values are inserted directly into the result.

List cleaning remains order-preserving. The change removes protocol dispatch and intermediate tuple handling from every nested map visited during ingestion.

### 4. Fast-path BigQuery-safe field names

Most production keys already satisfy BigQuery's standard field-name rules. Previously, every key still passed through the complete rewrite, including binary reconstruction and post-processing.

The new predicate checks length, leading digits, reserved prefixes, and the allowed ASCII byte set. Safe keys return the original binary. Only unsafe keys enter the existing rewrite path, preserving its handling of dashes, non-alphanumeric bytes, invalid UTF-8, reserved prefixes, leading digits, and the 128-byte limit.

### 5. Fuse ingest cleaning and key normalization

The ingestion path previously walked the payload once to remove nil/empty values and again to normalize every key for BigQuery.

A dedicated `:clean_to_bigquery_column_spec` transformation now performs both operations during one recursive traversal. It cleans map and list values, drops containers that become empty, normalizes retained map keys, and preserves list order. Database reconstruction keeps the existing cleaning-only path so its behavior does not change.

Named `mapper_from_db/2` and `mapper_for_ingest/2` entry points make those lifecycle differences explicit. They inject the appropriate cleaner into the shared mapper instead of relying on a defaulted `:from_db | :ingest` mode argument.

### 6. Scan safe keys in four-byte chunks

The safe-key predicate originally consumed one byte per recursive call. It now matches and validates four bytes at a time before falling back to a single-byte tail clause.

The byte predicate is expressed as a guard, so the common ASCII-safe path performs fewer function calls without allocating substrings or changing accepted characters.

## Benchmarks

Compared `main@origin` at `fa242a0d` with the optimization stack at `b5d9111a` using `test/profiling/log_event_make_bench.exs`. The subsequent follow-ups remove a Dialyzer-confirmed unreachable pre-validation pattern and replace the mapper mode argument with named callback-based entry points.

Methodology:

- Three alternating baseline/candidate runs on the same Linux host
- 6 available cores, 11.65 GB memory
- Elixir 1.19.5, Erlang/OTP 27.3.4.6, JIT enabled
- Benchee: 2 s warmup, 5 s runtime, 3 s memory time, 3 s reduction time, parallelism 1
- Production-shaped OTEL trace (~15 leaf values) and edge log (~80 leaf values) fixtures
- Transform variants exercise copy, key-value enrichment, drop, and combined configurations
- Each cell below is the median of the three runs

Across all 12 scenarios, the stack delivers:

- **1.97–2.62x throughput** (**2.25x geometric mean**)
- **43–58% lower median latency**
- **50–57% less memory per operation**
- **51–59% fewer BEAM reductions**

| Scenario | Throughput, main → stack | Speedup | Median latency, main → stack | Memory change | Reduction change |
| OTEL trace | 120.0K → 267.4K ops/s | 2.23x | 7.42 → 3.54 µs | -57.2% | -56.8% |
| OTEL trace + copy | 114.6K → 249.6K ops/s | 2.18x | 7.75 → 3.79 µs | -54.9% | -54.0% |
| OTEL trace + drop | 107.7K → 250.1K ops/s | 2.32x | 7.83 → 3.79 µs | -53.1% | -54.8% |
| OTEL trace + KV | 88.8K → 192.3K ops/s | 2.17x | 9.33 → 4.83 µs | -55.2% | -52.7% |
| OTEL trace + copy + KV | 90.6K → 180.0K ops/s | 1.99x | 9.42 → 5.13 µs | -51.1% | -51.9% |
| OTEL trace + all | 80.7K → 158.8K ops/s | 1.97x | 10.58 → 5.54 µs | -49.7% | -51.4% |
| Edge log | 58.8K → 154.0K ops/s | 2.62x | 14.75 → 6.17 µs | -55.7% | -58.0% |
| Edge log + copy | 59.3K → 145.3K ops/s | 2.45x | 14.92 → 6.50 µs | -53.9% | -59.2% |
| Edge log + drop | 57.9K → 142.0K ops/s | 2.45x | 15.13 → 6.67 µs | -52.7% | -58.6% |
| Edge log + KV | 52.8K → 119.3K ops/s | 2.26x | 16.42 → 7.42 µs | -53.2% | -56.4% |
| Edge log + copy + KV | 50.9K → 121.7K ops/s | 2.39x | 16.96 → 7.79 µs | -52.9% | -56.8% |
| Edge log + all | 47.4K → 96.0K ops/s | 2.02x | 17.79 → 10.08 µs | -51.3% | -54.9% |

These measurements isolate `LogEvent.make/2`; they do not represent end-to-end ingestion throughput.


lib/logflare/logs/log_event.ex
```diff
@@ -234,12 +228,13 @@ defmodule Logflare.LogEvent do
...
-  defp transform(%LE{valid: false} = le, _source), do: le
+  @spec clean_ingest_body(map()) :: map()
+  defp clean_ingest_body(params),
+    do: IngestTransformers.transform(params, :clean_to_bigquery_column_spec)
 
-  defp transform(%LE{valid: true} = le, %Source{} = source) do
-    with {:ok, le} <- bigquery_spec(le),
-         {:ok, le} <- copy_fields(le, source),
+  @spec transform(LE.t(), Source.t()) :: LE.t()
+  defp transform(%LE{} = le, %Source{} = source) do
```
amokan, line 241: 🤖 

> lib/logflare/logs/log_event.ex:241-252 still produces %{le | valid: false, pipeline_error: ...}, but validate/2 at log_event.ex:209 now has only a %LE{valid: true} clause (the valid: false passthrough was removed).
>
> Currently unreachable — every clause of copy_fields/2, kv_enrich/2, and drop_fields/2 returns only {:ok, le}, and I confirmed Source.parse_*_config/1 can't produce an unmatched shape. But if any of them ever returns {:error, _}, make/2 raises FunctionClauseError in validate/2 instead of marking one event invalid — crashing the Broadway batch rather than degrading one event.
>
> Either delete the now-dead else too, or restore the one-line safety net:
defp validate(%LE{valid: false} = le, _source), do: le
I'd keep the passthrough — it's one line and this is a pipeline.

lib/logflare/logs/ingest_transformers.ex
```diff
@@ -25,6 +81,16 @@ defmodule Logflare.Logs.IngestTransformers do
   # digit rule is suppressed when a reserved prefix or dash has already added one.
   @spec to_bigquery_column_spec(term()) :: term()
   defp to_bigquery_column_spec(key) when is_binary(key) do
```
amokan, line 100: 🤖 

> :to_bigquery_column_spec has ~30 assertions and 6 properties in test/logflare/logs/ingest/ingest_transformer_test.exs, but after this PR it has zero lib/ callers — production goes through :clean_to_bigquery_column_spec (log_event.ex:233), which has no direct tests. The PR changes 4 files, all in lib/. Coverage today is only incidental via log_event_test.exs.
>
>Cheapest high-value fix — a property asserting exactly the invariant I fuzzed:
transform(m, :clean_to_bigquery_column_spec) ==
  m |> MetadataCleaner.deep_reject_nil_and_empty() |> transform(:to_bigquery_column_spec)
That locks in the refactor and keeps the existing suite meaningful.

lib/logflare/logs/ingest_transformers.ex
```diff
@@ -5,7 +5,24 @@ defmodule Logflare.Logs.IngestTransformers do
...
-  @spec transform(map, list(atom) | atom) :: map
+  @typep direct_transform :: :clean_to_bigquery_column_spec | :to_bigquery_column_spec
+  @typep transform_rule ::
+           :alphanumeric_only
+           | :alter_leading_numbers
+           | :dashes_to_underscores
+           | :strip_bq_prefixes
+           | {:field_length, max: pos_integer()}
+
+  defguardp nil_or_empty?(x) when x in [%{}, [], "", {}, nil]
```
amokan:
```suggestion
  defguardp is_nil_or_empty(x) when x in [%{}, [], "", {}, nil]
```

Nit. Guards should be prefixed with `is_` and the question mark suffix is generally for predicate functions.

lib/logflare/logs/ingest_transformers.ex
```diff
@@ -5,7 +5,24 @@ defmodule Logflare.Logs.IngestTransformers do
...
+
+  defguardp bq_safe_byte?(byte)
```
amokan:
```suggestion
  defguardp is_bq_safe_byte(byte)
```

Nit. Same as above.

amokan (changes requested):
Couple minor things we should probably address on the test front along with 2 tiny nits I can live with or without.

lib/logflare/logs/log_event.ex
djwhitt: Restored the `%LE{valid: false}` passthrough in `3c1ee8ad`, keeping the valid-event clause first for the hot path. Transform failures will remain isolated to the event instead of reaching a missing `validate/2` clause.

lib/logflare/logs/ingest_transformers.ex
djwhitt: Renamed to `is_bq_safe_byte` and updated its call sites in `3c1ee8ad`.

lib/logflare/logs/ingest_transformers.ex
djwhitt: Renamed to `is_nil_or_empty` and updated its call sites in `3c1ee8ad`.

lib/logflare/logs/ingest_transformers.ex
djwhitt: Added the equivalence property comparing the fused transform with `MetadataCleaner.deep_reject_nil_and_empty/1` followed by `:to_bigquery_column_spec`.

This uncovered an edge case where normalized-key collisions could select a different value after cleaning changed the map representation. The fused path now detects collisions and falls back to the sequential implementation. I also added a dedicated regression test. The focused suite passes with 101 tests and 7 properties, and benchmarking showed no meaningful regression.

amokan (approved):

PR #3750, opened 2026-07-29
fix: Add egress telemetry emission for syslog backend batches

Ziinc:
This PR adds telemetry to the syslog backend adaptor that will get emitted into system.metrics.

https://claude.ai/code/session_01CYPC36k6DNvHpXDq3So6aL


lib/logflare/backends/adaptor/syslog_adaptor/pipeline.ex
```diff
@@ -28,7 +28,10 @@ defmodule Logflare.Backends.Adaptor.SyslogAdaptor.Pipeline do
       ],
       context: %{
         source_id: source.id,
+        source_token: source.token,
         backend_id: backend.id,
+        backend_token: backend.token,
+        user_id: source.user_id,
```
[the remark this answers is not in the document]
Ziinc: Good catch — fixed in 30c5793aa by propagating `system_source` into the Broadway context and the egress telemetry metadata, matching the BigQuery pipeline's pattern. Verified `UserMonitoring.keep_metric_function/1` now filters these out correctly.

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/logflare/backends/adaptor/syslog_adaptor/pipeline.ex
```diff
@@ -42,13 +46,33 @@ defmodule Logflare.Backends.Adaptor.SyslogAdaptor.Pipeline do
...
       end
 
+    request_bytes = Enum.reduce(content, 0, fn frame, acc -> acc + IO.iodata_length(frame) end)
+
+    backend_meta =
+      for {k, v} <- backend_metadata || %{}, into: %{} do
+        {"backend.#{k}", v}
+      end
+
+    :telemetry.execute(
```
amokan: 🤖 

> `:telemetry.execute/3` fires before `Pool.send/2`, so bytes get counted for batches that never leave the box. If the endpoint is down or TLS handshake fails, `Pool.send` returns `{:error, _}`, `fail_batch/2` drops every message, and `logflare.backends.ingest.egress.request_bytes` still sums at full rate — a dashboard showing healthy egress during a total outage. Same over-count when `NimblePool.checkout!` times out. **Fix:** _move the `execute` into the `:ok ->` branch, which also matches the metric's own description. Worth noting in the PR body if this pre-emit is deliberately mirroring `HttpBased.EgressTracer:49`, since otherwise the two adaptors' identical metric names will quietly mean different things._

test/logflare/backends/adaptor/syslog_adaptor_test.exs
```diff
@@ -206,6 +206,55 @@ defmodule Logflare.Backends.Adaptor.SyslogAdaptorTest do
     assert String.starts_with?(telegraf_message, ~s[{"event_message":])
   end
 
+  test "emits egress telemetry for every batch sent" do
+    start_supervised!(Logflare.SystemMetrics.AllLogsLogged)
```
amokan: 🤖 

> copies the whole body of `start_syslog/1` (_lines 355–368_) verbatim just to add `metadata:`. That's a ~12-line near-exact clone and `mix test.slop` fails CI on a new clone at `--max-clones 29`. Adding an optional `backend_attrs` keyword to `start_syslog/2` keeps one setup path; run `mix test.slop` on the branch either way.

amokan (changes requested):
Could also use some doc updates or at least notes to avoid 🤖 silliness down the road.

> `lib/logflare/backends/user_monitoring.ex:33` and `docs/.../system-monitoring.md:48` — both still say the metric is HTTP-only. That description ships to users as the OTel metric description, so a syslog-only user would now conclude the non-zero number isn't theirs.

lib/logflare/backends/adaptor/syslog_adaptor/pipeline.ex
Ziinc: Good catch — fixed in 740bce56c by moving `:telemetry.execute/3` into `Pool.send/2`'s `:ok` branch, so failed/dropped batches no longer inflate `request_bytes`. This isn't mirroring `HttpBased.EgressTracer` deliberately — that was a straightforward bug, now fixed to match the metric's actual meaning.

---
_Generated by [Claude Code](https://claude.ai/code)_

test/logflare/backends/adaptor/syslog_adaptor_test.exs
Ziinc: Fixed in 740bce56c — added an optional `backend_attrs` keyword to `start_syslog/2` and reused it in the egress-telemetry test instead of duplicating the setup. Ran `mix test.slop` locally: 28/29 clone budget, no clone reported for this file anymore.

---
_Generated by [Claude Code](https://claude.ai/code)_

Ziinc:
@amokan Addressed both review findings in 740bce56c:
- Moved the egress `:telemetry.execute/3` call into `Pool.send/2`'s `:ok` branch so failed/dropped batches no longer over-count `request_bytes`.
- Refactored the egress-telemetry test to use an optional `backend_attrs` param on `start_syslog/2` instead of duplicating its setup — confirmed with `mix test.slop` (28/29 clone budget).
- Also updated the `request_bytes` metric description in `user_monitoring.ex` and `system-monitoring.md` to drop the stale "HTTP-only" wording.

CI is green on the current head. Would appreciate a re-review when you have a chance.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3755, opened 2026-07-31
perf: fix n+1 query on dashboard

Ziinc:
Fix N+1 query on dashboard that is causing issues where db latency is high

Part of O11Y-2231

### DIFF
--- a/lib/logflare/sources.ex
+++ b/lib/logflare/sources.ex
@@ -504,7 +504,9 @@ defmodule Logflare.Sources do
   @spec preload_for_dashboard(list(Source.t())) :: list(Source.t())
   def preload_for_dashboard(sources) do
     sources
-    |> Enum.map(&preload_defaults/1)
+    |> Repo.preload([:backends, [user: :team]])
+    |> Enum.map(&refresh_source_metrics/1)
+    |> Enum.map(&put_bq_table_id/1)
     |> Enum.map(&put_schema_field_count/1)
     |> Enum.sort_by(&{!&1.favorite, &1.name})
   end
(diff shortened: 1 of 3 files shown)


djwhitt (approved):
:robot:

> `preload_for_dashboard/1` still issues one `source_schemas` query per source when the schema cache is cold or expired. `put_schema_field_count/1` calls `SourceSchemas.Cache.get_source_schema_by/1` for every source, while this test only counts `users`, `teams`, and `backends`.
>
> When I included `source_schemas` in the telemetry filter, the test observed 4 queries for one source versus 23 for 20 sources. One possible fix would be to bulk-preload `:source_schema` here and have `get_schema_field_count/1` use the preloaded association, retaining the cache lookup as a fallback for other callers. Could we also include `source_schemas` in the regression assertion so the total dashboard query count remains constant?

djwhitt:
:robot:

> Non-blocking follow-up: while checking adjacent source-loading paths, I found two pre-existing N+1 patterns:
>
> - Multi-user `Sources.list_sources/1` calls `put_retention_days/1` for every source, resulting in per-user `users`, `billing_accounts`, and `plans` queries. A cold-cache probe increased from 4 total queries for one user to 61 for 20 users. The command palette uses this path but does not consume retention days, so it could skip that hydration.
> - Bulk `Users.preload_defaults/1` preloads users and billing accounts, but then `put_retention_days/1` re-fetches them through the caches for every user. A probe increased from 6 queries for one user to 63 for 20 users.
>
> Neither issue was introduced by this PR, so no changes are required here, but they may be worth tracking as follow-ups.

PR #3757, opened 2026-07-31
perf(bigquery): reduce acknowledgement allocations

djwhitt:
## Summary

- avoid concatenating successful and failed Broadway messages during acknowledgement
- replace `Enum.group_by/2` and per-reference message lists with tail-recursive contiguous-reference counting
- preserve correct in-flight accounting for interleaved producers and messages without an in-flight reference
- benchmark steady-state, producer-restart handoff, contiguous, and adversarial alternating reference layouts over the same 500 messages

Each contiguous run performs one atomic subtraction. A producer reference may appear in multiple runs or in both acknowledgement lists; those runs are subtracted independently and produce the same total count.

## Performance

Median of three local runs in the same six-scheduler Linux environment with `BENCH_TIME=2`. Every input contains 500 messages split into 450 successful and 50 failed messages.

| Producer-reference layout | Median latency | Memory | BEAM reductions |
| --- | ---: | ---: | ---: |
| Single producer (steady state) | 8.79 → 4.00 µs (**55% lower**) | 38.47 → 0.00 KB | 6.28K → 2.02K (**68% lower**) |
| Producer restart handoff | 10.04 → 4.04 µs (**60% lower**) | 41.68 → 0.00 KB | 6.29K → 2.03K (**68% lower**) |
| Four contiguous references | 9.79 → 4.04 µs (**59% lower**) | 44.61 → 0.00 KB | 6.31K → 2.03K (**68% lower**) |
| Four alternating references (adversarial) | 14.00 → 7.17 µs (**49% lower**) | 50.26 → 0.00 KB | 5.45K → 4.01K (**26% lower**) |

The single-reference input represents normal operation. The restart-handoff input uses two producer lifetimes in processor-demand-sized runs (`200 old, 100 new, 100 old, 100 new`). Fully alternating references remain an adversarial comparison rather than an expected steady-state layout.

### DIFF
--- a/lib/logflare/sources/source/bigquery/pipeline.ex
+++ b/lib/logflare/sources/source/bigquery/pipeline.ex
@@ -155,7 +155,11 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
   @spec decrement_in_flight([Message.t()]) :: :ok
   defp decrement_in_flight(messages), do: decrement_in_flight(messages, nil, 0)
 
-  defp decrement_in_flight([], nil, 0), do: :ok
+  @spec decrement_in_flight(
+          [Message.t()],
+          :atomics.atomics_ref() | nil,
+          non_neg_integer()
+        ) :: :ok
 
   # Flush the final run after consuming all messages.
   defp decrement_in_flight([], ref, run_count),
--- a/lib/logflare/sources/source/bigquery/pipeline.ex
+++ b/lib/logflare/sources/source/bigquery/pipeline.ex
@@ -134,7 +134,8 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
   def ack({queue, config}, successful, failed) do
     {sid, bid, _pipeline_ref} = queue
 
-    decrement_in_flight(successful ++ failed)
+    decrement_in_flight(successful)
+    decrement_in_flight(failed)
     finalize_acked_events({sid, bid}, successful)
     maybe_requeue_failed({sid, bid}, failed, config)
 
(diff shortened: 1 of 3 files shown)


test/logflare/bigquery/pipeline_test.exs
```diff
@@ -199,6 +199,44 @@ defmodule Logflare.BigQuery.PipelineTest do
       assert IngestEventQueue.list_recent_events({source.id, nil}, 10) == []
     end
 
+    test "ack decrements each producer's in-flight count without requiring grouped messages", %{
```
amokan, line 202: I think we'll want to split this test up so we exercise the `decrement_in_flight/1` stuff added to the pipeline.

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -147,18 +148,37 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
-    end)
+  defp decrement_in_flight(messages), do: decrement_in_flight(messages, nil, 0)
+
+  defp decrement_in_flight([], nil, 0), do: :ok
+
+  # Flush the final run after consuming all messages.
+  defp decrement_in_flight([], ref, run_count),
+    do: decrement_in_flight_ref(ref, run_count)
+
+  defp decrement_in_flight(
```
amokan, line 168: Nit. Would be nice to get a `@spec` in here for `decrement_in_flight/2`

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -147,18 +148,37 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
-    |> Enum.group_by(fn %{acknowledger: {_, _, ack_data}} ->
-      Map.get(ack_data, :in_flight_ref)
-    end)
-    |> Enum.each(fn
-      {nil, _msgs} -> :ok
-      {ref, msgs} -> :atomics.sub(ref, 1, length(msgs))
-    end)
+  defp decrement_in_flight(messages), do: decrement_in_flight(messages, nil, 0)
+
+  defp decrement_in_flight([], nil, 0), do: :ok
```
amokan: Could drop this clause if you wanted. Ultimately the clause below (L161) would return an `:ok` as well due to `decrement_in_flight_ref/2` having a pattern matched clause on that ref arg being `nil` 🤷‍♂️ 

```suggestion
```

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -147,18 +148,37 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
...
+    do: decrement_in_flight_ref(ref, run_count)
+
+  defp decrement_in_flight(
+         [%{acknowledger: {_, _, ack_data}} | messages],
+         current_ref,
+         run_count
+       ) do
+    ref = Map.get(ack_data, :in_flight_ref)
+
+    if ref == current_ref do
```
amokan, line 175: This could be moved to its own pattern-matched head.

Maybe something like

```elixir
defp decrement_in_flight([%{acknowledger: {_, _, %{in_flight_ref: in_flight_ref}} | messages], current_ref, run_count) when in_flight_ref == current_ref do
  decrement_in_flight(messages, current_ref, run_count + 1)
end
```

amokan (approved):
Approved. Some non-blocking notes added.

lib/logflare/sources/source/bigquery/pipeline.ex
djwhitt: 🤖 made an argument that the `if` here is clearer and I kinda buy it since you need something like this to replace the `if` (same as above, but showing both clauses):

```elixir
 defp decrement_in_flight(
        [%{acknowledger: {_, _, %{in_flight_ref: in_flight_ref}}} | messages],
        current_ref,
        run_count
      )
      when in_flight_ref == current_ref do
   decrement_in_flight(messages, current_ref, run_count + 1)
 end

 defp decrement_in_flight(
        [%{acknowledger: {_, _, ack_data}} | messages],
        current_ref,
        run_count
      ) do
   ref = Map.get(ack_data, :in_flight_ref)

   decrement_in_flight_ref(current_ref, run_count)
   decrement_in_flight(messages, ref, 1)
 end
```

lib/logflare/sources/source/bigquery/pipeline.ex
djwhitt: 🤖 was advocating for this as documentation. Not sure I care one way or another so I'm removing it.

PR #3758, opened 2026-07-31
fix(otlp): decode trace/span IDs and dedupe content-type header

Baishan:
This fixes two independent bugs in the OTLP backend adaptor's ProtobufFormatter, both discovered while debugging a customer's Dash0 log drain that was silently receiving zero traffic. First, trace_id/span_id values from the log event body are hex-encoded strings (the standard OTel textual representation), but the formatter was passing them straight through as the protobuf bytes field value instead of decoding them — sending a 32-byte ASCII string where a 16-byte TraceID is expected, which strict OTLP receivers (Dash0's included) reject outright with an unmarshal error. Second, the formatter unconditionally force-sets a lowercase content-type header via Tesla.put_header/3, which does a case-sensitive key match — so a backend with a user-configured Content-Type header (differing only in casing) ends up sending two separate content-type header lines, which some receivers reject as an unsupported/ambiguous media type. Both were confirmed against the real destination by capturing live Tesla egress traffic and replaying the actual failing request with each fix applied in isolation. The trace/span ID fix falls back to sending the raw value unchanged if it isn't valid hex, matching prior behavior rather than dropping the field, and the header fix strips any pre-existing content-type header case-insensitively before setting the correct one. Tests were added covering correct decoding, the invalid-hex fallback, and de-duplication of the content-type header.

### DIFF
--- a/lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
+++ b/lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
@@ -23,7 +23,7 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptor.ProtobufFormatter do
 
     response =
       env
-      |> Tesla.put_header("content-type", @protobuf_content_type)
+      |> put_content_type_header()
       |> Tesla.put_body(transform_batch(env.body, source))
       |> Tesla.run(next)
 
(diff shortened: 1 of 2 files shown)


djwhitt (approved):
:robot:

> Reviewed locally with no blocking findings. The focused OTLP adaptor tests pass, and all published CI checks are green.

PR #3759, opened 2026-08-01
perf(clickhouse): fuse mapper and RowBinary encoding

djwhitt:
## Summary

This PR fuses per-event mapping and ClickHouse RowBinary encoding in one normal NIF call while preserving the generic mapper contract.

```text
Before: ETS lookup → Mapper.map/3 → intermediate BEAM map → Elixir RowBinary iodata → gzip
After:  ETS lookup → native map + RowBinary binary → gzip
```

There is no cross-event native batching. Each event is still looked up, mapped, encoded, and passed to the existing incremental gzip stream independently.

## Design

- `MappingConfig.OutputFormat` is serialized with the mapping config and selects map or ClickHouse RowBinary output when compiled.
- `Mapper.map/3` remains the sole execution API. Map configurations return the existing map representation; RowBinary configurations return one encoded row binary.
- A typed `Mapper.OutputContext` carries per-event ClickHouse envelope metadata separately from the mapped document. Map output ignores that context, preserving full `%LogEvent{}` mapping semantics.
- Both outputs execute the same field-resolution core, covering existing defaults, transforms, coalescing, path caching, enum inference, filtering, JSON operations, and coercion.
- RowBinary layouts compile field indices once and reject missing or wire-incompatible required fields before event processing.
- Schema-specialized log, metric, and trace writers consume resolved values directly, including the existing log severity and trace-duration adjustments.
- `BinaryBuilder` writes directly into `rustler::OwnedBinary`, avoiding an intermediate Rust byte buffer and final copy into a BEAM binary.
- Map and list values stream through Rustler iterators rather than temporary decoded collections.
- The fused NIF remains normally scheduled for staging observation; it is not marked dirty.

## Performance

Linux aarch64, OTP 27, six online schedulers, production-shaped 1,000-row inputs. Every run first verifies byte-identical fused and separate RowBinary output.

### ETS lookup + mapping + RowBinary + gzip

| Event type | Separate | Fused | Latency change | BEAM memory |
|---|---:|---:|---:|---:|
| Log | 12.85 ms | 9.86 ms | -23% | 13.95 MB → 6.22 MB (-55%) |
| Metric | 23.60 ms | 18.93 ms | -20% | 25.76 MB → 10.85 MB (-58%) |
| Trace | 18.76 ms | 13.91 ms | -26% | 18.22 MB → 8.31 MB (-54%) |

### Mapping + RowBinary only

| Event type | Separate | Fused | Latency change | BEAM memory | Reduction count |
| Log | 11.65 ms | 5.55 ms | -52% | 10.33 MB → 2.60 MB (-75%) | 714.9K → 69.0K (-90%) |
| Metric | 27.94 ms | 13.51 ms | -52% | 19.76 MB → 4.88 MB (-75%) | 1.38M → 73.1K (-95%) |
| Trace | 16.64 ms | 8.93 ms | -46% | 13.53 MB → 3.68 MB (-73%) | 976.0K → 69.4K (-93%) |

Commands:

```sh
BENCH_SECTION=pipeline EVENT_TYPES=log,metric,trace BATCH_SIZES=1000 \
  BENCH_TIME=10 BENCH_WARMUP=3 BENCH_MEMORY_TIME=3 BENCH_REDUCTION_TIME=3 \
  MIX_ENV=test mix run --no-start bench/clickhouse_mapper_rowbinary_fusion.exs

BENCH_SECTION=encoding EVENT_TYPES=log,metric,trace BATCH_SIZES=1000 \
  BENCH_TIME=10 BENCH_WARMUP=3 BENCH_MEMORY_TIME=3 BENCH_REDUCTION_TIME=3 \
  MIX_ENV=test mix run --no-start bench/clickhouse_mapper_rowbinary_fusion.exs
```

## Validation

- mapper/config/output: 275 tests and 2 properties, 0 failures
- ClickHouse pipeline: 36 tests, 0 failures
- Rust: 15 tests passed
- Clippy with warnings denied and formatter checks passed
- 1,000-row log, metric, and trace output remained byte-identical


amokan (changes requested):
I really like this idea, but I think a different angle we should consider is to introduce an optional Keyword list arg in `MappingConfig.new/1` so we could do something like `MappingConfig.new(config, output: :clickhouse_row_binary)` at the configuration level.

This would allow us to continue invoking the mapper using `Mapper.map/3` instead of a dedicated `map_and_encode_clickhouse/3` function and open the door for other output format options down the road (_Arrow IPC, Parquet, etc_).

Another potential approach would be to refactor the structure of `Logflare.Mapper.MappingConfig` from

```elixir
typed_embedded_schema do
  embeds_many(:fields, Logflare.Mapper.MappingConfig.FieldConfig)
end
```

to

```elixir
typed_embedded_schema do
  embeds_many(:fields, Logflare.Mapper.MappingConfig.FieldConfig)
  embeds_one(:output, Logflare.Mapper.MappingConfig.OutputFormat)
end
```

_`OutputFormat` does not currently exist_

The advantage to this idea is that the output format would be part of the serialized configuration rather than a runtime decision.

native/mapper_ex/src/mapping.rs
```diff
@@ -11,13 +11,113 @@ use crate::string_filters::{CharClass, StringFilters};
 
 const ROOT_CACHE_MIN_REFERENCES: usize = 8;
 
+const CLICKHOUSE_LOG_FIELDS: &[&str] = &[
```
amokan: Ideally we can refactor a bit to move these ClickHouse attribute lists to a dedicated clickhouse module.

djwhitt: :robot:

> Accepted. The ClickHouse field definitions and layout compilation will be moved into the ClickHouse RowBinary module.

djwhitt:
:robot:

> Proposed revision:
>
> - Add a serialized `MappingConfig.OutputFormat` containing the output format and ClickHouse row type.
> - Default configurations without an output definition to map output.
> - Move ClickHouse field definitions and layout compilation into the ClickHouse RowBinary module.
> - Compile only the configured ClickHouse layout.
> - Dispatch output through `Mapper.map/3`.
> - Remove `map_and_encode_clickhouse` and runtime event-type selection.
>
> Clarification required: RowBinary encoding also requires event ID, source UUID, source name, `ingested_at`, and `mapping_config_id`. Should `Mapper.map/3` receive these values through an output-context option, or should they be included in the input document structure?

native/mapper_ex/src/clickhouse_rowbinary.rs
```diff
@@ -0,0 +1,824 @@
...
+impl<'values, 'env> RowValues<'values, 'env> {
+    fn new(values: &'values [Term<'env>], layout: &'values [usize]) -> Self {
+        Self {
+            values,
+            layout,
+            cursor: 0,
+        }
+    }
+
+    fn next(&mut self, name: &str) -> EncodeResult<Term<'env>> {
```
amokan, line 332: Non-blocking 🤖

> `RowValues::next(&mut self, name: &str)` ignores `name`; it returns `values[layout[cursor]]` and bumps the cursor, so column selection depends entirely on `append_log/append_metric/append_trace` calling `next` in the same order as the `*_FIELDS` const (_which `compile_layout` resolves by name_). Add a column between `sum` and `min` in `METRIC_FIELDS` without a matching `next` call and every later value shifts one column — adjacent `String/String` and `Float64/Float64` runs mean ClickHouse accepts the row and stores silently-wrong data. Nothing asserts `cursor == layout.len()` at the end either. Options: (a) `debug_assert!` + `Err` on cursor exhaustion in `append_row` — cheapest, catches missing calls in tests; (b) index by name via compile-time index constants — removes the ordering coupling entirely; (c) generate `append_*` from the field const with a macro. **Recommend (a) now, (b) if a fourth row type lands**.

lib/logflare/mapper.ex
```diff
@@ -55,10 +57,30 @@ defmodule Logflare.Mapper do
...
+    output_context = Keyword.get(opts, :output_context)
+
+    document
+    |> Native.map(compiled_mapping, {flat_keys, output_context})
+    |> unwrap_result()
+  end
+
+  defp unwrap_result({:ok, output}), do: output
+
+  defp unwrap_result({:error, reason}) do
```
amokan, line 92: Non-blocking (_but prob worth considering_) 🤖

> `unwrap_result({:error, reason})` raises `ArgumentError`, so a NIF error escapes `Pipeline.encode_message/6` (_`pipeline.ex:312`_) → `Enum.reduce` → `handle_batch/4` and fails the whole Broadway batch. One event missing `"timestamp"` or with a non-UUID `event.id` discards up to `batch_size` good events. Same as pre-PR behaviour, but this is the first version where the NIF returns a structured `{:error, reason}` the pipeline could turn into `Message.failed(message, reason)`. Options: (a) have `encode_message/6` call the non-bang path and mark just that message failed; (b) leave as-is and file a follow-up. **Recommend (a) — the plumbing already exists**.

native/mapper_ex/src/mapper.rs
```diff
@@ -43,12 +65,35 @@ pub fn map_single<'a>(
...
+    // Build the output map in a single allocation via enif_make_map_from_arrays.
+    // Duplicate field names are rejected at compile time so this should not fail.
+    Term::map_from_term_arrays(env, &keys, scratch.values()).unwrap_or_else(|_| Term::map_new(env))
+}
+
+/// Execute the shared mapping core into field-order storage.
+///
+/// ClickHouse's fused encoder consumes these values directly, avoiding an
+/// intermediate output map.
+pub fn map_values_into<'a>(
```
amokan, line 86: Non-blocking 🤖 

> `map_values_into` only appends to `scratch.values`, never clears it, while `MapScratch` is named/shaped for reuse. Correct today (_fresh scratch per document_), but the PR body calls out cross-event batching as the next step; hoisting one scratch would leave doc _N_'s values at `0..len`, failing `map_from_term_arrays`'s length check and silently returning an empty map. Fix: `values.clear()` at the top of `map_values_into`.

test/support/clickhouse_mapped_events.ex
```diff
@@ -127,16 +122,18 @@ defmodule Logflare.ClickHouseMappedEvents do
...
-      input_body
-      |> Mapper.map(compiled)
-      |> Map.put("mapping_config_id", config_id)
+    {mapped_body, config_id} = map_body(input_body, :trace)
+    mapped_body = Map.put(mapped_body, "mapping_config_id", config_id)
 
     %{event | body: mapped_body}
   end
 
+  defp map_body(input_body, event_type) do
```
amokan, line 135: Non-blocking 🤖

> `map_body/2` now calls `Mapper.compile!/1` on every invocation instead of the cached `MappingConfigStore.get_compiled/1`. It's hit ~45× across `ingester_test.exs`, `clickhouse_adaptor_test.exs`, `provisioner_test.exs`, `endpoints_test.exs`, `sql_test.exs`, `endpoints_live_test.exs` — each paying a full NIF compile of a 17–44-field config plus a `ResourceArc` that lives until GC. Decoupling from the GenServer is right; memoize per `event_type` (`:persistent_term`) instead. Also missing a `@spec`.

native/mapper_ex/src/mapping.rs
```diff
@@ -167,9 +175,38 @@ pub fn decode_mapping<'a>(env: Env<'a>, config: Term<'a>) -> Result<CompiledMapp
...
+    match format.as_str() {
+        "clickhouse_row_binary" => {
+            let row_type = get_string_key(env, output, "row_type")?
+                .ok_or_else(|| "ClickHouse RowBinary output row_type is required".to_string())?;
+            let fields_by_name = fields
+                .iter()
+                .enumerate()
+                .map(|(index, field)| (field.name.as_str(), (index, field.field_type)))
+                .collect();
+            let layout =
```
amokan: 🤖 Nit.

> `rustfmt --edition 2021 --check` reports a diff (_the `let layout = ...` split fits in 98 cols and rustfmt joins it_). CI only runs `cargo test -p mapper_ex`, so it won't fail the build

native/mapper_ex/src/lib.rs
```diff
@@ -1,18 +1,21 @@
...
+use rustler::{Binary, Encoder, Env, NewBinary, NifResult, Resource, ResourceArc, Term};
 
-use mapping::CompiledMapping;
+use mapping::{CompiledMapping, CompiledOutput};
 
 mod atoms {
     rustler::atoms! {
         ok,
         error,
+        nil,
```
amokan, line 17: Not a huge deal in my view - but worth noting we also bring in the same `nil` mod atom in `mapper.rs` (_line 17_)

My take is the compiler will do its magic here, but worth noting for longer term refactoring.

test/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults_test.exs
```diff
@@ -445,6 +451,11 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaultsTest do
     end
   end
 
+  defp compile_map_output(event_type) do
```
amokan: Nit. Lets move this function below the `describe` blocks.

Cargo.toml
```diff
@@ -2,3 +2,6 @@
 resolver = "2"
 
 members = ["native/arrowipc_ex", "native/sqlparser_ex", "native/mapper_ex"]
+
+[profile.release.package.mapper_ex]
+codegen-units = 1
```
amokan:
```suggestion
```

My thinking is that we should leave this out. We already set `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1` in both `Dockerfile` and `Dockerfile.base`, so this is a no-op for anything we actually ship. The only places it does take effect are local dev and the `elixir-ci` jobs, since Rustler defaults to `mode: :release` and those run without the Docker ENV. I believe all this ends up buying us is serialized codegen on a profile where we _probably_ don't care about optimization as much.

If we want the release tuning tracked in the manifest instead of Docker ENV, it may be worth doing that as its own change across all three crates 🤷‍♂️

amokan (approved):
Approved with some notes

native/mapper_ex/src/clickhouse_rowbinary.rs
djwhitt: :robot:

> Addressed in 9d849c63

test/support/clickhouse_mapped_events.ex
djwhitt: :robot:

> Addressed in ef15d488

native/mapper_ex/src/mapping.rs
djwhitt: :robot:

> Addressed in 9d849c63

test/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults_test.exs
djwhitt: :robot:

> Addressed in ef15d488

Cargo.toml
djwhitt: :robot:

> Addressed in 63971fca

lib/logflare/mapper.ex
djwhitt: :robot:

> Addressed in 6eca94e0

native/mapper_ex/src/mapper.rs
djwhitt: :robot:

> Deferred. `MapScratch` is currently allocated per document, so no state is reused across events. Clearing `values` alone would be incomplete because `QueryCache` also retains per-document state. Add a complete reset of both structures if scratch reuse or cross-event batching is introduced.

PR #3762, opened 2026-08-03
fix: improve float parsing and default handling

Ziinc:
improve integer handling for float runtime configs

discovered when an int was passed to a float config via helm.


amokan (changes requested):
This change appears to be a no-op given that `System.get_env/2` return a String or nil. So then when you call your new `Utils.parse_float!/1` - we're back to calling `String.to_float/1` again.

What about something like

```elixir
def parse_float!(v) when is_non_empty_binary(v) do
  case Float.parse(v) do
    {float, ""} -> float
    _ -> raise ArgumentError, "Could not parse to float: #{inspect(v)}"
  end
end
```

config/runtime.exs
```diff
@@ -462,7 +462,7 @@ if System.get_env("LOGFLARE_OTEL_ENDPOINT") do
             %{
               probability:
                 System.get_env("LOGFLARE_OTEL_SAMPLE_RATIO", "1.0")
```
amokan, line 464: Any reason we need to parse this again? We have it up on line 426 in the `default_sample_ratio` var.

Ziinc: adjusted

Ziinc:
Fixed and updated PR description, the main issue is when it receives `"1"` string and and we try to convert it to a float instead of an integer first.

lib/logflare/utils.ex
```diff
@@ -155,6 +155,36 @@ defmodule Logflare.Utils do
...
+    if String.contains?(v, ".") do
+      String.to_float(v)
+    else
+      String.to_integer(v) / 1
+    end
+  end
+
+  def parse_float!(v) when is_integer(v), do: v / 1
+  def parse_float!(v) when is_float(v), do: v
+  def parse_float!(v), do: raise("Could not parse to float: #{inspect(v)}")
```
amokan, line 186: The version in the PR has some edge case failures that could be covered.

```
"1.5"     -> 1.5
"1"       -> 1.0
"1e-4"    -> RAISES ArgumentError
"0.95\n"  -> RAISES ArgumentError
" 0.95"   -> RAISES ArgumentError
".95"     -> RAISES ArgumentError
```

```suggestion
  def parse_float!(v) when is_binary(v) do
    v
    |> String.trim()
    |> pad_leading_dot()
    |> Float.parse()
    |> case do
      {float, ""} -> float
      _ -> raise ArgumentError, "Could not parse to float: #{inspect(v)}"
    end
  end

  def parse_float!(v) when is_integer(v), do: v / 1
  def parse_float!(v) when is_float(v), do: v
  def parse_float!(v), do: raise("Could not parse to float: #{inspect(v)}")
  
  defp pad_leading_dot("." <> _rest = v), do: "0" <> v
  defp pad_leading_dot("-." <> rest), do: "-0." <> rest
  defp pad_leading_dot(v), do: v
```

The outputs for the suggested change are

```
"1.5"     -> 1.5
"1"       -> 1.0
"1e-4"    -> 0.0001
"0.95\n"  -> 0.95
" 0.95"   -> 0.95
".95"     -> 0.95
```

config/runtime.exs
```diff
@@ -458,20 +458,20 @@ config :libcluster,
 if System.get_env("LOGFLARE_OTEL_ENDPOINT") do
   default_sample_ratio =
     System.get_env("LOGFLARE_OTEL_SAMPLE_RATIO", "1.0")
-    |> String.to_float()
+    |> Utils.parse_float!()
```
amokan, line 461: 🤖 

> `config/runtime.exs:461`, `467`, `474` — **out-of-range OTEL ratios moved from a boot crash to a per-span crash**. `LOGFLARE_OTEL_INGEST_SAMPLE_RATIO=2` now parses to `2.0`; the Erlang sampler has no clause above `1.0`, and `open_telemetry_sampler.ex:103` calls `setup/1` inside `should_sample/7`. Add a `parse_ratio!/1` that bounds to `0.0..1.0` and raises, then use it at these three sites (_and `:67`, same class_)

amokan (dismissed):

chasers (approved):

PR #3764, opened 2026-08-03
fix: scope event age filtering back to ClickHouse backend

amokan:
## Why?

Event age filtering currently runs in `Backends.split_valid_events/2`, which drops old events before _any_ backend sees them. That bound generally exists for ClickHouse as late-arriving events land in old partitions and increase parts written per insert, but applying it globally has implications for other backends. The filtering did originally live in the ClickHouse pipeline up until #3030. The big difference between then and now is that we also support the ability to adjust that threshold (_or disable_) per ClickHouse backend.

## Key Changes

- Removed the stale-age bound from `split_valid_events/2`. The `+1h` future guard stays global, since a future timestamp is bad data for every backend.
- ClickHouse now implements the existing optional `Adaptor.pre_ingest/3` callback, which fires at each dispatch site immediately before `IngestEventQueue.add_to_table`. Stale events are never queued into the consolidated pipeline with this approach.
- New `max_event_age_hours` config for ClickHouse backends (_defaults to `24`, `0` disables the check_)
- `drop_stale` keeps its telemetry event name for dashboard continuity, and gains `backend_id`/`backend_type` metadata plus matching metric tags. A `keep:` guard discards emissions missing either key, and `source_id`/`source_token` are deliberately left untagged to avoid per-source cardinality.
- Events in the newly-accepted age range now count toward `Sources.Counters`, so recorded usage for sources sending old events will rise. This is the accurate number.

## Callout For Grafana Dashboards

- Any dashboards querying the bare `logflare.logs.ingest_logs.drop_stale` series will need `sum()` across the new labels as adding tags changes series identity.

### DIFF
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -33,7 +33,6 @@ defmodule Logflare.Backends do
 
   defdelegate child_spec(arg), to: __MODULE__.Supervisor
 
-  @max_event_age_us 24 * 3_600 * 1_000_000
   @max_future_event_us 1 * 3_600 * 1_000_000
   @max_pending_buffer_len_per_queue IngestEventQueue.max_queue_size()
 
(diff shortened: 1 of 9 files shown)


Baishan (approved):
Makes sense.

PR #3765, opened 2026-08-03
perf(logs): reuse clean BigQuery-safe payloads

djwhitt:
## Summary

- Scan nested maps and lists for empty values and BigQuery-unsafe keys before rebuilding them.
- Return the original payload when cleaning and key normalization would make no changes.
- Retain the fused cleaning traversal and collision fallback for payloads that require normalization.
- Extend coverage for clean nested payloads and nested slow-path cases.

## Stack

1. **#3765 — Reuse clean BigQuery-safe payloads**
2. #3767 — Transform bodies before constructing LogEvent
3. #3768 — Normalize integer timestamps without digit lists
4. #3769 — Fast-path common UTC ISO 8601 timestamps

### DIFF
--- a/lib/logflare/logs/ingest_transformers.ex
+++ b/lib/logflare/logs/ingest_transformers.ex
@@ -22,7 +22,11 @@ defmodule Logflare.Logs.IngestTransformers do
 
   @spec transform(map(), direct_transform() | [transform_rule()]) :: map()
   def transform(log_params, :clean_to_bigquery_column_spec) when is_map(log_params) do
-    clean_and_to_bigquery_column_spec(log_params)
+    if clean_and_bq_safe?(log_params) do
+      log_params
+    else
+      clean_and_to_bigquery_column_spec(log_params)
+    end
   catch
     :throw, :normalized_bigquery_column_collision ->
       log_params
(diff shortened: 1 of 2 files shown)


Ziinc (approved):

lib/logflare/logs/ingest_transformers.ex
```diff
@@ -22,7 +22,11 @@ defmodule Logflare.Logs.IngestTransformers do
 
   @spec transform(map(), direct_transform() | [transform_rule()]) :: map()
   def transform(log_params, :clean_to_bigquery_column_spec) when is_map(log_params) do
-    clean_and_to_bigquery_column_spec(log_params)
+    if clean_and_bq_safe?(log_params) do
```
Ziinc, line 25: This adds another recursive traversal to a hot ingest path, needs a bench run to verify that this doesn't affect performance.

Ziinc: Not sure if there's much benefit here in doing the if/do check since it doubles the traversal time

Ziinc (changes requested):
For some reason it registered as an approval

Ziinc:
Also, feels like we should be moving the BQ specific stuff to adaptor level and later in the ingest pipeline , since we're becoming more backend agnostic

PR #3773, opened 2026-08-04
ci: scope checks and caches by workload

djwhitt:
## Summary

- scope Mix and Cargo caches by workload and relevant source hashes
- detect backend, frontend, and Rust changes before scheduling expensive jobs
- narrow feature-test path triggers to feature and shared-support changes
- cache generated documentation with the corresponding Mix build artifacts

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -137,7 +137,7 @@ jobs:
             deps
             _build
             priv/native
-          key: mix-v5-${{ runner.os }}-${{ needs.versions.outputs.erlang }}-${{ needs.versions.outputs.elixir }}-e2e-${{ hashFiles('mix.lock', '**/Cargo.lock') }}-${{ hashFiles('mix.exs', 'Cargo.toml', 'config/**', 'lib/**', 'native/**', 'priv/**', 'test/support/**') }}
+          key: mix-v5-${{ runner.os }}-${{ needs.versions.outputs.erlang }}-${{ needs.versions.outputs.elixir }}-e2e-${{ hashFiles('mix.lock', '**/Cargo.lock') }}-${{ hashFiles('mix.exs', 'Cargo.toml', 'config/**', 'native/**', 'priv/**', 'test/support/**') }}
           restore-keys: |
             mix-v5-${{ runner.os }}-${{ needs.versions.outputs.erlang }}-${{ needs.versions.outputs.elixir }}-e2e-${{ hashFiles('mix.lock', '**/Cargo.lock') }}-
             mix-v5-${{ runner.os }}-${{ needs.versions.outputs.erlang }}-${{ needs.versions.outputs.elixir }}-e2e-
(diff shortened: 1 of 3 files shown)


.github/workflows/elixir-ci.yml
```diff
@@ -39,19 +40,69 @@ jobs:
...
+          fetch-depth: 0
+
       - name: Read versions from .tool-versions
         id: v
         run: |
           echo "elixir=$(grep '^elixir '  .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
           echo "erlang=$(grep '^erlang '  .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
           echo "rust=$(grep   '^rust '    .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
 
+      - name: Detect affected checks
```
amokan, line 61: Looks like we're missing a clause for the `docs/docs.logflare.com/docs/**` path in here.

amokan (dismissed):

.github/workflows/elixir-ci.yml
```diff
@@ -39,19 +42,79 @@ jobs:
...
+                test/e2e/features/*)
+                  quality=true
+                  ;;
+                test/e2e/*)
+                  ;;
+                .github/scripts/*|config/*|docs/docs.logflare.com/docs/*|lib/*|native/*|priv/*|test/*|*.exs|*.lock|Dockerfile.base|Dockerfile.runner|Dockerfile.multi-step)
+                  backend=true
+                  ;;
+              esac
+            done < <(git diff --name-only "$BASE_SHA" "$HEAD_SHA")
```
amokan: So if for some reason this `git diff` fails it looks like CI would be skipped and things would be green.

What about inverting this to default things to true?

```shell
backend=true; frontend=true; rust=true; quality=true

if [[ "$EVENT_NAME" != "workflow_dispatch" ]] \
   && git cat-file -e "${BASE_SHA}^{commit}" 2>/dev/null \
   && changed="$(git diff --name-only "$BASE_SHA...$HEAD_SHA")"; then
  backend=false; frontend=false; rust=false; quality=false
  while IFS= read -r path; do
    case "$path" in ... esac
  done <<< "$changed"
fi
```

.github/workflows/e2e-tests.yml
```diff
@@ -132,10 +137,15 @@ jobs:
             deps
             _build
             priv/native
-          key: mix-v3-${{ runner.os }}-${{ needs.versions.outputs.erlang }}-${{ needs.versions.outputs.elixir }}-${{ hashFiles('mix.lock', '**/Cargo.lock', 'native/**') }}
+          key: mix-v5-${{ runner.os }}-${{ needs.versions.outputs.erlang }}-${{ needs.versions.outputs.elixir }}-e2e-${{ hashFiles('mix.lock', '**/Cargo.lock') }}-${{ hashFiles('mix.exs', 'Cargo.toml', 'config/**', 'lib/**', 'native/**', 'priv/**', 'test/support/**') }}
```
amokan: I think the inclusion of `lib/**` in the second `hashFiles` call will cause the key to miss on every commit.

Ziinc (approved):

PR #3774, opened 2026-08-04
ci: share Supabase integration images across browsers

djwhitt:
## Summary

- build the Logflare image once for the Supabase integration workflow
- upload the uncompressed image as a short-lived workflow artifact
- load the same image in each browser test job instead of rebuilding it three times
- use the GitHub Actions BuildKit cache for the shared build

## Impact

Representative hosted runs from the original draft used about 40% fewer runner-seconds while retaining approximately ten-minute wall time.


amokan (approved):
👍

PR #3775, opened 2026-08-04
test: speed up backend adaptor integration tests

djwhitt:
## Summary

- replace fixed ClickHouse waits with observable retry assertions
- inject deterministic ClickHouse connection failures instead of waiting on unreachable ports
- shorten HTTP adaptor producer and batch intervals only in the test environment
- exercise HTTP adaptor pipelines through the full `Backends.ingest_logs/2` path
- start source supervision synchronously instead of waiting for adaptor startup

## Stack

- [#3775](https://github.com/Logflare/logflare/pull/3775) — Backend adaptor integration tests **(this PR)**
- [#3777](https://github.com/Logflare/logflare/pull/3777) — Lifecycle and side-effect synchronization
- [#3778](https://github.com/Logflare/logflare/pull/3778) — LiveView setup reuse

### DIFF
--- a/test/logflare/backends/adaptor/clickhouse_adaptor/provisioner_test.exs
+++ b/test/logflare/backends/adaptor/clickhouse_adaptor/provisioner_test.exs
@@ -67,23 +67,17 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.ProvisionerTest do
 
   describe "connection test failure handling" do
     @tag capture_log: true
-    test "fails initialization when ClickHouse is unavailable" do
-      {_source, invalid_backend} =
-        setup_clickhouse_test(
-          config: %{
-            url: "http://localhost",
-            username: "invalid_user",
-            password: "invalid_pass",
-            port: 19_999
-          }
-        )
+    test "fails initialization when ClickHouse is unavailable", %{backend: backend} do
+      expect(ClickHouseAdaptor, :test_connection, fn ^backend ->
+        {:error, :grant_check_unknown_failure}
+      end)
 
-      pid = start_supervised!({Provisioner, invalid_backend}, restart: :transient)
+      pid = start_supervised!({Provisioner, backend}, restart: :transient)
       ref = Process.monitor(pid)
 
       assert_receive {:DOWN, ^ref, :process, ^pid,
                       {:shutdown, {:error, :grant_check_unknown_failure}}},
-                     5_000
+                     1_000
     end
   end
 
--- a/test/logflare/backends/adaptor/clickhouse_adaptor/provisioner_test.exs
+++ b/test/logflare/backends/adaptor/clickhouse_adaptor/provisioner_test.exs
@@ -3,6 +3,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.ProvisionerTest do
 
   alias Logflare.Backends.Adaptor.ClickHouseAdaptor
   alias Logflare.Backends.Adaptor.ClickHouseAdaptor.Provisioner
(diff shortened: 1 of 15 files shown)


test/logflare/backends/adaptor/clickhouse_adaptor/provisioner_test.exs
```diff
@@ -67,23 +67,17 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.ProvisionerTest do
...
-            password: "invalid_pass",
-            port: 19_999
-          }
-        )
-
-      pid = start_supervised!({Provisioner, invalid_backend}, restart: :transient)
+    test "fails initialization when ClickHouse is unavailable", %{backend: backend} do
+      expect(ClickHouseAdaptor, :test_connection, fn ^backend ->
+        {:error, :grant_check_unknown_failure}
+      end)
```
Ziinc: should not stub this as internal functions are not tested, like check_ingest_grants/2

Ziinc: generally should not stub internal context modules.

test/support/data_case.ex
```diff
@@ -118,6 +121,32 @@ defmodule Logflare.DataCase do
...
   end
 
+  @doc """
+  Enqueues already-built log events directly into a source's configured backend pipeline.
+
+  Adaptor tests use this to exercise the adaptor without starting the unrelated full
+  source supervision tree through `Backends.ingest_logs/2`.
+  """
+  def enqueue_backend_logs(events, source) do
+    [backend | _] = BackendsCache.list_backends(source_id: source.id)
```
Ziinc: should insert into all backends if present

test/logflare/backends/adaptor/otlp_adaptor_test.exs
```diff
@@ -165,7 +164,7 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptorTest do
           timestamp: ts_us
         )
 
-      assert {:ok, _} = Backends.ingest_logs([log_event], source)
+      assert {:ok, _} = enqueue_backend_logs([log_event], source)
```
Ziinc: i don't see the benefit of this if the AdaptorSupervisor is already started, and increases risk since we aren't going through actual ingest path

test/support/data_case.ex
```diff
@@ -118,6 +121,32 @@ defmodule Logflare.DataCase do
     end)
   end
 
+  @doc """
+  Enqueues already-built log events directly into a source's configured backend pipeline.
+
+  Adaptor tests use this to exercise the adaptor without starting the unrelated full
+  source supervision tree through `Backends.ingest_logs/2`.
+  """
+  def enqueue_backend_logs(events, source) do
```
Ziinc: should add a guard for the events list to restrict to only LogEvent
```suggestion
  def enqueue_backend_logs([%LogEvent{} | _ ] = events, source) do
```

Ziinc (approved):

Ziinc:
needs some adjustments before merge

test/logflare/backends/adaptor/clickhouse_adaptor/provisioner_test.exs
djwhitt: Adjusted this one to stub at the external dep level instead to keep it deterministic. Worth noting we do have coverage for this in `ClickHouseAdaptorTest` too.

test/logflare/backends/adaptor/otlp_adaptor_test.exs
djwhitt: Fair point. I benchmarked and it's not a huge difference either. I switched it back.

PR #3776, opened 2026-08-04
test: build and synchronize queue fixtures efficiently

djwhitt:
## Summary

- build one valid queue event and clone it with unique IDs for saturation fixtures
- enqueue large fixtures in batches instead of performing factory and ETS work per event
- replace worker sleeps with monitors, messages, state barriers, and retry assertions
- preserve queue-cap, spillover, memory-monitor, and buffer-limiter coverage

## Stack

- [#3775](https://github.com/Logflare/logflare/pull/3775) — Backend adaptor integration tests
- [#3776](https://github.com/Logflare/logflare/pull/3776) — Queue fixtures and worker synchronization **(this PR)**
- [#3777](https://github.com/Logflare/logflare/pull/3777) — Lifecycle and side-effect synchronization
- [#3778](https://github.com/Logflare/logflare/pull/3778) — LiveView setup reuse

PR #3777, opened 2026-08-04
test: synchronize lifecycle and side-effect assertions

djwhitt:
## Summary

- synchronize source and backend lifecycle assertions with process state and monitors
- await ingestion, telemetry, cache, and controller side effects instead of sleeping
- exercise endpoint cache timer behavior directly while retaining timer-delivery coverage
- remove external DNS dependencies and flush Logger synchronously in utility checks
- isolate user-monitoring log capture from unrelated asynchronous logs

## Stack

- [#3775](https://github.com/Logflare/logflare/pull/3775) — Backend adaptor integration tests
- [#3777](https://github.com/Logflare/logflare/pull/3777) — Lifecycle and side-effect synchronization **(this PR)**
- [#3778](https://github.com/Logflare/logflare/pull/3778) — LiveView setup reuse

### DIFF
--- a/test/logflare/backends/user_monitoring_test.exs
+++ b/test/logflare/backends/user_monitoring_test.exs
@@ -11,7 +11,6 @@ defmodule Logflare.Backends.UserMonitoringTest do
   alias Logflare.Backends.QueryError
   alias Logflare.Backends.SourceSup
   alias Logflare.Backends.UserMonitoring
-  alias Logflare.Backends.UserMonitoring.IngestPipeline
   alias Logflare.SystemMetrics.AllLogsLogged
   alias Logflare.LogEvent
   alias Logflare.Backends.Adaptor.ClickHouseAdaptor
@@ -32,7 +31,9 @@ defmodule Logflare.Backends.UserMonitoringTest do
     [source: source, source_b: source_b, user: user]
   end
 
-  def start_otel_exporter(_context) do
+  def start_user_monitoring_pipeline(_context) do
+    start_supervised!(AllLogsLogged)
+
     UserMonitoring.get_otel_exporter()
     |> Enum.each(&start_supervised!/1)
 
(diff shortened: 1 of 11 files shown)


test/logflare/utils_test.exs
```diff
@@ -14,6 +14,8 @@ defmodule Logflare.Utils.FlagTest do
   alias Logflare.User
   alias Logflare.Utils
 
+  setup :set_mimic_global
```
Ziinc: should not set this for async tests

test/logflare/backends/user_monitoring_test.exs
```diff
@@ -230,6 +232,8 @@ defmodule Logflare.Backends.UserMonitoringTest do
                  )
                )
       end)
+
+      stop_supervised!(IngestPipeline)
```
Ziinc: why? should get auto stopped

test/logflare/backends/user_monitoring_test.exs
```diff
@@ -308,6 +312,8 @@ defmodule Logflare.Backends.UserMonitoringTest do
         refute attr["source_id"] in [source_id, metrics_source_id]
         refute attr["my_label"] == "test"
       end
+
+      stop_supervised!(IngestPipeline)
```
Ziinc: should get auto stopped

Ziinc (changes requested):

test/logflare/utils_test.exs
djwhitt: Addressed in 83862cb83f58a4f30a1c02c685715a38951fcaed

test/logflare/backends/user_monitoring_test.exs
djwhitt: Teardown was automatic, but there was an ordering issue. Addressed this in 776d67bca130297fd0f448e1c9ee448bc9193617.

Ziinc (approved):

PR #3778, opened 2026-08-04
test: reduce repeated LiveView setup

djwhitt:
## Summary

- include the signed-in user's selected team when mounting test LiveViews
- follow redirects only when a resource selects a genuinely different team
- add focused coverage for team query-parameter handling
- reuse each search LiveView across shorthand timestamp filter cases

## Stack

- [#3775](https://github.com/Logflare/logflare/pull/3775) — Backend adaptor integration tests
- [#3777](https://github.com/Logflare/logflare/pull/3777) — Lifecycle and side-effect synchronization
- [#3778](https://github.com/Logflare/logflare/pull/3778) — LiveView setup reuse **(this PR)**

### DIFF
--- a/test/support/conn_case.ex
+++ b/test/support/conn_case.ex
@@ -215,11 +215,12 @@ defmodule LogflareWeb.ConnCase do
   application redirect when the resource belongs to another team.
 
   Supplying the common `t=` parameter up front avoids mounting the LiveView
-  twice solely to discover the default team.
+  twice solely to discover the default team. Set `bypass_team_param: true` when
+  the first mount must exercise team selection or authorization without it.
   """
   defmacro live_with_redirect(conn, path \\ nil, opts \\ []) do
     quote bind_quoted: [conn: conn, path: path, opts: opts] do
-      path = LogflareWeb.ConnCase.put_default_team_param(conn, path)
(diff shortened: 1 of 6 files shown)


test/support/conn_case.ex
```diff
@@ -205,20 +211,27 @@ defmodule LogflareWeb.ConnCase do
...
-  Call live/3 and automatically follow the first live redirect.
+  Calls `live/3` with the signed-in user's selected team and follows the first
+  application redirect when the resource belongs to another team.
 
-  Useful for the common case of a live view redirecting to add the `t=` param.
+  Supplying the common `t=` parameter up front avoids mounting the LiveView
+  twice solely to discover the default team.
   """
   defmacro live_with_redirect(conn, path \\ nil, opts \\ []) do
     quote bind_quoted: [conn: conn, path: path, opts: opts] do
```
Ziinc, line 222: i worry that making this global might make authz bugs harder to catch and debug, would suggest adding an escape hatch like `live_with_redirect macro(bypass: false)` so that we exclude team-related auth tests.

Ziinc (changes requested):

Ziinc (approved):

PR #3781, opened 2026-08-04
feat: measure routing rule evaluation time

djwhitt:
## Summary

- measure routing-rule matching time without including sink/backend dispatch
- aggregate measurements once per routed batch and export an untagged microsecond histogram
- export the evaluated-event count so matching time can be normalized per event

## Performance

Benchmarked the adjacent baseline and this change with one scheduler, an attached `OtelMetricExporter`, and an empty router to isolate instrumentation overhead:

| Batch size | Baseline | Instrumented | Added per batch | Added per event |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 0.0446 µs | 1.23 µs | 1.19 µs | 1.19 µs |
| 100 | 1.73 µs | 6.81 µs | 5.08 µs | 0.051 µs |
| 1,000 | 16.19 µs | 58.50 µs | 42.31 µs | 0.042 µs |

The fixed cost is the single telemetry event and two exporter updates per batch. The per-event cost is the matching timer and aggregation.

## Validation

- `../bin/test test/logflare/sources/source_router_test.exs test/logflare/telemetry_test.exs`
- `MIX_ENV=test ../bin/x mix format --check-formatted`

### DIFF
--- a/lib/logflare/sources/source_router.ex
+++ b/lib/logflare/sources/source_router.ex
@@ -16,18 +16,51 @@ defmodule Logflare.Sources.SourceRouter do
   @spec route_to_sinks_and_ingest(LE.t() | [LE.t()], Source.t(), module()) :: LE.t() | [LE.t()]
   def route_to_sinks_and_ingest(events, source, router \\ @default_router)
 
-  def route_to_sinks_and_ingest(events, source, router) when is_list(events),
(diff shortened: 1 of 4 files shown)


djwhitt:
Closing. Too much a perf hit on this one.

PR #3784, opened 2026-08-04
feat(telemetry): tag ingestion metrics by backend type

djwhitt:
## Summary

- tag Broadway batcher, batch-processor, processor, and per-message duration metrics with explicit low-cardinality telemetry metadata
- propagate concrete `Backend.type` values as `backend_type`; identify pipelines without a backend through a separate module-valued `pipeline` tag
- tag dynamic-pipeline counts and retry requeue lookup misses by backend type
- emit direct backend dispatch counts on the event consumed by the existing dispatch-count metric

## Motivation

Shared Broadway duration metrics currently aggregate every ingestion pipeline, which prevents isolating ClickHouse saturation from upstream or unrelated backend work. The additional dimensions make those stages filterable without introducing backend-ID cardinality.

Direct backend dispatches, including backend-routing rules, currently publish their count under a different event name from the registered metric. Aligning the event name reports existing fan-out accurately; it does not change routing, source counters, `AllLogsLogged`, or billing behavior. The metric counts queue-dispatch copies, so one input event routed to N backends contributes N observations.

## Tagging contract

Broadway pipelines publish a dedicated `context.telemetry_tags` map:

- Pipelines with a `Backend` provide only its concrete `backend_type`.
- Pipelines without a backend provide only their module as `pipeline`.
- Missing or invalid telemetry metadata reports `pipeline=unknown`.
- The legacy nil-backend Webhook pipeline reports the valid `backend_type=webhook` fallback.

The metric extractor reads only this namespace and returns only the registered `backend_type` and `pipeline` attributes. This keeps `backend_type` consistent with dispatch and dynamic-pipeline metrics while retaining separate visibility into internal pipelines.

## Operational note

The four existing Broadway histograms will fan out by `backend_type` or `pipeline`. Existing panels and alerts must aggregate over or filter these attributes if they expect one series per node.

## Validation

- telemetry-tag configuration assertions are colocated with each owning pipeline test, run asynchronously without database or external-service setup, and share a process-local Broadway expectation helper
- the 11 isolated configuration cases complete together in under one second locally
- the central manually maintained pipeline registry test has been removed
- the real Axiom pipeline uses deterministic process readiness and verifies all four Broadway duration events map to `backend_type=axiom` under one shared five-second deadline
- exporter tests cover backend series, internal-pipeline series, invalid metadata, unknown fallback, and tag-map isolation
- direct-dispatch tests cover regular and consolidated ClickHouse queues
- latest affected test groups: 8 doctests and 220 tests, 0 failures
- formatting, warnings-as-errors compilation, and `mix lint.all` pass

The complete `backends_test.exs` file was attempted locally; the changed dispatch tests pass, while an unrelated existing `SourceSupWorker` timing assertion at line 745 remains flaky in this environment. A prior PR revision passed the complete GitHub Actions test suite.


test/logflare/backends/broadway_context_test.exs
```diff
@@ -0,0 +1,144 @@
...
+
+  defmodule QueueStub do
+    def resolve(name), do: {:ok, name}
+  end
+
+  defmodule StorageStub do
+  end
+
+  setup :set_mimic_global
+  setup :verify_on_exit!
```
amokan:
```suggestion
```

Shouldn't need these as we already have Broadway stubbed. By setting this globally here, I'm pretty sure it would also cause any other tests running that did something like a `Broadway.start_link` to return an ok tuple even if it wasn't actually started 🤷‍♂️

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -327,6 +327,7 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
           http: [concurrency: 6, batch_size: 250]
         ],
         context: %{
+          backend_type: (args.backend && args.backend.type) || :webhook,
```
amokan:
```suggestion
          backend_type: backend_type(args.backend),
```

reads a bit better if you pop in a custom function like:

```elixir
@spec backend_type(Backend.t() | nil) :: atom()
defp backend_type(%Backend{type: type}) when is_atom_value(type), do: type
defp backend_type(_backend), do: :webhook
```

(_and add an import to bring in the custom guard at the top of the module: `import Logflare.Utils.Guards, only: [is_atom_value: 1]`_)

test/logflare/backends/broadway_context_test.exs
```diff
@@ -0,0 +1,144 @@
...
+        Application.put_env(:logflare, :spool, original_spool_config)
+      else
+        Application.delete_env(:logflare, :spool)
+      end
+    end)
+
+    :ok
+  end
+
+  test "Broadway pipelines expose their effective backend type to duration metrics" do
```
amokan: This could probably be simplified by doing something like this (_I get macros are sometimes frowned upon, but I think this seems reasonable_)...

```elixir
@pipelines [
  clickhouse: :clickhouse,
  bigquery: :bigquery,
  postgres: :postgres,
  http_based: :axiom,
  s3: :s3,
  syslog: :syslog,
  webhook_with_backend: :datadog,
  webhook_without_backend: :webhook,
  spool_consumer: :spool_consumer,
  spool_producer: :spool_producer,
  user_monitoring: :user_monitoring
]

describe "Effective backend type exposed to Broadway duration metrics" do
  for {pipeline, expected_type} <- @pipelines do
    test "#{pipeline} pipeline reports #{inspect(expected_type)}", ctx do
      assert_backend_type(unquote(expected_type), unquote(pipeline), ctx)
    end
  end
end
```

amokan (approved):
Approved with a couple nits

test/logflare/backends/broadway_context_test.exs
Ziinc: macros in test are completely fine imo
its the macros in prod code that could be done with regular funcs that is where the frowning starts

test/logflare/backends/adaptor/axiom_adaptor_test.exs
```diff
@@ -143,6 +153,22 @@ defmodule Logflare.Backends.Adaptor.AxiomAdaptorTest do
...
+
+      assert_receive {:telemetry_event, [:logflare, :backends, :pipeline, :handle_batch],
+                      _measurements, %{backend_type: :axiom}}
+
+      metrics = Logflare.Telemetry.metrics()
+
+      for event <- broadway_duration_events do
+        assert_receive {:telemetry_event, ^event, %{duration: duration},
+                        %{context: %{backend_type: :axiom}} = metadata},
+                       5_000
```
Ziinc: this extends the test quite significantly, perhaps wrap the for/do block with a TestUtils.retry_assert instead

lib/logflare/backends/user_monitoring/ingest_pipeline.ex
```diff
@@ -35,7 +35,8 @@ defmodule Logflare.Backends.UserMonitoring.IngestPipeline do
           batch_size: batch_size,
           batch_timeout: batch_timeout
         ]
-      ]
+      ],
+      context: %{backend_type: :user_monitoring}
```
Ziinc: this is not a valid backend type, should not set this. also, should not set context when it is not used.

lib/logflare/backends/spool/producer_pipeline.ex
```diff
@@ -65,6 +65,7 @@ defmodule Logflare.Backends.Spool.ProducerPipeline do
         ]
       ],
       context: %{
+        backend_type: :spool_producer,
```
Ziinc: this is not a valid backend type, should not set this. also, should not set context when it is not used.

lib/logflare/backends/spool/consumer_pipeline.ex
```diff
@@ -45,7 +45,8 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline do
           batch_timeout: 1_000,
           concurrency: 1
         ]
-      ]
+      ],
+      context: %{backend_type: :spool_consumer}
```
Ziinc: this is not a valid backend type, should not set this. also, should not set context when it is not used.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -110,7 +110,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
           batch_timeout: @batch_timeout
         ]
       ],
-      context: %{backend_id: backend.id}
+      context: %{backend_id: backend.id, backend_type: :clickhouse}
```
Ziinc: unused

lib/logflare/backends/spool/consumer_pipeline.ex
Ziinc: unused, and not a valid backend adaptor

lib/logflare/backends/spool/producer_pipeline.ex
Ziinc: unused, and not a valid backend adaptor

lib/logflare/sources/source/bigquery/pipeline.ex
```diff
@@ -87,6 +87,7 @@ defmodule Logflare.Sources.Source.BigQuery.Pipeline do
             ]
           ],
           context: %{
+            backend_type: :bigquery,
```
Ziinc: unused, also unnecessary to set it here if we're hardcoding it at :611

lib/telemetry.ex
```diff
@@ -603,6 +623,17 @@ defmodule Logflare.Telemetry do
...
+  defp broadway_backend_tags(%{context: %{backend_type: backend_type}})
+       when is_atom_value(backend_type),
+       do: %{backend_type: backend_type}
+
+  defp broadway_backend_tags(%{context: %{backend: %{type: backend_type}}})
+       when is_atom_value(backend_type),
+       do: %{backend_type: backend_type}
+
+  defp broadway_backend_tags(_metadata), do: %{backend_type: :unknown}
+
```
Ziinc: doing this is too implicit and make the telemetry tagging pipeline hard to understand, should favour explicit tags as opposed to building it implicitly from a broadway context

test/logflare/backends/broadway_context_test.exs
```diff
@@ -0,0 +1,144 @@
...
+      ConsumerPipeline.start_link(name: :spool_consumer_context_test)
+    end)
+
+    assert_backend_type(:spool_producer, tag_values, fn ->
+      ProducerPipeline.start_link(name: :spool_producer_context_test)
+    end)
+
+    assert_backend_type(:user_monitoring, tag_values, fn ->
+      UserMonitoringPipeline.start_link(metric_store_name: :user_monitoring_context_test)
+    end)
```
Ziinc: these are Broadway pipelines but not actual backend Adaptor types. this would be confusing down the line.

Ziinc: would be more appropriate to do something like `pipeline: UserMonitoringPipeline` instead to tag by the pipeline name

test/logflare/backends/broadway_context_test.exs
```diff
@@ -0,0 +1,144 @@
...
+  end
+
+  defp assert_backend_type(expected, tag_values, start_pipeline) do
+    assert {:ok, _pid} = start_pipeline.()
+    assert_receive {:broadway_context, context}
+    assert tag_values.(%{context: context}) == %{backend_type: expected}
+  end
```
Ziinc: not a fan of how implicit this is, and also means that each additional backend adaptor needs to get registered here. if we're going to enforce Adaptor-wide behaviour then it should be done in an appropriate Adaptor macro for/do test that automatically discovers adaptors using this Adaptor behaviour and tests it for a specific implementation, or to use a required callback to enforce implementation

Ziinc (changes requested):

Ziinc:
i mentioned many as unused as i went down the LOCs, but that also highlights the issue : changes/usage is not collocated so it is hard for reader to follow the why in which the context value is set.

djwhitt:
@Ziinc Going through your feedback I agree about overloading backend type and the code being too implicit, but the additions you call out as being unused are used on broadway stop telemetry.

lib/logflare/backends/adaptor/http_based/pipeline.ex
```diff
@@ -68,7 +69,7 @@ defmodule Logflare.Backends.Adaptor.HttpBased.Pipeline do
       [:logflare, :backends, :pipeline, :handle_batch],
       %{batch_size: batch_info.size, batch_trigger: batch_info.trigger},
       %{
-        backend_type: :http_based
+        backend_type: context.telemetry_tags.backend_type
       }
```
Ziinc, line 73: Could just pass in `context.telemetry_tags`

lib/logflare/backends/spool/producer_pipeline.ex
```diff
@@ -65,6 +65,7 @@ defmodule Logflare.Backends.Spool.ProducerPipeline do
         ]
       ],
       context: %{
+        telemetry_tags: %{pipeline: __MODULE__},
```
Ziinc, line 68: This will be the fully qualified module, would be quite verbose. Will suggest a simple atom like `spool_consumer`

Ziinc (approved):
Would be good to have an Adaptor level test that discovers all modules implementing the Adaptor and tests that they emit the correct pipeline telemetry for conformance, or to implement a required callback for Adaptor require telemetry declaration

Ziinc:
@djwhitt approved to unblock, can you resolve comments and conflicts before merge thanks 🙏

PR #3787, opened 2026-08-05
Support multiple OTLP endpoints for metrics export

djwhitt:
## Summary
- `OtelMetricExporter` only supports one `otlp_endpoint` per instance, and running a second instance to reach a second collector would duplicate the entire metrics ETS table and double telemetry handler invocations on the hot path (every tracked event gets recorded twice, not just exported twice).
- Adds `Logflare.Telemetry.MultiEndpointMetricsExporter`, an `export_callback` that fans a single already-built batch out to the primary endpoint plus any endpoints configured via `LOGFLARE_OTEL_METRICS_EXTRA_ENDPOINTS` (JSON array of `{"endpoint": "...", "headers": {...}}`), so there's still only one `MetricStore`/ETS table/telemetry subscription — only the periodic export step does N requests instead of one.
- No-op when no extra endpoints are configured: `lib/telemetry.ex` only installs the callback if `otel_metrics_extra_endpoints` is non-empty, so current single-endpoint behavior is unchanged.
- If any endpoint fails, the whole batch is reported as failed (identifying which endpoint) so it's retried on the next export cycle, mirroring the retry-on-error behavior the library already has for a single endpoint.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -441,10 +441,27 @@ if System.get_env("LOGFLARE_OTEL_ENDPOINT") do
       value -> String.to_float(value)
     end
 
+  # A JSON array of `{"endpoint": "https://...", "headers": {"x-api-key": "..."}}`
+  # objects, sent to in addition to LOGFLARE_OTEL_ENDPOINT for metrics only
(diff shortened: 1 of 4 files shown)


djwhitt:
Closing. We will take a different approach.

PR #3788, opened 2026-08-06
Fix/otlp protobuf formatting

Baishan:
Sets the service.name and service.namespace to reference the project ref and supabase component the log originated from.

Sets severity number according to the postgres severity/http status or metadata.level.

Removes node and cluster from the resource attributes.

closes O11Y-2277 O11Y-2278 and O11Y-2279


lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
```diff
@@ -54,41 +55,70 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptor.ProtobufFormatter do
   @spec reserved_headers() :: [String.t()]
   def reserved_headers, do: ["content-type"]
 
-  defp transform_batch(events, source) do
+  defp transform_batch(events, source, project_ref) do
```
Ziinc, line 58: We should not have any references to project ref, logflare should not know anything about supabase or multi tenancy

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
```diff
@@ -54,41 +55,70 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptor.ProtobufFormatter do
...
+          [
+            "service.name": name,
+            "service.version": Application.spec(:logflare, :vsn) |> to_string()
+          ]
+
+        {:source, name} ->
+          # We have no reliable version for the customer's own service, so
+          # service.version is omitted rather than reporting Logflare's version
+          # under a name that no longer refers to Logflare.
+          ["service.name": name]
```
Ziinc, line 87: Just a note that we could try to check if the event has a version field set in the future as more products get otel compliant.

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
```diff
@@ -109,11 +139,82 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptor.ProtobufFormatter do
...
+      "debug" -> :SEVERITY_NUMBER_DEBUG
+      "info" -> :SEVERITY_NUMBER_INFO
+      "notice" -> :SEVERITY_NUMBER_INFO
+      "warn" -> :SEVERITY_NUMBER_WARN
+      "warning" -> :SEVERITY_NUMBER_WARN
+      "error" -> :SEVERITY_NUMBER_ERROR
+      "critical" -> :SEVERITY_NUMBER_FATAL
+      "fatal" -> :SEVERITY_NUMBER_FATAL
+      "panic" -> :SEVERITY_NUMBER_FATAL
+      _unrecognized -> nil
```
Ziinc, line 210: Default to info?

Ziinc (changes requested):

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
Baishan: Customers need a way to differentiate resources if they have multiple projects. Arguably project is already baked into the ui in places and will be set to nil if its empty. I'm open to alternatives.

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
Baishan: I prefer to omit this if we can't determine it. Its easier to identify something is wrong and set the severity_number than if we mark important errors as info

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
Baishan: I think we should be setting service.name/version explicitly in the log events but we need to plan where these things should live.

Ziinc (approved):

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
Ziinc: i think having a `tenant` concept within logflare is fine, but managing this will be tricky. 

> Arguably project is already baked into the ui in places and will be set to nil if its empty. 

this actually should not have leaked in, `project` concept makes no sense in the logflare-only world if we were to look at it in a standalone view, which is how all supabase products should function.

PR #3789, opened 2026-08-06
Fix/otlp body format

Baishan:
Screenshot 2026-08-06 at 15 56 10

Sets the LogRecord.body to be the human readable text event_message as this is what is displayed on dash0 and other provider uis.

Sets the attributes to be a flattened list of keys corresponding to the metadata in the logevent body. See the screenshot above. This allows dash0 and likely others that don't supported nested json to query and filter attributes.

closes O11Y-2280

### DIFF
--- a/lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
+++ b/lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
@@ -18,14 +18,15 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptor.ProtobufFormatter do
   @protobuf_content_type "application/x-protobuf"
 
   @impl true
-  def call(env, next, _opts) do
+  def call(env, next, opts) do
     source = env.opts[:source] || %{}
     project_ref = env.opts |> Keyword.get(:metadata, %{}) |> Map.get("backend.project_ref")
+    flatten_attributes? = opts[:flatten_attributes] || false
 
     response =
       env
       |> put_content_type_header()
-      |> Tesla.put_body(transform_batch(env.body, source, project_ref))
+      |> Tesla.put_body(transform_batch(env.body, source, project_ref, flatten_attributes?))
       |> Tesla.run(next)
 
     with {:ok, %{status: 200} = env} <- response do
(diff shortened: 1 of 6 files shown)


Ziinc (changes requested):
Should make this configurable instead of setting it as a blanket change. Should add a boolean option to set event_message on body and move all other fields to flattened attributes

Baishan:
> Should make this configurable instead of setting it as a blanket change. Should add a boolean option to set event_message on body and move all other fields to flattened attributes

These are opt in now via a boolean on the backend config.

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
```diff
@@ -18,14 +18,15 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptor.ProtobufFormatter do
   @protobuf_content_type "application/x-protobuf"
 
   @impl true
-  def call(env, next, _opts) do
+  def call(env, next, opts) do
     source = env.opts[:source] || %{}
     project_ref = env.opts |> Keyword.get(:metadata, %{}) |> Map.get("backend.project_ref")
+    flatten_attributes? = opts[:flatten_attributes] || false
```
Ziinc: something more self-documenting and explicit would be better, like `flatten_to_attributes` or `body_to_attributes`, as this changes both the body AND the attributes

lib/logflare_web/open_api_schemas.ex
```diff
@@ -435,7 +434,13 @@ defmodule LogflareWeb.OpenApiSchemas do
       endpoint: %Schema{type: :string},
       protocol: %Schema{type: :string, nullable: true},
       gzip: %Schema{type: :boolean},
-      headers: %Schema{type: :object}
+      headers: %Schema{type: :object},
+      flatten_attributes: %Schema{
+        type: :boolean,
+        nullable: true,
+        description:
+          "Sends the log body as a plain message and flattens everything else into individual attributes instead of one nested body. Defaults to false."
```
Ziinc:
```suggestion
          "Sends the log event_message as a text string in the body field and flattens other fields into individual attributes, instead of the default nested body. Defaults to false."
```

Ziinc (approved):

lib/logflare/backends/adaptor/otlp_adaptor/protobuf_formatter.ex
Baishan: Changed to flatten_to_attributes

lib/logflare_web/open_api_schemas.ex
Baishan: Updated

PR #3791, opened 2026-08-06
fix: cover every Datadog region in Finch pool config and metrics

djwhitt:
Stacked on #3790 — review that first; this PR's base is its branch.

`Logflare.Networking.all_datadog_pools/0` and the URL list `SystemMetrics.Cluster.finch/0` polls were both hand-maintained copies of the Datadog adaptor's region map, and both had drifted: neither listed `AP2`, added several regions ago.

A missing entry is not a connectivity problem. Finch treats the `:default` entry as a catch-all *configuration* ([docs](https://hexdocs.pm/finch/Finch.html#start_link/1)), so an unlisted intake host still gets its own pool with the same protocol and sizing. The only difference is `start_pool_metrics?`, which gates `Finch.get_pool_status/2`. So the effect was silently absent pool telemetry for the omitted regions — worth noting since adding pool entries alone wouldn't have fixed it, as `Cluster.finch/0` is what actually emits.

Both lists now derive from the adaptor map via `DatadogAdaptor.intake_origins/0`, so a new region can't be metered inconsistently. Also drops the `# TODO(ziinc): add in datadog pools` comment, which sat above a list that already contained Datadog pools.

### DIFF
--- a/test/logflare/application_test.exs
+++ b/test/logflare/application_test.exs
@@ -2,6 +2,8 @@ defmodule Logflare.ApplicationTest do
   @moduledoc false
   use ExUnit.Case, async: true
 
+  alias Logflare.Backends.Adaptor.DatadogAdaptor
+
   describe "datadog_connection_pools/0" do
     setup do
       original_config = Application.get_env(:logflare, :http_connection_pools)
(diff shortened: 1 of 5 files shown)


test/logflare/application_test.exs
```diff
@@ -24,17 +24,19 @@ defmodule Logflare.ApplicationTest do
           else: Application.delete_env(:logflare, :http_connection_pools)
 
         pools = get_pools()
-        assert map_size(pools) == 5
+        assert map_size(pools) == 9
```
amokan:
```suggestion
        assert MapSet.new(Map.keys(pools)) == MapSet.new(DatadogAdaptor.intake_origins())
```

This would avoid a magic count situation 🤷

amokan (changes requested):
One thing we're missing is an update to `test/logflare/networking_test.exs` to include the 4 new regions. We could also assert against `DatadogAdaptor.intake_origins()` if we didn't want to add them verbosely.

amokan (approved):

PR #3802, opened 2026-08-07
chore: trim whitespace in log filter fields

amokan:
Trims any whitespace around log filter fields on the LV backend as well as the presentation side via a JS hook.

This is a QOL improvement for people copy/pasting values in.

### DIFF
--- a/assets/js/test/lql_editor_wrapper_hook.test.js
+++ b/assets/js/test/lql_editor_wrapper_hook.test.js
@@ -1,0 +1,55 @@
+import { describe, expect, it } from "vitest";
+
+const { collectTrimmedFields, trimFieldInput } = await import(
+  "../lql_editor_wrapper_hook.js"
+);
+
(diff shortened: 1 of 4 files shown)


assets/js/test/lql_editor_wrapper_hook.test.js
```diff
@@ -0,0 +1,55 @@
...
+      "metadata.organization_slug": "abcdef",
+      "metadata.project_ref": "ghijkl",
+      "metadata.request_id": "",
+    });
+  });
+
+  it("returns an empty map when there are no inputs", () => {
+    expect(collectTrimmedFields([])).toEqual({});
+  });
+});
```
djwhitt, line 77: Small nit: might be worth adding a lifecycle test to ensure the handler is registered and removed correctly in `mounted` and `destroyed`.

djwhitt (approved):

PR #3805, opened 2026-08-10
perf: Reduce BigQuery schema update pressure

ruslandoga:
Adds focused Benchee and tprof coverage for schema helpers, queue-length admission strategies, and end-to-end schema update planning.

Bounds outstanding Schema casts with per-process Registry atomics, refreshes counters on restart, pins casts to the checked PID, and skips planning during cooldown or when schema updates are disabled.

Replaces generic DeepMerge and flattened-schema diffing with indexed, change-aware merges that preserve the original schema on no-ops and sort only real updates.

Against the checked-in Apple M2 baseline, median planning latency improves by about 1.8x to 4.5x and reductions fall by 40% to 69%; 46 focused tests and 14 doctests pass.

This draft overlaps #3589's change-flag work and should be split or stacked before review.

### DIFF
--- a/lib/logflare/sources/source/bigquery/schema.ex
+++ b/lib/logflare/sources/source/bigquery/schema.ex
@@ -17,6 +17,7 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
   alias Logflare.Sources.Source
   alias Logflare.LogEvent
   alias Logflare.AccountEmail
+  alias Logflare.Backends
   alias Logflare.Mailer
   alias Logflare.TeamUsers
   alias Logflare.SingleTenant
@@ -31,13 +32,20 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
     )
   end
 
+  @spec update_counter() :: :atomics.atomics_ref()
+  def update_counter do
+    :atomics.new(1, signed: false)
+  end
+
   @spec update(pid() | tuple(), LogEvent.t(), Source.t()) :: :ok
   def update(pid, %LogEvent{} = log_event, %Source{} = source)
       when is_pid(pid) or is_tuple(pid) do
-    if accept_update?(pid) do
-      GenServer.cast(pid, {:update, log_event, source})
-    else
-      :ok
+    case checkout_update(pid) do
+      {:ok, counter} ->
+        GenServer.cast(pid, {:update, counter, log_event, source})
+
+      :drop ->
+        :ok
     end
   end
 
@@ -90,20 +98,21 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
     {:noreply, %{state | field_count: count_fields(schema)}}
   end
 
-  def handle_cast(
-        {:update, %LogEvent{}, %Source{lock_schema: true}},
-        state
-      ),
-      do: {:noreply, state}
+  def handle_cast({:update, counter, log_event, source}, state) do
+    try do
+      handle_update(log_event, source, state)
+    after
+      checkin_update(counter)
+    end
+  end
 
-  def handle_cast(
-        {:update, %LogEvent{}, _source},
-        %{field_count: fc, field_count_limit: limit} = state
-      )
-      when fc > limit,
-      do: {:noreply, state}
+  defp handle_update(%LogEvent{}, %Source{lock_schema: true}, state), do: {:noreply, state}
 
-  def handle_cast({:update, %LogEvent{body: body, id: event_id}, _source}, state) do
+  defp handle_update(%LogEvent{}, _source, %{field_count: fc, field_count_limit: limit} = state)
+       when fc > limit,
+       do: {:noreply, state}
+
+  defp handle_update(%LogEvent{body: body, id: event_id}, _source, state) do
     LogflareLogger.context(source_id: state.source_token, log_event_id: event_id)
 
     source_schema = SourceSchemas.Cache.get_source_schema_by(source_id: state.source_id)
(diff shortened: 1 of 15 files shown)


ruslandoga:
Post-rebase benchmark on Apple M2, Elixir 1.19.5 / OTP 27.3.4.6, constrained to two schedulers:

`ERL_FLAGS='+S 2:2 +A 1' PROFILE=1 TPROF_TYPE=time MIX_ENV=test mix run test/profiling/bq_schema_update_internals_bench.exs`

| Path | Median | IPS | Memory | Reductions |
| --- | ---: | ---: | ---: | ---: |
| edge log / noop | 5.96 us | 134.32 K | 10.85 KB | 963 |
| edge log / update | 10.13 us | 77.56 K | 17.16 KB | 1,840 |
| lists / noop | 2.38 us | 306.22 K | 4.53 KB | 421 |
| lists / update | 4.38 us | 177.57 K | 7.09 KB | 830 |
| scalars / noop | 3.42 us | 218.26 K | 5.12 KB | 472 |
| scalars / update | 6.50 us | 94.01 K | 8.32 KB | 956 |
| rate limited | 0.083-0.084 us | 2.84-5.16 M | 0 KB | 9 |

Compared with the checked-in Apple M2 upstream snapshot, median latency improves by 4.46x/3.17x for edge-log noop/update, 3.12x/1.87x for lists, and 3.08x/1.82x for scalars. Reductions fall by 67%/43%, 68%/42%, and 69%/40%, respectively.

`tprof` confirms that the rate-limited path never enters `SchemaBuilder`, and caching backend eligibility removes `SingleTenant`, `Application.get_env`, and ETS config lookups from eligible calls. No-op profiles no longer reach sorting or the update-only `MapSet` work; real updates remain dominated by recursive `merge_field_schema/2`, list/map construction and traversal, `MapSet`, and `deep_sort_by_fields_name/1`. Per-function times are noisy at these microsecond durations under tracing, so the call graph/counts plus Benchee medians/reductions are the useful signals.

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -9,20 +9,22 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
...
   alias Logflare.SingleTenant
+  alias Logflare.SourceSchemas
+  alias Logflare.Sources
+  alias Logflare.Sources.Source
+  alias Logflare.Sources.Source.BigQuery.SchemaBuilder
+  alias Logflare.TeamUsers
 
   def start_link(args) when is_list(args) do
     {name, args} = Keyword.pop(args, :name)
+    name = put_update_counter(name)
```
ruslandoga, line 27: Allocate the counter in `start_link/1`, rather than while constructing the supervisor child spec. Each restart therefore registers a fresh zeroed counter instead of inheriting outstanding work from the mailbox that disappeared with the old PID.

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -31,20 +33,32 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
...
+    :atomics.new(1, signed: false)
+  end
+
+  @spec update(pid() | tuple(), LogEvent.t(), Source.t()) :: :ok
   def update(pid, %LogEvent{} = log_event, %Source{} = source)
       when is_pid(pid) or is_tuple(pid) do
-    GenServer.cast(pid, {:update, log_event, source})
+    case checkout_update(pid) do
+      {:ok, update_pid, counter} ->
+        GenServer.cast(update_pid, {:update, counter, log_event, source})
```
ruslandoga, line 46: The Registry lookup returns the PID and its counter together, then the cast is pinned to that PID. This removes a second Registry lookup and prevents a concurrent restart from pairing the new process with the old process counter.

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -31,20 +33,32 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
...
 
   # Public for profiling: benchmarks can target the GenServer's pure schema
   # planning work without measuring mailbox scheduling or BigQuery side effects.
   @doc false
   def plan_update(body, db_schema, state) do
-    schema = try_schema_update(body, db_schema)
-
-    if schema_needs_update?(db_schema, schema, state) do
-      {:update, schema}
+    if schema_update_allowed?(state) do
```
ruslandoga, line 57: Check the existing cooldown/backend eligibility before constructing a candidate schema. Events received during the cooldown were already discarded after diffing; this avoids the full recursive build for a patch that cannot be attempted.

lib/logflare/sources/source/bigquery/schema_builder.ex
```diff
@@ -268,51 +302,50 @@ defmodule Logflare.Sources.Source.BigQuery.SchemaBuilder do
...
-    @spec maybe_find_with_name(list(TFS.t()), TFS.t()) :: TFS.t() | nil
-    def maybe_find_with_name(enumerable, %TFS{name: name}) do
-      Enum.find(enumerable, &(&1.name === name))
+    if field_changed? do
+      new_field_names = MapSet.new(new_fields, & &1.name)
+      old_only_fields = Enum.reject(old_fields, &MapSet.member?(new_field_names, &1.name))
+
+      {%{new | fields: merged_new_fields ++ old_only_fields}, true}
+    else
+      {old, false}
```
ruslandoga, line 341: If payload order or a nested-field subset produces no real change, retain the original table field schema. This fixes false-positive updates and avoids allocating the old-only `MapSet`/list or deep-sorting the full schema on the no-op path.

ruslandoga:
Closing this draft because the experiment has grown too broad for a reviewable PR. Continuing benchmark- and profile-driven exploration locally; any eventual submissions will be split into small, self-contained patches.

PR #3806, opened 2026-08-10
fix(bigquery): bound schema sample queues

djwhitt:
## Summary

- bound pending processor-side BigQuery schema samples to 8 per source/backend
- keep existing asynchronous sampling semantics while using the freshest available rate estimate
- skip schema cache/planning work while updates are throttled or disabled
- remove redundant full-schema conversions from exact equality checks
- add low-cardinality admission, queue, and schema-operation telemetry
- isolate admission state in each Schema Registry registration so Schema restarts leave pipelines running

## Problem

`Pipeline.process_data/3` sampled events for schema discovery and cast every selected sample to the Schema GenServer. When schema processing fell behind, these casts could grow the mailbox without a bound and retain large event payloads.

Admitted samples also continued fetching and rebuilding schemas during the update cooldown, even though they could not patch BigQuery. Every allowed planning attempt additionally converted both schemas to flat typemaps before requiring exact structural equality.

## Implementation

- store a fresh `:atomics` admission counter and configured limit in each Schema process's Registry value
- look up the registered `{pid, counter}` pair, reserve before casting, and release when Schema dequeues the sample
- use compare-and-exchange reservations so rejected concurrent attempts never inflate the pending count
- use `max(average_rate, last_rate)` for sampling probability
- keep Schema and Pipeline independently supervised with `:one_for_one` restart semantics
- check cooldown/Postgres eligibility before schema cache access and planning
- compare schemas directly while preserving exact structural-equality behavior
- expose fixed-cardinality node-level selected/admitted/rejected/handled counters
- report bounded Schema queue aggregates from the existing top-10 process sample
- time schema patch, refresh, persist, and notification operations
- enrich ErlSysMon queue diagnostics with source registry keys

```mermaid
flowchart LR
    P[Pipeline processors] --> S[Sample selected]
    S --> R[Registry lookup]
    R --> A{Reserve atomic slot?}
    A -- Full --> X[Record rejection]
    A -- Available --> C[Cast to Schema PID]
    C --> Q[Bounded Schema mailbox, max 8]
    Q --> U[Dequeue and release slot]
    U --> W[Cooldown check and schema work]

    Z[Schema restart creates new PID, counter, and capacity] -.-> R
```

Schema patch attempts are limited to 6 per minute per Schema process, producing a 10-second cooldown after each successful or failed attempt. Sampling and bounded admission continue during that interval, but dequeued samples release their slot and return before schema-cache access or planning. This keeps ingestion asynchronous while avoiding redundant schema work.

Sampling remains best-effort: samples selected while all 8 reservations are occupied are rejected. A Schema restart can drop a sample already reserved against the old PID, but the replacement starts with independent admission capacity and pipelines remain running.

### Why 8

Sampling targets roughly one event per second, so 8 slots absorb about eight seconds of normal sampled backlog while retaining several opportunities to discover sparse fields. A limit of 32 retained roughly 32 seconds of samples, but after a successful patch most queued samples immediately hit the update cooldown; zero-rate/startup bursts are also likely to contain redundant neighboring events.

Because queued samples retain full event payloads, reducing the limit from 32 to 8 cuts bounded synthetic queue memory from 0.15 MiB to 0.04 MiB. Eight is intentionally more conservative than 4 so slow operations still retain useful sample diversity. Rollout can be evaluated with the selected/admitted/rejected counters, queue maximum, and schema-operation latency already added by this PR.

### Admission bound

Synthetic run with 50,000 messages containing 50 fields:

| Mode | Queue length | Process memory | Admission/enqueue time |
| --- | ---: | ---: | ---: |
| Unbounded | 50,000 | 233.84 MiB | 96.20 ms |
| Bounded | 8 | 0.04 MiB | 0.344 ms |

- queue reduction: 6,250x
- memory reduction: 5,847.07x
- saturated admission gate: 6.88 ns/attempt
- accepted reserve/release cycle: 17.18 ns
- instrumented rejection: 80.82 ns/attempt

### Schema planning follow-up

Adjacent same-machine before/after runs:

| Workload | Before | After |
| --- | ---: | ---: |
| Cooldown, 500 fields (median) | 3.52 ms | 42 ns |
| Cooldown memory | 4.32 MiB | 0 B |
| Cooldown reductions | 471.27K | 8 |

Across the existing edge-log, list, and scalar fixtures:

- no-op planning median improved **51-67%**
- update planning median improved **57-75%**
- planning allocations fell **57-65%**
- planning reductions fell **35-39%**

The profiling script now retains a dedicated cooldown workload.

## Validation

- [x] 40 focused tests, 0 failures (3 excluded)
- [x] concurrent reservation, Schema restart-race, and pipeline-stability coverage
- [x] `MIX_ENV=test mix compile --warnings-as-errors`
- [x] formatting check
- [x] `MIX_ENV=test mix lint.all`
- [x] admission and schema-planning benchmarks
- [x] independent correctness/restart-race review

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -41,7 +41,7 @@ config :logflare, :clickhouse_backend_adaptor,
 
 config :logflare, Logflare.Sources.Source.BigQuery.Schema,
   updates_per_minute: 6,
-  max_pending_samples: 32
+  max_pending_samples: 8
 
 # Configures the endpoint
 config :logflare, LogflareWeb.Endpoint,
--- a/config/config.exs
+++ b/config/config.exs
@@ -39,7 +39,9 @@ config :logflare, :clickhouse_backend_adaptor,
   pool_size: 3,
   native_pool_size: 10
 
-config :logflare, Logflare.Sources.Source.BigQuery.Schema, updates_per_minute: 6
+config :logflare, Logflare.Sources.Source.BigQuery.Schema,
+  updates_per_minute: 6,
+  max_pending_samples: 32
 
 # Configures the endpoint
 config :logflare, LogflareWeb.Endpoint,
--- a/lib/logflare/sources/source/bigquery/schema.ex
+++ b/lib/logflare/sources/source/bigquery/schema.ex
@@ -59,9 +59,9 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
     :ok
   end
 
-  def update(pid, %LogEvent{} = log_event, %Source{} = source)
-      when is_pid(pid) or is_tuple(pid) do
-    GenServer.cast(pid, {:update, log_event, source})
+  def update(server, %LogEvent{}, %Source{}) do
+    raise ArgumentError,
+          "expected the Schema server to use a Registry name, got: #{inspect(server)}"
   end
 
   @doc false
@@ -106,7 +106,14 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
     {:via, Registry, {registry, key, {@admission_value_tag, sample_counter, max_pending_samples}}}
   end
 
-  defp register_admission_counter(name, _max_pending_samples), do: name
+  defp register_admission_counter({:via, Registry, _name}, max_pending_samples) do
+    raise ArgumentError,
+          "expected :max_pending_samples to be a positive integer, got: #{inspect(max_pending_samples)}"
+  end
+
+  defp register_admission_counter(name, _max_pending_samples) do
+    raise ArgumentError, "expected :name to be a Registry name, got: #{inspect(name)}"
+  end
 
   # Public for profiling: benchmarks can target the GenServer's pure schema
   # planning work without measuring mailbox scheduling or BigQuery side effects.
@@ -162,10 +169,6 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
     release_update_slot(sample_counter)
 
     SchemaMetrics.record_handled()
-    handle_update(log_event, source, state)
-  end
-
-  def handle_cast({:update, %LogEvent{} = log_event, %Source{} = source}, state) do
     handle_update(log_event, source, state)
   end
 
--- a/lib/logflare/sources/source/bigquery/schema.ex
+++ b/lib/logflare/sources/source/bigquery/schema.ex
@@ -24,6 +24,8 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
 
   @admission_value_tag :schema_admission
 
+  @type admission_counter :: :atomics.atomics_ref()
+
   def start_link(args) when is_list(args) do
     {name, args} = Keyword.pop(args, :name)
     {max_pending_samples, args} = Keyword.pop(args, :max_pending_samples)
@@ -63,8 +65,8 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
   end
 
   @doc false
-  def reserve_update_slot(sample_counter, limit)
-      when is_reference(sample_counter) and is_integer(limit) and limit > 0 do
+  @spec reserve_update_slot(admission_counter(), pos_integer()) :: :ok | :full
+  def reserve_update_slot(sample_counter, limit) when is_integer(limit) and limit > 0 do
     current = :atomics.get(sample_counter, 1)
 
     if current >= limit do
@@ -78,8 +80,15 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
   end
 
   @doc false
-  def pending_update_slots(sample_counter) when is_reference(sample_counter) do
+  @spec pending_update_slots(admission_counter()) :: non_neg_integer()
+  def pending_update_slots(sample_counter) do
     :atomics.get(sample_counter, 1)
+  end
+
+  @spec release_update_slot(admission_counter()) :: :ok
+  defp release_update_slot(sample_counter) do
+    :atomics.sub(sample_counter, 1, 1)
+    :ok
   end
 
   defp configured_max_pending_samples do
@@ -150,7 +159,7 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
         {:update, %LogEvent{} = log_event, %Source{} = source, sample_counter},
         state
       ) do
-    :atomics.sub(sample_counter, 1, 1)
+    release_update_slot(sample_counter)
 
     SchemaMetrics.record_handled()
     handle_update(log_event, source, state)
--- a/lib/logflare/sources/source/bigquery/schema.ex
+++ b/lib/logflare/sources/source/bigquery/schema.ex
@@ -22,10 +22,15 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
   alias Logflare.TeamUsers
   alias Logflare.SingleTenant
 
-  @sample_counter_scale 1_000_000
+  @admission_value_tag :schema_admission
 
   def start_link(args) when is_list(args) do
     {name, args} = Keyword.pop(args, :name)
+    {max_pending_samples, args} = Keyword.pop(args, :max_pending_samples)
+
+    max_pending_samples = max_pending_samples || configured_max_pending_samples()
+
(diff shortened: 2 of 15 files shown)


ruslandoga:
👋 

Sorry for unsolicited comments!

One thing I considered was storing the counters in the Registry for the Schema process keys, since it's unique already. And "just" incrementing and decrementing that counter in a `try/after` block.

djwhitt:
> 👋
> 
> Sorry for unsolicited comments!
> 
> One thing I considered was storing the counters in the Registry for the Schema process keys, since it's unique already. And "just" incrementing and decrementing that counter in a `try/after` block.

Unsolicited comments appreciated! I probably won't circle back to this till the weekend (it's a bit of a side quest), but will consider this when I do.

ruslandoga:
I'll post a few code snippets to illustrate the idea better then!

djwhitt:
> I'll post a few code snippets to illustrate the idea better then!

Go for it. Feel free to share any other ideas you have on this as well! I'll probably try to keep the initial efforts fairly tightly scoped, but there are a bunch of things about this that need to be improved.

ruslandoga:
My snippets/ideas:

I think we can store the counter alongside the registered name, either in `start_link/1` when the name was provided or `init/1`

https://github.com/Logflare/logflare/blob/de7e9d47d81c2713c1cf9c52ae02b8d9291c9a3d/lib/logflare/sources/source/bigquery/schema.ex#L24-L32 could become
```elixir
{name, args} = Keyword.pop(args, :name)

name =
  case name do
    {:via, Registry, {registry, key}} ->
      queue_counter = :atomics.new(1, signed: false)
      {:via, Registry, {registry, key, queue_counter}}
      
    nil ->
      # should we allow it?
      nil
      
    _other ->
      raise ArgumentError, "expected the name to be a Registry name, etc. etc."
  end

GenServer.start_link(__MODULE__, args,
  name: name,
  hibernate_after: 5_000,
  spawn_opt: [fullsweep_after: 500]
)
```

and then on Schema update we could do something like this

https://github.com/Logflare/logflare/blob/de7e9d47d81c2713c1cf9c52ae02b8d9291c9a3d/lib/logflare/sources/source/bigquery/schema.ex#L34-L38
```elixir
@max_queue_size 32

defp reserve(counter) do
  if :atomics.add_get(counter, 1, 1) <= @max_queue_size do
    :ok
  else
    :atomics.sub(counter, 1, 1)
    :full
  end
end

# or we can call it something like update_gated
def update({:via, Registry, {registry, key}}, log_event, source) do
  with [{pid, counter}] <- Registry.lookup(registry, key), :ok <- reserve(counter) do
    GenServer.cast(pid, {:update, log_event, source, counter})
  else
    _ -> :ok
  end
end

# some tests use pids, I think, and they would need to be updated to use Registry-ed processes
def update(pid, _log_event, _source) when is_pid(pid) do
  raise ArgumentError, "... where we explain that Schema needs to be registered ..."
end
```

and then we would also need to decrement it inside the Schema process in both functions, but maybe then can be made one https://github.com/Logflare/logflare/blob/de7e9d47d81c2713c1cf9c52ae02b8d9291c9a3d/lib/logflare/sources/source/bigquery/schema.ex#L95-L125

```elixir
def handle_cast({:update, log_event, source, counter}, state) do
  try do
    handle_update(log_event, source, state)
  after
    :atomics.sub(counter, 1, 1)
  end
end
```

And I think this removes the need for restart-generation tracking and `:rest_for_one`.

djwhitt:
@ruslandoga I made some adjustments. If you get a chance, take another look and let me know what you think.

config/config.exs
```diff
@@ -36,9 +36,12 @@ config :logflare, :bigquery_pipeline, max_retries: 0
...
 config :logflare, :clickhouse_backend_adaptor,
   engine: "MergeTree",
-  pool_size: 3
+  pool_size: 3,
+  native_pool_size: 10
 
-config :logflare, Logflare.Sources.Source.BigQuery.Schema, updates_per_minute: 6
+config :logflare, Logflare.Sources.Source.BigQuery.Schema,
+  updates_per_minute: 6,
+  max_pending_samples: 8
```
ruslandoga, line 44: Since it's in config.exs and not runtime.exs, I wonder if there is benefit to actually hardcode it, and skip one ETS lookup?

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -31,20 +41,79 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
...
+      SchemaMetrics.record_admission(:admitted)
+      GenServer.cast(pid, {:update, log_event, source, sample_counter})
+    else
+      _reason -> SchemaMetrics.record_admission(:rejected)
+    end
+
+    :ok
+  end
+
   def update(pid, %LogEvent{} = log_event, %Source{} = source)
```
ruslandoga: I wonder if we really need this clause? It potentially allows for "un-gated" cast.

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -31,20 +41,79 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
...
+  end
+
+  @spec release_update_slot(admission_counter()) :: :ok
+  defp release_update_slot(sample_counter) do
+    :atomics.sub(sample_counter, 1, 1)
+    :ok
+  end
+
+  defp configured_max_pending_samples do
+    Application.get_env(:logflare, __MODULE__, [])
```
ruslandoga, line 95: This is the lookup I think we could skip, not that it's expensive, but just it's already sort-of hardcoded in config.exs.

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -31,20 +41,79 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
...
+         {:via, Registry, {registry, key}},
+         max_pending_samples
+       )
+       when is_integer(max_pending_samples) and max_pending_samples > 0 do
+    sample_counter = :atomics.new(1, signed: false)
+
+    {:via, Registry, {registry, key, {@admission_value_tag, sample_counter, max_pending_samples}}}
+  end
+
+  defp register_admission_counter(name, _max_pending_samples), do: name
```
ruslandoga: I'd raise here as well to avoid silently using un-gated casts.

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -87,61 +156,84 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
...
-        {:update, %LogEvent{}, _source},
-        %{field_count: fc, field_count_limit: limit} = state
-      )
-      when fc > limit,
-      do: {:noreply, state}
+    SchemaMetrics.record_handled()
+    handle_update(log_event, source, state)
+  end
+
+  def handle_cast({:update, %LogEvent{} = log_event, %Source{} = source}, state) do
```
ruslandoga: And I would personally remove un-gated handle_cast clauses too. It doesn't seem like we would actually be still using them in "production" code, only tests? Or maybe I am wrong, I'll need to check.

lib/logflare/sources/source/bigquery/schema.ex
```diff
@@ -87,61 +156,84 @@ defmodule Logflare.Sources.Source.BigQuery.Schema do
   end
 
   def handle_cast(
-        {:update, %LogEvent{}, %Source{lock_schema: true}},
+        {:update, %LogEvent{} = log_event, %Source{} = source, sample_counter},
         state
-      ),
-      do: {:noreply, state}
+      ) do
+    release_update_slot(sample_counter)
```
ruslandoga, line 169: I wonder what your thoughts are on decrementing the counter after handling the update? It's probably practically the same since the work is probably fast, but I find it semantically more meaningful, with `after` keyword in `try/after`

```elixir
try do
  SchemaMetrics.record_handled()
  handle_update(log_event, source, state)
after
  release_update_slot(sample_counter)
end
```

PR #3810, opened 2026-08-11
fix(syslog): fix possible socket leak in the pool

ruslandoga:
This fixes a Syslog socket leak when the borrower dies after opening and transferring a connection but before check-in. It's probably quite unlikely (why would a broadway worker just die during `Pool.send/2`) but probably still worth doing.


test/test_helper.exs
```diff
@@ -55,6 +55,7 @@ Mimic.copy(Logflare.Backends.Adaptor.ClickHouseAdaptor.CircuitBreaker)
 Mimic.copy(Logflare.Backends.Adaptor.ClickHouseAdaptor.ConnectionManager)
 Mimic.copy(Logflare.Backends.Adaptor.HttpBased.Client)
 Mimic.copy(Logflare.Backends.Adaptor.SlackAdaptor.Client)
+Mimic.copy(Logflare.Backends.Adaptor.SyslogAdaptor.Socket)
```
ruslandoga, line 58: I'm not sure how best to test it since we are not waiting for anything during `Socket.send/2` so we can't properly time the process exit.

Ziinc (approved):
testing logic looks fine to me 👍

PR #3813, opened 2026-08-11
fix(syslog): Reject port zero for syslog backends

ruslandoga:
Rejects syslog destination ports outside 1–65,535 and aligns the config type and form bounds with that range.

Port 0 is only meaningful when binding a local socket and cannot identify a remote syslog service.

Adds boundary coverage for valid and invalid ports.

Validated with `mix test test/logflare/backends/adaptor/syslog_adaptor_test.exs` (23 tests, 0 failures).


lib/logflare/backends/adaptor/syslog_adaptor.ex
```diff
@@ -80,7 +80,10 @@ defmodule Logflare.Backends.Adaptor.SyslogAdaptor do
   def validate_config(changeset) do
     changeset
     |> validate_required([:host, :port])
-    |> validate_inclusion(:port, 0..65_535)
+    |> validate_number(:port,
```
ruslandoga: I switched to `validate_inclusion/3` since `errors_on/2` test helper was failing on `Range.t` from `validate_inclusion/3`. I'll open a separate PR for that.

ruslandoga: https://github.com/Logflare/logflare/pull/3815

Ziinc (approved):

PR #3815, opened 2026-08-11
fix(test): fix changeset error interpolation test helper

ruslandoga:
## Summary

`Logflare.DataCase.errors_on/1` now interpolates only metadata referenced by `%{...}` placeholders, preventing inclusion errors from trying to stringify unrelated values such as ranges.
This preserves the existing error map and numeric placeholder behavior.
The Syslog config test now covers an invalid port through the real inclusion-validation path.

## Validation

- `mix format`
- `git diff --check`
- `mix test test/logflare/backends/adaptor/syslog_adaptor_test.exs` (22 tests, 0 failures)

## Further work

We might also want to update `LogflareWeb.Utils.stringify_changeset_errors/1` since it has similar logic to the prev `errors_on` version that tried to stringify all metadata values.

### DIFF
--- a/test/support/data_case.ex
+++ b/test/support/data_case.ex
@@ -112,8 +112,8 @@ defmodule Logflare.DataCase do
   """
   def errors_on(changeset) do
     Ecto.Changeset.traverse_errors(changeset, fn {message, opts} ->
-      Enum.reduce(opts, message, fn {key, value}, acc ->
-        String.replace(acc, "%{#{key}}", to_string(value))
+      Regex.replace(~r"%{(\w+)}", message, fn _, key ->
+        opts |> Keyword.get(String.to_existing_atom(key), key) |> to_string()
       end)
     end)
   end
--- a/test/logflare/backends/adaptor/syslog_adaptor_test.exs
+++ b/test/logflare/backends/adaptor/syslog_adaptor_test.exs
@@ -372,6 +372,13 @@ defmodule Logflare.Backends.Adaptor.SyslogAdaptorTest do
   end
 
   describe "config validation" do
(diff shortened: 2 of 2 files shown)


test/support/data_case.ex
```diff
@@ -112,8 +112,8 @@ defmodule Logflare.DataCase do
   """
   def errors_on(changeset) do
     Ecto.Changeset.traverse_errors(changeset, fn {message, opts} ->
-      Enum.reduce(opts, message, fn {key, value}, acc ->
-        String.replace(acc, "%{#{key}}", to_string(value))
+      Regex.replace(~r"%{(\w+)}", message, fn _, key ->
+        opts |> Keyword.get(String.to_existing_atom(key), key) |> to_string()
```
ruslandoga, line 116: Copied from a new `phx.new` project

https://github.com/phoenixframework/phoenix/blob/5506184f9180ef21cb99dbbcb08645329e0784a3/installer/templates/phx_ecto/data_case.ex.eex#L42-L56

ruslandoga:
We might also want to update `LogflareWeb.Utils.stringify_changeset_errors/1`

Ziinc (approved):

Ziinc:
@ruslandoga yes can we adjust to use the Utils function? sorry merged too fast before i saw the comment

ruslandoga:
@Ziinc sure, I'll open another PR to update the utils function and switch to it in the `errors_on` helper, but I think it still makes sense to keep that function even if it just delegates the logic to make the tests easier to read/write / as a convention.

ruslandoga:
Follow-up: https://github.com/Logflare/logflare/pull/3818

PR #3818, opened 2026-08-12
fix(changesets, forms): centralize changeset error interpolation

ruslandoga:
## Summary

Follow-up to #3815: centralize safe Ecto error interpolation in `LogflareWeb.Utils.changeset_errors/1`, converting only metadata referenced by message placeholders.
Both `stringify_changeset_errors` variants now consume the shared error map, which also fixes placeholder interpolation in `/2`, while `Logflare.DataCase.errors_on/1` remains the test-facing convention as a thin delegate.
Regression coverage includes unreferenced `Range` metadata and numeric interpolation.

## Validation

- `mix format --check-formatted`
- `git diff --check`
- `mix test test/logflare_web/utils_test.exs` (16 doctests, 17 tests, 0 failures)
- `mix test test/logflare/backends/adaptor/syslog_adaptor_test.exs` (22 tests, 0 failures)


test/logflare_web/utils_test.exs
```diff
@@ -210,55 +250,96 @@ defmodule LogflareWeb.UtilsTest do
...
+        })
+
+      assert Utils.stringify_changeset_errors(changeset, "Backend configuration is invalid") ==
+               "Backend configuration is invalid: name: has invalid format, " <>
+                 "should be at least 3 character(s); port: is invalid; " <>
+                 "endpoint: has invalid format"
     end
   end
 
+  defp example_config_changeset(overrides) do
```
ruslandoga, line 280: The idea behind this test helper is to try and get into some Logflare-like changeset validation error cases and make sure that it stringifies correctly.

ruslandoga:
The failure (query error vs backend error in a liveview test) seems to have already been fixed in https://github.com/Logflare/logflare/pull/3825

test/logflare_web/utils_test.exs
```diff
@@ -210,55 +250,96 @@ defmodule LogflareWeb.UtilsTest do
...
-        data: %{},
-        types: %{email: :string}
-      }
-
-      assert Utils.stringify_changeset_errors(changeset) ==
-               "email: can't be blank & is invalid format"
+      assert Utils.stringify_changeset_errors(changeset) == """
+             name: has invalid format & should be at least 3 character(s)
+             port: is invalid
+             query_timeout: must be greater than or equal to 100\
```
Ziinc, line 259: Trailing slash?

Ziinc: Don't see anywhere obvious that it would be coming out from. Odd....

ruslandoga: It removes the trailing newline in `"""`

```elixir
iex(1)> """
        a
        b
        """
"a\nb\n"
iex(2)> """
        a
        b\
        """
"a\nb"
```

Ziinc (approved):

PR #3821, opened 2026-08-12
fix: Multiple read clusters liveview component for ClickHouse backends

amokan:
Fixes the most pressing issues with the `LogflareWeb.Backends.ReadClusterUrlsComponent` LiveView component.

_Usability fixes coming after this merges_


djwhitt:
:robot:

> `test/logflare_web/live/backends/backends_live_test.exs:700` derives row order with `Map.to_list/1`, while `ReadClusterUrlsComponent` now sorts `read_only_urls` by label. Map enumeration order is not guaranteed. Sort the entries using the component’s ordering before selecting the first and last labels.

djwhitt:
:robot:

> Add follow-up browser coverage in `test/e2e/features/backends_test.exs` that edits an existing row and fills a newly added row before a component rerender, then verifies that the values remain. Include a failed-save path to cover value preservation across error rerenders. This is non-blocking for the current state-reset fix.

djwhitt (approved):

PR #3823, opened 2026-08-13
fix: incidentio dedup-key as a min 3 hour window

Ziinc:
Remove minutely deduplication key in incidentio adaptor
Fixes the issue where an alert constantly pages due to the changing minute

### DIFF
--- a/lib/logflare/backends/adaptor/incidentio_adaptor.ex
+++ b/lib/logflare/backends/adaptor/incidentio_adaptor.ex
@@ -30,6 +30,7 @@ defmodule Logflare.Backends.Adaptor.IncidentioAdaptor do
       |> Map.put(:source_url, url(~p"/alerts/#{alert_query.id}"))
       |> Map.put(:title, alert_query.name)
       |> Map.put(:description, alert_query.description)
+      |> Map.put(:alert_query_id, alert_query.id)
 
     updated_backend = %{backend | config: updated_config}
 
@@ -51,13 +52,14 @@ defmodule Logflare.Backends.Adaptor.IncidentioAdaptor do
       end)
 
     alert_name = Map.get(config, :title, "unknown")
+    alert_query_id = Map.get(config, :alert_query_id, "unknown")
     window = div(DateTime.to_unix(DateTime.utc_now()), 3 * 60 * 60)
 
     metadata = Map.get(config, :metadata, %{})
     merged_metadata = Map.merge(metadata, %{"data" => batch})
 
     %{
-      "deduplication_key" => "#{alert_name}-#{window}",
+      "deduplication_key" => "#{alert_query_id}-#{alert_name}-#{window}",
       "description" => Map.get(config, :description),
       "metadata" => merged_metadata,
       "source_url" => Map.get(config, :source_url),
--- a/lib/logflare/backends/adaptor/incidentio_adaptor.ex
+++ b/lib/logflare/backends/adaptor/incidentio_adaptor.ex
@@ -51,15 +51,15 @@ defmodule Logflare.Backends.Adaptor.IncidentioAdaptor do
         other -> other
       end)
 
-    alert_name = Map.get(config, :title, "unknown")
     alert_query_id = Map.get(config, :alert_query_id, "unknown")
+    # for simplicity, we use unix epoch for window alignment
     window = div(DateTime.to_unix(DateTime.utc_now()), 3 * 60 * 60)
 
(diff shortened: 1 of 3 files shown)


lib/logflare/backends/adaptor/incidentio_adaptor.ex
```diff
@@ -50,14 +51,15 @@ defmodule Logflare.Backends.Adaptor.IncidentioAdaptor do
         other -> other
       end)
 
-    now = DateTime.utc_now()
-    hash = :erlang.phash2(batch)
+    alert_name = Map.get(config, :title, "unknown")
+    alert_query_id = Map.get(config, :alert_query_id, "unknown")
+    window = div(DateTime.to_unix(DateTime.utc_now()), 3 * 60 * 60)
```
djwhitt, line 56: :robot:

> Dividing Unix time into fixed buckets does not guarantee three hours between new keys. Firings immediately before and after a UTC bucket boundary can generate different keys seconds apart. If the requirement is a minimum rolling three-hour interval, persist per-alert expiry state and rotate only after three elapsed hours. Otherwise, document this explicitly as an epoch-aligned three-hour bucket.

lib/logflare/backends/adaptor/incidentio_adaptor.ex
```diff
@@ -50,14 +51,15 @@ defmodule Logflare.Backends.Adaptor.IncidentioAdaptor do
...
 
     metadata = Map.get(config, :metadata, %{})
     merged_metadata = Map.merge(metadata, %{"data" => batch})
 
     %{
-      "deduplication_key" => "#{hash}-#{now.minute}",
+      "deduplication_key" => "#{alert_query_id}-#{alert_name}-#{window}",
```
djwhitt: :robot:

> The key now contains the unique alert-query ID, so including the mutable `alert_name` makes the deduplication identity unstable. Renaming a firing alert changes its key and creates another Incident.io alert within the same window. Use only the immutable query ID and the chosen window identity; retain the name as display data.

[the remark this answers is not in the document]
Ziinc: intentional

lib/logflare/backends/adaptor/incidentio_adaptor.ex
```diff
@@ -50,14 +51,15 @@ defmodule Logflare.Backends.Adaptor.IncidentioAdaptor do
         other -> other
       end)
 
-    now = DateTime.utc_now()
-    hash = :erlang.phash2(batch)
+    alert_query_id = Map.get(config, :alert_query_id, "unknown")
```
djwhitt, line 54: I think incident.io can be used as a rules backend too. In that case, I'm pretty sure this will cause all incident.io alerts sent via any incident.io backend to get the same ID.

Maybe something like this would work:

```elixir
deduplication_identity =
  case Map.fetch(config, :alert_query_id) do
    {:ok, alert_query_id} -> "alert-query:#{alert_query_id}"
    :error -> "backend:#{Map.fetch!(config, :source_url)}"
  end

# For simplicity, use Unix epoch for window alignment.
window = div(DateTime.to_unix(DateTime.utc_now()), 3 * 60 * 60)

metadata = Map.get(config, :metadata, %{})
merged_metadata = Map.merge(metadata, %{"data" => batch})

%{
  "deduplication_key" => "#{deduplication_identity}-#{window}",
```

Ziinc: addressed

djwhitt (approved):

lib/logflare/backends/adaptor/incidentio_adaptor.ex
```diff
@@ -50,14 +51,20 @@ defmodule Logflare.Backends.Adaptor.IncidentioAdaptor do
...
-    now = DateTime.utc_now()
-    hash = :erlang.phash2(batch)
+    deduplication_identity =
+      case Map.fetch(config, :alert_query_id) do
+        {:ok, alert_query_id} -> "alert-query:#{alert_query_id}"
+        :error -> "backend:#{Map.fetch!(config, :source_url)}"
+      end
+
+    # for simplicity, we use unix epoch for window alignment
+    window = div(DateTime.to_unix(DateTime.utc_now()), 3 * 60 * 60)
```
djwhitt: As discussed elsewhere, I wonder if 1 hour might be better since we often ack without resolving. I'm fine either way.

PR #3824, opened 2026-08-13
perf: improve consumer throughput

Baishan:
Removed a blocking path in the QueueProducer.
Added max in flight throttling.
Reduced batch sizes and increased concurrency to allow rule matching to spread across threads.
Added an exponential backoff rather than jumping to the 1 second interval if the queue is empty. I think this was causing a race when keeping up with production which then slowed the consumers down and they fall behind. 
Added section to the dev/dashboard to highlight when its throttling.

Also turning back on the spool for prod so we can do another canary test.


lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -217,9 +236,13 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
...
-  # Only when we have a current file and no prefetch already running.
-  defp maybe_start_prefetch(%{prefetch: nil, current: %{}} = state) do
+  # Starts a background Task to fetch the next message/file whenever nothing
+  # is already in flight — regardless of whether a file is currently being
+  # streamed (current: %{}) or not (current: nil, e.g. cold start or an idle
+  # queue). This is the only place safe_fetch_next runs; it never runs inline
+  # in this process, so a slow or long-polling queue_mod.receive call (SQS/
+  # PubSub) can never block handle_demand/2, handle_info/2, or :sys introspection.
+  defp maybe_start_prefetch(%{prefetch: nil} = state) do
     if over_limit?() do
```
djwhitt, line 328: I think we might need something to extend the visibility timeout on the message (SQS terminology, but I assume GCP has something equivalent). I think you extended the default for our GCP queue, but you could also add something that bumps it periodically in the code. Wouldn't consider that blocking for this PR, but seems worth considering in a follow up before we fully enable it in production.

lib/logflare/backends/spool/consumer_pipeline.ex
```diff
@@ -14,52 +14,69 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline do
...
+
   @spec start_link(keyword()) :: {:ok, pid()} | {:error, term()}
   def start_link(args) do
     {name, _args} = Keyword.pop!(args, :name)
 
     spool_config = Application.get_env(:logflare, :spool, [])
     bucket = Keyword.fetch!(spool_config, :bucket)
-    concurrency = Keyword.get(spool_config, :consumer_concurrency, 4)
-    batch_size = Keyword.get(spool_config, :consumer_batch_size, 5_000)
+    concurrency = System.schedulers_online()
```
djwhitt: Nit: how about keeping the config but making `System.schedulers_online()` the default.

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -217,9 +236,13 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
...
-  # Start a background Task to fetch the next file while we stream the current one.
-  # Only when we have a current file and no prefetch already running.
-  defp maybe_start_prefetch(%{prefetch: nil, current: %{}} = state) do
+  # Starts a background Task to fetch the next message/file whenever nothing
+  # is already in flight — regardless of whether a file is currently being
+  # streamed (current: %{}) or not (current: nil, e.g. cold start or an idle
+  # queue). This is the only place safe_fetch_next runs; it never runs inline
+  # in this process, so a slow or long-polling queue_mod.receive call (SQS/
+  # PubSub) can never block handle_demand/2, handle_info/2, or :sys introspection.
+  defp maybe_start_prefetch(%{prefetch: nil} = state) do
```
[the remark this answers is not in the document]
djwhitt: This seems fine since we trust the queue bucket, but also, are there any downsides to using `[:safe]`? If not, might as well. 🤷

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -248,7 +271,10 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
   defp emit_from_buffer(%{demand: 0} = state), do: {[], state}
 
   defp emit_from_buffer(state) do
-    {to_emit, remaining} = Enum.split(state.current.lines, state.demand)
+    to_take = min(state.demand, available_in_flight(state))
+    {to_emit, remaining} = Enum.split(state.current.lines, to_take)
+
+    if to_emit != [], do: :atomics.add(state.in_flight_ref, 1, length(to_emit))
```
djwhitt, line 359: Might be worth adding a regression test for this condition.

djwhitt (approved):

djwhitt:
:robot:

> Non-blocking allocation reductions:
>
> **`ConsumerPipeline.handle_batch/4`**
>
> `Enum.group_by/3` avoids the intermediate data list and groups each batch in one traversal:
>
> ```elixir
> messages
> |> Enum.group_by(
>   fn message -> record_source_id(message.data) end,
>   & &1.data
> )
> ```
>
> **`QueueProducer.register_sources/2`**
>
> Building the source ID set in one reduction avoids the intermediate lists and repeated traversals from `map`, `reject`, and `uniq`:
>
> ```elixir
> source_ids =
>   Enum.reduce(lines, MapSet.new(), fn line, source_ids ->
>     case record_source_id(line) do
>       nil -> source_ids
>       source_id -> MapSet.put(source_ids, source_id)
>     end
>   end)
>
> new_source_ids = MapSet.difference(source_ids, state.registered_sources)
>
> Enum.each(new_source_ids, &MemoryMonitor.register_source/1)
>
> %{
>   state
>   | registered_sources: MapSet.union(state.registered_sources, source_ids)
> }
> ```

test/logflare/backends/spool/consumer_pipeline/queue_producer_test.exs
```diff
@@ -328,15 +328,18 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducerTest do
...
-      # how many consumers are asking for demand.
+      # Empty results back off from @min_empty_backoff_ms (100ms) up to
+      # @max_empty_backoff_ms (1000ms), doubling each time: ~100, 200, 400,
+      # 800, 1000, 1000... so a healthy, de-duplicated producer sees on the
+      # order of 5-6 receive calls in a 2500ms window of sustained empty
+      # results, regardless of how many consumers are asking for demand. If
+      # every concurrent handle_demand call spawned its own poll chain (the
+      # pre-fix bug), 5 subscribers would each independently drive that same
+      # backoff schedule, and the call count would grow roughly with
+      # subscriber_count * expected_count instead of staying flat.
```
Ziinc, line 339: should move this to module doc of queue producer

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -80,45 +115,53 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
...
   end
 
   # Records the result and, if demand is waiting with nothing buffered, kicks
-  # an immediate poll rather than waiting for the periodic loop. Unlike
-  # handle_demand/2, this is a real "something just became available" signal
-  # (the background prefetch Task just completed), not a guess — worth
-  # reacting to right away so a producer/consumer running at roughly matched
-  # rates doesn't pay a needless @poll_interval stall at every file boundary
-  # while the next file's prefetch has already landed.
+  # a poll rather than waiting for the periodic loop — a real "something just
```
Ziinc, line 198: proper `@doc` or move to module doc

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -9,6 +9,20 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
 
   @poll_interval 1_000
   @throttle_interval 100
+  @min_empty_backoff_ms 100
+  @max_empty_backoff_ms 1_000
```
Ziinc: should add in some telemetry on usual backoff level, so that we know resting average.

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -9,6 +9,20 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
...
+  # How soon to retry once max_in_flight capacity frees up, instead of
+  # waiting out the full @poll_interval.
+  @min_in_flight_retry_ms 100
```
Ziinc: is this always the same as `@min_empty_backoff_ms`?

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -80,45 +115,53 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
         |> emit_from_buffer()
       end
 
-    {:noreply, events, schedule_poll(new_state, next_poll_delay())}
+    {:noreply, events, schedule_poll(new_state, @poll_interval)}
```
Ziinc: is this always the same as `@max_empty_backoff_ms`?

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -33,29 +48,49 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
...
+          emitted_state =
+            if capped_by_in_flight?(emitted_state),
+              do: schedule_poll(emitted_state, @min_in_flight_retry_ms),
+              else: emitted_state
 
-        true ->
+          {events, emitted_state}
+
+        match?({:ready, _}, new_state.prefetch) ->
           {[], schedule_poll(new_state, 0)}
```
Ziinc, line 166: should pattern match on state first before cond/do, will be faster.

lib/logflare/backends/spool/consumer_pipeline.ex
Ziinc: should set a min number for self hosters in case their systems are small.

lib/logflare/backends/spool/consumer_pipeline.ex
```diff
@@ -14,52 +14,69 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline do
...
+             bucket: bucket,
+             storage_mod: storage_mod,
+             queue_mod: queue_mod,
+             max_in_flight: max_in_flight
+           ]},
         transformer: {__MODULE__, :transform, []}
       ],
       processors: [
-        default: [concurrency: concurrency, max_demand: 100]
+        default: [concurrency: concurrency, min_demand: 50, max_demand: 100]
```
Ziinc: can also consider increasing max demand to reduce message passing and increase producer emission

lib/logflare/backends/spool/consumer_pipeline.ex
```diff
@@ -14,52 +14,69 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline do
...
   @spec start_link(keyword()) :: {:ok, pid()} | {:error, term()}
   def start_link(args) do
     {name, _args} = Keyword.pop!(args, :name)
 
     spool_config = Application.get_env(:logflare, :spool, [])
     bucket = Keyword.fetch!(spool_config, :bucket)
-    concurrency = Keyword.get(spool_config, :consumer_concurrency, 4)
-    batch_size = Keyword.get(spool_config, :consumer_batch_size, 5_000)
+    concurrency = System.schedulers_online()
+    batch_size = Keyword.get(spool_config, :consumer_batch_size, 100)
```
Ziinc: why not bump up consumer batch size? 100 events per queue insert is very small

lib/logflare/backends/spool/consumer_pipeline.ex
Ziinc, line 131: can consider partitining batcvher by source id so that we don't need to do this on each batch

Ziinc (approved):

lib/logflare/backends/spool/consumer_pipeline.ex
Baishan: Its somethign I want to test. I noticed edge_logs came in avg batches of 10-50 on prod and suspect having this at 5000 causes pauses with allocations. I will increase it if I can debunk that.

lib/logflare/backends/spool/consumer_pipeline.ex
Baishan: I was thinking of this and might do it. It might lead to low volume sources getting more latency, it also makes the max_inflight a little harder to calculate

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
Baishan: Good idea

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
Baishan: Not necessarily, I'm still trying to fine tune things

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
Baishan: Not necessarily but it might be, I'm just leaving them seperate in case I need to max_empty greater. Its certainly a trial an error phase

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
Baishan: 👍

lib/logflare/backends/spool/consumer_pipeline.ex
Baishan: Yeah I've tried that but want to try reducing batch size first then can increase it.

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
Baishan: I've extended it to 600 seconds already. Will need to find a sensible number once things are stable.

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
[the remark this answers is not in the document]
Baishan: I think the problem was that new atoms wont be created if the consumer is a seperate node to the producer so it failed to decode some things like source uuids that are atoms.

lib/logflare/backends/spool/consumer_pipeline.ex
Baishan: Done

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
Baishan: Added telemetry for these

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
Baishan: Cleaned these up with just a min_backoff/max_backoff. I can seperate them again if needed.

lib/logflare/backends/spool/consumer_pipeline.ex
Baishan: Increased it to 500 to try that.

PR #3826, opened 2026-08-13
fix: use DB_SCHEMA for Oban tables

### LINKED ISSUE #3251: Make Oban table creation configurable to use custom schema (avoid public schema pollution)
#### Problem
PR #3167 introduced Oban into Logflare (merged on 2026-03-02), and currently Oban creates its tables in the **public** database schema by default. This causes unnecessary schema pollution of the public namespace, which is undesirable for database organization and isolation of Logflare's analytics-related tables.

#### Expected Solution
Allow configuring the database schema for Oban tables via **environment variables** and/or direct config, with the intended target schema being `_analytics` for Logflare's analytics workloads.
1. Set the Oban prefix to `_analytics` by default for Logflare: `config :logflare, Oban, prefix: "_analytics"`
2. Add environment variable support for flexible schema control:
   - Primary: `LOGFLARE_OBAN_SCHEMA` (takes precedence if set)
   - Fallback: Reuse existing `DB_SCHEMA` if `LOGFLARE_OBAN_SCHEMA` is not provided

Thank you for considering this configuration improvement!

ElemTran:
## Summary

- configure Oban to use `DB_SCHEMA`, falling back to `public`
- create Oban structures in the configured schema for both fresh and existing installations
- preserve existing jobs and Oban objects in `public` during upgrades
- document the upgrade behavior and cover both migration paths with PostgreSQL integration tests

Fixes #3251

## Testing

- `mix test test/logflare/oban_schema_migration_test.exs --trace` (2 tests, 0 failures)
- verified the migration tests fail when the custom-schema prefix is removed
- `mix format --check-formatted`
- `mix credo`
- `git diff --check`
- verified default, fresh custom-schema, and existing-installation upgrade scenarios against PostgreSQL 15 and PostgreSQL 18.4


test/logflare/oban_schema_migration_test.exs
```diff
@@ -0,0 +1,114 @@
+defmodule Logflare.ObanSchemaMigrationTest do
```
Ziinc, line 1: test repo would already have migrations run against it, so this test module is redundant.

can instead do the following:
- change application config for db schema in test setup
- provision the schema and run the migration
- assert that the oban tables are created in the randomized schema.
- assert no public tables created.

priv/repo/migrations/20260814000000_add_oban_tables_to_configured_schema.exs
```diff
@@ -0,0 +1,8 @@
+defmodule Logflare.Repo.Migrations.AddObanTablesToConfiguredSchema do
+  use Ecto.Migration
+
+  def up, do: Oban.Migration.up(version: 12, prefix: oban_prefix())
+  def down, do: :ok
+
+  defp oban_prefix, do: System.get_env("DB_SCHEMA", "public")
```
Ziinc: should not use env var directly, should reference the db schem via application config with get_env

priv/repo/migrations/20260216120000_add_oban_jobs_table.exs
```diff
@@ -1,6 +1,8 @@
 defmodule Logflare.Repo.Migrations.AddObanJobsTable do
   use Ecto.Migration
 
-  def up, do: Oban.Migration.up(version: 12)
-  def down, do: Oban.Migration.down(version: 1)
+  def up, do: Oban.Migration.up(version: 12, prefix: oban_prefix())
+  def down, do: Oban.Migration.down(version: 1, prefix: oban_prefix())
+
+  defp oban_prefix, do: System.get_env("DB_SCHEMA", "public")
```
Ziinc: shoudl not use env var directly

config/runtime.exs
```diff
@@ -488,6 +488,7 @@ config :syn,
 enable_alerting? = Env.get_boolean("LOGFLARE_ALERTS_ENABLED", true)
 
 config :logflare, Oban,
+  prefix: System.get_env("DB_SCHEMA", "public"),
```
Ziinc: can bind as module variable and reference for both db config and oban config.

docs/docs.logflare.com/docs/self-hosting/index.md
```diff
@@ -36,7 +36,7 @@ All browser authentication will be disabled when in single-tenant mode.
...
-| `DB_SCHEMA`                                    | String, defaults to `nil`                                               | Allows configuration of the database schema to scope Logflare operations.                                                                                                                                                                                            |
+| `DB_SCHEMA`                                    | String, defaults to `nil`                                               | Allows configuration of the database schema to scope Logflare operations, including Oban jobs. Defaults to the `public` schema when unset. When upgrading an existing installation, Oban structures are created in the configured schema, but existing jobs and Oban objects in `public` are not copied or removed automatically. Stop all Logflare nodes before manually migrating jobs that must be preserved. |
```
Ziinc: We can move the upgrade note to a an Upgrading section with a version number  + date.

Ziinc (changes requested):

test/logflare/oban_schema_migration_test.exs
ElemTran: The CI setup runs `mix ecto.setup` before the test suite, so `public.oban_jobs` already exists when this test starts. For the "assert no public tables created" step, would you prefer that we snapshot and assert the existing public Oban tables/jobs remain unchanged, or temporarily remove and restore the public Oban structures around this test?

docs/docs.logflare.com/docs/self-hosting/index.md
ElemTran: Which release version should the new Upgrading section use? I can date the note `2026-08-14`, but I don't want to guess the target release version.

ElemTran:
Addressed the review feedback in commit `fa5cfdb7b`:

- Reused the existing `Logflare.Repo` in the integration tests and switched the test repo to `:auto` sandbox mode so migration tasks can check out connections.
- Asserted that existing `public` Oban tables and legacy jobs remain unchanged.
- Read the schema from the configured `Logflare.Repo` application config in both migrations.
- Read `DB_SCHEMA` once in `runtime.exs` and reused it for the Repo schema, search path, and Oban prefix.
- Moved the upgrade note into the versioned `Upgrading` section (`v1.50.2 — 2026-08-14`).

Validation: `2 tests, 0 failures`, `mix format --check-formatted`, `mix credo`, and `git diff --check` pass.

docs/docs.logflare.com/docs/self-hosting/index.md
```diff
@@ -57,6 +57,12 @@ All browser authentication will be disabled when in single-tenant mode.
 
 Additional environment variable configurations for the OpenTelemetry libraries used can be found [here](https://hexdocs.pm/opentelemetry_exporter/readme.html).perf/bq-pipeline-sharding
 
+### Upgrading
+
+#### v1.50.2 — 2026-08-14
```
Ziinc:
```suggestion
#### v1.51.0
```

test/logflare/oban_schema_migration_test.exs
Ziinc: i see, good point. this current method works as well, we can keep it as is 😄

Ziinc (approved):

Ziinc:
@ElemTran can you resolve the checks and also bump the `VERSION` to v1.51.0? 🙏 
much appreciated.

docs/docs.logflare.com/docs/self-hosting/index.md
Ziinc: i've removed the date as it is rather hard to predict our next release dates.

ElemTran:
@Ziinc Fixed the CI failures from the branch update: the merge with `main` introduced `20260814000000_add_enabled_to_backends.exs`, which collided with our `20260814000000` Oban schema migration (`Ecto.MigrationError: migration version 20260814000000 is duplicated`).

Renamed our migration to `20260831120000_add_oban_tables_to_configured_schema.exs` (`7b32c5eb`) so it sorts after all current migrations. Verified locally on PostgreSQL 15: full `ecto.migrate` chain completes without duplication errors, and `test/logflare/oban_schema_migration_test.exs` passes (`2 tests, 0 failures`).

The new CI runs are waiting for workflow approval.

Ziinc:
Will need to defer merging this in until the pglogical replication PR is merged. Thanks for patience

PR #3828, opened 2026-08-14
feat: single tenant mode with Clickhouse backend

msmithstubbs:
Added single-tenant ClickHouse backend support.

* Start with `make start.st.ch`
* Uses `CLICKHOUSE_BACKEND_URL` connection string
* Single tenant Clickhouse uses a special case backend with id 0. 
* Updated development doc

While the PR touches a lot of files, many of the changes are minor refinements, such as allowing `backend.id` to be zero or `not SingleTenant.postgres_backend?()` becoming `Backends.bigquery_default_backend?()`.

UI support (search, alters, etc) will be handled in later PRs.

Part of  O11Y-1222


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -1229,7 +1240,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
     end
   end
 
-  @spec maybe_start_query_connection_manager(pid() | nil, pos_integer(), String.t() | nil) ::
+  @spec maybe_start_query_connection_manager(pid() | nil, non_neg_integer(), String.t() | nil) ::
           :ok | {:error, term()}
   defp maybe_start_query_connection_manager(nil, backend_id, label)
        when is_pos_integer(backend_id) do
```
djwhitt: Looks like this guard needs to be changed to `non_neg_integer()` as well.

lib/logflare/networking.ex
```diff
@@ -3,17 +3,17 @@ defmodule Logflare.Networking do
...
   alias Logflare.Backends.Adaptor.DatadogAdaptor
-  alias Logflare.SingleTenant
+  alias Logflare.Backends
 
   def pools do
-    if SingleTenant.postgres_backend?() do
-      finch_pools(true)
+    if Backends.bigquery_default_backend?() do
+      bigquery_finch_pools() ++ grpc_pools()
     else
```
djwhitt, line 11: This branch is selected for single-tenant ClickHouse, but `base_finch_pools_with_default` doesn't start the pool needed for `WebhookAdaptor`.

djwhitt (changes requested):
Overall looks good, but I found a couple issues (see inline comments).

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
msmithstubbs: Fixed guard.

lib/logflare/networking.ex
msmithstubbs: Moved `FinchDefaultHttp1` to `base_finch_pools/0` and is now always started.

msmithstubbs:
Thanks for reviewing. Feedback addressed.

config/runtime.exs
```diff
@@ -306,15 +307,46 @@ socket_options_for_url = fn
     nil
 end
 
+single_tenant? = Env.get_boolean("LOGFLARE_SINGLE_TENANT")
+postgres_backend_url = System.get_env("POSTGRES_BACKEND_URL")
+clickhouse_backend_url = System.get_env("CLICKHOUSE_BACKEND_URL")
+
```
Ziinc, line 346: we should also error on multiple backend env vars being declared - user should only provide 1 configuration for the default system backend

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -512,8 +512,19 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
   @spec clickhouse_ingest_table_name(Backend.t(), TypeDetection.event_type()) :: String.t()
+  def clickhouse_ingest_table_name(%Backend{single_tenant_default?: true}, :log),
+    do: "otel_logs_single_tenant"
+
+  def clickhouse_ingest_table_name(%Backend{single_tenant_default?: true}, :metric),
+    do: "otel_metrics_single_tenant"
+
+  def clickhouse_ingest_table_name(%Backend{single_tenant_default?: true}, :trace),
+    do: "otel_traces_single_tenant"
+
```
Ziinc, line 650:
```suggestion
  Single-tenant default system table use `default` suffix in place of `<token>`
 """
@spec clickhouse_ingest_table_name(Backend.t(), TypeDetection.event_type()) :: String.t()
  def clickhouse_ingest_table_name(%Backend{single_tenant_default?: true}, :log),
    do: "otel_logs_default"

  def clickhouse_ingest_table_name(%Backend{single_tenant_default?: true}, :metric),
    do: "otel_metrics_default"

  def clickhouse_ingest_table_name(%Backend{single_tenant_default?: true}, :trace),
    do: "otel_traces_default"

```

lib/logflare/backends/backend.ex
```diff
@@ -42,6 +42,7 @@ defmodule Logflare.Backends.Backend do
     field :consolidated_ingest?, :boolean, virtual: true, default: false
     field :metadata, :map
     field :default_ingest?, :boolean, source: :default_ingest, default: false
+    field :single_tenant_default?, :boolean, virtual: true, default: false
```
Ziinc, line 47: hmm instead of storing this in the struct (which could add to memory footpright ever so slightly), can we just fetch it from application env when we need via a utility function?

lib/logflare/sources/source/supervisor.ex
```diff
@@ -161,6 +153,10 @@ defmodule Logflare.Sources.Source.Supervisor do
     end
   end
 
+  defp maybe_init_table(source_token) do
+    if Backends.bigquery_default_backend?(), do: init_table(source_token), else: :ok
+  end
```
Ziinc, line 158: i think we should implement this source setup as part of system backend source setup as an adaptor callback

lib/logflare/application.ex
```diff
@@ -235,7 +237,7 @@ defmodule Logflare.Application do
       SingleTenant.create_supabase_endpoints()
       SingleTenant.ensure_supabase_sources_started()
 
-      unless SingleTenant.postgres_backend?() do
+      if Backends.bigquery_default_backend?() do
```
Ziinc, line 240: as above, we should start moving this to an adaptor callback

Ziinc: i think we can have a different behaviour called SystemBackend that implements these additional callbacks

lib/logflare/backends.ex
```diff
@@ -35,15 +35,42 @@ defmodule Logflare.Backends do
...
   @spec max_buffer_queue_len() :: non_neg_integer()
   def max_buffer_queue_len, do: @max_pending_buffer_len_per_queue
 
+  @doc "Returns whether BigQuery is the deployment's default backend"
+  @spec bigquery_default_backend?() :: boolean()
+  def bigquery_default_backend?,
+    do: not SingleTenant.single_tenant?() or SingleTenant.bigquery_backend?()
+
+  @doc false
+  @spec ingest_queue_key(Source.t(), Backend.t() | nil) :: ingest_queue_key()
```
Ziinc, line 57: should add docs for when it should be used - for LLMs

lib/logflare/backends.ex
```diff
@@ -35,15 +35,42 @@ defmodule Logflare.Backends do
...
+    {:consolidated, backend_id}
+  end
+
+  def ingest_queue_key(%Source{id: source_id}, %Backend{id: backend_id}) do
+    {source_id, backend_id}
+  end
+
+  def ingest_queue_key(%Source{id: source_id}, nil), do: {source_id, nil}
+
+  @doc false
```
Ziinc, line 68: need docs

lib/logflare/backends.ex
```diff
@@ -208,29 +235,53 @@ defmodule Logflare.Backends do
...
+    user = SingleTenant.get_default_user()
+
+    if SingleTenant.clickhouse_backend?() && user,
+      do: single_tenant_clickhouse_backend(user)
+  end
+
+  defp single_tenant_clickhouse_backend(%User{} = user) do
+    %Backend{
+      id: @single_tenant_clickhouse_backend_id,
+      token: "00000000-0000-0000-0000-000000000000",
```
Ziinc, line 278: hmm i think we leave this blank for bigquery/postgres as well

i think these functions should go to the SingleTenant module

lib/logflare/backends.ex
```diff
@@ -795,28 +848,25 @@ defmodule Logflare.Backends do
   defp dispatch_to_backends(source, nil, log_events) do
     backends = __MODULE__.Cache.list_backends(source_id: source.id)
 
-    for backend <- [nil | backends] do
+    for backend <- [lookup_system_default_backend(source) | backends] do
       dispatch_to_default_backend(source, backend, log_events)
     end
   end
 
+  defp lookup_system_default_backend(%Source{user_id: user_id}) do
```
Ziinc, line 884: as above, should be in the SingleTenatn module instead of making the Backend module fatter.

lib/logflare/networking.ex
```diff
@@ -79,6 +77,8 @@ defmodule Logflare.Networking do
     base = System.schedulers_online()
 
     [
+      {Finch,
+       name: Logflare.FinchDefaultHttp1, pools: %{default: [protocols: [:http1], size: 50]}},
```
Ziinc, line 84: Perhaps we have separate function to dynamically add children for each backend selected, otherwise if using postgres backend this pool is unused

Ziinc (changes requested):

lib/logflare/single_tenant.ex
```diff
@@ -377,26 +419,29 @@ defmodule Logflare.SingleTenant do
   end
 
   @doc """
-  Returns true if postgres backend is enabled for single tenant
+  Returns the selected single-tenant backend type.
   """
+  @spec backend_type :: backend_type()
+  def backend_type do
+    Application.get_env(:logflare, :single_tenant_backend) || :bigquery
```
Ziinc, line 426: I think we should always set this at runtime.exs otherwise having a default here can get confusing

lib/logflare/single_tenant.ex
```diff
@@ -377,26 +419,29 @@ defmodule Logflare.SingleTenant do
...
-  @spec backend_type :: :postgres | :bigquery
-  def backend_type do
-    if postgres_backend?() do
-      :postgres
-    else
-      :bigquery
-    end
+  @doc "Returns true if ClickHouse is selected in single-tenant mode"
+  @spec clickhouse_backend? :: boolean()
+  def clickhouse_backend? do
```
Ziinc, line 443: Nit: can be single line with do:

test/logflare/backends/adaptor/clickhouse_adaptor/single_tenant_ingestion_test.exs
```diff
@@ -0,0 +1,99 @@
...
+    on_exit(fn ->
+      cleanup_single_tenant_tables(backend)
+
+      if Process.alive?(supervisor_pid) do
+        Process.exit(supervisor_pid, :shutdown)
+      end
+
+      if Process.alive?(connection_manager_pid) do
+        Process.exit(connection_manager_pid, :shutdown)
+      end
```
Ziinc, line 46: start_link_supervised should be much cleaner way for these to be handled instead of explicit exit

test/logflare/backends/adaptor/clickhouse_adaptor/single_tenant_ingestion_test.exs
```diff
@@ -0,0 +1,99 @@
...
+      assert {:ok, {[%{"count" => 1}], _bytes}} =
+               ClickHouseAdaptor.execute_ch_query(
+                 backend,
+                 "SELECT count(*) AS count FROM #{table_name} WHERE event_message = {message:String}",
+                 %{message: message}
+               )
+    end)
+  end
+
+  defp cleanup_single_tenant_tables(backend) do
```
Ziinc, line 75: Perhaps moving these to TestUtils and sharing them with Clickhouse adaptor tests would be good as a follow up PR

Ziinc: Generally the current setup/cleanup code for Clickhouse is quite verbose, anything to trim stuff down is great

test/logflare/backends/adaptor/clickhouse_adaptor_test.exs
```diff
@@ -57,6 +57,19 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptorTest do
       assert ClickHouseAdaptor.clickhouse_ingest_table_name(backend, :trace) ==
                "otel_traces_#{stringified_backend_token}"
     end
+
+    test "uses fixed table names for the synthetic single-tenant backend", %{backend: backend} do
+      backend = %{backend | id: 0, single_tenant_default?: true}
```
Ziinc, line 62: I think i mentioned above but I think having a hard coded static name for each table might bite us in the future as this isn't the same behaviour for other databases

Ziinc (changes requested):

config/runtime.exs
msmithstubbs: Now raises if more than one of `bigquery_service_account`, `postgres_backend_url`, `clickhouse_backend_url` set.

test/logflare/backends/adaptor/clickhouse_adaptor/single_tenant_ingestion_test.exs
msmithstubbs: O11Y-2518

lib/logflare/networking.ex
msmithstubbs: This refers to Logflare.FinchDefaultHttp1 ?

It is used by WebhookAdaptor and friends. Or should we dynamically manage them as users configure backends?

test/logflare/backends/adaptor/clickhouse_adaptor_test.exs
msmithstubbs: Table name suffix can be customised with LOGFLARE_CLICKHOUSE_TABLE_SUFFIX

lib/logflare/backends.ex
msmithstubbs: Moved to SingleTenant

lib/logflare/application.ex
msmithstubbs: Added SystemBackend

lib/logflare/backends/system_backend.ex
```diff
@@ -0,0 +1,45 @@
...
+  @spec on_supabase_start() :: :ok
+  def on_supabase_start, do: invoke(:on_supabase_start, [])
+
+  @spec invoke(atom(), list()) :: :ok
+  defp invoke(callback, args) do
+    type = if SingleTenant.single_tenant?(), do: SingleTenant.backend_type(), else: :bigquery
+    adaptor = Adaptor.get_adaptor(%Backend{type: type})
+    Code.ensure_loaded!(adaptor)
+
+    if function_exported?(adaptor, callback, length(args)) do
```
msmithstubbs, line 39: This matches implementation in other modules. Alternatively this could be macro with defoverridable.

lib/logflare/networking.ex
Ziinc: > It is used by WebhookAdaptor and friends. Or should we dynamically manage them as users configure backends?
i see, lets keep the default pool for now then, ideally we want to have each backend adaptor having their own pool and scaling it dynamically

msmithstubbs: OK, I'll leave the existing pool functions as they are in this branch: one list for BigQuery, another for Clickhouse and postgres. Kudos to @djwhitt for flagging this in the first review.

Created O11Y-2544 for dynamic pools per system backend.

msmithstubbs:
Addressed all feedback.

PR #3830, opened 2026-08-14
feat(toggle): Add UI/API toggle controls

ruslandoga:
PR 4/4 for [O11Y-2315](https://linear.app/supabase/issue/O11Y-2315/add-destination-level-toggle-controls), following the runtime-only release in [#3900](https://github.com/Logflare/logflare/pull/3900).

## Summary

- Adds destination enable/disable controls to the backend API and Backends LiveView, including acting-team-member authorization.
- Stops admitting new destination copies at attached/default fan-out, rule and explicit routing, buffer limiting, and Incident.io alert dispatch; existing processes stay alive so queued or in-flight work can drain.
- Reuses the existing unfiltered backend-list cache key and filters in memory, preserves the existing replica-backed cache/WAL propagation, and skips source/pipeline restarts for enabled-only updates; no new cache protocol or cluster RPC is introduced.

## Consistency

The toggle follows the existing WAL/cache eventual-consistency contract: cache reads already in flight during invalidation may temporarily retain prior state, and queued or in-flight deliveries may finish.

> [!WARNING]
> 
> Backend cache misses deliberately retain the existing replica-backed read path. After invalidation, a lagging replica can refill the prior enabled state and Cachex may retain it for approximately the backend-cache TTL (~20 minutes or longer with cache gossip ttl resets), extending convergence beyond the underlying replica lag. If tighter propagation becomes necessary, primary-backed Backend cache misses would prevent that replica-lag amplification at the cost of additional primary traffic and a special read path.

## Rollout

1. The migration is merged in [#3855](https://github.com/Logflare/logflare/pull/3855) and should be deployed first.
2. Deploy the schema-only release from [#3856](https://github.com/Logflare/logflare/pull/3856) to every node.
3. Deploy the runtime routing release from [#3900](https://github.com/Logflare/logflare/pull/3900) to every node while toggle writes and controls remain unavailable.
4. Deploy this PR to expose the API and LiveView toggle writes only after every node runs the routing-aware release; re-enable disabled destinations before rolling back below it.

## Related fixes and follow-ups

- [#3864](https://github.com/Logflare/logflare/pull/3864) removed the unreachable Cachex deletion key and is already merged.
- [#3865](https://github.com/Logflare/logflare/pull/3865) fixes exact child selection in Supervisor.which_children/1 and remains independently reviewable.
- (not implemented) Backend PATCH already accepts partial updates in the controller, including an `enabled`-only toggle, but its shared OpenAPI request schema still marks the create-only `name`, `type`, and `config` fields as required. Splitting the create and update schemas is intentionally left to a follow-up PR.

## Validation

All affected full test files pass across backend routing/cache, alerting, buffer limiting, API/OpenAPI, and LiveView coverage. mix format --check-formatted, mix compile --warnings-as-errors, and mix credo pass; Credo reports only existing design suggestions.

Linear: [O11Y-2315](https://linear.app/supabase/issue/O11Y-2315/add-destination-level-toggle-controls)


ruslandoga:
### Deployment notes

Keeping this PR as the end-state reference; production should roll out in three phases:

1. **Migration only:** add `enabled BOOLEAN NOT NULL DEFAULT true`. PostgreSQL 11+ uses the constant-default fast path, though the statement still needs a brief exclusive lock.
2. **Dormant logic:** deploy the schema, routing/supervisor guards, reconciliation, and cache handling everywhere, but do not accept API writes or show the UI toggle. Mixed-version code must preserve/version the cluster RPC contract and treat cached backend structs without `enabled` as enabled.
3. **Activation:** expose API and LiveView writes only after every node runs the logic release.

The explicit `enabled: true` argument creates a new cache-key variant alongside the legacy key. Coexistence is safe but causes temporary cold misses and duplicate entries; keep eviction for both variants. Backend-ID busting clears non-empty lists, while exact per-source deletion is required for cached empty lists when changing `false → true`. Optionally wait one backend-cache TTL (~20 minutes) after the logic rollout; rollback only to a logic-aware release while any destinations are disabled.

ruslandoga:
## Scope checkpoint

This PR has grown because it moved from “add a destination toggle” to “guarantee cluster-wide, race-free lifecycle transitions under replica lag, cache gossip, WAL delivery, and process failures.”

The product contract determines the appropriate scope:

- **Simple routing toggle:** disabled destinations are excluded from new routing. Existing backend processes may remain alive but idle.
- **Strong pause/resume:** disabling immediately stops delivery everywhere, while enabling cannot expose a destination until every required pipeline is ready.

The current implementation has been converging on the second contract. That is what pulled in cache generations, tombstones, WAL ordering, RPC reconciliation, locks, retries, and supervisor lifecycle changes.

If the simple contract is sufficient, the PR can be reduced substantially:

- Keep the migration, schema, API/OpenAPI, and UI.
- Filter disabled destinations at the actual routing and dispatch boundaries.
- Leave backend children and consolidated pipelines running but idle.
- Remove synchronous lifecycle reconciliation, cluster RPC orchestration, worker rewrites, and most cache-protocol changes.
- Keep #3864 and #3865 as separate fixes.
- Document that propagation follows the existing cache/WAL semantics.

If an immediate, hard cluster-wide cutoff is required, the current complexity is real, but it should be presented and split as distributed lifecycle infrastructure rather than treated as part of a small UI toggle.

No pruning decision has been applied yet; this comment records the scope decision we should make before continuing.

ruslandoga:
This PR now implements the smaller admission-control contract discussed above.

ruslandoga:
@codex review

PR #3831, opened 2026-08-14
ci: isolate Docker build caches

djwhitt:
## Summary

- isolate GitHub Actions cache objects by image role and CPU architecture
- cancel superseded Docker image builds for the same pushed ref
- preserve every main SHA image while keeping test-, benchmark-, and documentation-only rebuilds cacheable
- serialize deployment targets, retain queued versioned releases, and skip stale main deployment jobs
- exclude non-production and local build artifacts from the Docker context
- run the Docker build check when `.dockerignore` or Cargo workspace manifests change

## Motivation

Docker's GHA cache backend defaults every build to the `buildkit` scope, where each exporter overwrites the previous cache. The application, base, runner, PR-check, amd64, and arm64 builds currently share that default scope.

In main run 31831544506, the arm64 Buildx step took 13m47s, including 5m30s recompiling Mix dependencies and 2m57s exporting the cache. That run followed a test-only PR, but tests were still included in the Docker context and invalidated `COPY . ./`.

## Validation

- `actionlint` passes for the changed workflows after excluding pre-existing legacy-action/custom-runner diagnostics and the pinned 1.7.12 schema's missing support for GitHub's documented `concurrency.queue`
- a local BuildKit context fixture confirms root Markdown, docs, benchmarks, and tests are excluded while production source and `VERSION` remain included
- Docker Build Check passed ([run 32084439455](https://github.com/Logflare/logflare/actions/runs/32084439455))
- fresh independent review found no remaining actionable deployment, workflow, or context-exclusion issues

## Stack

1. **This PR:** cache isolation, build cancellation, deployment ordering, and context filtering
2. #3832: cache native Rust release compilation in a dedicated Docker layer


.github/workflows/docker-ci-amd-and-arm.yml
```diff
@@ -4,10 +4,18 @@ on:
   workflow_dispatch:
   push:
     branches: [main]
+    paths-ignore:
```
amokan: 🤖 

> This is the only workflow that publishes `supabase/logflare:dev-<sha>`, and the release path resolves that tag from `git HEAD: Makefile:8` sets `SHA_IMAGE_TAG ?= dev-$(shell git rev-parse --short=7 HEAD)`, and `docker-retag-versioned.yml` → `make tag-versioned` (`Makefile:394`) does `docker buildx imagetools create -t supabase/logflare:$(VERSION) ... supabase/logflare:$(SHA_IMAGE_TAG)`. Scenario: `VERSION` is bumped, then a test-only or `README.md`-only PR merges to main. That push is skipped, no `dev-<sha>` is published, and the release retag fails with a bare "manifest unknown". The pre-existing guard at `Makefile:388-390` won't catch it — the line continuation after echo "Dev image must be built on CI: ..." makes `docker manifest inspect ...` an argument to echo, so it always prints "OK".
Options: (a) add a cheap job that retags the previous `dev-<sha>` onto the new SHA when the build is skipped; (b) have `tag-versioned` walk back through main's history to the newest existing `dev-<sha>`; (c) at minimum fix the `Makefile:389` guard so the failure is loud and early. Recommend (a) — it preserves the "every main SHA has an image" invariant that both release and staging deploy assume.

.github/workflows/docker-ci-amd-and-arm.yml
```diff
@@ -4,10 +4,18 @@ on:
...
+      - "test/**"
+      - "bench/**"
+      - "*.md"
 
 permissions:
   contents: read
 
+concurrency:
```
amokan: 🤖 

> Group key is `workflow + ref` with no `event_name`, and this workflow calls `trigger-cloudbuild.yml`, whose `workflow_dispatch` branch runs `make deploy.prod.versioned` and `deploy.staging.versioned` (_`trigger-cloudbuild.yml:48`, matrix `[prod, staging]`; `github.event_name` in a called workflow is the caller's event_). Scenario: someone manually dispatches Docker CI to cut a prod deploy; a routine merge to main lands, and that push run (`cancel-in-progress: true`) cancels the in-flight prod deploy. The same key also cancels pending runs regardless of the flag. On plain push-vs-push, `deploy.staging.main` (`Makefile:286-304`) is three sequential gcloud builds submit calls — cancelling mid-way leaves the image built but instance groups un-updated, and the server-side Cloud Build keeps running after the runner dies, so SHA A's deploy can land after SHA B's.
Options: (a) move concurrency to the build job only, so publish/deploy jobs are never cancellable (_recommended — the savings are all in the 13-minute buildx step_); (b) keep workflow-level but add `-${{ github.event_name }}` to the group; (c) drop cancel-in-progress.

.github/workflows/docker-test.yml
```diff
@@ -6,6 +6,8 @@ on:
     branches: [main]
     paths:
       - 'native/**'
+      - 'Cargo.toml'
```
amokan, line 9: Non-blocking 🤖 

> `Cargo.toml/Cargo.lock` were added, but `mix.exs`, `mix.lock`, `VERSION` are still missing, even though `Dockerfile.multi-step:6-10` copies all three. A dependency-bump PR — the most likely way to break the prod image build — still skips the Docker Build Check.

.github/workflows/docker-test.yml
```diff
@@ -14,6 +16,7 @@ on:
       - 'Dockerfile.multi-step'
       - 'Dockerfile.base'
       - 'Dockerfile.runner'
+      - '.dockerignore'
       - '!assets/**'
```
amokan: Non-blocking 🤖 

>  `'!assets/**'` sits after - `'assets/**'` (_line 14_) and cancels it, so asset-only PRs never run the build check despite `npm run --prefix assets deploy` running in the image build. Pre-existing, but the PR edits this exact list

amokan (changes requested):
Two things worth investigating and a couple non-blockers

.github/workflows/docker-ci-amd-and-arm.yml
djwhitt: :robot:

> Addressed in fd23375aa825

.github/workflows/docker-ci-amd-and-arm.yml
djwhitt: :robot:

> Addressed in fa0ce2efcbd8

.github/workflows/docker-test.yml
djwhitt: :robot:

> Addressed in 750801725b8f

.github/workflows/docker-test.yml
djwhitt: :robot:

> Addressed in 750801725b8f

.github/workflows/trigger-cloudbuild.yml
```diff
@@ -13,6 +13,9 @@ jobs:
     strategy:
       matrix:
         env: [prod, staging]
+    concurrency:
```
amokan, line 17: 🤖 

> The new `cloudbuild-deploy-${{ matrix.env }}` group is global (_no ref, no caller_), and the prod matrix job is created on push too — both run steps are skipped, but the job still checks out, auths against `GCP_PROD_CREDENTIALS`, and holds the `cloudbuild-deploy-prod` slot for ~1 min. GitHub cancels the pending member of a group when a new one joins, even with `cancel-in-progress: false`. Scenario: a release runs `docker-retag-versioned.yml` → `trigger_cloudbuild` (_`workflow_dispatch` → `make deploy.prod.versioned`_), it queues behind a push's no-op prod job, the next merge to main enqueues its own no-op prod job, and the pending release prod deploy is cancelled. Same class as the original finding, relocated.
Options: (a) add `if: github.event_name == 'workflow_dispatch' || matrix.env == 'staging'` to the job — skipped jobs don't occupy a concurrency group, so push runs never touch `cloudbuild-deploy-prod` (_recommended; also stops burning prod credentials on every merge_); (b) `add -${{ github.event_name }}` to the group; (c) split prod/staging into separate jobs.

amokan (changes requested):
One new issue - but fix seems pretty small

amokan (approved):

PR #3837, opened 2026-08-15
perf(clickhouse): encode RowBinary rows in processors

djwhitt:
## Summary

- move fused mapping and ClickHouse RowBinary encoding from the batch callback into runtime-scaled Broadway processors
- replace claimed generation-store events with shared `EncodedRow` values so retries reuse encoded bytes without remapping
- leave batch processors responsible only for gzip and insertion, reducing concurrent inserts from 32 to 4
- independently retain the previous 64-batch, 3.84-million-message per-backend in-flight capacity
- add repeatable throughput, concurrency, phase-decomposition, and memory benchmarks

## Stack

This PR is stacked on #3759 (`perf/mapper-rowbinary-fusion`). Review it against that branch; it contains only processor-side changes.

## Performance

Linux aarch64, OTP 27, six online schedulers, production-shaped 60,000-row batches. Throughput results cover three measured batches after one warmup and validate byte-identical output first.

### End-to-end comparison with current main

Current `main` at `ee61fab1` was measured in a separate clean workspace on the same host. It uses the production-equivalent ETS lookup → `Mapper.map/3` → Elixir RowBinary → streaming gzip path. The same fixtures and fixed-work settings were used, and compressed output byte counts matched exactly.

| Event type | Current main | Fused batch-side (#3759) | Processor-side (#3837) | #3837 elapsed change vs main |
|---|---:|---:|---:|---:|
| Log | 75,520 rows/s | 78,693 rows/s | 168,077 rows/s | -55% |
| Metric | 37,309 rows/s | 49,452 rows/s | 102,807 rows/s | -64% |
| Trace | 38,807 rows/s | 66,893 rows/s | 137,989 rows/s | -72% |

### Incremental effect of #3837 on #3759

| Event type | Fused batch-side (#3759) | Processor-side (#3837) | Change | Gzip only |
|---|---:|---:|---:|---:|
| Log | 78,693 rows/s | 168,077 rows/s | -53% elapsed | 230,489 rows/s |
| Metric | 49,452 rows/s | 102,807 rows/s | -52% elapsed | 178,773 rows/s |
| Trace | 66,893 rows/s | 137,989 rows/s | -52% elapsed | 219,139 rows/s |

Six-processor encoding-only throughput was 379k log, 230k metric, and 334k trace rows/s. Gzip is therefore the remaining serial local bottleneck.

### Processor concurrency

| Event type | 4 processors | 6 processors | 8 processors | 12 processors |
|---|---:|---:|---:|---:|
| Log | 158,534 | 170,951 | 171,329 | 174,227 |
| Metric | 99,290 | 107,397 | 106,749 | 105,733 |
| Trace | 120,384 | 135,506 | 136,458 | 138,754 |

Six processors remain the balanced setting on the six-scheduler benchmark host. Production concurrency is calculated at pipeline startup as `max(System.schedulers_online() - 4, 6)`, leaving nominal capacity for four fixed gzip/HTTP batch processors. This gives 6 processors on 6 schedulers, 8 on 12, and 28 on 32. Concurrency remains per backend pipeline, so multiple active ClickHouse backends multiply the count.

### Steady-state VM memory

| Event type | Full-event state | Encoded-row state | Change |
|---|---:|---:|---:|
| Log | 328.2 MB | 229.3 MB | -30.1% |
| Metric | 472.4 MB | 275.0 MB | -41.8% |
| Trace | 402.8 MB | 249.6 MB | -38.0% |

The ETS value and Broadway message share the same reference-counted RowBinary payload. Maximum RSS remained within 1.6% because the benchmark also observes the transient event-to-binary handoff.

Four batch processors bound concurrent gzip/HTTP inserts. During downstream stalls, encoded partial and completed batches can accumulate in Broadway's `:ch` batcher up to the independent 3.84-million-message cap; additional backlog remains in `IngestEventQueue`.

Detailed commands and results are recorded in `bench/clickhouse_processor_rowbinary.md`.

## Validation

- pipeline, `IngestEventQueue`, and consolidated supervisor: 146 tests, 0 failures, 3 excluded
- formatter check passed
- processor output matched the batch-side baseline byte-for-byte for every benchmark run

### DIFF
--- a/bench/clickhouse_processor_rowbinary.exs
+++ b/bench/clickhouse_processor_rowbinary.exs
@@ -1,5 +1,6 @@
 # Fixed-work benchmark for moving fused ClickHouse mapping/RowBinary production
-# from Broadway batch processors into processors.
+# from Broadway batch processors into processors. It excludes Broadway message
+# handling, generation replacement, HTTP, and ClickHouse execution.
 #
 # Examples:
 #
--- a/bench/clickhouse_processor_rowbinary.exs
+++ b/bench/clickhouse_processor_rowbinary.exs
@@ -1,0 +1,185 @@
+# Fixed-work benchmark for moving fused ClickHouse mapping/RowBinary production
+# from Broadway batch processors into processors.
+#
+# Examples:
+#
+#   MIX_ENV=test SCENARIO=baseline EVENT_TYPE=log BATCH_SIZE=10000 BATCHES=20 \
+#     mix run --no-start bench/clickhouse_processor_rowbinary.exs
+#
+#   MIX_ENV=test SCENARIO=processor CONCURRENCY=6 EVENT_TYPE=log \
+#     BATCH_SIZE=10000 BATCHES=20 \
+#     mix run --no-start bench/clickhouse_processor_rowbinary.exs
+#
+# Scenarios:
+#   baseline  ETS lookup -> fused mapper/RowBinary -> streaming gzip, serially
+#   processor ETS lookup -> fused mapper/RowBinary in parallel processor chunks,
+#             then streaming gzip serially
+#   encode    processor-side lookup and fused mapper/RowBinary only
+#   compress  gzip already-encoded rows only
+
+Code.require_file("support/clickhouse_pipeline_bench_data.exs", __DIR__)
+
+defmodule Logflare.Bench.ClickHouseProcessorRowBinary do
+  @moduledoc false
+
+  alias Logflare.Backends.IngestEventQueue
+  alias Logflare.LogEvent
+  alias Logflare.Mapper
+  alias Logflare.Mapper.OutputContext
+
+  @spec encode_rows(
+          [{term(), :ets.tid()}],
+          reference(),
+          binary(),
+          pos_integer(),
+          pos_integer()
+        ) :: [binary()]
+  def encode_rows(id_tid_pairs, compiled, mapping_config_id, concurrency, processor_chunk_size) do
+    id_tid_pairs
+    |> Enum.chunk_every(processor_chunk_size)
+    |> Task.async_stream(
+      fn chunk -> Enum.map(chunk, &encode_pointer(&1, compiled, mapping_config_id)) end,
+      max_concurrency: concurrency,
+      ordered: true,
+      timeout: :infinity
+    )
+    |> Enum.flat_map(fn {:ok, rows} -> rows end)
+  end
+
+  @spec encode_rows_serial([{term(), :ets.tid()}], reference(), binary()) :: [binary()]
+  def encode_rows_serial(id_tid_pairs, compiled, mapping_config_id) do
+    Enum.map(id_tid_pairs, &encode_pointer(&1, compiled, mapping_config_id))
+  end
+
+  @spec compress_rows([binary()]) :: binary()
+  def compress_rows(rows) do
+    gzip_stream(fn z -> Enum.map(rows, &:zlib.deflate(z, &1)) end)
+  end
+
+  @spec baseline([{term(), :ets.tid()}], reference(), binary()) :: binary()
+  def baseline(id_tid_pairs, compiled, mapping_config_id) do
+    gzip_stream(fn z ->
+      Enum.map(id_tid_pairs, fn pair ->
+        :zlib.deflate(z, encode_pointer(pair, compiled, mapping_config_id))
+      end)
+    end)
+  end
+
+  defp encode_pointer({id, tid}, compiled, mapping_config_id) do
+    %LogEvent{} = event = IngestEventQueue.lookup_event(tid, id)
+    output_context = OutputContext.clickhouse_row_binary(event, mapping_config_id)
+    Mapper.map(event.body, compiled, output_context: output_context)
+  end
+
+  defp gzip_stream(fun) do
+    z = :zlib.open()
+
+    try do
+      :zlib.deflateInit(z, :default, :deflated, 31, 8, :default)
+      chunks = fun.(z)
+      IO.iodata_to_binary([chunks, :zlib.deflate(z, "", :finish)])
+    after
+      :zlib.deflateEnd(z)
+      :zlib.close(z)
+    end
+  end
+end
+
+alias Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester
+alias Logflare.Bench.ClickHousePipelineData, as: Data
+alias Logflare.Bench.ClickHouseProcessorRowBinary, as: Bench
+
+scenario = System.get_env("SCENARIO", "processor") |> String.to_existing_atom()
+type = System.get_env("EVENT_TYPE", "log") |> String.to_existing_atom()
+batch_size = System.get_env("BATCH_SIZE", "10000") |> String.to_integer()
+batches = System.get_env("BATCHES", "20") |> String.to_integer()
+warmup_batches = System.get_env("WARMUP_BATCHES", "2") |> String.to_integer()
+concurrency = System.get_env("CONCURRENCY", "6") |> String.to_integer()
+processor_chunk_size = System.get_env("PROCESSOR_CHUNK_SIZE", "1000") |> String.to_integer()
+
+{compiled, config_id} = Data.compiled_output(type)
+mapping_config_id = Ingester.encode_mapping_config_id(config_id)
+events = Data.batch(type, batch_size, :realistic)
+tid = Data.setup_processing_ets(events)
+id_tid_pairs = Enum.map(events, &{&1.id, tid})
+encoded_rows = Bench.encode_rows_serial(id_tid_pairs, compiled, mapping_config_id)
+
+baseline = Bench.baseline(id_tid_pairs, compiled, mapping_config_id)
+candidate = Bench.compress_rows(encoded_rows)
+
+if baseline != candidate do
+  raise "processor-side output differs from the batch-side baseline"
+end
+
+run =
+  case scenario do
+    :baseline ->
+      fn -> Bench.baseline(id_tid_pairs, compiled, mapping_config_id) end
+
+    :processor ->
+      fn ->
+        id_tid_pairs
+        |> Bench.encode_rows(compiled, mapping_config_id, concurrency, processor_chunk_size)
+        |> Bench.compress_rows()
+      end
+
+    :encode ->
+      fn ->
+        Bench.encode_rows(
+          id_tid_pairs,
+          compiled,
+          mapping_config_id,
+          concurrency,
+          processor_chunk_size
+        )
+      end
+
+    :compress ->
+      fn -> Bench.compress_rows(encoded_rows) end
+  end
+
+if warmup_batches > 0 do
+  for _ <- 1..warmup_batches, do: run.()
+end
+
+:erlang.garbage_collect()
+:erlang.statistics(:runtime)
+:erlang.statistics(:wall_clock)
+:erlang.statistics(:reductions)
+
+{elapsed_us, result_bytes} =
+  :timer.tc(fn ->
+    Enum.reduce(1..batches, 0, fn _, total ->
+      result = run.()
+      bytes = if is_binary(result), do: byte_size(result), else: Enum.sum_by(result, &byte_size/1)
+      total + bytes
+    end)
+  end)
+
+{_, runtime_ms} = :erlang.statistics(:runtime)
+{_, wall_ms} = :erlang.statistics(:wall_clock)
+{_, reductions} = :erlang.statistics(:reductions)
+rows = batch_size * batches
+
+IO.puts("scenario=#{scenario}")
+IO.puts("event_type=#{type}")
+IO.puts("schedulers_online=#{System.schedulers_online()}")
+IO.puts("concurrency=#{concurrency}")
+IO.puts("processor_chunk_size=#{processor_chunk_size}")
+IO.puts("batch_size=#{batch_size}")
+IO.puts("batches=#{batches}")
+IO.puts("rows=#{rows}")
+
+IO.puts(
+  "rowbinary_bytes_per_row=#{Float.round(Enum.sum_by(encoded_rows, &byte_size/1) / batch_size, 2)}"
+)
+
+IO.puts("result_bytes_total=#{result_bytes}")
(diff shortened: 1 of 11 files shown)


bench/clickhouse_processor_rowbinary.exs
```diff
@@ -0,0 +1,185 @@
...
+  end
+
+  defp encode_pointer({id, tid}, compiled, mapping_config_id) do
+    %LogEvent{} = event = IngestEventQueue.lookup_event(tid, id)
+    output_context = OutputContext.clickhouse_row_binary(event, mapping_config_id)
+    Mapper.map(event.body, compiled, output_context: output_context)
+  end
+
+  defp gzip_stream(fun) do
+    z = :zlib.open()
```
ruslandoga, line 76: 👋 ... it's me again ...

I wonder if the new [`:zstd`](https://www.erlang.org/doc/apps/stdlib/zstd.html) could be a perf-improvment too?

Also wanted to add in my experience building RowBinary iodata "in full" then calling `:zstd.compress/1` once was faster than doing it incrementally: https://github.com/plausible/ch/pull/303#discussion_r3052460058

ruslandoga: My old experiment: https://github.com/plausible/ch/pull/303#discussion_r3052460058

ruslandoga: I wonder if encoding to RowBinary could happen ahead of batching (kind of like in https://github.com/plausible/analytics/pull/3623) e.g. when the event is accepted/buffered, and then a single event is turned into a single binary (to aid in storing/message passing by binary ref) or a flat list (with `:erlang.iolist_to_iovec/1`), and compression would happen right before the HTTP request.

---

I took a closer look at the PR and I see it already uses a single binary per event :) Nice!

And Logflare would need to upgrade to OTP 28 to get `:zstd` https://github.com/Logflare/logflare/blob/ee61fab1060828a28496b48da1cb297ef160d42a/.tool-versions#L2

djwhitt: Yeah, the single binary yields a nice latency reduction and spreads the work across processors a bit more evenly. I'm going to keep `:zstd` out of this PR but hope to get to it soon in a follow up.

ruslandoga: **UPDATE:** I asked Codex to verify if my old measurements still hold, with one-shot compression being faster than stream, and it seems like it doesn't, at least for the RowBinary shapes/sizes in the PR's benchmark (the difference is only visible on sub-512B binaries).

<details>
<summary>Codex summary</summary>

Codex reran both gzip and Zstd against the exact 60,000-row :realistic fixtures from PR #3837 at commit 0e403c12.

The controlled comparison used:

- Erlang/OTP 29.0.3 and Elixir 1.20.2
- 8 schedulers
- Zstd’s default compression level 3
- 3 warmups and 12 balanced-order samples
- mapping and RowBinary encoding completed before timing
- codec context setup/cleanup and final output materialization included

The compared call shapes were essentially:

```elixir
# One-shot
:zlib.gzip(rows)
:zstd.compress(rows)

# Streaming
Enum.map(rows, &:zlib.deflate(gzip_context, &1))

Enum.flat_map(leading_rows, &stream_zstd(zstd_context, &1))
:zstd.finish(zstd_context, last_row)
```

The Zstd stream pledged the total source size and passed the last row directly to `finish/2`, making it comparable to `:zstd.compress/1`.

### One-shot versus streaming

| Fixture | Mean row | Gzip stream | Gzip once | Zstd stream | Zstd once |
| Log | 1,585 B | 393.7 ms | 381.8 ms | 66.5 ms | 66.4 ms |
| Metric | 2,347 B | 602.7 ms | 604.2 ms | 99.5 ms | 98.9 ms |
| Trace | 1,923 B | 516.5 ms | 507.6 ms | 76.3 ms | 77.7 ms |

That translates to:

- Logs: gzip one-shot was 3.0% faster; Zstd was tied.
- Metrics: both codecs were tied.
- Traces: gzip one-shot was 1.7% faster; Zstd streaming was 1.8% faster.

There is therefore no meaningful or consistent one-shot advantage for the PR’s row shapes with either codec.

All four methods decompressed to the exact original RowBinary payload. One-shot and streaming output were also byte-for-byte identical within each codec.

### Gzip versus Zstd

| Fixture | Gzip size | Zstd size | Zstd size change | Zstd speedup |
|---|---:|---:|---:|---:|
| Log | 4,282,960 B | 3,163,809 B | 26.1% smaller | about 5.7–5.9× |
| Metric | 6,235,840 B | 6,135,438 B | 1.6% smaller | about 6.0–6.1× |
| Trace | 5,698,516 B | 3,960,582 B | 30.5% smaller | about 6.5–6.8× |

So the substantial result is not one-shot versus streaming—it is Zstd versus gzip. On these fixtures, Zstd compressed roughly 6× faster and produced output that was the same size or smaller.

### Why the old one-shot result does not transfer

Both implementations pass iodata through `:erlang.iolist_to_iovec/1`. With OTP 29, the Zstd benchmark showed an exact boundary:

| Binary row size | Result |
|---:|---|
| 511 B | one-shot about 20% faster |
| 512 B | one-shot about 18% faster |
| 513 B | tied |
| 768 B+ | tied |

For 10,000 binaries of 512 bytes or less, `iolist_to_iovec/1` coalesced the input into one entry. At 513 bytes, it retained 10,000 separate entries.

That explains the old [ch benchmark](https://github.com/plausible/ch/pull/303#discussion_r3052460058): it supplied nested, small RowBinary iodata, allowing one-shot compression to coalesce the input and avoid a large amount of per-row overhead.

PR #3837 already represents every event as one materialized `EncodedRow` binary averaging 1.5–2.35 KiB. Those remain separate iovec entries, so one-shot and streaming perform approximately the same work.

Gzip showed the same broader size dependence, although without such a sharp boundary:

- 65,536 × 64-byte rows: one-shot was 25–69% faster, depending on compressibility.
- 60,000 × 512-byte event-like rows: one-shot was 11.2% faster.
- 4,096 × 1 KiB rows: only 2.5–4.7% faster.
- The PR’s 1.5–2.35 KiB rows: between a small regression and a 3% improvement.

### Fixture caveat

The PR’s `:realistic` fixtures are rich synthetic/stress records, not evidence of the production row-size distribution. Simpler two-attribute OTEL-shaped records encoded to approximately:

- Log: 451 B
- Metric: 461 B
- Trace: 551 B

Those are around the coalescing boundary, so real traffic containing many small rows could still benefit from one-shot compression. That would need to be verified against a production RowBinary size histogram and would trade the streaming memory profile for transient input coalescing/copying.

### Takeaway

- The old one-shot measurement was valid for small or nested RowBinary iodata.
- It does not reproduce for the already-materialized, kilobyte-sized rows in this PR, with either gzip or Zstd.
- Switching either codec from streaming to one-shot is not supported by the PR fixture results.
- Zstd itself looks worth a separate follow-up after the OTP upgrade: it was approximately 6× faster than gzip and produced equal or smaller output here.
- A codec decision should also measure peak RSS, actual production row sizes, end-to-end ClickHouse latency, and HTTP/ClickHouse compatibility—not just compression CPU.

</details>

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -422,32 +507,82 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
+
+        {:error, :not_initialized} ->
+          IngestEventQueue.reinsert_pointer({:consolidated, backend_id}, pointer)
+      end
+
+    case result do
+      :ok ->
+        :requeued
+
+      {:error, :already_exists} ->
```
amokan: 🤖 

> `reinsert_retry_pointer/2` maps `{:error, :already_exists}` → `:deduplicated`, but `requeue_retriable/2` only reads `:lookup_miss` and `:queue_unavailable`. The retried `EncodedRow` is neither requeued _nor_ deleted, and the colliding pointer carries a different `gen_event_id`, so the row is orphaned until rotation — no counter moves. Fix: emit a `:deduplicated` counter + `delete_id` the orphan (_alternatives: fold into `:queue_unavailable`, or into `requeue_lookup_miss`_).

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -422,32 +507,82 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
 
     :ok
   end
 
-  # A miss here means the pointer's generation was already dropped by
-  # GenerationJanitor before the retry could look it up — the event is silently
-  # lost rather than requeued. Bounded and rare in practice, but worth surfacing
-  # since it's otherwise invisible.
+  @spec requeue_payload(pos_integer(), EncodedRow.t() | LogEventPointer.t()) :: requeue_result()
+  defp requeue_payload(backend_id, %EncodedRow{pointer: pointer} = encoded) do
```
amokan, line 570: 🤖 

> retries no longer refresh the event's generation.
Old path did `delete_id` + `add_to_table`, re-homing the event in the current generation and resetting its age clock. New path keeps the original generation via `replace_event/3`, so `GenerationJanitor` (_4 min max age_) can drop it before the requeued pointer is re-claimed. Worst under sustained CH slowness — exactly when retries fire. Fix: on requeue, write into the _current_ generation and repoint `pointer.tid`/`gen_event_id`.

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -1007,21 +1026,58 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+  @spec reinsert_pointer(LogEventPointer.t()) ::
+          :ok | {:error, :already_exists | :not_initialized}
   def reinsert_pointer(%LogEventPointer{} = pointer) do
     row =
       {pointer.id, pointer.tid, pointer.gen_event_id, pointer.size, pointer.retries,
        pointer.event_type, pointer.day_bucket}
 
-    :ets.insert(pointer.queue_tid, row)
-    :ok
+    if :ets.insert_new(pointer.queue_tid, row) do
```
amokan: 🤖 

> `reinsert_pointer/1` switched `:ets.insert` → `:ets.insert_new`, but both pre-existing callers ignore the new error returns. `buffer_producer.ex:121` and `spool/producer_pipeline.ex:155` discard it inside `Enum.each/2`. In `seed_from_startup/4` the pointer was already `:ets.taken`, so a dropped reinsert strands the event with nothing logged. Fix: handle `{:error, _}` in both callers, or keep the shared function at `:ets.insert` and add a separate `_new` variant for the CH retry path.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -39,18 +40,34 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
+  @max_in_flight_batches 64
+  @max_in_flight @batch_size * @max_in_flight_batches
+
+  @doc false
+  @spec processor_concurrency() :: pos_integer()
+  def processor_concurrency, do: processor_concurrency(System.schedulers_online())
+
+  @doc false
+  @spec processor_concurrency(pos_integer()) :: pos_integer()
+  def processor_concurrency(schedulers_online)
```
amokan, line 72: Not a hill to die on now, but this is ultimately unbound. Not a big deal in the current form, but if we started opening up ClickHouse log drains - I'd say it's worth a comment to indicate this should be revisited.

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -422,32 +507,82 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
+          :ok
+
+        {:error, :already_exists} = error ->
+          error
+
+        {:error, :not_initialized} ->
+          IngestEventQueue.reinsert_pointer({:consolidated, backend_id}, pointer)
+      end
+
+    case result do
```
amokan: Nitpick. Wouldn't hate it if we moved this second case to a set of pattern matched defp functions.

```elixir
defp handle_reinsert_pointer_result(:ok), do: :requeued
defp handle_reinsert_pointer_result({:error, :already_exists}), do: :deduplicated
# etc etc
```

amokan (changes requested):
I like the direction this is going - but worth checking some of these 🤖 callouts

lib/logflare/backends/adaptor/clickhouse_adaptor/pipeline.ex
```diff
@@ -111,10 +134,27 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Pipeline do
...
         ]
       ],
-      context: %{backend_id: backend.id}
+      context: processor_context(backend.id)
     )
   end
 
+  @doc false
+  @spec processor_context(pos_integer()) :: map()
+  def processor_context(backend_id) do
```
amokan: _Super nit_ ™️ 

Thinking the name of this function should be `build_processor_context/1` - but not the end of the world in the current form.

lib/logflare/backends/ingest_event_queue.ex
```diff
@@ -1001,29 +1021,233 @@ defmodule Logflare.Backends.IngestEventQueue do
...
+
+  Active producer queues are tried from least to most pending, followed by the startup
+  queue. A candidate that disappears during routing is skipped. The backing generation
+  row is not copied; the routed pointer continues to reference it directly.
+  """
+  @spec insert_new_pointer(
+          queues_key() | consolidated_queues_key() | spool_producer_queues_key(),
+          LogEventPointer.t()
+        ) :: :ok | {:error, :already_exists | :not_initialized}
+  def insert_new_pointer(queues_key, %LogEventPointer{} = pointer) do
```
amokan: Looks like `insert_new_pointer/2` is only called from a single test (`test/logflare/backends/ingest_event_queue_test.exs:1060`).

Let's dump it.

amokan (approved):
Approved. I think the only thing I'd push for is dumping the dead `insert_new_pointer/2` code.

PR #3839, opened 2026-08-17
feat(backends): add SigNoz backend adaptor

Nageshbansal:
## Summary

- Adds SigNoz as an ingest-only backend, sending logs over OTLP/HTTP. It reuses `HttpBased.Pipeline` and `OtlpAdaptor.ProtobufFormatter`.
-  Adds a docs page at `docs/backends/signoz.mdx`.

Config is two fields:

- `endpoint` (required): the ingestion URL, for example `https://ingest.us.signoz.cloud:443`. The `/v1/logs` path is appended. Validation rejects an endpoint that already ends in `/v1/logs`, since pasting the full logs URL is an easy mistake to make and it would otherwise 404 at ingest time instead of failing on save.
- `ingestion_key` (optional): sent as the `signoz-ingestion-key` header. The header is left off entirely when no key is set, so a self-managed collector with no auth works too.

## Media

Demo of logs draining from a local Logflare into SigNoz Cloud, filtered on `service.name`:

image

image

image


docs/docs.logflare.com/docs/backends/signoz.mdx
```diff
@@ -0,0 +1,33 @@
...
+- Authentication is handled via the `signoz-ingestion-key` header, which is omitted when no ingestion key is configured.
+
+### Log record mapping
+
+SigNoz indexes the OTLP log record body and attributes, and does not read `event_name`. The adaptor therefore always sends flattened records:
+
+- `event_message` becomes the record body, so it shows as the log message in SigNoz.
+- Every other field is flattened into individual attributes using dotted keys, for example `metadata.parsed.error_severity`, so each one is filterable.
+- Severity is derived from the source's own level fields when the event carries no explicit `severity_number`.
+- The source name is sent as the `service.name` resource attribute. For log drains, the project ref is sent as `service.namespace`.
```
Ziinc:
```suggestion
- The source name is sent as the `service.name` resource attribute.
```
Should not include supabase specific aspects here.

lib/logflare/backends/adaptor/signoz_adaptor.ex
```diff
@@ -0,0 +1,110 @@
...
+    |> Ecto.Changeset.validate_format(:endpoint, ~r/https?\:\/\/.+/)
+    |> validate_endpoint_without_logs_path()
+  end
+
+  # Pasting the full logs URL is an easy mistake, and appending to it would silently 404.
+  @spec validate_endpoint_without_logs_path(Ecto.Changeset.t()) :: Ecto.Changeset.t()
+  defp validate_endpoint_without_logs_path(changeset) do
+    endpoint = Ecto.Changeset.get_field(changeset, :endpoint)
+
+    if is_non_empty_binary(endpoint) and
```
Ziinc: is_non_empty_binary is redudant

lib/logflare/backends/adaptor/signoz_adaptor.ex
```diff
@@ -0,0 +1,110 @@
...
+
+  @impl Adaptor
+  def cast_config(params, existing_config \\ %{}) do
+    types = %{
+      endpoint: :string,
+      ingestion_key: :string
+    }
+
+    {existing_config, types}
+    |> Ecto.Changeset.cast(params, Map.keys(types))
```
Ziinc:
```suggestion
    alias Ecto.Changeset
    
    
    ...
    |> Changeset.cast(params, Map.keys(types))
```
alias preferred

Ziinc (changes requested):
Thanks for the contribution, looking good!
some minor comments. Please also include a screencap demo showing the insertion into signoz just to cover all bases. 💚

docs/docs.logflare.com/docs/backends/signoz.mdx
Ziinc: @Baishan i think we can come up with some namespace rules for this if we want to include the project ref in `service` permanently, and would need to document that we would look at specific fields to populate this `service.namespace`. Realtime looks at `tenant` and `project` and uses the first non-nil, we can do the same to keep it non-supabase-specific.

lib/logflare/backends/adaptor/signoz_adaptor.ex
Nageshbansal: Switched to Changeset.validate_change/3 in 130dd2fba, following the shape of endpoint_validator in the S3 adaptor. It skips empty values itself, so the guard is gone entirely.

lib/logflare/backends/adaptor/signoz_adaptor.ex
Nageshbansal: Done in 130dd2fba, and used the alias for the other Ecto.Changeset calls in the module too

Nageshbansal:
@Ziinc We have updated the PR and re-requesting your review

vikrantgupta25:
Hi @Ziinc , SigNoz maintainer here. Can we please review the above PR. Thank you :)

Ziinc (approved):

Nageshbansal:
@Ziinc Please rerun the checks, Fixed the failing lint.

PR #3842, opened 2026-08-17
fix: fail closed when serializing backend config

djwhitt:
## Summary

- add an explicit per-adaptor safe response policy for backend configuration
- fail closed when adaptor lookup, module loading, redaction callbacks, or config shapes are unsafe
- omit credential components, arbitrary headers and metadata, unsafe URLs, and other non-allowlisted configuration
- separate write-time backend schemas from secret-free response schemas
- cover every adaptor plus direct and nested Management API response paths

## Security behavior

Backend config serialization now returns an empty config when it cannot prove that serialization is safe. URL fields are returned only for plain origins without userinfo, path capabilities, query parameters, or fragments. Top-level backend metadata is no longer serialized.

Source- and alert-level Slack/webhook fields are tracked separately in O11Y-2355.

## Validation

- `../bin/test test/logflare/backends/backend_response_config_test.exs test/logflare_web/controllers/api/backend_controller_test.exs test/logflare_web/controllers/api/source_controller_test.exs test/logflare_web/controllers/api/alert_controller_test.exs test/logflare_web/open_api_test.exs` — 96 tests, 0 failures
- `../bin/x mix compile`

Linear: https://linear.app/supabase/issue/O11Y-2189/make-management-api-backend-config-redaction-comprehensive-and-fail


djwhitt:
This is overkill. Closing. Superseded by #3846 and #3844.

PR #3849, opened 2026-08-18
feat: stop spool consumer during drain

Baishan:
Implement prepare_for_draining to stop consumer pipline reading from sqs


djwhitt:
:robot:

> `prepare_for_draining/1` only prevents starting a new prefetch; it does not stop an already-running prefetch from being processed with the producer's stale `demand`.
>
> Broadway cancels consumers immediately after this callback. If the prefetch completes afterward, `handle_info({:prefetch_result, ...})` schedules a poll, the file is emitted into GenStage with no consumers, and a subsequent poll can ack the queue handle. Those buffered events are discarded when the topology exits, causing data loss. This was reproduced with an ack occurring after cancelling the producer's only consumer.
>
> Treat draining as a hard freeze: clear demand, guard post-drain demand/poll handling, and nack or leave late-prefetched handles unacknowledged for redelivery. The test should also exercise Broadway topology shutdown with a prefetch completing after consumer cancellation; the current test subscribes after calling `prepare_for_draining/1`, which is the reverse of production ordering.

Baishan:
> 🤖
> 
> > `prepare_for_draining/1` only prevents starting a new prefetch; it does not stop an already-running prefetch from being processed with the producer's stale `demand`.
> > Broadway cancels consumers immediately after this callback. If the prefetch completes afterward, `handle_info({:prefetch_result, ...})` schedules a poll, the file is emitted into GenStage with no consumers, and a subsequent poll can ack the queue handle. Those buffered events are discarded when the topology exits, causing data loss. This was reproduced with an ack occurring after cancelling the producer's only consumer.
> > Treat draining as a hard freeze: clear demand, guard post-drain demand/poll handling, and nack or leave late-prefetched handles unacknowledged for redelivery. The test should also exercise Broadway topology shutdown with a prefetch completing after consumer cancellation; the current test subscribes after calling `prepare_for_draining/1`, which is the reverse of production ordering.

Updated this to nack anything thats in flight once prepare to drain is called.

djwhitt (approved):
:robot:

> Non-blocking: consider making settlement of a running prefetch independent of the producer’s lifetime. Because the prefetch uses an unlinked `Task.start/1`, Broadway can terminate the producer before the task returns. Its result is then sent to a dead PID, leaving the queue message invisible until the provider deadline expires rather than immediately nacking it.
>
> The current test keeps the standalone producer alive, so a future follow-up could cover this through `Broadway.stop/3` and make the task settle its eventual handle when the parent exits.

PR #3851, opened 2026-08-19
feat: add Splunk backend adaptor

bblaszkow06:
Adds an ingest-only backend sending logs to Splunk via the HTTP Event Collector API, built on HttpBased.Pipeline + HttpBased.Client.


lib/logflare/backends/adaptor/splunk_adaptor/hec_formatter.ex
```diff
@@ -0,0 +1,38 @@
...
+    %{
+      "event" => body,
+      "time" => body["timestamp"] / 1_000_000,
+      "sourcetype" => Map.get(config, :sourcetype) || @default_sourcetype,
+      "source" => Map.get(config, :source) || log_event.source_name,
+      "index" => Map.get(config, :index),
+      "host" => Map.get(config, :host)
+    }
+    |> Enum.reject(fn {_key, value} -> is_nil(value) end)
+    |> Map.new()
```
Ziinc: since only index and host can possibly be empty, we can instead conditionally put them instead of converting map to list, iterating, then back to map. would be faster too.

test/logflare/backends/adaptor/splunk_adaptor_test.exs
```diff
@@ -0,0 +1,254 @@
...
+    end
+  end
+
+  describe "client_opts/1" do
+    test "authenticates with the HEC scheme" do
+      backend = build(:backend, type: :splunk, config: @valid_config)
+      opts = @subject.client_opts(backend)
+
+      assert opts[:url] == @url
+      assert opts[:headers] == [{"authorization", "Splunk test-token"}]
```
Ziinc, line 85: only this is specific to hec auth

test/logflare/backends/adaptor/splunk_adaptor_test.exs
```diff
@@ -0,0 +1,254 @@
...
+      assert opts[:json] == true
+      assert opts[:gzip] == true
+      assert opts[:ssrf] == true
```
Ziinc:
```suggestion
```
nit

Ziinc (changes requested):

lib/logflare/backends/adaptor/splunk_adaptor.ex
```diff
@@ -0,0 +1,143 @@
...
+
+  @impl HttpBased.Client
+  def client_opts(%Backend{config: config}) do
+    [
+      url: config.url,
+      headers: [{"authorization", "Splunk #{config.token}"}],
+      formatter: {HecFormatter, config: config},
+      json: true,
+      gzip: true,
+      ssrf: true
```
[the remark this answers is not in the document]
bblaszkow06: This affects other adaptors as well, and needs to be solved at a lower level

Ziinc (approved):

PR #3859, opened 2026-08-20
fix(clickhouse): wrap endpoint limits around set operations

djwhitt:
## Summary

- normalize direct and parenthesized ClickHouse set operations through a wrapper before applying endpoint `max_limit`
- apply set-level `ORDER BY`, `LIMIT`, `OFFSET`, and `LIMIT BY` to the completed result when ordering references exposed output columns
- preserve compatible behavior for unexposed or unknown ordering expressions while applying the endpoint cap outside them
- keep query-level `SETTINGS` and `FORMAT` outermost through all generated wrappers
- preserve negative and fractional limit semantics with a secondary outer cap
- cover all supported set operations and execute parenthesized, ordered, offset, `LIMIT BY`, `SETTINGS`, negative, fractional, and compatibility-fallback regressions against ClickHouse
- bump the application version to `1.50.6`

## Testing

- `../bin/test test/logflare/sql/dialect_transformer/clickhouse_test.exs test/logflare/backends/adaptor/clickhouse_adaptor_test.exs` — 157 tests, 0 failures


lib/logflare/sql/dialect_transformer/clickhouse.ex
```diff
@@ -55,9 +55,12 @@ defmodule Logflare.Sql.DialectTransformer.ClickHouse do
   defp apply_limit_to_statements(_statements, _query, _max_rows),
     do: {:error, "Expected one ClickHouse query"}
 
+  # Set operations need an outer query so ClickHouse applies and parses the cap
+  # against the completed result instead of a UNION/INTERSECT/EXCEPT branch.
   # LIMIT BY is per group, while negative, fractional, and arbitrary expressions
   # do not have ordinary row-count semantics. Cap their completed result from an
   # outer query instead of comparing the original expression numerically.
+  defp requires_outer_limit?(%{"body" => %{"SetOperation" => _operation}}), do: true
```
amokan: 🤖 

> The new guard only matches a top-level `SetOperation` body. A parenthesized set operation parses as `body => %{"Query" => ...}` and falls through to the plain path. Verified: input `(SELECT 1 UNION ALL SELECT 2)` is valid ClickHouse (returns `1,2`), but `apply_limit/2` emits `(SELECT 1 UNION ALL SELECT 2) LIMIT 10`, which ClickHouse rejects — `Code: 62 ... Syntax error: failed at position 31 (LIMIT)`. A working endpoint query becomes a hard error once `max_limit` is set. Pre-existing, not introduced here, but it's the same bug class the PR claims to close (_"cover all supported set operations"_).
Options: (a) add a clause matching a nested `%{"body" => %{"Query" => _}}` body → `true`; (b) recurse into nested query bodies to detect set operations at any depth; (c) invert the guard so only bodies known to be plain Select/Values take the fast path and everything else wraps. Recommend (c) — fail-safe by default, since wrapping is always semantically valid and the only cost is a subquery; (a) is the minimal patch if you want to keep the change tight.

test/logflare/sql/dialect_transformer/clickhouse_test.exs
```diff
@@ -42,6 +42,28 @@ defmodule Logflare.Sql.DialectTransformer.ClickHouseTest do
...
+            "SELECT 1 INTERSECT SELECT 2",
+            "SELECT 1 EXCEPT SELECT 2"
+          ] do
+        expected = "SELECT * FROM (#{query}) LIMIT 3"
+
+        assert {:ok, ^expected} = ClickHouse.apply_limit(query, 3)
+      end
+    end
+
+    test "preserves a set operation's stricter limit and offset inside the global cap" do
```
amokan: 🤖 

> The test name asserts a guarantee the code does not provide. The inner `LIMIT 4 OFFSET 2` binds to the last `UNION ALL` branch, so the wrapped query returns 8 rows, not ≤4 — nothing "stricter" is preserved. Worse, because `max_rows` (10) exceeds the 8 rows the query can produce, the outer cap is not load-bearing: this test would still pass if the wrapping regressed.
Options: (a) rename to something truthful (e.g. _"keeps an existing branch limit and offset inside the subquery"_) and drop the _"stricter/global cap"_ claim; (b) keep the intent but set `max_rows` to 3 so the outer `LIMIT` actually constrains the result, and add an adaptor-level integration case asserting exactly 3 rows come back; (c) change the transform to hoist the top-level limit/offset out to the wrapper so the name becomes true (see finding 3). Recommend (a)+(b) for this PR.

lib/logflare/sql/dialect_transformer/clickhouse.ex
amokan, line 145: 🤖 Unrelated to the changes in the PR - but noting 

> `wrap_with_limit/2` leaves the top-level `ORDER BY` inside the subquery, and ClickHouse does not preserve subquery ordering. Verified: `SELECT * FROM (SELECT number FROM numbers(100) UNION ALL SELECT number+100 FROM numbers(100) ORDER BY number DESC) LIMIT 5 → 0,1,2,3,4`, while hoisting the sort out `(SELECT * FROM (...) ORDER BY number DESC LIMIT 5)` → `199,198,197,196,195`. Concrete impact: a "newest N" union endpoint `(... UNION ALL ... ORDER BY timestamp DESC)` silently returns an arbitrary slice. It's plan-dependent — the smaller numbers(20) CTE case does come back ordered — so it will look fine in tests and go wrong on real data. Not a regression (the pre-fix branch-local `ORDER BY` plus decoded-row truncation was equally arbitrary), but the fix passes up the chance to make it right.
Options: (a) when the body is a set operation, move `order_by` (and limit/offset) to the outer wrapper and cap with least(existing, max_rows) — matches both sqlparser/standard semantics and user intent; (b) hoist order_by only, leave limit/offset in place; (c) leave as-is and note the limitation in the @doc. Recommend (a), but as a follow-up PR — it changes user-visible results for existing set-operation endpoints and deserves its own integration coverage

amokan (changes requested):
The one item in the transformer module is worth tweaking I think.

lib/logflare/sql/dialect_transformer/clickhouse.ex
```diff
@@ -10,6 +10,8 @@ defmodule Logflare.Sql.DialectTransformer.ClickHouse do
   alias Logflare.Sql.Parser
   alias Logflare.User
 
+  @set_operation_modifiers ~w(order_by limit offset format_clause)
```
amokan: 🤖 

> `limit_by` is not hoisted with `limit`, producing invalid SQL. `ClickHouse.LIMIT n BY cols` is one syntactic unit, but `@set_operation_modifiers` takes `limit` and leaves `limit_by`. Verified end to end:
> - Input `SELECT 1 AS n UNION ALL SELECT 2 AS n LIMIT 1 BY n` — valid, returns `1,2` on CH 26.2
> - New output `SELECT * FROM (SELECT 1 AS n UNION ALL SELECT 2 AS n BY n) LIMIT least(1, 3)` — `Code: 62 ... Syntax error: failed at position 54 (BY)`
> - `main` produced `SELECT * FROM (... LIMIT 1 BY n) LIMIT 3`, which runs fine
> 
> Options: (a) add `limit_by` to @set_operation_modifiers — I verified the resulting shape `SELECT * FROM (SELECT * FROM (union) LIMIT 1 BY n) LIMIT 3` executes correctly, and it composes with `requires_outer_limit?/1`'s existing `limit_by` clause; (b) skip hoisting `limit` when `limit_by` is non-empty and fall back to the plain wrap; (c) bail out of set-operation hoisting entirely when `limit_by` is present. Recommend (a) plus a regression test — it's also the semantically right answer (_per-group cap over the completed set operation_).

lib/logflare/sql/dialect_transformer/clickhouse.ex
```diff
@@ -47,13 +63,52 @@ defmodule Logflare.Sql.DialectTransformer.ClickHouse do
...
+    modifiers = Map.take(query_ast, @set_operation_modifiers)
+
+    inner_query_ast =
+      Enum.reduce(@set_operation_modifiers, query_ast, fn key, query_ast ->
+        Map.put(query_ast, key, nil)
+      end)
+
+    statement = put_in(statement, ["Query"], inner_query_ast)
+
+    with {:ok, inner_query} <- Parser.to_string(statement),
```
amokan: 🤖 

> set operation + query-level `SETTINGS` now fails to re-parse, so the endpoint returns an error instead of rows. Same root cause: `settings` stays on the inner query after limit is removed, and sqlparser only accepts `SETTINGS` after a set operation when a `LIMIT` precedes it. Confirmed the mechanism directly:
> - `Parser.parse("clickhouse", "SELECT * FROM (SELECT 1 UNION ALL SELECT 2 SETTINGS max_threads = 1)")` → `{:error, "sql parser error: Expected: joined table, found: 1 at Line: 1, Column: 23"}`
> - So `apply_limit("SELECT 1 UNION ALL SELECT 2 LIMIT 100 SETTINGS max_threads = 1", 3)` now returns that `{:error, ...}`; on `main` it returned working SQL, and the input itself is valid CH (returns `1,2`)
> 
> `SETTINGS` is a supported, allowlisted feature here (`clickhouse_adaptor_test.exs` has "limits endpoint queries with allowed query-level settings"), so this is a live break for any union endpoint using it. Options: (a) add `settings` to `@set_operation_modifiers` — verified `SELECT * FROM (union) LIMIT least(100, 3) SETTINGS max_threads = 1` executes, and query-level `settings` arguably belong on the whole query anyway; (b) bail out of hoisting when `settings` is present; (c) build the outer AST directly instead of `Parser.to_string/1` → `Parser.parse/2`, which removes this whole class of _"serializes to something we can no longer parse"_ fragility. Recommend (a) now, (c) as a follow-up.

lib/logflare/sql/dialect_transformer/clickhouse.ex
```diff
@@ -47,13 +63,52 @@ defmodule Logflare.Sql.DialectTransformer.ClickHouse do
...
+  defp redundant_query_wrapper?(query_ast) do
+    query_ast
+    |> Map.delete("body")
+    |> Enum.all?(fn {_key, value} -> is_nil(value) or value == [] end)
+  end
+
+  # ClickHouse binds trailing modifiers to a set-operation branch. Move modifiers
+  # onto a SELECT over the completed result so ordering and limits are global.
+  defp wrap_set_operation(%{"Query" => query_ast} = statement) do
+    modifiers = Map.take(query_ast, @set_operation_modifiers)
```
amokan: 🤖 

> hoisting `ORDER BY` breaks queries that sort on a column the set operation doesn't project. Verified: `SELECT name FROM system.tables UNION ALL SELECT name FROM system.tables ORDER BY database LIMIT 3` runs today, but the transformed `SELECT * FROM (...) ORDER BY database LIMIT 3` → `Code: 47 UNKNOWN_IDENTIFIER: Unknown expression identifier 'database'`. The log-shaped version of this is plausible: `SELECT event_message FROM a UNION ALL SELECT event_message FROM b ORDER BY timestamp DESC`. A related silent case: when a branch aliases over a source column name, the hoisted sort key rebinds to the output column.
>
> Options: (a) accept and document — those queries were already returning near-garbage (_sort applied to one branch only_), and a loud error beats silent wrong data; (b) hoist `ORDER BY` only when every identifier it references resolves to an output column/alias of the set operation, otherwise leave it inside and just wrap; (c) hoist `LIMIT/OFFSET` only and never `ORDER BY`. I lean (b) for zero breakage on existing endpoints, but (a) is defensible — your call, since it's a product decision about failing loudly vs. preserving currently-accepted SQL.

amokan (changes requested):

amokan (approved):

PR #3865, opened 2026-08-21
fix: stop the requested backend child

ruslandoga:
## Summary

- Stop only the backend child whose ID was requested.
- Cover stopping a non-first backend child and handling a missing backend ID without disturbing existing children.

## Why

The piped `Enum.find_value/3` call passed the apparent predicate in the default-value position and `&elem(&1, 0)` as the mapper. The mapper returned the first supervisor child ID, which was truthy, so enumeration stopped before comparing the requested backend ID.


ruslandoga:
Converting to draft until I simplify the test...

**UPDATE:** done.

djwhitt (approved):

PR #3866, opened 2026-08-21
feat: list sources for ingest credentials

djwhitt:
## Summary

- add `GET /api/ingest-sources` for minimal source token/name discovery
- authorize account-wide, legacy, private, and resource-scoped ingest credentials, plus existing partner impersonation, while preserving ownership isolation
- support ordered JSON and RFC 4180 CSV responses selected through HTTP `Accept` negotiation, and document the endpoint in OpenAPI and access-token docs

## Why a dedicated endpoint

The existing `GET /api/sources` route is a management API protected by private scope. It returns the full source representation and performs dashboard-oriented backend and metrics preloads. Allowing ingest credentials onto that route would either expose management fields or make its response shape depend on the credential type, complicating the existing API contract and increasing regression risk for management clients.

A dedicated endpoint keeps `GET /api/sources` unchanged while providing ingest credentials with a small, stable `token`/`name` contract. It also permits a lightweight authentication and query path that applies source-scoped filtering without the management preloads or the single-source ingestion pipeline.

## Alternative implementation

See #3869 for the alternative implementation that reuses `GET /api/sources` and varies its response by credential type. The two PRs are intended for direct comparison; only one approach should merge.

### DIFF
--- a/lib/logflare_web/controllers/api/ingest_source_controller.ex
+++ b/lib/logflare_web/controllers/api/ingest_source_controller.ex
@@ -1,0 +1,90 @@
+defmodule LogflareWeb.Api.IngestSourceController do
+  use LogflareWeb, :controller
+  use OpenApiSpex.ControllerSpecs
+
+  alias Logflare.Sources
+  alias LogflareWeb.Api.FallbackController
+  alias LogflareWeb.OpenApi.BadRequest
+  alias LogflareWeb.OpenApi.List
+  alias LogflareWeb.OpenApi.Unauthorized
+  alias LogflareWeb.OpenApiSchemas.IngestSource
+  alias OpenApiSpex.MediaType
+  alias OpenApiSpex.Response
+  alias OpenApiSpex.Schema
+
+  action_fallback(FallbackController)
+
+  tags(["ingest"])
+
+  operation(:index,
+    summary: "List sources available for ingestion",
+    parameters: [
+      format: [
+        in: :query,
+        description: "Response format",
+        schema: %Schema{type: :string, enum: ["csv"]}
+      ]
+    ],
+    responses: %{
+      200 => %Response{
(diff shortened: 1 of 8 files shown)


lib/logflare_web/controllers/api/ingest_source_controller.ex
```diff
@@ -0,0 +1,82 @@
...
+  defp csv(sources) do
+    [
+      "token,name\r\n"
+      | Enum.map(sources, fn %{token: token, name: name} ->
+          [csv_field(token), ",", csv_field(name), "\r\n"]
+        end)
+    ]
+  end
+
+  defp csv_field(value) do
```
Ziinc: `csv/1` should be moved to a render function in a View module
https://tmbb.github.io/phoenix/views.html

lib/logflare_web/router.ex
```diff
@@ -504,6 +507,12 @@ defmodule LogflareWeb.Router do
     delete "/key-values", Api.KeyValueController, :delete
   end
 
+  scope "/api", LogflareWeb do
+    pipe_through([:ingest_source_discovery_auth])
+
+    get("/ingest-sources", Api.IngestSourceController, :index)
```
Ziinc, line 513: we could change the path to be something more generic like:
```
/api/discover/ingest -- ingestable sources
/api/discover/query -- queryable endpoints
```

```
/api/ingest-sources
/api/query-endpoints
```
works as well, just a matter of preference.. i'm fine with either

Ziinc (approved):

lib/logflare_web/router.ex
djwhitt: `/api/discover`... felt a little to RPC like (as opposed to RESTful) to me, so I kept as is.

PR #3870, opened 2026-08-23
feat: Add optional read-only user/password config options for ClickHouse backends

amokan:
Adds the ability to configure a dedicated read-only user for all ClickHouse query operations


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -288,6 +294,22 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
 
   defp strip_credentials(url), do: EndpointUtils.strip_credentials(url)
 
+  @spec preserve_blank_query_password(map(), map()) :: map()
+  defp preserve_blank_query_password(params, %{query_password: password})
```
Ziinc: wouldn't it be cleaner to do this as part of the changeset transformation function?

Ziinc (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
amokan: Yep 🫠

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -990,6 +1083,13 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
     end)
   end
 
+  @spec redact_query_password(map()) :: map()
+  defp redact_query_password(%{query_password: password} = config)
+       when is_non_empty_binary(password),
+       do: %{config | query_password: "REDACTED"}
```
djwhitt, line 1089: :robot:

> Non-blocking: `query_password` is not safe to round-trip through Management API responses. `GET /api/backends/:token` emits `"REDACTED"` here, and an unchanged `PUT` accepts and persists that literal because blank-password preservation only handles `nil`; subsequent read connections authenticate with the sentinel.
>
> Do not add sentinel interpretation on writes. In a follow-up, make backend secrets write-only: omit `password` and `query_password` from serialized responses and preserve absent fields. A non-empty input rotates the secret, while clearing both query fields revokes the dedicated credentials. The primary ClickHouse `password` has the same existing behavior, so this is non-blocking for this PR but should be addressed consistently at the shared API contract.

djwhitt (approved):

PR #3871, opened 2026-08-24
chore: improve ClickHouse read connection pool settings / behavior

amokan:
It became obvious when configuring labeled read clusters that the connection pool situation was not flexible enough to support us in production.

This PR makes some sane adjustments to that area:
- Lowers the `@ch_idle_interval` from 5s -> 3s to reduce churn
- Reworks the pool sizing for queries entirely:
  - `pool_size` config removed from ClickHouse backends as it was only used for queries and was not labeled appropriately in the UX
  - Adds `read_pool_size` config (default of 50) option. This is the single source of truth for read pool sizing for every scenario _other than labeled read pools_
  - Adds `labeled_read_pool_size` config (default 32) for labeled read clusters. The exception is that the default labeled read cluster will always use the `read_pool_size` value.
- Adds new `:pool_exhausted` error handling
- Slow-checkout log now includes a `read_cluster` with the relevant label for a value
- Adds new `[:logflare, :clickhouse, :read_pool, :checkout]` telemetry event
- Removes stale `pool_size: 3` from `config.exs`

_Also adds a version bump to 1.50.7_


lib/logflare/backends/adaptor/clickhouse_adaptor/connection_manager.ex
```diff
@@ -29,6 +29,9 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.ConnectionManager do
   @ch_queue_target :timer.seconds(5)
   @recycle_interval :timer.minutes(10)
   @recycle_spread :timer.seconds(60)
+  @ch_idle_interval :timer.seconds(3)
+  @read_pool_size 50
+  @labeled_read_pool_size 32
```
Ziinc:
```suggestion
  @default_read_pool_size 50
  @default_labeled_read_pool_size 32
```

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -640,22 +646,29 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
+      %{pool_time_ms: pool_ms},
+      %{backend_id: backend_id, read_cluster: label}
+    )
+
     if pool_ms >= slow_pool_checkout_ms() do
       Logger.warning(
         "ClickHouse slow connection checkout: waited #{pool_ms}ms for a pool connection",
-        backend_id: backend_id
+        backend_id: backend_id,
+        read_cluster: label
```
Ziinc: would be good to nest this under `clickhouse.read_cluster`

```suggestion
        clickhouse: %{read_cluster: label} 
```

lib/logflare_web/live/display_helpers.ex
```diff
@@ -15,7 +15,7 @@ defmodule LogflareWeb.Live.DisplayHelpers do
   """
   def sanitize_backend_config(config) when is_map(config) do
```
Ziinc, line 16: for future, probably want to let the Adaptor handle this declaration based on their config keys

Ziinc (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
amokan: So this gets a little funky on tests (_and consistency_) due to the dev/test env using the standard Logger which effectively drops the key because it returns `nil` for any value that does not have a `String.Chars` impl (_like a map_).

So what I did as a compromise is prefix these with `clickhouse_` - so rather than `read_cluster: label` -> `clickhouse_read_cluster: label`.

Ziinc: 👍 ypu that works

Ziinc (approved):

PR #3878, opened 2026-08-27
feat: add optional backend adaptor callback for UI config sanitization

amokan:
Replaces the clunky old `lib/logflare_web/live/display_helpers.ex` module that required every 'safe' key to be added to that module for display in the UX with an adaptor-level callback.

_The only thing I don't love is the way `mask_config_values` looks in actual implementations as, at first glance, it reads like it's doing the opposite of what it does. Open to suggestions on better naming or shape to make that clearer._


lib/logflare/backends/adaptor.ex
```diff
@@ -131,6 +133,39 @@ defmodule Logflare.Backends.Adaptor do
...
+  Sanitizes a backend's configuration for display in the UI.
+
+  Delegates to the adaptor's optional `c:sanitize_config_for_display/1` callback.
+  Masks every config value when the adaptor does not implement the callback.
+  """
+  @spec sanitize_config_for_display(Backend.t()) :: map()
+  def sanitize_config_for_display(%Backend{config: config} = backend) when is_map(config) do
+    adaptor = get_adaptor(backend)
+
+    if function_exported?(adaptor, :sanitize_config_for_display, 1) do
```
djwhitt: :robot:

> `function_exported?/3` only detects functions on modules that are already loaded. In a fresh BEAM, this takes the mask-all fallback instead of invoking the adaptor callback.
>
> This is reproducible with `../bin/test test/logflare/backends/adaptor_test.exs`: the new test fails in isolation with 5 tests, 1 failure, while the full suite passes because another test loads `PostgresAdaptor` first.
>
> Ensure the adaptor is loaded before checking the optional callback, for example with `Code.ensure_loaded?/1`.

lib/logflare/backends/adaptor/otlp_adaptor.ex
```diff
@@ -80,6 +80,11 @@ defmodule Logflare.Backends.Adaptor.OtlpAdaptor do
     Map.update!(config, :headers, &redact_headers/1)
   end
 
+  @impl Adaptor
+  def sanitize_config_for_display(config) do
+    Adaptor.mask_config_values(config, [:endpoint, :protocol, :gzip, :flatten_to_attributes])
```
djwhitt: :robot:

> Preserving `:endpoint` verbatim can expose credentials in the backend detail page. The adaptor accepts endpoints such as `https://user:secret@collector.example.com/v1/logs`, and this callback returns that value unchanged. Before this change, `endpoint` was not allowed by the global sanitizer and was masked.
>
> The same issue exists in `S3Adaptor.sanitize_config_for_display/1`. Keep these endpoints masked, or display only parsed non-sensitive components such as scheme, host, and port. Add regression cases for URI userinfo and credential-bearing query parameters.

djwhitt:
> The only thing I don't love is the way mask_config_values looks in actual implementations as, at first glance, it reads like it's doing the opposite of what it does. Open to suggestions on better naming or shape to make that clearer.

What about doing something like this?

```elixir
Adaptor.mask_config_values(
  config,
  except: [:domain, :dataset_name]
)
```

djwhitt (approved):

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -182,6 +183,13 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
     |> Map.update(:url, nil, &redact_url_userinfo/1)
   end
 
+  @impl Logflare.Backends.Adaptor
+  def sanitize_config_for_display(config) do
+    config
+    |> Adaptor.mask_config_values(except: [:url, :http, :gzip])
+    |> Map.update(:url, nil, &redact_url_userinfo/1)
+  end
+
```
Ziinc, line 192: when would we deviate redacting the same fields as `redact_config/1`? looks like we're duplicating logic.
`redact_config/1` is already used for the json api as well, so whatever we mask from UI should apply to json payload and vice versa

Ziinc (approved):

PR #3879, opened 2026-08-27
fix: EndpointsLive handles invalid description

msmithstubbs:
Fixes a bug where the EndpointsLive would crash when an endpoint is created with a description exceeding the maximum character length.

* Adds migration for `endpoint_queries.description` to text
* Add form validation test coverage to Endpoints new

CleanShot 2026-08-27 at 13 34 05@2x

### DIFF
--- a/priv/repo/migrations/20260827222658_alter_endpoint_query_description_to_text.exs
+++ b/priv/repo/migrations/20260827222658_alter_endpoint_query_description_to_text.exs
@@ -1,0 +1,9 @@
+defmodule Logflare.Repo.Migrations.AlterEndpointQueryDescriptionToText do
+  use Ecto.Migration
+
+  def change do
+    alter table(:endpoint_queries) do
+      modify :description, :text, from: :string
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


lib/logflare/endpoints/endpoint_query.ex
```diff
@@ -171,6 +171,7 @@ defmodule Logflare.Endpoints.EndpointQuery do
     |> validate_required([:name, :query, :user, :language])
     |> unique_constraint(:name, name: :endpoint_queries_name_index)
     |> unique_constraint(:token)
+    |> validate_length(:description, max: 255)
```
Ziinc: hmm we should use `:text` though, no reason to restrict it to just 255. can think of situations where documentation becomes length etc.

Ziinc (changes requested):

lib/logflare/endpoints/endpoint_query.ex
msmithstubbs: Gotcha. This was just to reflect the db schema. Removed it and added migration instead.

Ziinc (approved):

PR #3882, opened 2026-08-27
feat: add NDJSON output format, generalize CH mapper

bblaszkow06:
This PR adds NDJSON output format to the Rust Mapper and moves OTEL mapping modules out of the ClickHouse namespace


native/mapper_ex/src/ndjson.rs
```diff
@@ -0,0 +1,223 @@
...
+                column.name
+            )
+        })?;
+
+        // Try derived value, then enum mapping, fallback to generic conversion
+        match derived_value(layout, column.index, values)? {
+            Some(number) => object.serialize_entry(&column.name, &number),
+            None => match enum_label(column, value) {
+                Some(label) => object.serialize_entry(&column.name, label),
+                None => object.serialize_entry(&column.name, &JsonTerm { value, nil }),
```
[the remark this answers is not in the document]
bblaszkow06: The mappings aren't user-provided; they live in our codebase.

lib/logflare/mapper/output_context.ex
```diff
@@ -27,6 +27,30 @@ defmodule Logflare.Mapper.OutputContext do
...
+        %LogEvent{
+          id: id,
+          source_uuid: source_uuid,
+          source_name: source_name,
+          ingested_at: ingested_at
+        },
+        mapping_config_id
+      )
+      when is_binary(mapping_config_id) do
+    ingested_at = if ingested_at, do: DateTime.to_unix(ingested_at, :microsecond)
```
Ziinc, line 51: DateTime conversion functions can get expensive, would be worth generate it at ingest time
also, i would expect ingested_at to always be present

native/mapper_ex/src/lib.rs
```diff
@@ -113,29 +150,37 @@ fn map_clickhouse_output<'a>(
...
+    const CONTEXT_ERROR: &str =
+        "ClickHouse RowBinary output requires a ch_row_binary output_context";
+    let (flat_keys, format, mapping_config_id, envelope) =
+        decode_output_options(options).map_err(|_| CONTEXT_ERROR.to_string())?;
+    if format != atoms::ch_row_binary() {
+        return Err(CONTEXT_ERROR.to_string());
+    }
+    let mapping_config_id = mapping_config_id
+        .decode::<Binary>()
+        .map_err(|_| "mapping_config_id must be a pre-encoded 16-byte UUID binary".to_string())?;
```
Ziinc, line 162: error message differes from :103

Ziinc: also, if this is standard then should move into decode_output_options

test/logflare/mapper/config_store_test.exs
```diff
@@ -0,0 +1,59 @@
...
+  use ExUnit.Case, async: true
+
+  alias Logflare.Mapper.ConfigStore
+  alias Logflare.Mapper.OtelDefaults
+
+  setup_all do
+    case ConfigStore.start_link([]) do
+      {:ok, _pid} -> :ok
+      {:error, {:already_started, _pid}} -> :ok
+    end
```
Ziinc, line 11: start_link_supervised

test/logflare/mapper/mapping_config_test.exs
```diff
@@ -386,13 +386,42 @@ defmodule Logflare.Mapper.MappingConfigTest do
...
 
+  describe "MappingConfig.apply_timestamp_precision/1" do
+    test "output with and without a timestamp_precision" do
+      fields = [
+        Field.datetime64("timestamp", path: "$.timestamp", precision: 9),
+        Field.array_datetime64("times", path: "$.times", precision: 9),
+        Field.string("event_message", path: "$.event_message")
+      ]
+
+      config = MappingConfig.new(fields, output: OutputFormat.ndjson(:log))
```
Ziinc, line 404: some food for thought - what if we wanted ndjson but nanosecond precision?

test/logflare/mapper/ndjson_output_test.exs
```diff
@@ -0,0 +1,227 @@
...
+    end
+  end
+
+  defp encode(event_type, body, ndjson) do
+    event = raw_event(event_type, body)
+    context = OutputContext.ndjson(event, OtelDefaults.config_id(event_type))
+    Mapper.map(event.body, ndjson[event_type], output_context: context)
+  end
+
+  defp decode(event_type, body, ndjson) when is_atom(event_type) do
```
Ziinc, line 196: probably better to call this mapped_and_decoded for readability sake since it is not just decoding

test/logflare/mapper/ndjson_output_test.exs
```diff
@@ -0,0 +1,227 @@
...
+            Field.uint64("a", path: "$.a"),
+            Field.enum8("kind", paths: ["$.kind"], values: %{"One" => 1})
+          ],
+          output: OutputFormat.ndjson(:metric)
+        )
+
+      event = raw_event(:metric, %{"b" => "x", "a" => 1, "kind" => "one"}, ingested_at: nil)
+      decoded = decode(event, Mapper.compile!(config))
+
+      assert Map.drop(decoded, @envelope_keys) == %{"b" => "x", "a" => 1, "kind" => "one"}
```
Ziinc, line 186: the label_enums threw me off with this assertion
i think the label_enums adds little value, can we remove?

test/support/clickhouse_mapped_events.ex
```diff
@@ -126,11 +126,37 @@ defmodule Logflare.ClickHouseMappedEvents do
       |> deep_merge_opts(opts[:body] || %{})
 
     {mapped_body, config_id} = map_body(input_body, :trace)
-    mapped_body = Map.put(mapped_body, "mapping_config_id", config_id)
+
+    mapped_body =
+      mapped_body
+      |> Map.put("mapping_config_id", config_id)
+      |> apply_derived_fields(:trace)
```
Ziinc, line 133: hmm feels weird that we run it through the mapper but still need to do this, probably better to generate based on a desired config + output

Ziinc (approved):

test/logflare/mapper/mapping_config_test.exs
bblaszkow06: It's possible: timestamp_precision can be set to 9 in the output config. OutputFormat.ndjson/1 only encodes the Iceberg default of 6. It's not needed for now, but it could accept `:timestamp_precision` option with an additional `ndjson/2` clause.

test/logflare/mapper/config_store_test.exs
bblaszkow06: Turns out this can be skipped as config store is started anyway, even in tests

test/logflare/mapper/mapping_config_test.exs
bblaszkow06: This is supported by OutputFormat, but not exposed via `ndjson/1` helper to not add an unused API. We could have `ndjson/2` variant accepting a precision option. ATM it can be hacked by manually modifying the struct coming from `ndjson/1`

test/support/clickhouse_mapped_events.ex
bblaszkow06: The :map output has no row_type, so the NIF can't apply the derive rules there. The Elixir apply_derived_fields is the deliberate reference implementation that mapper_output_test checks the Rust derive.rs against

lib/logflare/mapper/output_context.ex
bblaszkow06: Made `ingested_at` required. 
I'd keep the conversion at map time, though. DateTime.to_unix/2 on an existing %DateTime{} is integer arithmetic with no parsing. Precomputing it would add another field to every LogEvent, even not going to this backend.
This could be done in a separate PR, but I think it should be benchmarked first.

test/logflare/mapper/ndjson_output_test.exs
bblaszkow06: Removed. metric_type is now excluded from the map-vs-NDJSON comparison, because :map emits the enum integer and NDJSON emits the label. The dedicated "metric_type labels" test covers the label output directly.

bench/support/clickhouse_pipeline_bench_data.exs
amokan:
```suggestion
      body: Map.put(body, "mapping_config_id", OtelDefaults.config_id(type))
```

Looks like we have some additional missed spots in the project:

- `bench/clickhouse_encode_e2e.exs`
- `bench/clickhouse_row_encoding.exs`
- `bench/broadway_metric_sampling.exs`

`git grep -n -E 'MappingDefaults|MappingConfigStore' -- bench` will help show the places clearer than I can.

native/mapper_ex/src/ndjson.rs
```diff
@@ -0,0 +1,223 @@
...
+            let alt = decode_u64(values[severity_alt], "severity_number_alt")?;
+            let mapped = decode_u64(values[severity_number], "severity_number")?;
+            Ok(Some(crate::derive::severity_number(alt, mapped)))
+        }
+        Derived::Trace {
+            duration,
+            start_time,
+            end_time,
+        } if index == duration => {
+            let explicit = decode_u64(values[duration], "duration")?;
```
amokan, line 217: 🤖 _but seems legit_

> `start_time`/`end_time` are rewritten to precision 6 by `apply_timestamp_precision/1`, so the derived `end` - `start` is microseconds, but an explicit `$.duration` passes through `derive::duration` unscaled in nanoseconds. The test at `ndjson_output_test.exs:115-116` locks this in (`{%{}, 1}` for a 1500 ns span vs `{%{"duration" => 42}, 42}`). Suggest storing the compiled precision on `Derived::Trace` in `compile_layout` and scaling the explicit value before calling `derive::duration`, e.g. `coerce::scale(explicit as i64, 9, precision) as u64`. _RowBinary is unaffected (9 to 9)_. Update the test to `{%{"duration" => 42_000}, 42}`.

native/mapper_ex/src/mapper.rs
amokan: 🤖 

> `resolve_enum8` passes any integer through, so `{"metric_type": 7}` reaches the NDJSON writer with no label and `ndjson.rs:165` falls back to `JsonTerm`, emitting `"metric_type":7` while other rows emit `"metric_type":"sum"`. Same column, mixed string/int, from user-supplied HTTP JSON. Suggest only passing integers through when they are a configured value (_precompute a `HashSet<i8>` on `Enum8Data`_), otherwise fall to infer rules then default. Also have `decode_enum8_data` reject a default outside the configured values, and change the `None` arm at `ndjson.rs:165` to `serialize_none()` rather than the raw integer. Add a test for `"metric_type" => 7` expecting `"gauge"`.

amokan: 🤖 

> `JsonTerm` serializes any non-UTF-8 string as null, and non-UTF-8 map keys are dropped at line 1350. This PR makes `JsonTerm` pub and routes every top-level string column through it, so a log's `event_message` can now vanish silently. Ingest doesn't guarantee UTF-8 (_BERT endpoint, protobuf string fields_) and RowBinary writes the raw bytes with no loss. Suggest `String::from_utf8_lossy` for both values and keys (_borrowed `Cow` on the valid path, so no alloc cost_), and update `ndjson_output_test.exs:142` to assert `"\uFFFD\uFFFD"` instead of `nil`.

native/mapper_ex/src/ndjson.rs
```diff
@@ -0,0 +1,223 @@
...
+            },
+            None,
+        ),
+        _ => return Err(format!("unsupported NDJSON row type '{row_type}'")),
+    };
+
+    let columns = fields
+        .iter()
+        .enumerate()
+        .filter(|(_, field)| Some(field.name.as_str()) != hidden)
```
amokan, line 110: 🤖 

> Nothing rejects a config field named `id`, `source_uuid`, `source_name`, `mapping_config_id` or `ingested_at`. `write_row` emits the envelope then every column unconditionally, and serde_json does not dedupe, so a collision produces duplicate keys and last-wins parsers drop the envelope value. `MappingConfig.from_json` is a real config-driven entry point, so this isn't limited to `OtelDefaults`. Suggest extending this filter to return `Err("field name '<x>' is reserved for the NDJSON envelope")`, mirroring the existing duplicate-field-name error. Compile-time only, no runtime cost.

lib/logflare/mapper/output_context.ex
```diff
@@ -9,24 +9,36 @@ defmodule Logflare.Mapper.OutputContext do
...
+  @doc """
+  Builds the per-row context required by NDJSON output.
 
-    {:clickhouse_row_binary, mapping_config_id, {id, source_uuid, source_name || "", ingested_at}}
+  `mapping_config_id` is emitted as the UUID string. `ingested_at` is emitted
+  as Unix **microseconds**, matching the microsecond precision of the mapped
+  timestamp fields in NDJSON output.
+  """
+  @spec ndjson(LogEvent.t(), String.t()) :: t()
+  def ndjson(%LogEvent{} = event, mapping_config_id) when is_binary(mapping_config_id) do
```
amokan: 🤖 Nit.

> `ch_row_binary/2` (line 18) and `ndjson/2` have identical `is_binary` guards but expect different `mapping_config_id` encodings (_16 raw bytes vs 36-char UUID string_), and the NDJSON side only checks UTF-8 (`ndjson.rs:142`). Passing the 16-byte form here fails per-row for most UUIDs and silently emits control characters for the rest. Suggest having `ch_row_binary/2` take the string UUID and do the 16-byte encoding internally so both constructors share one contract. At minimum, guard `ndjson/2` with `byte_size(id) == 36`

native/mapper_ex/src/ndjson.rs
```diff
@@ -0,0 +1,223 @@
...
+        .map(|(index, field)| Column {
+            name: field.name.clone(),
+            index,
+            enum_labels: field.enum8_data.as_ref().map(|data| {
+                data.value_map
```
amokan, line 115: 🤖 _This could be a follow-up probably. Worth adding a note for though._

> `enum_labels` inverts `value_map` (`HashMap<String, i8>`), so two labels sharing one value make the emitted label depend on hash iteration order, which varies per process. `decode_enum_values` (`mapping.rs:831-860`) doesn't check value uniqueness. OTEL defaults are unaffected, but a custom config with `%{"warn" => 13, "warning" => 13}` emits different labels on different nodes. Suggest a reverse-uniqueness check in `decode_enum_values` returning `Err` from compile.

native/mapper_ex/src/ndjson.rs
```diff
@@ -0,0 +1,223 @@
...
+        &mut buffer,
+        layout,
+        values,
+        nil,
+        envelope,
+        mapping_config_id,
+    )?;
+    buffer.push(b'\n');
+
+    let mut binary = OwnedBinary::new(buffer.len())
```
amokan, line 144: 🤖

> `encode_row` serializes into a `Vec`, then allocates an `OwnedBinary` and copies the whole row. The RowBinary path already has `BinaryBuilder` (`clickhouse_rowbinary.rs:248-313`) growing an OwnedBinary in place with no copy. Suggest moving `BinaryBuilder` to a shared module, adding `impl std::io::Write for BinaryBuilder` (~10 lines), and passing `&mut builder` to the `serde_json` serializer. Unify the 2048 vs 3072 initial capacity constants while there. Worth a benchmark number before and after.

lib/logflare/mapper/mapping_config.ex
```diff
@@ -135,15 +137,40 @@ defmodule Logflare.Mapper.MappingConfig do
   end
 
   @spec to_nif_map(t()) :: map()
-  def to_nif_map(%__MODULE__{fields: fields, output: output}) do
-    config = %{"fields" => Enum.map(fields, &field_to_nif_map/1)}
+  def to_nif_map(%__MODULE__{} = config) do
+    %__MODULE__{fields: fields, output: output} = apply_timestamp_precision(config)
```
amokan: 🤖 

> Timestamp precision now has two owners: `OutputFormat.timestamp_precision` and per-field `FieldConfig.precision`, reconciled by this hidden rewrite inside `to_nif_map/1`. Effects: every `precision: 9` in `otel_defaults.ex` is dead for configs with an output; `to_json` emits `"precision": 9` beside `"timestamp_precision": 6`; an explicit `precision: false` is silently overwritten; `for_type(t, :map)` compiles at 9 while `:ndjson` compiles at 6, which is why `ndjson_output_test.exs:64-67` needs the `apply_timestamp_precision` workaround. Suggest one owner: either `for_type(type, :ndjson)` rewrites fields explicitly and `OutputFormat` drops `timestamp_precision`, or `otel_defaults.ex` drops per-field precision and the output is the sole source. Either way the test setup collapses to `Mapper.compile!(OtelDefaults.for_type(&1, :map))`.

I know you are dealing with an inconsistent API in spots here @bblaszkow06 when you extended the mapper - but it would be good to tighten this up. I'm open to either suggestion the agent mentioned.

native/mapper_ex/src/lib.rs
```diff
@@ -155,9 +190,15 @@ fn decode_clickhouse_envelope<'a>(
...
     })?;
     let ingested_at = ingested_at
-        .decode::<Option<i64>>()
-        .map_err(|_| "ingested_at must be nil or Unix microseconds".to_string())?;
+        .decode::<i64>()
+        .map_err(|_| "ingested_at must be an integer Unix timestamp".to_string())?;
+    Ok((id, source_uuid, source_name, ingested_at))
+}
 
+fn decode_clickhouse_envelope<'a>(
```
amokan: 🤖 

> Three representations of the same envelope: `ndjson.rs:47-53` and `clickhouse_rowbinary.rs:240-246` declare identical `RowEnvelope` structs, and the tuple alias at `lib.rs:181` gets re-packed here and at `lib.rs:108-113`. `map_ndjson_output` and `map_clickhouse_output` also duplicate the same decode/scratch prelude, `decode_u64` exists in both `ndjson.rs:208` and `clickhouse_rowbinary.rs:721`, and `EncodeResult<T>` is declared twice. Suggest one shared `RowEnvelope` returned directly by `decode_envelope`, one prelude helper, and shared `decode_u64/EncodeResult`. A third output format would otherwise mean a fourth envelope copy.

amokan (changes requested):
@bblaszkow06 I apologize about the 🤖 review comments, but they seemed like decent call outs.

native/mapper_ex/src/lib.rs
bblaszkow06: Fixed in https://github.com/Logflare/logflare/pull/3882/commits/ef6a41404c90f6fbe081a2d4ead9e5d750b36476

lib/logflare/mapper/mapping_config.ex
bblaszkow06: Went with a rewrite of per-field precision using `with_timestamp_precision/2`, dropped the precision on OutputFormat (https://github.com/Logflare/logflare/pull/3882/commits/d215b3ab21014f99c08641c6c7a922ddbb4de284)

native/mapper_ex/src/ndjson.rs
```diff
@@ -0,0 +1,242 @@
...
+            let mapped = decode_u64(values[severity_number], "severity_number")?;
+            Ok(Some(crate::derive::severity_number(alt, mapped)))
+        }
+        Derived::Trace {
+            duration,
+            start_time,
+            end_time,
+            precision,
+        } if index == duration => {
+            let explicit = scale_duration(decode_u64(values[duration], "duration")?, precision);
```
amokan: Scaling before that `derive::duration/3` check down below could cause small durations to be recomputed from start/end due to precision.

Something like a `duration_us: 400` with a precision of 6 would end up being 0, right?

native/mapper_ex/src/ndjson.rs
```diff
@@ -0,0 +1,242 @@
...
+            let end = values[end_time].decode::<i64>();
+            Ok(Some(crate::derive::duration(explicit, start, end)))
+        }
+        _ => Ok(None),
+    }
+}
+
+/// Explicit OTEL durations arrive in nanoseconds; scale them down to the
+/// layout's timestamp precision (0..=9, validated at compile time).
+fn scale_duration(nanos: u64, precision: u8) -> u64 {
```
amokan: Only thought on this is that if we ever decided to send the config specified in otel_default.ex over this formatter - it _could_ be a bit problematic due to mismatched types. Will need to think about this one more.

Thinking mostly about the duration config 🤔 

```
Field.uint64("duration", paths: ["$.duration", "$.duration_ns", "$.duration_ms", "$.duration_us"], default: 0)
```

native/mapper_ex/src/mapping.rs
```diff
@@ -803,9 +813,11 @@ fn decode_filters<'a>(env: Env<'a>, field: Term<'a>) -> Result<Option<StringFilt
 
 pub fn decode_enum8_data<'a>(env: Env<'a>, field: Term<'a>) -> Result<Enum8Data, String> {
     let value_map = decode_enum_values(env, field)?;
+    let values: HashSet<i8> = value_map.values().copied().collect();
```
amokan, line 831: Can we do something like this to guard against an enum8 accepting something without a dictionary?

```suggestion
    if value_map.is_empty() {
        return Err("enum8 field requires a non-empty 'enum_values' map".to_string());
    }
    
    let values: HashSet<i8> = value_map.values().copied().collect();
```

amokan (changes requested):
Super close @bblaszkow06 

The duration stuff is a bit tough. Please tell me if I'm off on any of these.

native/mapper_ex/src/ndjson.rs
bblaszkow06: While the BinaryBuilder is now moved to output, the writer solution seems to be slower - the time saved on memcopy is lost on slower allocation (malloc with thread-local cache vs enif_alloc + enif_realloc on finish).

Some optimizations might be viable, e.g., creating a subbinary instead of realloc at the cost of larger memory usage (the unused bytes cannot be GCd until the subbinary goes away)

lib/logflare/mapper/output_context.ex
bblaszkow06: Went with binary-size patterns (16 bytes for RowBinary, 36 for NDJSON) instead of encoding per row. The pipeline pre-encodes the UUID once per process as an optimization; didn't want to break that

native/mapper_ex/src/ndjson.rs
bblaszkow06: This is actually a bug in the current implementation of clickhouse_rowbinary that silently assumes the duration is in ns even though it's extracted from `duration_ms` or `duration_us` field.

🤖 suggested dropping the _ms and _us from paths as a temporary workaround.
Currently, cffe736 unifies both to assume duration precision from start_time precision, not sure that makes sense though

native/mapper_ex/src/ndjson.rs
bblaszkow06: That's right, fixed

native/mapper_ex/src/derive.rs
```diff
@@ -0,0 +1,90 @@
...
+/// already coerced to that precision, when both decode and the span is
+/// positive.
+pub fn duration<E>(
+    explicit: u64,
+    start_time: Result<i64, E>,
+    end_time: Result<i64, E>,
+    duration_precision: u8,
+) -> u64 {
+    if explicit != 0 {
+        return explicit / 10u64.pow(u32::from(9_u8.saturating_sub(duration_precision)));
```
amokan:
```suggestion
        return (explicit / 10u64.pow(u32::from(9_u8.saturating_sub(duration_precision)))).max(1);
```

Clamps the scaled result so an explicit sub-unit duration doesn't land on `0`

native/mapper_ex/src/mapper.rs
```diff
@@ -1342,20 +1344,16 @@ impl Serialize for JsonTerm<'_> {
...
                 }
             }
             entries.sort_unstable_by(|(left, _), (right, _)| left.as_slice().cmp(right.as_slice()));
 
             let mut map = serializer.serialize_map(Some(entries.len()))?;
             for (binary, value) in entries {
-                let key = std::str::from_utf8(binary.as_slice()).expect("validated UTF-8 key");
                 map.serialize_entry(
-                    key,
+                    &String::from_utf8_lossy(binary.as_slice()),
```
amokan: This collapses distinct non-UTF-8 keys and serde will not dedupe.

What about sorting on the converted key and then dedupe? (_Lossy-equal keys aren't adjacent under the current byte sort_)

```rust
entries.push((String::from_utf8_lossy(binary.as_slice()), value));
...
entries.sort_unstable_by(|(left, _), (right, _)| left.cmp(right));
entries.dedup_by(|(left, _), (right, _)| left == right);
```

Worth noting that `term_to_string_term` (`mapper.rs:1259`) still blanks non-UTF-8 scalars - so that should probably be covered as well.

native/mapper_ex/src/mapping.rs
amokan, line 424:
```suggestion
    };

    if let (Some(data), DefaultValue::Str(label)) = (&enum8_data, &default) {
        if !data.value_map.contains_key(&label.to_lowercase()) {
            return Err(format!(
                "enum8 field '{name}' default '{label}' is not one of its enum_values"
            ));
        }
    }
```

Catches a mislabeled enum8 default at compile time

amokan (changes requested):

native/mapper_ex/src/ndjson.rs
amokan: > This is actually a bug in the current implementation of clickhouse_rowbinary that silently assumes the duration is in ns even though it's extracted from duration_ms or duration_us field.

Yeah. This is a tough one as I don't even remember if we mapped those for a legit reason or if it was just inferred based on documented potential paths 🫠 

My thinking is that rather than have you die on that hill, I'd follow up this PR after it merges with some adjustments on that front.

lib/logflare/mapper/output_context.ex
amokan: Fair point for this effort. I'll probably do some follow-up after we merge this.

native/mapper_ex/src/mapping.rs
bblaszkow06: Needed to modify mapping_config.ex as well, as this was no-op: string labels were coerced to 0 before reaching the NIF

native/mapper_ex/src/ndjson.rs
bblaszkow06: Created a new issue for that:
https://linear.app/supabase/issue/O11Y-2543/mapper-trace-duration-coalesces-millisecond-and-microsecond-inputs

amokan (approved):

PR #3883, opened 2026-08-27
feat: read replica with URI

Ziinc:
Adds read replica support for postgres URIs, and adjust cachex fetch paths to prefer read replicas when available.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -537,15 +537,20 @@ config :logflare, :context_cache_gossip, %{
   max_nodes: cache_gossip_max_nodes
 }
 
-# LOGFLARE_READ_REPLICAS: Comma-separated list of PostgreSQL read replica hostnames to distribute
+# LOGFLARE_READ_REPLICAS: Comma-separated list of PostgreSQL read replicas to distribute
 # context cache queries across. If unset or empty, all queries go to the primary database.
-# Example: "replica1.example.com,replica2.example.com"
+# Each entry is either a bare hostname (inheriting the primary's port, credentials, database
+# and SSL settings) or a full URI, in which case only the parts present in the URI override
+# the primary's config: postgres://user:pass@host:port/database?ssl=true&pool_size=5
+# Example: "replica1.example.com,postgres://user:pass@replica2.example.com:5432/logflare"
 read_replicas =
   "LOGFLARE_READ_REPLICAS"
   |> System.get_env("")
   |> String.split(",", trim: true)
   |> Enum.map(&String.trim/1)
+  |> Enum.reject(&(&1 == ""))
   |> Enum.uniq()
+  |> Enum.map(&Logflare.Repo.Replicas.parse!/1)
 
 config :logflare, :read_replicas, read_replicas
 
--- a/docs/docs.logflare.com/docs/self-hosting/index.md
+++ b/docs/docs.logflare.com/docs/self-hosting/index.md
@@ -165,7 +165,7 @@ The configuration follows the [Erlang Security Working Group recommendations](ht
 
 ## Read Replicas
 
-`LOGFLARE_READ_REPLICAS` is a comma-separated list of PostgreSQL read replicas to distribute some context cache queries across. If unset or empty, all queries go to the primary database.
+`LOGFLARE_READ_REPLICAS` is a comma-separated list of PostgreSQL read replicas to distribute some ingest-path data fetching queries across. If unset or empty, all queries go to the primary database.
 
 Each entry is either a **bare hostname** or a **connection URI** (`postgres://user:pass@host:port/database?ssl=true&pool_size=5`). In both cases, only the parts given override the primary's `DB_*` settings - anything omitted (port, database, credentials, SSL, ...) is inherited from the primary. Query params: `ssl` (`true`/`false`), `pool_size` (positive integer).
 
--- a/docs/docs.logflare.com/docs/self-hosting/index.md
+++ b/docs/docs.logflare.com/docs/self-hosting/index.md
@@ -38,7 +38,7 @@ All browser authentication will be disabled when in single-tenant mode.
 | `DB_POOL_SIZE`                                 | Integer, defaults to `10`                                               | Overrides the Ecto connection pool size for Logflare's internal PostgreSQL database connection.                                                                                                                                                                      |
 | `DB_SCHEMA`                                    | String, defaults to `nil`                                               | Allows configuration of the database schema to scope Logflare operations.                                                                                                                                                                                            |
 | `DB_SSL`                                       | Boolean, defaults to `false`                                            | Enables SSL/TLS connection to the internal Logflare database. Requires certificate files when enabled. See [Database SSL Configuration](#database-ssl-configuration).                                                                                                |
-| `LOGFLARE_READ_REPLICAS`                       | String, defaults to `nil`                                               | Comma-separated list of read replica hostnames. Each replica is assumed to use the same port, credentials, and database name as the primary. If unset, all queries go to the primary. Example: `replica1.example.com,replica2.example.com`                           |
+| `LOGFLARE_READ_REPLICAS`                       | String, defaults to `nil`                                               | Comma-separated list of PostgreSQL read replicas. If unset, all queries go to the primary. See [Read Replicas](#read-replicas).                                                                                                                                     |
 | `LOGFLARE_LOG_LEVEL`                           | String, defaults to `info`. <br/>Options: `error`,`warning`, `info`     | Allows runtime configuration of log level.                                                                                                                                                                                                                           |
 | `LOGFLARE_NODE_HOST`                           | string, defaults to `127.0.0.1`                                         | Sets node host on startup, which affects the node name `logflare@<host>`                                                                                                                                                                                             |
 | `LOGFLARE_METADATA_CLUSTER`                    | string, defaults to `nil`                                               | Sets global logging/tracing metadata for the cluster name and affects the release node name (e.g., `logflare-production@<host>`). Useful for filtering logs by cluster name and distinguishing nodes in multi-cluster setups. See the [metadata](#Metadata) section. |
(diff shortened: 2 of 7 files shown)


lib/logflare/context_cache.ex
```diff
@@ -51,7 +51,7 @@ defmodule Logflare.ContextCache do
     cache_key = {fun, args}
 
     fetch(cache, cache_key, fn ->
-      Logflare.Repo.apply_with_random_repo(context, fun, args)
+      Logflare.Repo.apply_with_replica(context, fun, args)
```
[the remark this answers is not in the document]
djwhitt: @Ziinc I think this one is a legit issue. Might be rare, but my read is that WAL based cache invalidation can happen before the change is replicated. A stale read could then repopulate the cache with the old key.

lib/logflare/repo/replicas.ex
```diff
@@ -43,13 +56,127 @@ defmodule Logflare.Repo.Replicas do
...
+  end
+
+  defp build_key(config) do
+    "#{config[:hostname]}-#{:erlang.phash2(config)}"
+  end
+
+  defp redact(entry) do
+    if String.contains?(entry, "://") do
+      uri = URI.parse(entry)
+      URI.to_string(%{uri | userinfo: if(uri.userinfo, do: "REDACTED")})
```
djwhitt, line 141: :robot:

> **Medium:** Hostless replica URIs are intentionally supported—for example, `postgres:///db` inherits the primary hostname—but redaction fails when a hostless URI contains credentials.
>
> This can realistically occur when an operator overrides the replica credentials or database while inheriting `DB_HOSTNAME`, such as `postgres://user:password@/replica_db`. It can also result from an empty Helm, Terraform, or Docker interpolation—for example, `postgres://user:password@${REPLICA_HOST}/db` becoming `postgres://user:password@/db`.
>
> In that case, `Replicas.parse!/1` raises an error whose message still contains the password. When `host` is nil, changing only `URI.userinfo` does not replace the original `authority` serialized by `URI.to_string/1`, potentially exposing credentials in startup logs.
>
> Reconstruct or redact the authority as well, and add a regression test using a hostless URI with userinfo.

lib/logflare/context_cache.ex
djwhitt, line 54: :robot:

> **High:** This routes every `ContextCache` miss to a replica whenever one is configured. Primary-WAL invalidation can run before the replica applies the corresponding change, allowing the next miss to retrieve and cache the pre-change value. The stale result then survives after replication catches up because there is no subsequent invalidation.
>
> This affects more than OAuth tokens: a rotated legacy API key can remain accepted for the `Users.Cache` TTL, an endpoint changed from public to authenticated can remain anonymously accessible, and an early lookup for a newly created source can cache `nil` and reject ingestion.
>
> Keep primary/strong consistency as the default and require callers to opt explicitly into eventual replica reads, for example:
>
> ```elixir
> ContextCache.fetch_or_apply(
>   Sources,
>   :get_by,
>   [kv],
>   consistency: :eventual
> )
> ```
>
> Eventual paths should confirm negative or recently tombstoned results against the primary before caching them and fall back to the primary when the selected replica is unavailable. Add regression coverage for stale authorization, stale API keys, endpoint authorization changes, and newly created sources.

lib/logflare/context_cache.ex
djwhitt, line 1: :robot:

> **Non-blocking:** `apply_fun/3` obscures that this is a read-through cache operation which dynamically invokes a function only on a miss. Rename it to `fetch_or_apply/3` (and `/4` if consistency options are added):
>
> ```elixir
> ContextCache.fetch_or_apply(context, fun, args)
> ```
>
> This also makes the distinction from the lower-level `fetch/3` clearer while retaining the connection to `Kernel.apply/3`.

djwhitt: Not that important to address this. I just found the name a little confusing when reviewing.

lib/logflare/repo/replicas.ex
```diff
@@ -43,13 +56,127 @@ defmodule Logflare.Repo.Replicas do
...
+            |> Ecto.Repo.Supervisor.parse_url()
+
+          _ ->
+            # no database set - use a placeholder to satisfy Ecto's URL parser,
+            # then drop it so the primary's database is inherited instead
+            URI.to_string(%{uri | path: "/placeholder"})
+            |> Ecto.Repo.Supervisor.parse_url()
+            |> Keyword.delete(:database)
+        end
+        |> Keyword.delete(:scheme)
```
djwhitt, line 118: :robot:

> **Medium:** URI parsing currently accepts invalid replica connection strings instead of failing configuration validation. For example, `mysql://user:pass@host/db` is accepted because any value containing `://` enters `parse_uri/1` and the parsed scheme is subsequently discarded. Likewise, `postgres://user:pass@host:abc/db` is accepted with no parsed port, causing it to inherit the primary port silently.
>
> These configuration errors can direct reads to an unintended endpoint or fail only when the replica is first used. Restrict schemes to `postgres` and `postgresql`, validate malformed authorities and ports before building the config, and add rejection tests for arbitrary schemes and nonnumeric ports while preserving bracketed IPv6 support.

lib/logflare/repo/replicas.ex
```diff
@@ -43,13 +56,127 @@ defmodule Logflare.Repo.Replicas do
...
+      entry
+    end
+  end
+
+  defp redact(message, nil), do: message
+  defp redact(message, userinfo), do: String.replace(message, userinfo, "REDACTED")
+
+  defp resolve_ssl(config) do
+    case Keyword.get(config, :ssl) do
+      true -> Keyword.put(config, :ssl, primary_ssl_opts(config[:hostname]))
```
djwhitt, line 152: :robot:

> **Medium:** Replica pools can inherit transport options that were derived from the primary hostname rather than the replica hostname. A DNS replica leaves `:socket_options` unset, so an IP primary’s `[:inet]` or `[:inet6]` setting survives Ecto’s configuration merge. Similarly, `resolve_ssl/1` adjusts SNI only when the replica URI explicitly contains `ssl=true`; when SSL is inherited as documented, it retains the primary host’s SNI configuration.
>
> This means an IP-primary/DNS-replica configuration can inherit a forced address family and `server_name_indication: :disable`, while a DNS-primary/IP-replica can inherit TLS options without the required IP-literal SNI suppression. Both cases can cause valid mixed DNS/IP or IPv4/IPv6 replica configurations to fail connection or TLS setup.
>
> Build the effective transport configuration for each replica hostname, explicitly overriding inherited socket options and recalculating SSL/SNI behavior. Add tests for IP-primary/DNS-replica and DNS-primary/IP-replica configurations with inherited SSL.

lib/logflare/repo.ex
```diff
@@ -59,7 +48,7 @@ defmodule Logflare.Repo do
   defp pick_replica_repo(replicas), do: replicas |> Enum.random() |> resolve_repo()
 
   defp resolve_repo(repo) when is_atom(repo), do: repo
-  defp resolve_repo(hostname) when is_binary(hostname), do: Replicas.lookup!(hostname)
+  defp resolve_repo({key, _config}), do: Replicas.lookup!(key)
```
djwhitt, line 51: :robot:

> **Medium:** A configured replica is treated as available solely because its pool is registered. Ecto/Postgrex pool processes can remain alive while their database connections are reconnecting, so `Replicas.lookup!/1` can succeed even though queries against the selected replica fail. A pool restart can also temporarily make `lookup!/1` raise `unknown replica`.
>
> Because context-cache misses now always select a replica when any are configured, these conditions cause requests to fail despite the primary database remaining healthy.
>
> Add health-aware selection or retry the operation against the primary when a replica lookup or connection fails. Keep that fallback inside the explicitly read-only eventual-consistency path rather than generically retrying arbitrary MFAs in `apply_with_replica/3`. Add regression coverage using an unavailable and restarting replica, asserting that the primary result is returned.

docs/docs.logflare.com/docs/self-hosting/index.md
```diff
@@ -163,6 +163,14 @@ The SSL connection is configured with:
 
 The configuration follows the [Erlang Security Working Group recommendations](https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/ssl).
 
+## Read Replicas
+
+`LOGFLARE_READ_REPLICAS` is a comma-separated list of PostgreSQL read replicas to distribute ingest-path data fetching queries across. If unset or empty, all queries go to the primary database.
```
djwhitt, line 168: :robot:

> **Non-blocking:** “Ingest-path data fetching queries” overstates and obscures the routing scope. `LOGFLARE_READ_REPLICAS` affects selected context-cache and key-value cache misses; ordinary `Logflare.Repo` reads continue using the primary.
>
> Describe the exact read categories routed to replicas and document their eventual-consistency and primary-fallback behavior once that contract is finalized. This prevents operators from assuming that all ingest reads—or all read-only queries—are distributed across replicas.

config/runtime.exs
```diff
@@ -537,15 +537,20 @@ config :logflare, :context_cache_gossip, %{
...
+# the primary's config: postgres://user:pass@host:port/database?ssl=true&pool_size=5
+# Example: "replica1.example.com,postgres://user:pass@replica2.example.com:5432/logflare"
 read_replicas =
   "LOGFLARE_READ_REPLICAS"
   |> System.get_env("")
   |> String.split(",", trim: true)
   |> Enum.map(&String.trim/1)
+  |> Enum.reject(&(&1 == ""))
   |> Enum.uniq()
+  |> Enum.map(&Logflare.Repo.Replicas.parse!/1)
```
djwhitt, line 553: :robot:

> **Non-blocking:** The repository tests pre-parse individual entries and write `Application.put_env(:logflare, :read_replicas, entries)` directly, so they do not exercise the production `LOGFLARE_READ_REPLICAS` contract implemented here: comma splitting, whitespace trimming, empty-entry removal, deduplication, and failure behavior.
>
> Consider extracting this pipeline into a testable `Replicas.parse_entries!/1` function and calling it from `runtime.exs`. Add cases for unset and whitespace-only input, mixed hostname/URI entries, duplicate entries, malformed URIs, and credential-redacted startup errors.

test/logflare/repo_test.exs
```diff
@@ -4,51 +4,53 @@ defmodule Logflare.RepoTest do
...
     end
 
-    test "uses replica repo during execution and reverts afterward" do
+    test "always uses a replica when replicas are configured" do
       start_read_replicas(["127.0.0.1", "::1"])
 
-      # since random repo choice includes both primary and replicas,
-      # we may need to perform several calls to hit a replica
-      repos = for _ <- 1..30, do: Repo.apply_with_random_repo(Repo, :get_dynamic_repo, [])
+      repos = for _ <- 1..30, do: Repo.apply_with_replica(Repo, :get_dynamic_repo, [])
```
djwhitt, line 50: :robot:

> **Non-blocking:** These repeated random calls prove only that the primary was not selected. A regression that always selects the first configured replica would still pass, so the test does not establish that all configured replica pools participate in selection.
>
> Avoid relying on probabilistic iteration. Inject or expose the replica-selection step so the test can deterministically exercise each configured entry, or test resolution of every configured key independently. This would also make the hostname-plus-URI case prove that both pool forms are selectable rather than merely started.

djwhitt (approved):
:robot:

> Approved to unblock testing. Outstanding findings remain documented in the open review threads for follow-up.

PR #3888, opened 2026-08-28
feat: upload parquet via PutObject-only so the S3 backend needs no longer requires GetObject permission

amokan:
The S3 backend previously required `s3:GetObject` (_and `s3:ListBucket`_) because Explorer's `CloudWriter` issues a `HeadObject` after every cloud write. With a write-only policy the bytes landed in S3, but the NIF reported failure and `test_connection/1` broke.

This PR keeps Explorer for the parquet *format* but drops it for *transport*: dataframes are serialized in-memory with `DataFrame.dump_parquet/1` and uploaded via `ExAws.S3.put_object` with per-request config overrides. This allows us to lower the permission requirements on the S3 backend to only `s3:PutObject`.

## Key Changes

- **`S3Adaptor`**: both call sites (`test_connection/1`, `push_log_events_to_s3/2`) now go through a new `put_parquet/3` (_in-memory parquet + `PutObject`_); `fss_s3_config/1` removed; `new_s3_filename/2` → `new_s3_key/1` (_bare key, ExAws takes bucket separately_).
- **Endpoint path prefixes preserved**: endpoints that carry a path (_e.g. Supabase Storage's `https://<ref>.supabase.co/storage/v1/s3`_) are folded into the request path, matching the old `object_store` behavior - `{endpoint}/{bucket}/{key}` since ExAws config has no base-path concept.
- **Retry visibility**: our S3 requests emit a custom telemetry event carrying `backend_id` and bucket; a handler attached at app start logs a warning whenever ExAws retries an attempt, so a flaky provider is traceable to the specific drain.
- **Tests**: mocked describes rewritten to stub `ExAws.request/2` and assert on the real operation struct (_probe key, key shape, parquet magic bytes, credential/endpoint overrides, path-bearing endpoints, telemetry options, 403 path, panic rescue_) plus retry-logger tests.
- **Docs**: `s3.mdx` now documents the PutObject-only IAM policy, SSE-KMS note, key layout, and region values for non-AWS providers.


docs/docs.logflare.com/docs/backends/s3.mdx
```diff
@@ -22,9 +22,30 @@ The following values are required when creating a S3 backend:
...
+      "Action": "s3:PutObject",
+      "Resource": "arn:aws:s3:::YOUR_BUCKET/*"
+    }
+  ]
+}
+```
+
+For buckets encrypted with SSE-KMS, the credential additionally needs `kms:GenerateDataKey` on the KMS key.
+
+Objects are written to keys of the form `<source_token>/<unix_microseconds>.parquet`, where `source_token` is the source UUID with dashes replaced by underscores. The backend connection test also writes a small sentinel file to `_connection_test.parquet` at the bucket root, so a policy scoped to a key prefix must still allow writing that root key.
```
Ziinc:
```suggestion
Objects are written to keys of the form `<source uuid>/<unix_microseconds>.parquet`, where the source UUID has dashes replaced by underscores. The backend connection test also writes a small sentinel file to `_connection_test.parquet` at the bucket root, so a policy scoped to a key prefix must still allow writing that root key.
```

lib/logflare/backends/adaptor/s3_adaptor.ex
```diff
@@ -249,13 +254,38 @@ defmodule Logflare.Backends.Adaptor.S3Adaptor do
...
+        put_parquet(df, config, s3_key, backend_id)
       rescue
         error -> {:error, error}
       end
     end
   end
 
+  @doc false
+  @spec attach_request_logger() :: :ok | {:error, :already_exists}
+  def attach_request_logger do
```
Ziinc: for runtime debugging? might want a detach function as well, and to add a @doc that it is specifically for runtime usage.

Ziinc (approved):

lib/logflare/backends/adaptor/s3_adaptor.ex
amokan: Removed. Will revisit a better long-term solution later.

lib/logflare/backends/adaptor/s3_adaptor.ex
```diff
@@ -287,12 +289,46 @@ defmodule Logflare.Backends.Adaptor.S3Adaptor do
...
+      region: config.storage_region
+    ] ++ endpoint_opts(config[:endpoint])
+  end
+
+  @spec endpoint_opts(String.t() | nil) :: keyword()
+  defp endpoint_opts(nil), do: []
+
+  defp endpoint_opts(endpoint) when is_non_empty_binary(endpoint) do
+    %URI{scheme: scheme, host: host, port: port} = URI.parse(endpoint)
+    [scheme: "#{scheme}://", host: host, port: port]
```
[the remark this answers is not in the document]
amokan: Overstated impact and suggested fix is broken.

> "sends the access key, secret key ... over plaintext"

secret access key is never transmitted under SigV4

PR #3891, opened 2026-08-31
fix: Require hostname in read replica URIs and improve error handling

Ziinc:
Require hostname to be provided, addressing comments in #3883

### DIFF
--- a/lib/logflare/repo/replicas.ex
+++ b/lib/logflare/repo/replicas.ex
@@ -110,27 +110,25 @@ defmodule Logflare.Repo.Replicas do
   end
 
   defp parse_uri_with_host(entry, uri) do
-    try do
-      config =
-        case uri.path do
-          v when v not in [nil, "", "/"] ->
-            entry
-            |> Ecto.Repo.Supervisor.parse_url()
+    config =
+      case uri.path do
+        v when v not in [nil, "", "/"] ->
+          entry
+          |> Ecto.Repo.Supervisor.parse_url()
 
-          _ ->
-            # no database set - use a placeholder to satisfy Ecto's URL parser,
-            # then drop it so the primary's database is inherited instead
-            URI.to_string(%{uri | path: "/placeholder"})
-            |> Ecto.Repo.Supervisor.parse_url()
-            |> Keyword.delete(:database)
-        end
-        |> Keyword.delete(:scheme)
-        |> maybe_put_socket_options()
+        _ ->
+          # no database set - use a placeholder to satisfy Ecto's URL parser,
+          # then drop it so the primary's database is inherited instead
+          URI.to_string(%{uri | path: "/placeholder"})
+          |> Ecto.Repo.Supervisor.parse_url()
+          |> Keyword.delete(:database)
+      end
+      |> Keyword.delete(:scheme)
+      |> maybe_put_socket_options()
 
-      {:ok, {build_key(config), config}}
-    rescue
-      e in Ecto.InvalidURLError -> {:error, redact(e.message, uri.userinfo)}
-    end
+    {:ok, {build_key(config), config}}
+  rescue
+    e in Ecto.InvalidURLError -> {:error, redact(e.message, uri.userinfo)}
   end
 
   defp maybe_put_socket_options(config) do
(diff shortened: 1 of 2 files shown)


lib/logflare/repo/replicas.ex
```diff
@@ -136,12 +143,7 @@ defmodule Logflare.Repo.Replicas do
   end
 
   defp redact(entry) do
-    if String.contains?(entry, "://") do
-      uri = URI.parse(entry)
-      URI.to_string(%{uri | userinfo: if(uri.userinfo, do: "REDACTED")})
-    else
-      entry
-    end
+    Regex.replace(~r{^([a-z][a-z0-9+.-]*://)[^/@]*@}i, entry, "\\1")
```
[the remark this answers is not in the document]
djwhitt: @Ziinc This one looks like a legit issue to me.

lib/logflare/repo/replicas.ex
```diff
@@ -172,12 +178,7 @@ defmodule Logflare.Repo.Replicas do
   end
 
   defp redact(entry) do
-    if String.contains?(entry, "://") do
-      uri = URI.parse(entry)
-      URI.to_string(%{uri | userinfo: if(uri.userinfo, do: "REDACTED")})
-    else
-      entry
-    end
+    Regex.replace(~r{^([a-z][a-z0-9+.-]*://)[^/]*@}i, entry, "\\1")
```
amokan, line 181: 🤖 

> `redact/1`'s regex requires `@` before the first `/`, so an unescaped `/` in a password defeats both redaction layers. `postgres://u:pa/ss@host/db` parses as `host: "u"`, `userinfo: nil`, so the host guard passes, Ecto raises `path should be a database name` from `config/runtime.exs:585`, and the full password lands in startup logs. **Fix:** validate that the entry round-trips (`URI.to_string(URI.parse(entry)) == entry`) and reject with a fixed string that never echoes the entry, plus a fallback `~r{^([a-z][a-z0-9+.-]*://).*@}I` in `redact/1` when `userinfo == nil`.

lib/logflare/repo/replicas.ex
```diff
@@ -137,29 +137,35 @@ defmodule Logflare.Repo.Replicas do
...
-            |> Ecto.Repo.Supervisor.parse_url()
-            |> Keyword.delete(:database)
-        end
-        |> Keyword.delete(:scheme)
-        |> maybe_put_socket_options()
-
-      {:ok, {build_key(config), config}}
-    rescue
-      e in Ecto.InvalidURLError -> {:error, redact(e.message, uri.userinfo)}
+    if uri.host in [nil, ""] do
```
amokan, line 140: Could invert the `if` and use the `is_non_empty_binary` guard.

lib/logflare/repo/replicas.ex
```diff
@@ -137,29 +137,35 @@ defmodule Logflare.Repo.Replicas do
...
+      {:error, "hostname is required"}
+    else
+      parse_uri_with_host(entry, uri)
     end
   end
```
amokan, line 145:
```suggestion
    cond do
      uri.host in [nil, ""] ->
        {:error, "hostname is required"}
  
      URI.to_string(uri) != downcase_scheme(entry) ->
        {:error,
         "malformed URI - reserved characters in the username or password must be percent-encoded"}
  
      true ->
        parse_uri_with_host(entry, uri)
    end
  end

  defp downcase_scheme(entry) do
    case String.split(entry, "://", parts: 2) do
      [scheme, rest] -> String.downcase(scheme) <> "://" <> rest
      _ -> entry
    end
  end
```

This has to land with a `redact/1` change or it just moves the leak, since `parse!/1` interpolates `redact(entry)` and the current regex still can't match these entries. At `replicas.ex:181`:

```elixir
defp redact(entry) do
  case Regex.replace(~r{^([a-z][a-z0-9+.-]*://)[^/?#]*@}i, entry, "\\1") do
    ^entry -> Regex.replace(~r{^([a-z][a-z0-9+.-]*://).*@}i, entry, "\\1")
    redacted -> redacted
  end
end
```

lib/logflare/repo/replicas.ex
amokan, line 181: `[^/]*@` is greedy and runs past the authority into the query when there's no path: `postgres://u:pass@host?options=a@b&pool_size=abc` → `"postgres://b&pool_size=abc"`. Creds are gone but so is the host, so you can't tell which replica failed. One-char fix: `[^/?#]*@`.

lib/logflare/repo/replicas.ex
amokan, line 9: Module doc could use an update with these changes indicating _"hostname is required for URI entries"_ or something similar.

Same sort of treatment to `docs/docs.logflare.com/docs/self-hosting/index.md:172` and `config/runtime.exs:575`

amokan (changes requested):

PR #3892, opened 2026-08-31
perf: sample Broadway message duration metric

djwhitt:
## Summary

`broadway.processor.message.stop.duration` currently records every processed message. A dominant production histogram bucket exceeds **80,000 observations/second on each node**.

Observations in that bucket atomically update the same row in `OtelMetricExporter.MetricStore`. Benchmarks and OTP lock-counting profiles show contention rising sharply around 60,000–80,000 retained writes/second.

This PR samples the metric at approximately **1% by default**, reducing an 80,000 writes/second bucket to roughly **800 writes/second/node**.

## Changes

- Sample `broadway.processor.message.stop.duration` using a stateless hash of Broadway's `telemetry_span_context`.
- Apply one startup-configured sampling rate to all Broadway topologies on a node.
- Leave the processor, batcher, and batch-processor duration metrics unsampled.
- Allow full retention or complete removal of the per-message metric.
- Ignore sampled events that do not contain the expected span context rather than detaching the telemetry handler.
- Add configuration and sampling tests plus reproducible S3, ClickHouse, and no-op benchmarks.

The exported histogram count represents **sampled observations**, not total processed messages. No count scaling is applied.

## Why 1%?

The acceptance budget was no more than 5% median throughput loss and 5% median reductions/event growth relative to omitting the per-message handler.

| Workload | 1% throughput Δ | 1% reductions/event Δ |
|---|---:|---:|
| ClickHouse log | -2.46% | +2.34% |
| ClickHouse metric | -2.92% | +1.31% |
| ClickHouse trace | +1.48% | +2.66% |
| S3 spool, three runs | -1.88% to +11.54% | +3.32% to +3.57% |

Representative ClickHouse and S3 workloads remained within budget at 1%. The S3 workload exceeded the reductions budget at 10%, making 1% the highest tested rate that passed.

The no-op workload intentionally magnifies fixed handler and predicate costs and showed approximately 14% throughput loss at 1%. It is an upper bound rather than a representative production pipeline.

OTP `:lcnt` profiling also showed the contention threshold directly in the no-op workload:

| Retention | Retained writes/s | `db_hash_slot` collisions/block | Lock wait/block |
|---:|---:|---:|---:|
| 1% | ~16,100 | 19 | 4 µs |
| 10% | ~61,700 | 83,398 | 2.76 s |
| 100% | ~80,400 | 990,944 | 32.34 s |

The collisions were attributed to the metric store's named ETS table; its table-coordination lock remained uncontended. Absolute throughput and the contention threshold are machine-specific, and the lock-counting emulator is used for attribution rather than production-cost estimates.

## Configuration

`LOGFLARE_BROADWAY_MESSAGE_SAMPLE_DENOMINATOR` is read once while metric definitions are built:

- unset or blank: `100` (approximately 1%)
- `1`: retain every observation
- `N` from `2` through `4294967296`: retain approximately one observation per `N`
- `disabled`: omit the metric and its handler

Nodes contributing to the same aggregated histogram should use the same denominator.

## Deployment expectations

After all nodes use the default denominator of `100`:

- The target histogram's count, sum, and bucket counts should fall by approximately 99%; they no longer represent total message volume.
- High-volume latency distributions should remain broadly comparable, while low-volume series may become noisier.
- Other Broadway duration metrics should be unchanged.
- Metric-store contention and scheduler work should decrease, although the end-to-end CPU and throughput improvement is workload-dependent.

During a rolling deployment, upgraded and unupgraded nodes will contribute at different sampling rates. Aggregated counts and distribution weighting should not be interpreted until all nodes use the same denominator.

Post-deployment checks should compare scheduler utilization, Broadway queueing or latency, and the sampled histogram shape against the pre-deployment baseline. Any dashboards or alerts that use this histogram's count or sum must account for the approximately 100× reduction. The rate can be changed at startup; `1` restores full retention and `disabled` removes the metric entirely.

## Validation

Tests cover:

- custom, full-retention, and disabled configurations
- invalid denominator rejection
- deterministic sampling from the span context
- omission of malformed sampled events
- preservation of the other Broadway duration metrics

Detailed methodology, limitations, and reproduction instructions are in `bench/broadway_metric_sampling.md` and the accompanying benchmark scripts.


config/runtime.exs
```diff
@@ -73,6 +73,24 @@ cache_stats =
     val -> val |> String.downcase() |> String.to_existing_atom()
   end
 
+case System.get_env("LOGFLARE_BROADWAY_MESSAGE_SAMPLE_DENOMINATOR") do
+  nil ->
+    :ok
+
+  "disabled" ->
```
amokan: 🤖 

> only lowercase "disabled" parses (_the adjacent `LOGFLARE_CACHE_STATS` block downcases first_), and a blank or whitespace-padded value raises at boot — and `.template.env` ships the var blank, so a docker-compose `env_file` copy won't boot.

Could trim/downcase the value and then just treat `""` as unset 🤷‍♂️

test/logflare/telemetry_test.exs
```diff
@@ -17,6 +17,11 @@ defmodule Logflare.TelemetryTest do
     :batch_size
   ]
   @clickhouse_batch_metric_string "logflare.backends.clickhouse.pipeline.handle_batch.batch_size"
+  @broadway_processor_message_event [:broadway, :processor, :message, :stop]
+  @broadway_processor_message_exporter :logflare_broadway_processor_message_metrics_test
+  @broadway_processor_message_metric_name [:broadway, :processor, :message, :stop, :duration]
+  @broadway_processor_message_metric_string "broadway.processor.message.stop.duration"
+  @broadway_processor_message_sample_denominator 100
```
amokan: This is hardcoded to 100, but in `telemetry.ex` we fetch `:broadway_message_sample_denominator` from env vars. Aside from this,  `runtime.exs` populates it from `LOGFLARE_BROADWAY_MESSAGE_SAMPLE_DENOMINATOR`.

So lots of diverging paths on this to where a shell-exported value would fail the sampling assertion. If set to disabled, we'd end up producing match errors.

bench/run_broadway_metric_sampling.sh
```diff
@@ -0,0 +1,136 @@
...
+    ORDER_INDEX="$order_index" \
+    EVENTS="$EVENTS" \
+    WARMUPS="$WARMUPS" \
+    TRIALS="$TRIALS" \
+    PROCESSOR_CONCURRENCY="$PROCESSOR_CONCURRENCY" \
+    BATCHER_CONCURRENCY="$BATCHER_CONCURRENCY" \
+    BATCH_SIZE="$BATCH_SIZE" \
+    BATCH_TIMEOUT_MS="$BATCH_TIMEOUT_MS" \
+    PAYLOAD_SHAPE="$PAYLOAD_SHAPE" \
+    ../bin/x mix run --no-start "$BENCH_FILE" > "$output" 2>&1
```
amokan: I don't think this will work on another machine. Not sure if it is expected to, but just calling that out.

lib/telemetry.ex
```diff
@@ -170,12 +195,27 @@ defmodule Logflare.Telemetry do
...
+    broadway_processor_message_metrics =
+      case broadway_processor_message_sample_denominator!() do
+        :disabled ->
+          []
+
+        denominator ->
+          [
+            distribution("broadway.processor.message.stop.duration",
+              unit: {:native, :millisecond},
+              keep: &sample_broadway_processor_message_duration?(&1, denominator)
```
amokan, line 213: 🤖 

> `denominator=1` still installs the always-true phash2 `keep:` closure, while both benchmarks rewrote that case to `keep: nil`. The published "1" column therefore measured a cheaper variant than production runs (the predicate alone is ~16% reductions/event in the noop scenario). Fix: add a `1 ->` clause that emits the distribution without `keep:`.

amokan (changes requested):

lib/telemetry.ex
```diff
@@ -86,6 +87,30 @@ defmodule Logflare.Telemetry do
...
+
+  defp sample_broadway_processor_message_duration?(_metadata, _denominator), do: false
+
+  defp broadway_processor_message_sample_denominator! do
+    case Application.fetch_env!(:logflare, :broadway_message_sample_denominator) do
+      :disabled ->
+        :disabled
+
+      denominator
+      when is_integer(denominator) and denominator > 0 and denominator <= @max_phash2_range ->
```
amokan: Nit. Just be sure to update the import statement at the top of the module to include `is_pos_integer` or drop the `only` clause and just import all the guards.

```suggestion
      when is_pos_integer(denominator) and denominator <= @max_phash2_range ->
```

config/runtime.exs
```diff
@@ -3,10 +3,36 @@ import Config
...
   def get_boolean(env, default \\ false) when is_boolean(default) do
     value = System.get_env(env)
     if value, do: value |> String.downcase() |> String.to_existing_atom(), else: default
   end
+
+  @spec parse_broadway_message_sample_denominator(String.t() | nil) ::
+          :default | :disabled | pos_integer()
+  def parse_broadway_message_sample_denominator(nil), do: :default
+
+  def parse_broadway_message_sample_denominator(value) when is_binary(value) do
```
amokan, line 17: 🤖 Probably not a terrible follow-up

> `config/runtime.exs:17` + `lib/telemetry.ex:99` — the parser lives in the `Env` module inside `config/runtime.exs`, and `test/logflare/telemetry_configuration_test.exs:30` calls it directly. That module isn't compiled into the app, so Credo/Dialyzer can't see it; the test only works because Mix evaluates `runtime.exs` first. The `4_294_967_296` bound and error message are duplicated in `lib/telemetry.ex:105` and can drift.

May be worth creating a `Logflare.Telemetry.Sampling` module 🤷‍♂️

test/logflare/telemetry_test.exs
```diff
@@ -94,6 +98,88 @@ defmodule Logflare.TelemetryTest do
...
+
+            distributions ->
+              distributions
+              |> Map.fetch!(%{})
+              |> Map.values()
+              |> Enum.sum_by(fn {count, _sum} -> count end)
+          end
+
+        malformed_metadata_count = if denominator == 1, do: 1, else: 0
+        assert observed_count == length(accepted_contexts) + malformed_metadata_count
```
amokan, line 179: 🤖 

> asserts an exact global observation count while `async: true` against a process-global `:telemetry` handler. Any concurrent test driving a Broadway pipeline inflates the count. **Fix:** `async: false`, or `assert >= expected`.

amokan (approved):
Approved. Noted a couple minor things, but nothing that I'd consider blocking.

PR #3893, opened 2026-08-31
feat: add ndjson payload format option to webhook backend

chasers:
## Human summary
- adds ndjson option to send ndjson payloads for webhook backends
- testing a connection tests the correct config
- gitignores my .plans folder

(custom header fields, the header-wipe fix, and the compression toggle moved to the stacked PR #3894)

## Summary

The webhook backend gains a `format` config option:

- `json` (default) — the current behavior, a JSON array of event bodies.
- `ndjson` — newline-delimited JSON, one event body per line, sent with the `application/x-ndjson` content type. A user-configured `content-type` header takes precedence.

## Design

The option is implemented through `transform_config/1`, the existing mechanism for payload shape and derived headers (the same pattern LokiAdaptor uses). When `format` is `ndjson`, `transform_config/1` sets `format_batch` and injects the content-type header. Because of this:

- The shared Broadway Pipeline is unchanged.
- `test_connection/1` resolves config the same way, so the connection test sends a format-correct probe (empty ndjson body with the `application/x-ndjson` header) instead of a JSON-array probe.
- Wrapper adaptors (Datadog, Elastic, Loki, and others) never cast `format`, so their behavior does not change.

The form's payload format select is sourced from `WebhookAdaptor.formats/0`, shared with `validate_config/1`. The webhook backend docs describe the `format` and `gzip` config options.

## Testing

`mix test test/logflare/backends/adaptor/webhook_adaptor_test.exs` — 8 doctests, 40 tests, 0 failures. New tests cover config cast/validation, the NDJSON body shape and content type through the full pipeline, precedence of a user-configured `content-type` header, and the ndjson connection probe.

Stacked follow-up: #3894 adds the custom header inputs, the header-wipe fix, the gzip checkbox, and poison-event drop-and-log.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,5 @@ flake.nix
 flake.lock
 
 /tmp
+
+.plans/
--- a/lib/logflare/backends/adaptor/webhook_adaptor.ex
+++ b/lib/logflare/backends/adaptor/webhook_adaptor.ex
@@ -34,6 +34,7 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
   alias Logflare.Backends.Adaptor
   alias Logflare.Backends.Backend
   alias Logflare.Backends.Adaptor.HttpBased.Headers
+  alias Logflare.Backends.Adaptor.HttpBased.NdjsonFormatter
   alias Logflare.LogEvent
   alias Logflare.Utils
   alias Logflare.Utils.SSRF
@@ -201,25 +202,43 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
 
   @impl Logflare.Backends.Adaptor
   @spec transform_config(Backend.t()) :: map()
-  def transform_config(%Backend{config: config}) do
-    if Map.get(config, :format) == "ndjson" do
-      headers =
-        (Map.get(config, :headers) || %{})
-        |> Map.put_new("content-type", "application/x-ndjson")
+  def transform_config(%Backend{config: %{format: "ndjson"} = config}) do
+    config
+    |> Map.put(:format_batch, &encode_ndjson/1)
+    |> Map.put(:headers, ndjson_headers(config))
+  end
 
-      config
-      |> Map.put(:format_batch, &encode_ndjson/1)
-      |> Map.put(:headers, headers)
-    else
-      config
+  def transform_config(%Backend{config: config}), do: config
+
+  # Normalizes the stored header names before adding the content type. Header names
+  # are case-insensitive, and config written before Headers.normalize_keys/1 (or
+  # through a partial update that never casts :headers) can still hold "Content-Type".
+  # Without the normalization both keys ship as separate headers.
+  @spec ndjson_headers(map()) :: map()
+  defp ndjson_headers(config) do
+    (Map.get(config, :headers) || %{})
+    |> Headers.normalize_keys()
+    |> Map.put_new("content-type", "application/x-ndjson")
+  end
+
+  @doc """
+  Builds the request body for a batch of log events.
+
+  Uses the config's `:format_batch` function when one is set, and otherwise falls
+  back to the list of event bodies.
+  """
+  @spec format_payload(map(), [LogEvent.t()]) :: term()
+  def format_payload(config, events) do
+    case Map.get(config, :format_batch) do
+      nil -> Enum.map(events, & &1.body)
+      format_batch -> format_batch.(events)
     end
   end
 
   @spec encode_ndjson([LogEvent.t()]) :: binary()
   defp encode_ndjson(events) do
     events
-    |> Enum.map(fn %LogEvent{body: body} -> Jason.encode_to_iodata!(body) end)
-    |> Enum.intersperse("\n")
+    |> NdjsonFormatter.encode()
     |> IO.iodata_to_binary()
   end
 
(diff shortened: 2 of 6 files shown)


lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -176,6 +193,36 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
...
+  @doc """
+  Payload formats this adaptor supports.
+  """
+  @spec formats() :: [String.t()]
+  def formats, do: @formats
+
+  @impl Logflare.Backends.Adaptor
+  @spec transform_config(Backend.t()) :: map()
+  def transform_config(%Backend{config: config}) do
+    if Map.get(config, :format) == "ndjson" do
```
amokan: 🤖 

> `transform_config/1` adds a lowercase "content-type" header via `Map.put_new` without checking case-insensitively. Trigger: a partial config update via the management API (_e.g. PATCH config: `{"format": "ndjson"}`_) that doesn't touch `:headers` — the changeset diff never includes `:headers`, so `normalize_header_keys/1` (_added in 9d4eae2d5, ~4 weeks ago_) never runs, and a pre-existing "Content-Type" key survives untouched. Since ndjson bodies are raw binaries, `Client.json_encodable?/1` is false, so `reserved_header_names/1` never marks content-type reserved — both differently-cased headers ship as separate HTTP headers, which Headers' own moduledoc warns some receivers can't parse.
Fix options: (a) make the `Map.put_new` case-insensitive against existing keys; (b) run `normalize_header_keys/1` unconditionally on the merged header set in `transform_config/1` instead of gating on whether `:headers` changed in this request.

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -176,6 +193,36 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
...
+      config
+      |> Map.put(:format_batch, &encode_ndjson/1)
+      |> Map.put(:headers, headers)
+    else
+      config
+    end
+  end
+
+  @spec encode_ndjson([LogEvent.t()]) :: binary()
+  defp encode_ndjson(events) do
```
amokan, line 239: 🤖 

> `encode_ndjson/1` reimplements logic that already exists in `lib/logflare/backends/adaptor/http_based/ndjson_formatter.ex:39-43` (_and a third variant in `bigquery_adaptor/google_api_client.ex:19-25`_) — and misses the content-type-collision fix the existing formatter already has. Fix: reuse/adapt `ndjson_formatter.ex`'s helper instead of a new implementation.

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -216,7 +263,15 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
   @impl Logflare.Backends.Adaptor
   @spec test_connection(Backend.t()) :: :ok | {:error, term()}
   def test_connection(%Backend{} = backend) do
-    test_connection(backend, [])
+    config = transform_config(backend)
+
+    body =
+      case Map.get(config, :format_batch) do
```
amokan: Prob should extract this logic to a private helper function as I see similar in `test_connection/1` and `Pipeline.handle_batch/4`

lib/logflare/backends/adaptor/webhook_adaptor.ex
```diff
@@ -176,6 +193,36 @@ defmodule Logflare.Backends.Adaptor.WebhookAdaptor do
...
 
+  @doc """
+  Payload formats this adaptor supports.
+  """
+  @spec formats() :: [String.t()]
+  def formats, do: @formats
+
+  @impl Logflare.Backends.Adaptor
+  @spec transform_config(Backend.t()) :: map()
+  def transform_config(%Backend{config: config}) do
```
amokan: Could get rid of some of that if logic by pattern matching...

`transform_config(%Backend{config: %{format: "ndjson"}} = ...)`

.gitignore
```diff
@@ -109,3 +109,5 @@ flake.nix
 flake.lock
 
 /tmp
+
+.plans/
```
amokan, line 113: Totally not against this, but one thing I do for these sort of things is add them to my global gitignore. Unless everyone or many of us are using something that writes to this path 🤷‍♂️

amokan (changes requested):

.gitignore
chasers: I figure lets just lean into it here why not.

lib/logflare/backends/adaptor/webhook_adaptor.ex
chasers: I found this too but wanted to keep this PR tightly scoped but can address it.

lib/logflare/backends/adaptor/webhook_adaptor.ex
chasers: Hmm, okay will check.

chasers: Handled

lib/logflare/backends/adaptor/webhook_adaptor.ex
chasers: Handled.

lib/logflare/backends/adaptor/webhook_adaptor.ex
chasers: Done

amokan (approved):

PR #3894, opened 2026-08-31
feat: add custom header inputs to the webhook backend form

chasers:
## Summary

- The webhook form gains custom header key/value rows. This allows a bearer token from the UI: key `authorization`, value `Bearer <token>`. The form renders one row per stored header, sorted by key, plus one blank row for a new entry. It keeps a minimum of 2 rows. Clearing a key removes the header.
- Stored header values render as the `REDACTED` sentinel, so the edit page HTML never carries a decrypted secret. Each row also carries a hidden `header<i>_stored_key`, so `transform_params/2` can resolve the sentinel by the key the row was rendered with. Renaming a key carries its stored value over. `WebhookAdaptor.unredact_headers/2` (merged in #3893) still covers the API path.
- The form gains a gzip checkbox. Some receivers do not decompress gzip request bodies, so the existing `gzip` config option is now visible in the UI. The box renders checked unless the value is explicitly false, so a backend created through the UI keeps the documented `true` default.
- `transform_params/1` derives the header indexes from the submitted params instead of a hardcoded `1..2` range, and only writes `headers` when the submitted form contains header fields. Before, every submit from a form without header inputs sent an empty headers map, which cleared headers set through the API.
- Poison-event handling: `NdjsonFormatter.encode/2` drops an event whose body cannot be JSON-encoded and logs the count per batch, with backend metadata. Before, one bad event raised and failed the whole batch. When every event in a batch is dropped, the pipeline skips the request instead of posting an empty body, and the formatter middleware returns `{:error, :all_events_dropped}` for the Google SecOps path.
- Docs: `webhook.mdx` documents the `headers` config option.

## Supporting changes

- `WebhookAdaptor.redacted_value/0` exposes the sentinel, so the form and the LiveView stop hardcoding the string.
- `Headers.normalize_key/1` for a single-key lookup. `normalize_keys/1` reuses it.
- `HttpBased.Client` gains `:formatter_opts`, so a formatter middleware can take options. It is omitted when unset, so a formatter that takes none stays a bare module.

## Review fixes

All four items from @amokan's review are fixed. Each one was reproduced with a probe test first.

| Item | Reproduced | Fix |
|---|---|---|
| Creating a webhook through the UI stored `gzip: false` | `CREATED VIA UI: %{gzip: false, ...}` | The checkbox renders checked unless the value is explicitly false. The edit path was already correct and rendered `checked`. |
| Renaming a header key dropped the credential | `HEADERS AFTER KEY RENAME: %{}` | Hidden `header<i>_stored_key` per row; `transform_params/2` resolves the sentinel against it. |
| A fully dropped batch posted an empty body | confirmed by reading both call paths | The pipeline skips the request and logs it. The middleware returns `{:error, :all_events_dropped}`. |
| The dropped-count log had no backend context | confirmed | The formatter takes metadata, from `encode/2` and from `:formatter_opts`. |

The OTLP and Loki header rows submit no `stored_key`, so `header_value/3` leaves them untouched and their behavior is unchanged.

**Why the original tests missed the first two.** Every test in the first round exercised the edit path and asserted only `updated.config.headers`. `backend_factory/1` writes config verbatim, so a stored `gzip: false` never failed an assertion. The header tests covered the unchanged, added, and cleared rows, but never a changed key.

## Testing

- `mix test test/logflare_web/live/backends/backends_live_test.exs` — 74 tests, 0 failures. New tests cover the gzip default on create, disabling gzip on create, keeping gzip off on edit, renaming a header key, and writing a header value the user actually changed.
- `mix test test/logflare/backends/adaptor/webhook_adaptor_test.exs` — 45 tests + 8 doctests, 0 failures. A new test asserts the request is skipped when every event is dropped.
- `mix test test/logflare/backends/adaptor/http_based/ndjson_formatter_test.exs` — 6 tests, 0 failures. New tests cover the log metadata, the all-dropped error, and an empty batch.
- `mix test test/logflare/backends/adaptor/google_secops_adaptor_test.exs test/logflare/backends/adaptor/http_based/ test/logflare/backends/adaptor/otlp_adaptor_test.exs` — 62 tests, 0 failures.

## Known limit

The OTLP header rows have the same secret-leak issue, but `OtlpAdaptor.cast_config` has no sentinel unredaction, so the same fix would corrupt stored OTLP headers on save. Left for a separate change.

https://claude.ai/code/session_01WPMHo5UmWRJnSVJcc36JJW

### DIFF
--- a/lib/logflare/backends/adaptor/http_based/ndjson_formatter.ex
+++ b/lib/logflare/backends/adaptor/http_based/ndjson_formatter.ex
@@ -15,11 +15,19 @@ defmodule Logflare.Backends.Adaptor.HttpBased.NdjsonFormatter do
   @content_type "application/json"
 
   @impl Tesla.Middleware
-  def call(env, next, _opts) do
-    env
-    |> put_content_type_header()
-    |> Tesla.put_body(encode(env.body))
-    |> Tesla.run(next)
+  def call(env, next, opts) do
+    metadata = Keyword.get(opts || [], :metadata, [])
+
+    case encode(env.body, metadata) do
+      [] when env.body != [] ->
+        {:error, :all_events_dropped}
+
+      body ->
+        env
+        |> put_content_type_header()
+        |> Tesla.put_body(body)
+        |> Tesla.run(next)
+    end
   end
 
   @doc """
@@ -43,10 +51,12 @@ defmodule Logflare.Backends.Adaptor.HttpBased.NdjsonFormatter do
   per line. Any other term passes through unchanged.
 
   An event whose body cannot be JSON-encoded is dropped. The count of dropped
-  events is logged per batch.
+  events is logged per batch, with `metadata` attached to the log entry.
   """
-  @spec encode([LogEvent.t()] | term()) :: iodata() | term()
-  def encode([%LogEvent{} | _] = events) do
+  @spec encode([LogEvent.t()] | term(), keyword()) :: iodata() | term()
+  def encode(events, metadata \\ [])
+
+  def encode([%LogEvent{} | _] = events, metadata) do
     {encoded, dropped_count} =
       Enum.reduce(events, {[], 0}, fn %LogEvent{body: body}, {acc, dropped_count} ->
         case Jason.encode_to_iodata(body) do
@@ -57,7 +67,8 @@ defmodule Logflare.Backends.Adaptor.HttpBased.NdjsonFormatter do
 
     if dropped_count > 0 do
       Logger.warning(
-        "Dropped #{dropped_count} log events from an NDJSON batch: JSON encoding failed"
+        "Dropped #{dropped_count} log events from an NDJSON batch: JSON encoding failed",
+        metadata
       )
     end
 
@@ -66,5 +77,5 @@ defmodule Logflare.Backends.Adaptor.HttpBased.NdjsonFormatter do
     |> Enum.intersperse("\n")
   end
 
-  def encode(term), do: term
+  def encode(term, _metadata), do: term
 end
--- a/lib/logflare/backends/adaptor/http_based/ndjson_formatter.ex
+++ b/lib/logflare/backends/adaptor/http_based/ndjson_formatter.ex
@@ -6,6 +6,8 @@ defmodule Logflare.Backends.Adaptor.HttpBased.NdjsonFormatter do
   `Logflare.Backends.Adaptor.HttpBased.Client.new/1` options.
   """
 
+  require Logger
+
   alias Logflare.LogEvent
 
   @behaviour Tesla.Middleware
@@ -39,11 +41,28 @@ defmodule Logflare.Backends.Adaptor.HttpBased.NdjsonFormatter do
   @doc """
   Encodes `Logflare.LogEvent`s as newline-delimited JSON iodata, one event body
   per line. Any other term passes through unchanged.
+
+  An event whose body cannot be JSON-encoded is dropped. The count of dropped
+  events is logged per batch.
   """
   @spec encode([LogEvent.t()] | term()) :: iodata() | term()
   def encode([%LogEvent{} | _] = events) do
-    events
-    |> Enum.map(fn %LogEvent{body: body} -> Jason.encode_to_iodata!(body) end)
+    {encoded, dropped_count} =
+      Enum.reduce(events, {[], 0}, fn %LogEvent{body: body}, {acc, dropped_count} ->
+        case Jason.encode_to_iodata(body) do
+          {:ok, iodata} -> {[iodata | acc], dropped_count}
+          {:error, _reason} -> {acc, dropped_count + 1}
+        end
+      end)
+
+    if dropped_count > 0 do
+      Logger.warning(
+        "Dropped #{dropped_count} log events from an NDJSON batch: JSON encoding failed"
+      )
+    end
+
+    encoded
+    |> Enum.reverse()
     |> Enum.intersperse("\n")
   end
 
--- a/lib/logflare_web/live/backends/components/backend_form.heex
+++ b/lib/logflare_web/live/backends/components/backend_form.heex
@@ -57,6 +57,23 @@
             <% end %>
             {text_input(f_config, :url, class: "form-control")}
           </div>
+          <% headers_number = 2
+
+          fields =
+            (input_value(f_config, :headers) || %{})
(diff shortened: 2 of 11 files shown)


lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -57,6 +57,40 @@
...
+          fields =
+            (input_value(f_config, :headers) || %{})
+            |> Stream.concat(Stream.repeatedly(fn -> {"", ""} end))
+            |> Enum.take(headers_number) %>
+          <%= for {{key, val}, i} <- Enum.with_index(fields, 1) do %>
+            <div class="form-group">
+              {label(f_config, "header#{i}_key", "Custom header #{i} - Key")}
+              {text_input(f_config, "header#{i}_key", value: key, class: "form-control")}
+              {label(f_config, "header#{i}_value", "Custom header #{i} - Value")}
+              {text_input(f_config, "header#{i}_value", value: val, class: "form-control")}
```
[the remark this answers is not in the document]
chasers: Fixed in 58c510354 (now part of the rebased branch): the value inputs render the REDACTED sentinel for stored values, and WebhookAdaptor.unredact_headers/2 restores the stored secret on submit. Test added asserting the raw token never reaches the HTML. Note: the OTLP header rows have the same leak, but OtlpAdaptor.cast_config has no sentinel unredaction, so the same fix would corrupt stored OTLP headers on save - left for a separate change.

lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -67,6 +88,13 @@
               Encoding of each batch. <code>json</code> sends a JSON array. <code>ndjson</code> sends newline-delimited JSON, one event per line.
             </small>
           </div>
+          <div class="form-group">
+            {label(f_config, :gzip, "Gzip")}
+            {checkbox(f_config, :gzip, id: "webhook-gzip")}
```
amokan: 🤖 

>The form's for is a plain map (`backend_form.heex:3-14`), so it's `%{}` for `:new` and adaptor defaults never reach it. `input_value(f_config, :gzip)` is `nil` → `phoenix_html_helpers` renders the box unchecked and emits its hidden companion input with `value="false"` (`form.ex:815-831`). The submit carries `config[gzip]="false"`, not an absent key, so `default_field_value(:gzip, true)` (`webhook_adaptor.ex:101`) doesn't fill it and the `Map.get(config, :gzip, true)` fallback at `webhook_adaptor.ex:449` never fires.
>
>Every webhook backend created through the UI after this ships uncompressed, contradicting docs/backends/webhook.mdx. Editing a backend whose stored config has no gzip key flips it to `false` on first save.
>
>This is visible in the PR's own tests: `backend_factory/1` writes config verbatim with no gzip key (`factory.ex:90-103`) and `LiveViewTest.form/3` submits the hidden input, so all five new edit tests set `gzip: false` and pass — they only assert `updated.config.headers`.

lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -57,6 +57,27 @@
...
+            |> Enum.sort_by(fn {key, _value} -> key end)
+            |> Stream.concat(Stream.repeatedly(fn -> {"", ""} end))
+            |> Enum.take(max(map_size(stored_headers) + 1, 2)) %>
+          <%= for {{key, val}, i} <- Enum.with_index(fields, 1) do %>
+            <div class="form-group">
+              {label(f_config, "header#{i}_key", "Custom header #{i} - Key")}
+              {text_input(f_config, "header#{i}_key", value: key, class: "form-control")}
+              {label(f_config, "header#{i}_value", "Custom header #{i} - Value")}
+              {text_input(f_config, "header#{i}_value",
+                value: if(val in [nil, ""], do: val, else: "REDACTED"),
```
amokan: 🤖

>`backend_form.heex:73` renders every stored value as "REDACTED" while the key input above it stays editable. `unredact_headers/2` matches on the sentinel value and looks the replacement up by the submitted key; `replace_header_with_existing/3` returns the accumulator untouched when there's no match (`webhook_adaptor.ex:150-155`).
>
>Stored `%{"authorization" => "Bearer abc"}`. User corrects the key to `x-api-key` and leaves the value alone — exactly what the help text says to do. Submit is `%{"x-api-key" => "REDACTED"}`, no match, headers becomes `%{}`. Success flash, credential gone.
>
>The drop is intentional in `unredact_headers/2` and was safe while nothing rendered the sentinel into an editable form. This PR makes it reachable.

lib/logflare/backends/adaptor/http_based/ndjson_formatter.ex
```diff
@@ -39,11 +41,28 @@ defmodule Logflare.Backends.Adaptor.HttpBased.NdjsonFormatter do
...
   per line. Any other term passes through unchanged.
+
+  An event whose body cannot be JSON-encoded is dropped. The count of dropped
+  events is logged per batch.
   """
   @spec encode([LogEvent.t()] | term()) :: iodata() | term()
   def encode([%LogEvent{} | _] = events) do
-    events
-    |> Enum.map(fn %LogEvent{body: body} -> Jason.encode_to_iodata!(body) end)
+    {encoded, dropped_count} =
```
amokan, line 60: Prob an edge case but if every event failed to encode for some reason, `encoded` will be `[]` and would turn into a `""` body. Upstream it looks like it would then try to potentially compress and post the empty body.

Prob should return and handle something like a `{:error, :all_dropped}` so `handle_batch/4` could skip the request 🤷‍♂️

lib/logflare/backends/adaptor/http_based/ndjson_formatter.ex
```diff
@@ -39,11 +41,28 @@ defmodule Logflare.Backends.Adaptor.HttpBased.NdjsonFormatter do
...
+      Enum.reduce(events, {[], 0}, fn %LogEvent{body: body}, {acc, dropped_count} ->
+        case Jason.encode_to_iodata(body) do
+          {:ok, iodata} -> {[iodata | acc], dropped_count}
+          {:error, _reason} -> {acc, dropped_count + 1}
+        end
+      end)
+
+    if dropped_count > 0 do
+      Logger.warning(
+        "Dropped #{dropped_count} log events from an NDJSON batch: JSON encoding failed"
```
amokan: I think we'll want some metadata in here with `backend_id` at least to understand which one fired the error.

amokan (dismissed):

amokan (approved):

PR #3896, opened 2026-09-01
chore: improve ClickHouse read-pool telemetry to support the slow-checkout investigation.

amokan:
- The `[:logflare, :clickhouse, :read_pool, :checkout]` event now emits `idle_time` and `connection_time` alongside `pool_time` (raw native units, converted at the metric declaration)
- New `read_pool.query_error` counter: one count per user query (_after any failover retry_), tagged `backend_id` / `read_cluster` / `error_kind`, excluding user SQL errors (`:invalid_query`)
- New `read_pool.failover` counter, tagged with the unhealthy cluster failed over from
- `to_query_error/1` now classifies `Mint.TransportError` / `Mint.HTTPError` as `:connection_error` (_enabling failover for node-down reads_), except recv timeouts which map to `:timeout` so a 60s query timeout doesn't re-run the statement on the default cluster
- `ConnectionManager` start failures now return a logged `%QueryError{}` instead of a raw `{:error, reason}`, so total-outage errors are counted and eligible for failover
- `logflare.clickhouse.read_pool.checkout` (ms) is renamed to `...checkout.pool_time` and re-united to **microseconds**
- `connection_time` (_query execution latency_) is published as `read_pool.connection_time`, outside the `checkout.*` namespace


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -695,20 +701,25 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
...
-             user_id: backend.user_id,
-             backend_id: backend.id,
-             backend_token: backend.token,
-             clickhouse_read_cluster: label,
-             host: ConnectionManager.read_host(backend, label)
-           )}
+          {:error, error |> to_query_error() |> log_query_error(backend, label)}
       end
+    else
+      {:error, reason} ->
```
djwhitt, line 707: :robot:

> [P1] Propagate connection-pool startup errors.
>
> This branch does not handle failures from `ConnectionManager.ensure_pool_started/2` because `ensure_pool_and_notify/2` discards that result and always returns `:ok`. The code then calls `Ch.query/4` with no registered pool and exits from `DBConnection.Holder.checkout/3`, bypassing the new `QueryError` telemetry and failover behavior.
>
> Propagate `ensure_pool_started/2`'s result and notify activity only after success. Add a regression where the requested cluster cannot start but the default cluster is healthy, and assert that the query fails over instead of exiting.

lib/telemetry.ex
```diff
@@ -256,12 +307,44 @@ defmodule Logflare.Telemetry do
         description:
           "Sum of events dropped by a backend (timestamp older than its configured max event age)"
       ),
-      distribution("logflare.clickhouse.read_pool.checkout",
+      distribution("logflare.clickhouse.read_pool.checkout.pool_time",
```
djwhitt, line 356: :robot:

> The adaptor tests validate raw telemetry measurements but do not validate the metric definitions in `Logflare.Telemetry.metrics/0`. They would still pass if a new metric name, converted unit, tag set, or histogram bucket list were incorrect.
>
> Add focused assertions for the new metric definitions, especially `{:native, :microsecond}` versus `{:native, :millisecond}` and the explicit bucket lists.

djwhitt: Not convinced this is worth it, but leaving for consideration regardless.

djwhitt (approved):

PR #3856, opened 2026-08-20
feat(toggle): Add enabled field to backend schema

ruslandoga:
PR 2/4 for [O11Y-2315](https://linear.app/supabase/issue/O11Y-2315/add-destination-level-toggle-controls), following [#3855](https://github.com/Logflare/logflare/pull/3855) once the migration is executed.

This PR adds the `enabled` field to the Backend Ecto schema with a `true` default.

Reference: [#3830](https://github.com/Logflare/logflare/pull/3830).

PR #3900, opened 2026-09-02
feat(toggle): disabled backends filtering at runtime

ruslandoga:
PR 3/4 for [O11Y-2315](https://linear.app/supabase/issue/O11Y-2315/add-destination-level-toggle-controls), following [#3856](https://github.com/Logflare/logflare/pull/3856) once its schema-only release is deployed to every node and preceding the final API/LiveView activation in [#3830](https://github.com/Logflare/logflare/pull/3830); because these are cross-repository PRs from a fork, this PR targets `main` and temporarily includes #3856's one-line schema diff.

Adds runtime-only handling for the existing `enabled` field: explicit, attached, and drain-rule routes plus Incident.io alerts stop admitting new work for disabled destinations, while HTTP and spool-consumer backpressure ignore their queues.

Existing backend workers and pipelines remain alive so queued work can drain.

It reuses the existing unfiltered backend-cache/WAL eventual-consistency path and intentionally adds no toggle write casting, UI/API controls, cache-key generation, or cluster message.


ruslandoga:
Keeping it in draft until https://github.com/Logflare/logflare/pull/3856 is merged and deployed.

lib/logflare/backends.ex
```diff
@@ -751,6 +755,8 @@ defmodule Logflare.Backends do
   end
 
   # send to a specific backend
+  defp dispatch_to_backends(_source, %Backend{enabled: false}, _log_events), do: :ok
```
Ziinc: Hmm this shouldn't even be hit as we shouldn't load disabled bsxkdnds in the first place. Instead of returning OK perhaps raising is better

lib/logflare/backends/cache.ex
```diff
@@ -31,6 +31,17 @@ defmodule Logflare.Backends.Cache do
   end
 
   def list_backends(arg), do: apply_repo_fun(__ENV__.function, [arg])
+
+  @spec list_enabled_backends(keyword()) :: [Backends.Backend.t()]
+  def list_enabled_backends(arg) do
+    # We filter the existing cached list instead of creating a
+    # `{:list_backends, [..., enabled: true]}` key. That entry would omit disabled backends, so
```
Ziinc, line 38: Cache buster will clear the cache of these lists so its fine, I think adjusting it to list_enabled_backends is fine and more efficient as well so that we don't do the filtering all the time

lib/logflare/sources/source_router.ex
```diff
@@ -32,13 +32,18 @@ defmodule Logflare.Sources.SourceRouter do
 
   defp do_routing(%Rule{backend_id: backend_id} = rule, %LE{} = le, source)
```
Ziinc, line 33: Should instead adjust the routing rules fetched to only be for an enabled backend when fetching with ecto, the routing only works with enabled backends anyway

Ziinc (changes requested):

lib/logflare/backends/cache.ex
ruslandoga: I think I had some problems with cached enabled/disabled backends and decided to go this way since it was easier to understand. Especially around "negative" results and then also gossiping/deployments (because of the new cache key, but that might have been a non-issue).

Consider:
- backend A is disabled, so `list_enabled_backends` caches `[]` or some other backend list
- backend A is enabled, would cache buster remove `list_enabled_backends`? I don't think it would since there is no matching backend in the cached values.
- `list_enabled_backends` continues to return stale results

ruslandoga: And I also didn't want to cache overlapping sets (all backends vs enabled backends) under different keys (so, kind of, same data cached twice), but that's also probably a non-issue in practice.

lib/logflare/sources/source_router.ex
ruslandoga: This was an optimization to avoid `Enum.filter` from `list_enabled_backends`, so we load the cached backends, don't do any filtering (no new lists created) and pass it through those "fast" pattern matches to apply the dispatch.

ruslandoga: We'd also need backend updates to invalidate the affected sources' cached rule lists and routing trees, including empty results on re-enable. This adds invalidation complexity. The current enabled check is cheap and filtering earlier would mainly save matching work for disabled destinations rules but I don't know if it's worth the extra invalidation complexity.

lib/logflare/backends.ex
ruslandoga: Ah, right. I'll remove the redundant clause.

For the remaining function-head checks, I found it easier to understand the filtering behavior when it's next to the dispatch implementation. Code readers don't need to trace how the backends were loaded or whether a caller filtered them first.

And it was also a bit of an optimization to avoid building an intermediate enabled-only list with `Enum.filter/2` but just run through them with function head matching.

PR #3901, opened 2026-09-02
fix: remove the hardcoded 750ms LiveView latency simulator in dev

chasers:
## Problem

The dev root layout called `liveSocket.enableLatencySim(750)` on every page load:

```elixir
<%= if Application.get_env(:logflare, :env) == :dev do %>
  <script>
    window.liveSocket.enableLatencySim(750)
  </script>
<% end %>
```

LiveView then delayed **every** client message by 750ms:

```js
let latency = this.getLatencySim()
if(!latency){ cb(data) } else { setTimeout(() => cb(data), latency) }
```

So every click, every `phx-change`, and every form submit paid 750ms in dev.

## Evidence

The server was never slow. Measured on this codebase:

| Measurement | Result |
| --- | --- |
| `change_form_type` server render | 0.5ms median, 9 KB payload |
| LiveView mounts (dead render) | 6-21ms |
| Callbacks over 100ms | none |

Selecting a backend type renders in 0.5ms, then waits 750ms in the browser.

## Side effect this also fixes

`enableLatencySim` calls `enableDebug()` first, so dev also logged every LiveView event to the console.

## After merging

The value persists in `sessionStorage`, so removing the script does not clear it. Run this once in the browser console:

```js
liveSocket.disableLatencySim()
```

Opening a new tab also works.

A developer who wants the simulator can still call `liveSocket.enableLatencySim(750)` from the console.

https://claude.ai/code/session_015XJWSbFFxwiYB25iUAGmXc


chasers:
Closing for #3986 3986

PR #3902, opened 2026-09-02
fix(repo): open read replica connections read-only

djwhitt:
Replica connection pools now open their sessions read-only.

## Why

`LOGFLARE_READ_REPLICAS` serves selected cache reads through `Repo.apply_with_replica/3`, but the pools did not guard against accidental writes. Physical standbys reject writes themselves; logical replicas can accept them, causing divergence and later replication conflicts.

## Behavior

- Runs the primary's `:after_connect` hook first, preserving setup such as the `DB_SCHEMA` search path, then issues `SET default_transaction_read_only = on`. This guards accidental writes; it is a session default, not an immutable permission boundary.
- The `SET` pins PostgreSQL RDS Proxy sessions to database connections. Size proxy and database connection limits accordingly.

IAM authentication for primary and replica connections is added separately in stacked PR #3903.

## Validation

`test/logflare/repo_test.exs`: **14 tests pass**, including read-only sessions and hook composition. Formatting passes.

### DIFF
--- a/docs/docs.logflare.com/docs/self-hosting/index.md
+++ b/docs/docs.logflare.com/docs/self-hosting/index.md
@@ -166,13 +166,15 @@ The configuration follows the [Erlang Security Working Group recommendations](ht
 
 ## Read Replicas
 
-`LOGFLARE_READ_REPLICAS` is a comma-separated list of PostgreSQL read replicas to distribute ingest-path data fetching queries across. If unset or empty, all queries go to the primary database.
+`LOGFLARE_READ_REPLICAS` is a comma-separated list of PostgreSQL read replicas for selected context-cache and key-value-cache reads. If unset or empty, those reads use the primary database.
 
 Each entry is either a **bare hostname** or a **connection URI** (`postgres://user:pass@host:port/database?ssl=true&pool_size=5`). In both cases, only the parts given override the primary's `DB_*` settings - anything omitted (port, database, credentials, SSL, ...) is inherited from the primary. Query params: `ssl` (`true`/`false`), `pool_size` (positive integer).
 
 Example: `LOGFLARE_READ_REPLICAS=replica1.example.com,postgres://user:pass@replica2.example.com:5432/logflare`
 
-Replica connections open their sessions read-only (`SET default_transaction_read_only = on`), so a write that reaches a replica fails immediately rather than diverging from the primary. This matters for a logical replica, whose server would otherwise accept the write; on a physical standby it is redundant with the server's own read-only state.
+Logflare marks each replica session read-only with `SET default_transaction_read_only = on`. This makes accidental writes fail immediately on logical replicas; physical standbys already enforce read-only operation.
+
+PostgreSQL RDS Proxy [pins sessions that issue `SET`](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy-pinning.html), so each Logflare replica connection keeps one database connection for its lifetime. Size the proxy and database connection limits for that behavior.
 
 ## Database Encryption
 
--- a/docs/docs.logflare.com/docs/self-hosting/index.md
+++ b/docs/docs.logflare.com/docs/self-hosting/index.md
@@ -172,6 +172,8 @@ Each entry is either a **bare hostname** or a **connection URI** (`postgres://us
 
 Example: `LOGFLARE_READ_REPLICAS=replica1.example.com,postgres://user:pass@replica2.example.com:5432/logflare`
 
+Replica connections open their sessions read-only (`SET default_transaction_read_only = on`), so a write that reaches a replica fails immediately rather than diverging from the primary. This matters for a logical replica, whose server would otherwise accept the write; on a physical standby it is redundant with the server's own read-only state.
+
 ## Database Encryption
 
 Certain database columns that store sensitive data are encrypted with the `LOGFLARE_DB_ENCRYPTION_KEY` key.
--- a/lib/logflare/repo/replicas.ex
+++ b/lib/logflare/repo/replicas.ex
@@ -22,18 +22,13 @@ defmodule Logflare.Repo.Replicas do
   a function call using `apply_with_replica/3` on `Logflare.Repo`, which swaps
   the dynamic repo and restores it afterwards.
 
-  Every replica connection opens its session read-only, so a write that reaches a
-  replica fails immediately instead of diverging from the primary.
+  Every replica connection opens its session read-only.
   """
 
   @registry __MODULE__.Registry
 
-  # Replicas serve reads only: `Logflare.Repo.apply_with_replica/3` and
-  # `apply_with_random_repo/3` are the only ways to reach one, and both are read
-  # paths. Marking the session read-only turns a write that slips through into an
-  # immediate error rather than a silent divergence. That matters most when the
-  # replica is a logical subscriber, which - unlike a physical standby - has a
-  # server that will happily accept the write.
+  # A physical standby rejects a stray write on its own; a logical subscriber accepts it, and
+  # the divergence only surfaces later as a conflict that stalls replication.
   @read_only_statement "SET default_transaction_read_only = on"
 
   def child_spec(options) do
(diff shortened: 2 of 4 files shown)


docs/docs.logflare.com/docs/self-hosting/index.md
```diff
@@ -172,6 +172,8 @@ Each entry is either a **bare hostname** or a **connection URI** (`postgres://us
 
 Example: `LOGFLARE_READ_REPLICAS=replica1.example.com,postgres://user:pass@replica2.example.com:5432/logflare`
 
+Replica connections open their sessions read-only (`SET default_transaction_read_only = on`), so a write that reaches a replica fails immediately rather than diverging from the primary. This matters for a logical replica, whose server would otherwise accept the write; on a physical standby it is redundant with the server's own read-only state.
```
Ziinc:
```suggestion
Replica connections open their sessions read-only (`SET default_transaction_read_only = on`), preventing any possibility of stray writes to the read replica.
```

lib/logflare/repo/replicas.ex
```diff
@@ -21,10 +21,16 @@ defmodule Logflare.Repo.Replicas do
...
   a function call using `apply_with_replica/3` on `Logflare.Repo`, which swaps
   the dynamic repo and restores it afterwards.
+
+  Every replica connection opens its session read-only.
   """
 
   @registry __MODULE__.Registry
 
+  # A physical standby rejects a stray write on its own; a logical subscriber accepts it, and
+  # the divergence only surfaces later as a conflict that stalls replication.
```
Ziinc:
```suggestion
  # Logical replicas allow writes -- block writes.
```

lib/logflare/repo/replicas.ex
```diff
@@ -148,22 +184,35 @@ defmodule Logflare.Repo.Replicas do
   defp redact(message, userinfo), do: String.replace(message, userinfo, "REDACTED")
 
   defp resolve_ssl(config) do
-    case Keyword.get(config, :ssl) do
-      true -> Keyword.put(config, :ssl, primary_ssl_opts(config[:hostname]))
-      _ -> config
+    primary_ssl = Keyword.get(Logflare.Repo.config(), :ssl)
+    effective_ssl = Keyword.get(config, :ssl, primary_ssl)
```
Ziinc: i don't think the nested ssl boolean option is in use anymore

https://postgrex.hexdocs.pm/Postgrex.html#t:start_option/0
```
# what should be returned
iex> Logflare.Repo.config()
[..., ssl: [verify: :verify_peer, ... ], ... ]

# this should never happen due to runtime.exs:418
iex> Logflare.Repo.config()
[ssl: true, ...] 
```
which makes :193 dead code

lib/logflare/repo/replicas.ex
```diff
@@ -148,22 +184,35 @@ defmodule Logflare.Repo.Replicas do
...
+    end
+  end
 
-      _ ->
-        true
+  defp replica_ssl_opts(opts, hostname) do
+    opts = Keyword.delete(opts, :server_name_indication)
+
+    case :inet.parse_address(String.to_charlist(hostname || "")) do
+      {:ok, _ip} -> Keyword.put(opts, :server_name_indication, :disable)
```
Ziinc: this should already be done in runtime.exs:415, would be able to remove

lib/logflare/repo/replicas.ex
```diff
@@ -66,6 +78,30 @@ defmodule Logflare.Repo.Replicas do
...
+  composition would discard settings such as the `search_path` configured by `DB_SCHEMA`.
+  The inherited hook runs first so it can perform any required writes.
+  """
+  @spec after_connect(
+          DBConnection.conn(),
+          {module(), atom(), [term()]} | (DBConnection.t() -> any()) | nil
+        ) :: Postgrex.Result.t()
+  def after_connect(conn, primary_after_connect) do
+    run_after_connect(conn, primary_after_connect)
+    Postgrex.query!(conn, @read_only_statement, [])
```
Ziinc: hmm shouldn't use a bang query here, can bring server down.

lib/logflare/repo/replicas.ex
```diff
@@ -40,9 +45,16 @@ defmodule Logflare.Repo.Replicas do
     if entries == [] do
       :ignore
     else
+      primary_after_connect = Logflare.Repo.config()[:after_connect]
```
Ziinc, line 48: Logflare.Repo has no after_connect configured so we don't need to do this

lib/logflare/repo/replicas.ex
```diff
@@ -66,6 +78,30 @@ defmodule Logflare.Repo.Replicas do
...
+  Replica pools inherit the primary configuration. Replacing `:after_connect` without
+  composition would discard settings such as the `search_path` configured by `DB_SCHEMA`.
+  The inherited hook runs first so it can perform any required writes.
+  """
+  @spec after_connect(
+          DBConnection.conn(),
+          {module(), atom(), [term()]} | (DBConnection.t() -> any()) | nil
+        ) :: Postgrex.Result.t()
+  def after_connect(conn, primary_after_connect) do
+    run_after_connect(conn, primary_after_connect)
```
Ziinc, line 94: don't need a wrapping function

Ziinc: can remove `run_after_connect/2` since we don't set :after_connect for Logflare.Repo

lib/logflare/repo/replicas.ex
```diff
@@ -148,22 +184,35 @@ defmodule Logflare.Repo.Replicas do
...
-          {:error, _} -> opts
-        end
+      opts when is_list(opts) -> replica_ssl_opts(opts, hostname)
+      _ -> true
+    end
+  end
 
-      _ ->
-        true
+  defp replica_ssl_opts(opts, hostname) do
```
Ziinc: i think we don't need to do this, if `?ssl=true` then we just reuse the ssl configuration - we won't be using it anyways.
~~Full RR ssl support is not needed at the moment for our situation, if there's requests for it in the future then we can consider extending the SSL configs for it but this would be sufficient.~~

Reviewing #3903 now, this gets messy real fast though if we're using custom query params for denoting IAM SSL as well

lib/logflare/repo/replicas.ex
```diff
@@ -148,22 +184,35 @@ defmodule Logflare.Repo.Replicas do
...
+    primary_ssl = Keyword.get(Logflare.Repo.config(), :ssl)
+    effective_ssl = Keyword.get(config, :ssl, primary_ssl)
+    hostname = config[:hostname]
+
+    case effective_ssl do
+      true ->
+        Keyword.put(config, :ssl, primary_ssl_opts(hostname))
+
+      opts when is_list(opts) ->
+        Keyword.put(config, :ssl, replica_ssl_opts(opts, hostname))
```
Ziinc: its the inverse, only replica URI string can have `:ssl` as true, primary db config will use new postgrex :ssl config
https://postgrex.hexdocs.pm/Postgrex.html#t:start_option/0

lib/logflare/repo/replicas.ex
```diff
@@ -148,22 +184,35 @@ defmodule Logflare.Repo.Replicas do
   defp redact(message, userinfo), do: String.replace(message, userinfo, "REDACTED")
 
   defp resolve_ssl(config) do
```
Ziinc, line 186: we probably can add a 2nd argument for:
```
defp resolve_ssl(config, is_replica) do
...
```
since only replica are via  URIs that can have `?ssl=true`

Ziinc (changes requested):

lib/logflare/repo/replicas.ex
djwhitt: `DB_HOSTNAME` is used in `runtime.exs` this is for the replica. The results may be different for replicas.

lib/logflare/repo/replicas.ex
djwhitt: `Logflare.Repo` does configure `:after_connect` when `DB_SCHEMA` is set. `runtime.exs` installs a Postgrex query that sets `search_path`.

https://github.com/Logflare/logflare/blob/b895f8d18242784502aee443ddeb0bac0e8494dc/config/runtime.exs#L195-L198

lib/logflare/repo/replicas.ex
djwhitt: See earlier comment re `:after_connect`

Ziinc (approved):

PR #3903, opened 2026-09-02
feat(repo): authenticate PostgreSQL with AWS RDS IAM tokens

djwhitt:
Stacked on #3902. This diff adds AWS RDS IAM authentication for the primary database, read replicas, and the direct PostgreSQL clients used by the cluster strategy and cache invalidation.

## Configuration

Primary database:

```text
DB_AUTH=aws_iam
DB_AWS_REGION=eu-west-1
```

Read replica URI:

```text
LOGFLARE_READ_REPLICAS=postgres://logflare@my-proxy.proxy-abc.eu-west-1.rds.amazonaws.com:5432/logflare?auth=aws_iam&aws_region=eu-west-1
```

Optional RDS CA bundle override:

```text
RDS_CA_CERT_PATH=/path/to/rds-ca-bundle.pem
```

`auth` and `aws_region` are Logflare-specific query parameters and are removed before the options reach Postgrex. A password-bearing replica URI explicitly selects password authentication instead of inheriting primary IAM settings; combining a password with `auth=aws_iam` is rejected.

## Behavior

- Routes primary and replica configuration through the same `Logflare.Repo.init/2` connection-options pipeline.
- Routes the libcluster PostgreSQL strategy and Cainophile's epgsql client through resolved primary repository options. This gives those direct clients IAM token refresh plus consistent URL, TLS, and socket configuration without inheriting Ecto pool or registration options.
- Uses DBConnection's `:configure` callback to generate a fresh IAM token before every connection attempt while preserving an existing callback. The epgsql client likewise generates a fresh token for authentication and reconnects.
- Requires an explicit AWS region instead of deriving it from the hostname. IAM connections require a DNS hostname; IP literals are rejected, and the AWS-issued database, cluster, reader, or proxy endpoint should be used because the hostname is included in the signed token.
- Resolves credentials from static AWS environment variables, including `AWS_SESSION_TOKEN`, EKS Pod Identity, ECS task roles, or EC2 instance roles.
- Enforces peer and hostname verification for IAM connections. It rejects `ssl=false`, `verify: :verify_none`, and custom `:verify_fun` callbacks while preserving explicitly configured CA trust.
- Uses system roots plus the configured RDS CA bundle when no explicit trust store is supplied. A missing or empty bundle logs a warning and falls back to system roots.
- Container images download the commercial-region RDS CA bundle, verify its pinned SHA-256 digest, and install it at `/etc/ssl/certs/aws-rds-global-bundle.pem`. Other deployments and AWS partitions can set `RDS_CA_CERT_PATH`.
- Preserves primary `:after_connect` behavior and applies the read-only session callback only to replica pools.

The runtime role needs `rds-db:connect` permission for the database user.

## Validation

Focused coverage spans **40 tests** across `test/logflare/repo_test.exs`, `test/logflare/repo/aws_iam_test.exs`, and `test/logflare/repo/connection_options_test.exs`. It covers primary, replica, and direct-client configuration; authentication inheritance and override behavior; callback composition; explicit-region signing; temporary credentials; TLS enforcement; CA fallback; option consumption; credential redaction; and read-only sessions.

The repository test check, formatting, and `mix lint.all` pass. Live RDS/Aurora and RDS Proxy authentication has not been validated.


docs/docs.logflare.com/docs/self-hosting/index.md
```diff
@@ -168,7 +169,17 @@ The configuration follows the [Erlang Security Working Group recommendations](ht
...
 
-Each entry is either a **bare hostname** or a **connection URI** (`postgres://user:pass@host:port/database?ssl=true&pool_size=5`). In both cases, only the parts given override the primary's `DB_*` settings - anything omitted (port, database, credentials, SSL, ...) is inherited from the primary. Query params: `ssl` (`true`/`false`), `pool_size` (positive integer).
+Each entry is either a **bare host name or IP literal** or a **connection URI** (`postgres://user:pass@host:port/database?ssl=true&pool_size=5`). In both cases, only the parts given override the primary's `DB_*` settings - anything omitted (port, database, credentials, SSL, ...) is inherited from the primary. Query params: `ssl` (`true`/`false`), `pool_size` (positive integer), `auth` (`iam`).
+
+`auth=iam` replaces the password with an RDS IAM token minted for each connection. IAM connections always use verified TLS; `ssl=false` is rejected and omitted `ssl` is enabled automatically. The entry carries a username but no password:
+
+```
+LOGFLARE_READ_REPLICAS=postgres://logflare@my-proxy.proxy-abc.eu-west-1.rds.amazonaws.com:5432/logflare?auth=iam
```
Ziinc: probably want to make to make it clearer that it is vendor specific, so `?auth=aws_iam`

Also want to add a line that this is a custom query parameter for Logflare configuration consumption

docs/docs.logflare.com/docs/self-hosting/index.md
```diff
@@ -168,7 +169,17 @@ The configuration follows the [Erlang Security Working Group recommendations](ht
...
+```
+
+Use the AWS-issued database, cluster, reader, or proxy endpoint. IAM signing includes that hostname, so aliases and tunnels are unsupported. The runtime needs `rds-db:connect` for the database user. ExAws supports static environment credentials (including `AWS_SESSION_TOKEN`), EKS Pod Identity, ECS task roles, and EC2 instance roles.
```
Ziinc: should not mention ExAws in the docs.

Ziinc: can be a bullet point list of supported env credentials.

docs/docs.logflare.com/docs/self-hosting/index.md
```diff
@@ -168,7 +169,17 @@ The configuration follows the [Erlang Security Working Group recommendations](ht
...
+
+RDS Proxy uses ACM certificates and the image's system trust store. Direct RDS and Aurora endpoints require an [RDS CA bundle](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html). Container images install the commercial-region bundle at `/etc/ssl/certs/aws-rds-global-bundle.pem`; non-container deployments and non-commercial AWS partitions must set `RDS_CA_CERT_PATH`.
```
Ziinc: feels odd to add in aws-specific docs here. maybe a vendor subheader for helping people connect to different DB providers would be better.

lib/logflare/repo/replicas.ex
```diff
@@ -21,7 +21,8 @@ defmodule Logflare.Repo.Replicas do
   a function call using `apply_with_replica/3` on `Logflare.Repo`, which swaps
   the dynamic repo and restores it afterwards.
 
-  Every replica connection opens its session read-only.
+  Every replica connection opens its session read-only. An entry may use
+  `auth=iam` instead of a password; IAM connections always use verified TLS.
```
Ziinc:
```suggestion
  `auth=aws_iam` instead of a password; IAM connections always use verified TLS.
```

lib/logflare/repo/replicas.ex
```diff
@@ -30,6 +31,10 @@ defmodule Logflare.Repo.Replicas do
   # read-only default turns accidental writes into immediate errors.
   @read_only_statement "SET default_transaction_read_only = on"
 
+  @iam_token_expiry_seconds 900
+  @default_rds_ca_cert_path "/etc/ssl/certs/aws-rds-global-bundle.pem"
+  @rds_hostname ~r/\.([a-z]{2}(?:-[a-z]+)+-\d+)\.rds\.amazonaws\.com(?:\.cn)?$/i
```
Ziinc: might be behind a proxy or custom dns configuration, would be more reliable to get the user to declare it in the query params.

lib/logflare/repo/replicas.ex
```diff
@@ -102,6 +108,185 @@ defmodule Logflare.Repo.Replicas do
...
+    case Regex.run(@rds_hostname, hostname) do
+      [_, region] ->
+        region
+
+      nil ->
+        raise "cannot determine the AWS region for read replica #{inspect(hostname)}: " <>
+                "IAM authentication requires an RDS hostname"
+    end
+  end
+
```
Ziinc: as above, not so reliable to derive from the provided hostname

lib/logflare/repo/replicas.ex
```diff
@@ -102,6 +108,185 @@ defmodule Logflare.Repo.Replicas do
...
+  Builds an RDS IAM authentication token used in place of a password.
+  """
+  @spec rds_auth_token(String.t(), non_neg_integer(), String.t()) :: String.t()
+  def rds_auth_token(hostname, port, username) do
+    hostname = String.downcase(hostname)
+
+    config =
+      :rds
+      |> ExAws.Config.new(region: aws_region!(hostname))
+      |> maybe_put_env_session_token()
```
Ziinc: can use system tuples `{:system, ...}` then conditionally put in missing config in piped functions.

Ziinc: probably also want to let the self-hosting user override this on a per uri basis via query params - more choice for the user

lib/logflare/repo/replicas.ex
```diff
@@ -102,6 +108,185 @@ defmodule Logflare.Repo.Replicas do
...
+    end
+  end
+
+  defp put_iam_ssl!(config) do
+    hostname = config |> Keyword.fetch!(:hostname) |> String.downcase()
+    aws_region!(hostname)
+
+    case Keyword.get(config, :ssl) do
+      false ->
+        raise ArgumentError, "IAM authentication requires TLS; remove ssl=false"
```
Ziinc: can just log a warning

docs/docs.logflare.com/docs/self-hosting/index.md
Ziinc: should also consider that we will need to extend this configuration params to the primary in the future, so something like `DB_OPTS_AUTH` or something like that

Ziinc (changes requested):
we will need to extend IAM auth to primary as well, might as well do it for both now.

Baishan (approved):

lib/logflare/repo/aws_iam.ex
```diff
@@ -0,0 +1,218 @@
...
+       when is_binary(token) and token != "",
+       do: config
+
+  defp maybe_put_environment_session_token(config) do
+    with access_key_id when is_binary(access_key_id) and access_key_id != "" <-
+           System.get_env("AWS_ACCESS_KEY_ID"),
+         secret_access_key when is_binary(secret_access_key) and secret_access_key != "" <-
+           System.get_env("AWS_SECRET_ACCESS_KEY"),
+         session_token when is_binary(session_token) and session_token != "" <-
+           System.get_env("AWS_SESSION_TOKEN"),
```
Ziinc, line 178: hmm not a fan of doing env var reading outside of runtime.exs, though this will work fine for now.

lib/logflare/repo/aws_iam.ex
```diff
@@ -0,0 +1,218 @@
...
+         true <- config[:access_key_id] == access_key_id,
+         true <- config[:secret_access_key] == secret_access_key do
+      Map.put(config, :security_token, session_token)
+    else
+      _ -> config
+    end
+  end
+
+  defp drop_empty_session_token(%{security_token: token} = config) when token in [nil, ""],
```
Ziinc, line 187: this shouldn't be needed - should be put together with maybe_put_environment_session_token/1 ? seems odd to have a searate function for it

test/logflare/repo/aws_iam_test.exs
```diff
@@ -0,0 +1,709 @@
...
+      previous_security_token = Application.fetch_env(:ex_aws, :security_token)
+      previous_env = take_aws_credential_env()
+
+      Application.delete_env(:ex_aws, :access_key_id)
+      Application.delete_env(:ex_aws, :secret_access_key)
+      Application.delete_env(:ex_aws, :security_token)
+      System.put_env("AWS_ACCESS_KEY_ID", "ASIATEMPORARY")
+      System.put_env("AWS_SECRET_ACCESS_KEY", "temporary-secret")
+      System.put_env("AWS_SESSION_TOKEN", "session-token")
+
```
Ziinc, line 517: as mentioned, this should be handled fully by runtime.exs so testing it explicitly shouldn't be needed.

Ziinc (approved):

lib/logflare/repo/connection_options.ex
```diff
@@ -0,0 +1,177 @@
...
+    {config, password} = config |> prepare(:primary) |> prepare_epgsql_config()
+    hostname = Keyword.fetch!(config, :hostname)
+
+    %{}
+    |> put_present(:host, String.to_charlist(hostname))
+    |> put_present(:port, config[:port])
+    |> put_present(:username, config[:username])
+    |> put_present(:database, config[:database])
+    |> put_present(:password, password)
+    |> put_present(:tcp_opts, config[:socket_options])
```
Ziinc, line 67: alternatively can create a map and then reject nils.

```
%{
  host: ...
}
|> Map.filter(&(&1 != nil))

Ziinc: or Map.reject(&is_nil/1)

PR #3905, opened 2026-09-02
fix: keep database work out of the SourceSup partition critical section

chasers:
Only adds `SourceSup.prefetch/1` and a call to it from `start_source_sup/1`.

It warms the caches `init/1` reads, but the important part is *where* it runs: in the caller, before `start_child`. `init/1` runs inside the `SourcesSup` partition process, so a source that stalls resolving its config blocks every other source hashing to that partition. Prefetching first means the stall lands on the caller instead, and the rest keep starting.

`init/1` is unchanged, so a crash-restart still re-resolves.

https://claude.ai/code/session_015XJWSbFFxwiYB25iUAGmXc

### DIFF
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -926,6 +926,8 @@ defmodule Logflare.Backends do
   """
   @spec start_source_sup(Source.t()) :: :ok | {:error, :already_started}
   def start_source_sup(%Source{} = source) do
+    SourceSup.prefetch(source)
+
     case DynamicSupervisor.start_child(
            {:via, PartitionSupervisor, {SourcesSup, source.id}},
            SourceSup.child_spec(source)
--- a/lib/logflare/backends/source_sup.ex
+++ b/lib/logflare/backends/source_sup.ex
@@ -31,6 +31,35 @@ defmodule Logflare.Backends.SourceSup do
     Supervisor.start_link(__MODULE__, source, name: Backends.via_source(source, __MODULE__))
   end
 
+  @doc """
+  Resolves everything `init/1` reads, so the supervisor's critical section does no database work.
+
+  `DynamicSupervisor.start_child/2` runs `init/1` inside the `SourcesSup` partition process, and
+  every lookup there is cache-fronted with a Postgres fallback. On a node with a cold cache that
+  makes each source start several database round trips, serialized through one partition, which
+  blocks every other source hashing to it.
+
+  Calling this first moves that latency to the caller, where it parallelizes across callers.
+  `init/1` is left unchanged, so a crash-restart still re-resolves and picks up configuration
+  changes.
+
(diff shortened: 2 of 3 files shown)


chasers:
## Correction: the timeout path was dropping events

Pushed `0ef1b2414`. The safety claim in the description above was wrong, and I want it recorded rather than quietly edited.

**The claim:** on timeout the caller continues, and `dispatch_to_backends/3` writes to `IngestEventQueue`, so events are picked up once the supervisor starts.

**What actually happened:** `AdaptorSupervisor.init/1` is what creates a backend's startup queue (`adaptor_supervisor.ex:25`), and it only runs once `SourceSup` has started. Before that there is no table, so the write hits:

```elixir
def add_to_table({_, _, _} = sid_bid_pid, batch, _opts) do
  get_tid(sid_bid_pid)
  |> case do
    nil -> {:error, :not_initialized}
    tid -> add_to_table({sid_bid_pid, tid}, batch)
  end
end
```

The caller discards that error and answers `:ok`. `ingest_logs/4` then answers `{:ok, count}`, and the client gets a success response for a batch that was thrown away.

It is a race window rather than permanent loss — the `start_child` message is still queued on the partition, so `SourceSup` starts eventually. But the window is exactly the cold-boot period this PR targets.

## The fix

`SourceSup.prefetch/1` now creates each backend's startup queue and its generation, mirroring `AdaptorSupervisor.init/1`. Both calls are idempotent, and prefetch already runs before `start_child`, so the queue exists before any timeout can bite.

## Tests

Three added, and **all three fail without the change**:

| Test | Without the fix |
| --- | --- |
| `prefetch/1` creates the default backend startup queue | no tid |
| `prefetch/1` creates a startup queue for each attached backend | no tid |
| a timed out start still stores the events it was given | **0 pending, expected 1** |

`test/logflare/backends_test.exs` — 102 tests, 0 failures.

## Still worth a separate fix

The default `{sid, bid}` path in `add_to_table/3` drops with no telemetry at all. The consolidated ClickHouse path pipes through `maybe_emit_not_initialized/3` (`ingest_event_queue.ex:591`); the default path at line 651 has no equivalent. Not changed here.

chasers:
## Scoped down to one problem

Force-pushed `2d4b25e93`, replacing the previous four commits. This PR is now only `SourceSup.prefetch/1` and its call site — keeping `init/1` out of the database, nothing else.

The earlier scope is preserved locally on `tabled/source-sup-full-scope` at `76f193b97`, and the commits stay reachable in this PR's history.

**Removed from this branch:**

- the bounded `ensure_source_sup_started/1` timeout
- `Billing.CacheWarmer`
- the ingest startup queue creation, which only existed to make that timeout safe

The description above lists all six follow-ups. The first one carries a real hazard worth reading before anyone picks it up: a timeout on `ensure_source_sup_started/1` drops events silently and still returns a success response to the client, unless the startup queues are created first.

PR #3911, opened 2026-09-03
fix: complete SourceSup startup cache prefetch

djwhitt:
## Summary

Follow-up to #3905.

- skip prefetch for an already-registered `SourceSup` while always delegating the final start decision to `DynamicSupervisor`
- warm the source-by-id cache key used by initial buffer producers
- warm source schemas only when a BigQuery backend will start
- mirror `SourceSup.init/1` when filtering persisted and default backends
- assert exact startup cache keys and explicit start/prefetch delegation
- correct the documented OTP startup semantics

## Why

- **Safe already-started fast path:** avoids unnecessary database-capable prefetching for a live supervisor, but still calls `DynamicSupervisor.start_child/2`. This lets Registry registration clean up a stale dead-process entry instead of incorrectly leaving a source stopped after restart.
- **Source-by-ID prefetch:** `BufferProducer` reads this exact cache key during synchronous startup. Warming it first prevents a cold Postgres lookup while the `SourcesSup` partition is waiting.
- **Conditional BigQuery schema prefetch:** BigQuery's schema process reads `SourceSchemas.Cache` during startup. Warming it moves that miss outside the partition, while the condition avoids unnecessary schema queries for Postgres-only sources.
- **Matching backend selection:** persisted consolidated backends are filtered before the default backend is added, exactly as in `SourceSup.init/1`, preventing the two paths from drifting.
- **Documentation:** clarifies that initialization runs in the new child process while the partition waits synchronously, and removes the overly broad claim that the partition can never encounter a miss.
- **Tests:** exact cache-key assertions catch omitted or mismatched prefetches; explicit Mimic expectations prove new starts prefetch and already-registered starts still delegate without prefetching.

## Validation

- focused feedback regression tests — 5 tests, 0 failures
- `../bin/test test/logflare/backends_test.exs` — 101 tests, 0 failures, 3 excluded
- SourceSup consumer tests — 8 doctests and 524 tests, 0 failures, 8 excluded

## Scope

Automatic crash-restart policy, global prefetch concurrency limiting, and the separately identified cache-warmer key defects remain unchanged.


chasers:
## Review

The prefetch work itself looks right, and the new test is a real improvement over asserting `size > 0`. One finding needs a fix before merge.

### 1. `lib/logflare/backends.ex:929` — the fast path can leave a source unsupervised

`source_sup_started?/1` is `Registry.lookup/2`, which reads the ETS table directly and does **not** check whether the PID is alive. `SourceRegistry` is `keys: :unique`, and a unique registry only clears an entry when its partition process handles the `:DOWN` message. That is asynchronous.

So `restart_source_sup/1` has a window:

1. `stop_source_sup/1` runs. `DynamicSupervisor.terminate_child/2` returns as soon as the process is dead.
2. The registry partition has not necessarily handled its `:DOWN` yet.
3. `start_source_sup/1` sees the stale entry and returns `{:error, :already_started}`.
4. The SourceSup never restarts.

Before this PR that could not happen. `DynamicSupervisor.start_child/2` reached `Registry.register/3`, which detects the dead PID, deletes the stale entry, and retries.

This matters because `create_backend` (`backends.ex:201`) and `update_backend` (`backends.ex:326`) both call `restart_source_sup/1` for every attached source. A backend config update can leave sources with no supervision tree. Nothing recovers them until ingest traffic reaches `ensure_source_sup_started/1`, so a low-traffic source stays down.

Keeping the fast path as a prefetch skip avoids the race and preserves the win:

```elixir
def start_source_sup(%Source{} = source) do
  unless source_sup_started?(source), do: SourceSup.prefetch(source)
  do_start_source_sup(source)
end
```

### 2. `lib/logflare/backends/source_sup.ex:59` — `prefetch/1` and `init/1` filter the default backend differently

`prefetch/1` prepends `get_default_backend(user)` and *then* applies `Enum.reject(& &1.consolidated_ingest?)` to the whole list. `init/1` rejects consolidated backends from `ingest_backends` and `rules_backends` separately, then prepends `default_backend` unfiltered.

The two agree today only because `get_default_backend/1` returns an unpersisted `%Backend{}` whose virtual `consolidated_ingest?` defaults to `false`. If a default backend type ever becomes consolidated, `prefetch/1` stops warming the source schema while `init/1` still starts the BigQuery `Schema` child. That reintroduces the cold-cache round trip this PR removes, silently. Mirroring `init/1`'s shape would keep them in step.

### 3. `test/logflare/backends_test.exs:832` — lost coverage

The removed `start_source_sup/1 warms the caches before starting` test was the only assertion that `start_source_sup/1` calls `prefetch/1` at all. The new test covers only the skip path, so deleting the `prefetch/1` call from `start_source_sup/1` would leave the suite green.

### Unrelated, and unverified — worth its own issue

While reading the cache paths, two warmers look like they write keys the read path cannot hit:

- `lib/logflare/source_schemas/cache_warmer.ex:23` emits `{:cached, {key, ss}}` as the pair, which Cachex reads as key `:cached` with value `{key, ss}` — every row would collide on one key.
- `lib/logflare/sources/cache_warmer.ex:22-23` writes `{:get_by, [id: s.id]}`, while `ContextCache.apply_fun/3` builds `{:get_by, [[id: id]]}`.

I did not verify these, so treat them as a lead rather than a finding. If they hold, this PR's prefetch is carrying more weight than the docstring suggests.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

chasers:
Follow-up: I verified the two cache-warmer items I flagged as unverified. **Both are real.** Promoting them from "lead" to finding.

The key shape is settled by this PR's own new test, which asserts the live keys:

```elixir
Cachex.get(Logflare.SourceSchemas.Cache, {:get_source_schema_by, [[source_id: source_id]]})
Cachex.get(Logflare.Sources.Cache, {:get_by, [[id: source_id]]})
```

That double nesting is correct. `Logflare.SourceSchemas.Cache.get_source_schema_by/1` is `apply_fun(__ENV__.function, [kv])` (`source_schemas/cache.ex:38`), and `ContextCache.apply_fun/3` builds `cache_key = {fun, args}` (`context_cache.ex:51`). So the single argument `[source_id: id]` becomes the args list `[[source_id: id]]`. `Sources.Cache.get_by/1` (`sources/cache.ex:63`) is the same shape.

Against that:

- `source_schemas/cache_warmer.ex:23` writes the pair `{:cached, {{:get_source_schema_by, [source_id: ss.source_id]}, ss}}`. Cachex reads a warmer's list as `{key, value}`, so the **key is the atom `:cached`** and the whole intended pair is the value. All 1,000 rows collide on one key. The inner key is also single-nested, so it would miss even without the collision.
- `sources/cache_warmer.ex:23-24` has the pair shape right, but writes `{:get_by, [id: s.id]}` — single-nested. `[id: s.id]` is `[{:id, s.id}]`, one argument that is a **tuple**, where the real call passes one argument that is a **list**.

Net effect: neither warmer produces a key the read path can hit, so both warm nothing today. That makes the prefetch in this PR the only thing preventing those misses, rather than a backstop to the warmers.

Not a change request on this PR — it argues *for* the prefetch. It needs its own fix.

I also re-checked the finding in my previous comment against the Elixir 1.19.5 source, and it holds: `Registry.lookup/2` for a `:unique` registry is a plain ETS read with no liveness check (`registry.ex:660-671`), while `register_key(:unique, ...)` calls `Process.alive?/1`, deletes the stale object, and retries (`registry.ex:1206-1225`). `SourceSup.start_link/1` registers through `{:via, Registry, {SourceRegistry, {id, SourceSup}}}`, so the old path got that cleanup and the new fast path does not.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

chasers:
Re-checked at `bde8137e`. **All three findings are addressed.** Thanks for the quick turnaround.

**1. The restart race.** `start_source_sup/1` now delegates unconditionally, so `DynamicSupervisor.start_child/2` is the authority on duplicates again and `Registry.register/3` still clears a stale entry and retries. `restart_source_sup/1` can no longer no-op. The prefetch skip survives, so the original goal — keep the cold-cache round trips off the partition — is untouched.

**2. The backend filter.** `prefetch/1` now rejects `consolidated_ingest?` from `source_backends` and `rules_backends` separately, then prepends the default backend unfiltered. That mirrors `init/1` exactly, so the two cannot drift if a default backend type ever becomes consolidated.

**3. The coverage gap.** Covered better than I suggested. `start_source_sup/1 prefetches before starting` restores the dropped assertion, and `start_source_sup/1 skips prefetch but still delegates when already started` is a direct regression guard on finding 1 — the `reject(&SourceSup.prefetch/1)` plus `expect(SourceSup, :child_spec, ...)` pair proves the already-started path still reaches `start_child`.

CI is green on this head: 9 checks pass, 3 skipped, including the full test suite, dialyzer, credo, and compile warnings.

No further concerns from me. The cache warmer key-shape bug from my earlier comment is out of scope for this PR and is being picked up separately.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

lib/logflare/backends.ex
```diff
@@ -926,8 +926,11 @@ defmodule Logflare.Backends do
   """
   @spec start_source_sup(Source.t()) :: :ok | {:error, :already_started}
   def start_source_sup(%Source{} = source) do
-    SourceSup.prefetch(source)
+    unless source_sup_started?(source), do: SourceSup.prefetch(source)
```
chasers: `unless` is soft deprecated now in latest Elixir we should switch to `if not` everywhere

```suggestion
    if not source_sup_started?(source), do: SourceSup.prefetch(source)
```

chasers (changes requested):

chasers (approved):

PR #3913, opened 2026-09-03
fix: correct source and source schema cache warmers

djwhitt:
## Summary

- correct `SourceSchemas.CacheWarmer` entries so Cachex stores the exact read-through argument key and `{:cached, schema}` value expected by `ContextCache`
- correct `Sources.CacheWarmer` ID and token entries to use the same argument shape as `Sources.Cache.get_by/1`, including both internal atom and external binary UUID token forms
- preserve `Sources.get_by/1` value semantics by hydrating each warmed source's computed `retention_days`
- resolve source plans in bounded bulk queries, including preloaded billing accounts, rather than introducing per-source database lookups
- skip sources whose users disappear during plan hydration instead of aborting the entire warm pass

## Why

`ContextCache` keys function calls as `{function, arguments}`. Both source cache APIs accept a keyword list as one function argument, so their read-through keys contain a nested argument list such as `{:get_by, [[id: source_id]]}`.

The previous warmers emitted differently shaped entries. `SourceSchemas.CacheWarmer` also placed the key and schema inside the cached value. Cachex accepted those pairs, but normal reads could not reach them, so the first request still queried Postgres and inserted a separate correctly keyed entry.

Source tokens are stored as atoms by `Ecto.UUID.Atom`, while HTTP ingestion supplies binary UUIDs. Because `ContextCache` keys the exact arguments, the warmer now stores both token representations so internal and external reads reach the preloaded value.

Correcting the `Sources` keys exposed a second issue: the warmer loaded raw `Source` records, while the normal `Sources.get_by/1` path computes the virtual `retention_days` field. Without hydration, a warmed cache hit would return a different source shape from a cache miss. The warmer now bulk-loads the relevant users and billing accounts, resolves their plans once, and applies the same retention calculation before storing entries.

The source and user reads are separate statements, so an account deletion can remove a user after its source was selected. That source is now skipped with one aggregate warning instead of making the one-shot warmer discard every entry.

## Test coverage

- derive SourceSchemas and Sources cache keys through their public read paths instead of duplicating key construction in tests
- inspect raw Cachex values to verify the expected key and `{:cached, value}` contracts
- reject database fallback for warmed SourceSchemas entries and Sources ID, internal atom-token, and external binary-token entries
- verify warmed sources preserve plan-derived `retention_days`
- cover empty warmer input, unresolved users, and distinct retention outcomes across multiple user plans
- verify repository query count remains constant when growing from one to six source users

## Test plan

- `../bin/test test/logflare/source_schemas/cache_warmer_test.exs test/logflare/sources/cache_warmer_test.exs test/logflare/sources/sources_cache_test.exs`
- included in the final affected stack suite: 2 doctests, 1 property and 169 tests, 0 failures (3 excluded)

## Stack

Review each PR against its immediate base:

1. #3913 — Sources and SourceSchemas cache entry correctness, value parity, and source-user deletion tolerance during retention hydration (this PR; based on `main`)
2. #3915 — bounded, deduplicated, deterministic, and composable warmer queries
3. #3916 — best-effort SystemCache startup and foreground monitor-failure handling

The proposed KeyValues changes in #3914 were removed from this stack and remain deferred. The WAL fence follow-up in #3924 was closed because its coordination complexity was disproportionate and it did not solve post-invalidation replica-lag refills; any future stronger guarantee should use replica replay/LSN coordination.

### DIFF
--- a/lib/logflare/sources/cache_warmer.ex
+++ b/lib/logflare/sources/cache_warmer.ex
@@ -28,7 +28,8 @@ defmodule Logflare.Sources.CacheWarmer do
 
         [
           {{:get_by, [[id: s.id]]}, value},
-          {{:get_by, [[token: s.token]]}, value}
+          {{:get_by, [[token: s.token]]}, value},
+          {{:get_by, [[token: Atom.to_string(s.token)]]}, value}
         ]
       end
 
--- a/lib/logflare/sources/cache_warmer.ex
+++ b/lib/logflare/sources/cache_warmer.ex
@@ -20,8 +20,8 @@ defmodule Logflare.Sources.CacheWarmer do
         value = {:cached, s}
 
         [
-          {{:get_by, [id: s.id]}, value},
-          {{:get_by, [token: s.token]}, value}
+          {{:get_by, [[id: s.id]]}, value},
+          {{:get_by, [[token: s.token]]}, value}
         ]
       end
 
--- a/lib/logflare/sources/cache_warmer.ex
+++ b/lib/logflare/sources/cache_warmer.ex
@@ -1,9 +1,13 @@
 defmodule Logflare.Sources.CacheWarmer do
-  alias Logflare.Repo
-  alias Logflare.Sources.Source
+  use Cachex.Warmer
+
   import Ecto.Query
 
-  use Cachex.Warmer
+  alias Logflare.Billing
+  alias Logflare.Repo
+  alias Logflare.Sources
+  alias Logflare.Sources.Source
+  alias Logflare.User
   @impl true
   def execute(_state) do
     # Get sources that have been active in the last day, similar to ingesting users pattern
@@ -14,6 +18,7 @@ defmodule Logflare.Sources.CacheWarmer do
         limit: 1_000
       )
       |> Repo.all()
+      |> put_retention_days()
 
     get_kv =
       for s <- sources do
@@ -26,5 +31,25 @@ defmodule Logflare.Sources.CacheWarmer do
       end
 
     {:ok, List.flatten(get_kv)}
+  end
+
+  defp put_retention_days([]), do: []
+
+  defp put_retention_days(sources) do
+    user_ids = sources |> Enum.map(& &1.user_id) |> Enum.uniq()
+
+    users =
+      from(u in User,
+        where: u.id in ^user_ids,
+        preload: :billing_account
+      )
+      |> Repo.all()
+
+    plans_by_user_id = Billing.get_plans_by_users(users)
+
+    Enum.map(sources, fn source ->
+      plan = Map.fetch!(plans_by_user_id, source.user_id)
+      Sources.put_retention_days(source, plan)
+    end)
   end
 end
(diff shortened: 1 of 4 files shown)


test/logflare/source_schemas/cache_warmer_test.exs
```diff
@@ -0,0 +1,24 @@
...
+    user = insert(:user)
+    source = insert(:source, user: user, log_events_updated_at: NaiveDateTime.utc_now())
+    source_schema = insert(:source_schema, source: source)
+    cache_key = {:get_source_schema_by, [source_id: source.id]}
+
+    assert {:ok, pairs} = CacheWarmer.execute(nil)
+    assert {^cache_key, {:cached, %{id: source_schema_id}}} = List.keyfind(pairs, cache_key, 0)
+    assert source_schema_id == source_schema.id
+    assert {:ok, true} = Cachex.put_many(Cache, pairs)
+    assert {:ok, {:cached, %{id: ^source_schema_id}}} = Cachex.get(Cache, cache_key)
```
chasers: Should this call go through the SourceSchema.Cache module?

amokan: Yeah. Probably should rather than Cachex directly.

chasers (approved):

test/logflare/source_schemas/cache_warmer_test.exs
```diff
@@ -0,0 +1,28 @@
...
+
+    assert {:ok, pairs} = CacheWarmer.execute(nil)
+    assert {^cache_key, {:cached, %{id: source_schema_id}}} = List.keyfind(pairs, cache_key, 0)
+    assert source_schema_id == source_schema.id
+    assert {:ok, true} = Cachex.put_many(Cache, pairs)
+
+    # Check Cachex directly so read-through fallback cannot hide malformed warmer output.
+    assert {:ok, {:cached, %{id: ^source_schema_id}}} = Cachex.get(Cache, cache_key)
+
+    assert %{id: ^source_schema_id} = Cache.get_source_schema_by(source_id: source.id)
```
chasers: Don't think this will catch it.

chasers (dismissed):

amokan (approved):

lib/logflare/sources/cache_warmer.ex
```diff
@@ -14,17 +20,53 @@ defmodule Logflare.Sources.CacheWarmer do
...
+    {hydrated_sources, skipped_count} =
+      Enum.reduce(sources, {[], 0}, fn source, {hydrated_sources, skipped_count} ->
+        case Map.fetch(plans_by_user_id, source.user_id) do
+          {:ok, plan} ->
+            {[Sources.put_retention_days(source, plan) | hydrated_sources], skipped_count}
+
+          :error ->
+            {hydrated_sources, skipped_count + 1}
+        end
+      end)
```
Ziinc, line 62: I'm pretty sure I recently added a function to do this when I was refactoring the retention days warming on either the Users or Sources module, would be worth to check before merging.

Ziinc (approved):

lib/logflare/sources/cache_warmer.ex
djwhitt: Checked on this. The code here is using some of what you added. There's also `Users.preload_defaults/1`, but it would require some changes. Going to leave it off this PR, but I might do a follow up.

PR #3914, opened 2026-09-03
fix: fence key value cache warming

djwhitt:
> **Deferred:** This PR was removed from the active cache-warmer stack and closed for now.

## Summary

- fence KeyValues warming with a persistent atomic invalidation generation so concurrent updates and deletes cannot be reinserted from stale snapshots
- use primary reads and chunked streaming for warm queries, with conservative cleanup when a generation changes
- tie successful full-warm initialization to the current Cachex process PID so cache restarts trigger another full warm
- fail closed on Cachex write errors and recover from exceptions, throws, and exits while retaining bounded retry scheduling
- add deterministic regression coverage for invalidation races, cache restarts, write failures, and recovery

## Test plan

- `../bin/test test/logflare/key_values/cache_warmer_test.exs test/logflare/key_values/cache_test.exs`
- included in the final affected suite: 1 property and 155 tests, 0 failures

## Stack

This PR is no longer part of the active stack. The remaining series is #3913 → #3915 → #3916.

## Operational note

KeyValues warm queries now read the primary to close the pre-run replica-lag window. If an invalidation occurs during an initial full warm, the newly warmed cache is cleared and marked initialized; this favors correctness and avoids repeated full-scan retries at the cost of losing that one prefetch.

### DIFF
--- a/lib/logflare/key_values/cache.ex
+++ b/lib/logflare/key_values/cache.ex
@@ -3,6 +3,7 @@ defmodule Logflare.KeyValues.Cache do
 
   alias Logflare.ContextCache
   alias Logflare.KeyValues
+  alias Logflare.KeyValues.CacheWarmer
   alias Logflare.Repo
   alias Logflare.Utils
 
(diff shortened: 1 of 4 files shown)


chasers:
Key Values feature is not really used at all right now and and needs to more love separately which I would probably include in that workstream and not here.

PR #3915, opened 2026-09-03
fix: bound cache warmer source joins

djwhitt:
## Summary

- apply backend warmer limits in SQL instead of loading every ingesting backend before truncation
- pre-aggregate latest source activity per backend, then rank distinct filtered backend IDs so ingesting order composes with access and `has_sources_or_rules` filters
- make ingesting-user selection distinct before applying its limit
- add deterministic tie-breakers and regression coverage proving limits count unique backends and users
- use the shared positive-integer guard for backend limits

## Why

The original joins could return the same backend or user once per active source, so duplicates consumed the warmer limit and the application loaded an unbounded result before truncation.

Grouping ingesting activity directly in the outer backend query fixed the simple warmer call, but backend access and `has_sources_or_rules` filters already add `DISTINCT`. Combining those filters could either demote latest-activity ordering behind backend ID or produce PostgreSQL `42P10`.

The backend path now reduces ordinary filters to distinct backend IDs, joins them to a separately aggregated latest-activity query, and applies ordering and the limit only in the deduplication-free outer query.

## Test coverage

- limits unique ingesting backends by latest source activity
- preserves latest-activity ordering with `has_sources_or_rules` in either keyword order
- supports ingesting filters through `list_backends_by_user_access/2` without a `DISTINCT`/`ORDER BY` conflict
- deduplicates ingesting users before limiting
- deterministically breaks equal-activity ties by ID

## Test plan

- `../bin/test test/logflare/backends/cache_test.exs test/logflare/backends_test.exs test/logflare/users/users_test.exs test/logflare/users/cache_test.exs`
- included in the final affected stack suite: 2 doctests, 1 property and 169 tests, 0 failures (3 excluded)

## Stack

Review each PR against its immediate base:

1. #3913 — Sources and SourceSchemas cache entry correctness, value parity, and source-user deletion tolerance during retention hydration
2. #3915 — bounded, deduplicated, deterministic, and composable warmer queries (this PR; based on #3913)
3. #3916 — best-effort SystemCache startup and foreground monitor-failure handling

The proposed KeyValues changes in #3914 were removed from this stack and remain deferred. The WAL fence follow-up in #3924 was closed because its coordination complexity was disproportionate and it did not solve post-invalidation replica-lag refills; any future stronger guarantee should use replica replay/LSN coordination.

### DIFF
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -80,13 +80,17 @@ defmodule Logflare.Backends do
         join(q, :inner, [b], r in assoc(b, :rules), on: r.source_id == ^source_id)
 
       # filter down to backends with sources that have recently ingested.
-      # orders by the last active.
+      # orders unique backends by the last active source.
       {:ingesting, true}, q ->
         q
         |> join(:inner, [b], s in assoc(b, :sources),
           on: s.log_events_updated_at >= ago(1, "day")
         )
-        |> order_by([..., s], {:desc, s.log_events_updated_at})
+        |> group_by([b], b.id)
+        |> order_by([..., s], desc: max(s.log_events_updated_at))
+
+      {:limit, limit}, q when is_integer(limit) and limit > 0 ->
+        limit(q, ^limit)
 
       {:user_id, id}, q ->
         where(q, [b], b.user_id == ^id)
--- a/lib/logflare/backends.ex
+++ b/lib/logflare/backends.ex
@@ -67,60 +67,87 @@ defmodule Logflare.Backends do
   end
 
   defp filter_backends(query, filters) do
-    Enum.reduce(filters, query, fn
-      {:types, types}, q when is_list(types) ->
-        where(q, [b], b.type in ^types)
+    {ingesting?, filters} = Keyword.pop(filters, :ingesting, false)
+    {limit, filters} = Keyword.pop(filters, :limit)
 
-      # filter down to backends of this source
-      {:source_id, id}, q ->
-        join(q, :inner, [b], s in assoc(b, :sources), on: s.id == ^id)
+    query =
+      Enum.reduce(filters, query, fn
+        {:types, types}, q when is_list(types) ->
+          where(q, [b], b.type in ^types)
 
-      # filter down to backends with rules destinations
-      {:rules_source_id, source_id}, q ->
-        join(q, :inner, [b], r in assoc(b, :rules), on: r.source_id == ^source_id)
+        # filter down to backends of this source
+        {:source_id, id}, q ->
+          join(q, :inner, [b], s in assoc(b, :sources), on: s.id == ^id)
 
-      # filter down to backends with sources that have recently ingested.
-      # orders unique backends by the last active source.
-      {:ingesting, true}, q ->
-        q
-        |> join(:inner, [b], s in assoc(b, :sources),
-          on: s.log_events_updated_at >= ago(1, "day")
-        )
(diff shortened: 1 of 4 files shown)


chasers (approved):

lib/logflare/backends.ex
```diff
@@ -80,13 +80,17 @@ defmodule Logflare.Backends do
...
       # filter down to backends with sources that have recently ingested.
-      # orders by the last active.
+      # orders unique backends by the last active source.
       {:ingesting, true}, q ->
         q
         |> join(:inner, [b], s in assoc(b, :sources),
           on: s.log_events_updated_at >= ago(1, "day")
         )
-        |> order_by([..., s], {:desc, s.log_events_updated_at})
+        |> group_by([b], b.id)
```
amokan: 🤖 

>  `lib/logflare/backends.ex:84` (_medium, latent_) — the new `group_by`/`order_by max(...)` composes badly with the module's two existing dedup mechanisms, both verified empirically:
   - `list_backends(ingesting: true, has_sources_or_rules: true)` → `distinct_backends/1` adds `DISTINCT ON (b.id)` and Ecto prepends `b0."id"` to `ORDER BY`, silently demoting `max(...) DESC` to a secondary key. With `limit:` you get lowest-id backends, not most-recently-active.
   - `list_backends_by_user_access(user, ingesting: true)` → `Teams.filter_by_user_access/2` (`lib/logflare/teams.ex:164`) adds `distinct: true` → hard Postgres `42P10`, i.e. a 500.
>
> No caller combines these today, but `filter_backends/2` advertises them as freely composable. Options: (a) fold ingesting dedup into `distinct_backends/1` semantics, ranking via a subquery/lateral `max` with a matching leading `ORDER BY`; (b) rewrite ingesting as `where b.id in (subquery)` and order by a joined pre-aggregated `max` so no `GROUP BY` reaches the outer query; (c) at minimum document `:ingesting` as mutually exclusive with `:has_sources_or_rules` and unsupported in `list_backends_by_user_access/2`. Recommend (b) — keeps the ordering intent and leaves the outer query dedup-free.

lib/logflare/backends.ex
```diff
@@ -80,13 +80,17 @@ defmodule Logflare.Backends do
...
+        |> order_by([b, ..., s], desc: max(s.log_events_updated_at), asc: b.id)
+
+      {:limit, limit}, q when is_integer(limit) and limit > 0 ->
```
amokan:
```suggestion
      {:limit, limit}, q when is_pos_integer(limit) ->
```

amokan (commented):
The 🤖 is worth glancing at. The other is just a nit to use one of our guards.

lib/logflare/backends.ex
```diff
@@ -67,57 +67,88 @@ defmodule Logflare.Backends do
...
-        |> join(:inner, [b], s in assoc(b, :sources),
-          on: s.log_events_updated_at >= ago(1, "day")
-        )
-        |> order_by([..., s], {:desc, s.log_events_updated_at})
+    {ingesting?, filters} = Keyword.pop(filters, :ingesting, false)
+    {limit, filters} = Keyword.pop(filters, :limit)
 
-      {:user_id, id}, q ->
-        where(q, [b], b.user_id == ^id)
+    query =
```
Ziinc: Could just pipe this into the next two function calls, why do we need this binding?

Ziinc (approved):

PR #3923, opened 2026-09-04
feat: local disk wal

Baishan:
Messages in the producer pipeline can be lost if the container ooms or supervisor restarts before they are written to gcs/pubsub. Blocking ingest waiting on a group commit to gcs increased latency by an additional 150-200ms.

This aims to achieve the same blocking until the message is durable without the additional latency increase bringing us closer to at least once delivery. This is achieved by spooling messages to a local wal file that can survive container restarts. Requests wait until they are appended to the wal before sending the 200 response code. The wal is rotated and sealed every batch_timeout (1000ms) and then uploaded to gcs. On startup any sealed wal files are uploaded to gcs as part of recovery ensuring the messages weren't lost.

This also introduces a writehealth module that will signal the healthcheck controller that the spool has become unhealthy and simultaneously disable spooling of future requests.

A directory on the boot disk is mounted for spool usage. The expected spool should not exceed a few mb in normal operation.

This is currently configured with one active partition which was able to handle more than what we expect in prod and increases the size of chunks written to gcp.  It can be increased if the partition becomes a bottlneck.


cloudbuild/gce-startup.sh
```diff
@@ -64,6 +69,12 @@ wait_for_docker() {
   return 1
 }
 
+prepare_wal_dir() {
```
Ziinc, line 72: why not part of application startup sequence?

lib/logflare/backends/spool/buffer/mem.ex
```diff
@@ -0,0 +1,70 @@
...
+    %{
+      max_batch_bytes: Keyword.get(spool_config, :mem_max_batch_bytes, @default_max_batch_bytes),
+      pending: [],
+      pending_bytes: 0,
+      pending_count: 0
+    }
+  end
+
+  @impl true
+  def append(state, segment, raw_byte_size, event_count) do
```
Ziinc, line 37: running this through the genserver would result in bottlenecks, no? ets then doing a generation + traverse on roll would be non-blocking too

lib/logflare/backends/spool/buffer/wal.ex
```diff
@@ -0,0 +1,353 @@
...
+
+  @impl true
+  def init(opts) do
+    index = Keyword.fetch!(opts, :index)
+    wal_dir = Keyword.fetch!(opts, :wal_dir)
+    File.mkdir_p!(wal_dir)
+
+    active_path = active_path(wal_dir, index)
+    _offset = Framing.recover!(active_path)
+    {:ok, fd} = :file.open(active_path, [:append, :raw, :binary])
```
Ziinc, line 54: can consider moving file handling to handle_continue

lib/logflare/backends/spool/buffer/wal.ex
```diff
@@ -0,0 +1,353 @@
...
+      {:error, _reopen_reason, state} ->
+        {:error, original_reason, state}
+    end
+  end
+
+  defp try_write(fd, segment) do
+    with :ok <- :file.write(fd, segment) do
+      :file.datasync(fd)
+    end
+  end
```
Ziinc: very convoluted and hard to follow, try a main with/do for main happy path and recursive retry+pattern match and keep error handling together

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -613,22 +627,98 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
...
   # Decompression and parsing are both capable of raising on truncated or
-  # otherwise corrupt content (:zlib.gunzip/1 and :erlang.binary_to_term/2
-  # both crash rather than returning an error tuple) — caught here, close to
-  # the queue handle, so the caller can ack (drop) the poison message instead
-  # of losing the handle to safe_fetch_next's outer rescue and retrying forever.
+  # otherwise corrupt content (:zlib.gunzip/1, :ezstd.decompress/1, and
+  # :erlang.binary_to_term/2 all crash or return {:error, _} rather than a
+  # clean error tuple in every case) — caught here, close to the queue handle,
+  # so the caller can ack (drop) the poison message instead of losing the
+  # handle to safe_fetch_next's outer rescue and retrying forever.
```
Ziinc: inline comments need to get less verbose and if its critical should be moved to moduledoc.
inline comments should mostly be the guide the reader on intent not for such a huge block on implementation details.

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -613,22 +627,98 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
...
+        end
+
+      version ->
+        {:error, {:unsupported_version, version}}
+    end
   rescue
     e -> {:error, {:decode_failed, e}}
   catch
     kind, reason -> {:error, {:decode_failed, %RuntimeError{message: inspect({kind, reason})}}}
   end
```
Ziinc, line 717: hmm why nested 2 elem tuple instead of 3 elem?

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -613,22 +627,98 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
...
 
+  defp decode_legacy_content(file_key, raw) do
```
Ziinc: why do we need this thin wrapping function used only once

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
```diff
@@ -613,22 +627,98 @@ defmodule Logflare.Backends.Spool.ConsumerPipeline.QueueProducer do
...
+    {:ok, events} = parse_content(file_key, content)
+    events
+  end
+
+  defp decompress_by_extension(content, file_key) do
+    cond do
+      String.ends_with?(file_key, ".gz") -> :zlib.gunzip(content)
+      String.ends_with?(file_key, ".zst") -> decompress_zstd!(content)
+      true -> content
+    end
```
Ziinc, line 745: don't know if it will be faster, but can consider reversing the string and pattern matching on function clause

lib/logflare/backends/spool/committer.ex
```diff
@@ -0,0 +1,239 @@
...
+  @spec commit_async(
+          pid(),
+          (-> {:ok, binary()} | {:error, term()}),
+          non_neg_integer(),
+          atom(),
+          term(),
+          config()
+        ) :: {:ok, pid()}
+  def commit_async(acknowledger, body_thunk, total_count, trigger, context, config) do
+    Task.start(fn ->
```
Ziinc, line 75: hmm what's the frequency of these task spawning? if its consistent rate then would be better to have a persistent process worker doing it

lib/logflare/backends/spool/committer.ex
```diff
@@ -0,0 +1,239 @@
...
+
+  defp retry_delay_ms do
+    Application.get_env(:logflare, :spool, [])
+    |> Keyword.get(:retry_delay_ms, @default_retry_delay_ms)
+  end
+
+  defp emit_storage_put_telemetry(format, bytes, result, upload_us) do
+    :telemetry.execute(
+      [:logflare, :backends, :spool, :storage, :put],
+      %{count: 1, bytes: bytes, upload_duration: upload_us},
```
Ziinc:
```suggestion
      %{bytes: bytes, upload_duration: upload_us},
```
count can be derived with Telemetry.Metrics.counter

lib/logflare/backends/spool/committer.ex
```diff
@@ -0,0 +1,239 @@
...
+      %{format: format, result: if(match?({:ok, _}, result), do: :ok, else: :error)}
+    )
+  end
+
+  defp emit_batch_result(result, stage, batch_size) do
+    :telemetry.execute(
+      [:logflare, :backends, :spool, :producer, :batch],
+      %{count: batch_size},
```
Ziinc:
```suggestion
      %{batch_size: batch_size},
```
for clarity

lib/logflare/backends/spool/encoder.ex
```diff
@@ -0,0 +1,161 @@
...
+
+  @zstd_compression_level 3
+
+  # The wire format version this build writes (see Committer.file_key/1) —
+  # bump this, and add a matching decode path in
+  # ConsumerPipeline.QueueProducer, whenever the frame layout or encoding
+  # scheme changes in a way older code can't already read. Never remove an
+  # old version's decode path once files tagged with it might still be
+  # in-flight or sitting in a bucket.
+  @current_version 2
```
Ziinc, line 20: might want to include a ci script or mix task to ensure that this version gets bumped on changes to this file to avoid accidental commits without a version bump.

lib/logflare/backends/spool/encoder.ex
```diff
@@ -0,0 +1,161 @@
...
+  Extracts the version embedded in a spool file_key by `file_key_with_version/2`
+  (e.g. `"0/uuid.v2.ndjson"` -> `2`), or `:legacy` if none is present — the
+  pre-versioning main-branch producer wrote plain `"0/uuid.ndjson"` keys with
+  no version tag at all.
+  """
+  @spec file_key_version(String.t()) :: pos_integer() | :legacy
+  def file_key_version(file_key) do
+    case Regex.run(~r/\.v(\d+)\./, file_key) do
+      [_match, digits] -> String.to_integer(digits)
+      nil -> :legacy
```
Ziinc, line 35: i assume that we'll want to remove this legacy handling across the board once this is out on prod?

lib/logflare/backends/spool/encoder.ex
```diff
@@ -0,0 +1,161 @@
...
+  stays identical regardless of which one wrote a given file.
+  """
+  @spec build_file_key(non_neg_integer(), :ndjson | :etf, boolean(), :gzip | :zstd) :: String.t()
+  def build_file_key(index, format, compress, algorithm) do
+    ext = file_extension(format, compress, algorithm)
+    file_key_with_version("#{index}/#{generate_uuidv7()}", ext)
+  end
+
+  @spec generate_uuidv7() :: String.t()
+  def generate_uuidv7 do
```
Ziinc: why not use https://hex.pm/packages/uuidv7 ? 

if want to handroll this should move it to utils, though i think its unnecessary surface area if a lib already has the implementation

lib/logflare/backends/spool/buffer/mem.ex
Baishan: its held up fine in testing. Its a very fast write path. it can be scaled out via partition count but ive found it worked best with one.

lib/logflare/backends/spool/encoder.ex
Baishan: yes it was technically needed for the first rollout but we can actually do without it by letting the existing queues drain before deploy

lib/logflare/backends/spool/consumer_pipeline/queue_producer.ex
Baishan: I'll get 🤖 to do a pass on these and tidy it up

cloudbuild/gce-startup.sh
Baishan: it needs to be writeable and survive container restart.

PR #3924, opened 2026-09-04
fix: fence context cache writes against WAL invalidation

djwhitt:
## Summary

- add a session-scoped `ContextCache.WriteFence` that serializes cache publication with WAL invalidation
- route read-through misses, gossip receives, and first-party WAL-backed warmers through fenced writes
- preserve configured read-replica use for ContextCache misses
- remove cached `nil`/empty-list lookups when a new-record invalidation arrives

## Why

Cachex executes fallback getters and warmer callbacks before it performs their eventual cache write. A WAL invalidation could therefore scan and delete the cache while a stale database snapshot was still loading, after which Cachex would publish that stale value back into the cache.

Short-lived gossip tombstones narrow one instance of the race, but they do not order local read-through or warmer writes and cannot provide a durable happens-before relationship.

## Ordering

The write fence starts before the context caches and only releases writers after `CacheBuster` subscribes to WAL transaction broadcasts.

For each active database snapshot:

- if invalidation reaches the fence first, affected primary-key groups are omitted from the cache write
- if publication reaches the fence first, the following invalidation scan sees and removes the published entries
- if `CacheBuster` exits, active snapshots fail closed and new snapshots wait for its replacement
- if the fence restarts, tokens from the previous incarnation cannot commit

Keyword-based custom busts conservatively fence the whole affected context. Normal primary-key invalidations only discard the matching groups from a bounded warmer batch.

## Scope

Covered write paths:

- `ContextCache.fetch/3` read-through misses
- `ContextCache.Gossip.receive/3` miss publication
- Backends, Rules, SourceSchemas, Sources, and Users Cachex warmers

The KeyValues batch warmer remains separately deferred. SystemCache is not WAL-backed.

ContextCache misses continue to use configured read replicas. The fence prevents a cache population already in flight from surviving an overlapping invalidation; it does not prevent a later cache fill from observing ordinary replica lag after that invalidation has completed. A stronger guarantee would require replica replay/LSN coordination and is outside this PR.

## Tests

- deterministic invalidate-before-commit and commit-before-invalidate coverage
- mixed-batch, alias-group, custom keyword bust, readiness, restart, and duplicate-key coverage
- positive and negative read-through race coverage
- focused/affected suite: 1 property, 80 tests, 0 failures
- distributed gossip: 2 tests, 0 failures
- format check, warnings-as-errors compile, `mix lint.all`, and `mix test.typings`

## Stack

Stacked after #3916:

`#3913 → #3915 → #3916 → this PR`

### DIFF
--- a/lib/logflare/backends/cache_warmer.ex
+++ b/lib/logflare/backends/cache_warmer.ex
@@ -1,16 +1,31 @@
 defmodule Logflare.Backends.CacheWarmer do
+  require Logger
+
   alias Logflare.Backends
+  alias Logflare.ContextCache.WriteFence
 
   use Cachex.Warmer
+
   @impl true
   def execute(_state) do
+    {:ok, snapshot} = WriteFence.begin_snapshot(Backends)
     backends = Backends.list_backends(ingesting: true, limit: 1_000)
 
-    get_kv =
-      for b <- backends do
-        {{:get_backend, [b.id]}, {:cached, b}}
+    entry_groups =
+      for backend <- backends do
+        {backend.id, [{{:get_backend, [backend.id]}, {:cached, backend}}]}
       end
 
-    {:ok, get_kv}
+    case WriteFence.commit(snapshot, Backends.Cache, entry_groups) do
+      {:ok, %{skipped: skipped}} ->
+        if skipped > 0 do
+          Logger.warning("Backends cache warmer skipped #{skipped} invalidated backends")
+        end
+
+      {:error, reason} ->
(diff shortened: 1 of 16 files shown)


djwhitt:
:robot:

> Closing because the shared write-fence approach adds disproportionate coordination and lifecycle complexity for a low-frequency in-flight race, while preserving replica-backed misses means it still does not prevent stale cache refills caused by replica lag.
>
> The active stack will remain through #3916. A future consistency change should use replica replay/LSN coordination if production evidence justifies the stronger guarantee.

PR #3925, opened 2026-09-04
[WIP] ClickHouse Mapping Improvements

amokan:
## Overview

...

### Key Changes

- Better consistency on mapping for `resource_attributes` across all three types (_additional paths and addition of `resource_attributes.environment`_)
- Replaced hardcoded mapping configuration ids from benchmark scripts with calls to `MappingDefaults.config_id/1`
- Change the `log_attributes` (_logs_), `attributes` (_metrics_), and `span_attributes` (_trace_) config to elevate keys for both `$.metadata` as well as `$.attributes`.
- Drops redundant multi-elevate bailout from the NIF (`elevate.len() > 1`) so it uses the fused flatten path

### Risks / Considerations

- Existing queries referencing things like `log_attributes["attributes.foo"]` would fail as the attribute is now elevated to `log_attributes["foo"]`. Traces and metrics have the same concerns within their `attributes[...]` and `span_attributes` fields, but very little concern with those as they are generally not user-facing.


amokan:
Superseded by stack #3933 — split into six reviewable layers: #3927 → #3928 → #3929 → #3930 → #3931 → #3932. Same commits, same final diff.

PR #3927, opened 2026-09-04
ClickHouse OTEL mappings: consistent `resource_attributes`, elevate attribute keys

amokan:
## Overview

Brings the `resource_attributes` mapping in line across logs, metrics, and traces, and elevates attribute keys out of their `$.metadata` / `$.attributes` envelopes. First layer of the mapping improvements stack — config only, no NIF changes.

### Key Changes

- Unified the `resource_attributes` path set across all three event types, so the same key resolves the same way regardless of signal
- Added `resource_attributes.environment` to all three types (_with validation on the `prod` value_)
- Added fallback paths for `cluster`, `node`, and `region`, including the underscore-namespaced variants emitted by some collectors
- Changed `log_attributes` (_logs_), `attributes` (_metrics_), and `span_attributes` (_traces_) to elevate keys for both `$.metadata` and `$.attributes`
- Replaced hardcoded mapping configuration ids in the benchmark scripts with calls to `MappingDefaults.config_id/1`

### Risks / Considerations

- Existing queries referencing things like `log_attributes["attributes.foo"]` would fail as the attribute is now elevated to `log_attributes["foo"]`. Traces and metrics have the same concern within their `attributes[...]` and `span_attributes` fields, but very little concern with those as they are generally not user-facing.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
@@ -136,7 +136,13 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
              "$.metadata.app_name",
              "$.metadata.parsed.application_name"
            ]},
-          {"cluster", ["$.metadata.cluster", "$.cluster", "$.resource.cluster"]},
+          {"cluster",
+           [
+             "$.metadata.cluster",
+             "$.metadata.context.cluster",
+             "$.cluster",
+             "$.resource.cluster"
+           ]},
           {"environment",
            [
              "$.metadata.environment",
@@ -147,7 +153,13 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
           {"host", ["$.metadata.host", "$.metadata.context.host", "$.host", "$.resource.host"]},
           {"instance_id", ["$.metadata.instance_id"]},
           {"machine_id", ["$.machine_id"]},
-          {"node", ["$.metadata.context.vm.node", "$.resource.node"]},
+          {"node",
+           [
+             "$.metadata.node",
+             "$.metadata.context.vm.node",
+             "$.node",
+             "$.resource.node"
+           ]},
           {"organization_id", ["$.organization_id", "$.org_id"]},
           {"organization_slug", ["$.organization_slug"]},
           {"project",
@@ -159,7 +171,13 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
              "$.metadata.tenant",
              "$.metadata.tenantId"
            ]},
-          {"region", ["$.metadata.region", "$.region", "$.resource._project_region"]},
+          {"region",
+           [
+             "$.metadata.region",
+             "$.region",
+             "$.resource.region",
+             "$.resource._project_region"
+           ]},
           {"service_name",
            ["$.resource.service.name", "$.resource._service_name", "$.service_name"]},
           {"vector_file", ["$.metadata.vector_file"]},
@@ -287,7 +305,13 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
              "$.metadata.app_name",
              "$.metadata.parsed.application_name"
            ]},
-          {"cluster", ["$.metadata.cluster", "$.cluster", "$.resource.cluster"]},
+          {"cluster",
+           [
+             "$.metadata.cluster",
+             "$.metadata.context.cluster",
+             "$.cluster",
+             "$.resource.cluster"
+           ]},
           {"environment",
            [
              "$.metadata.environment",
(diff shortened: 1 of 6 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
```diff
@@ -135,11 +136,30 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
...
-          {"cluster", ["$.metadata.cluster", "$.cluster", "$.resource.cluster"]},
-          {"host", ["$.metadata.host", "$.metadata.context.host", "$.host"]},
+          {"cluster",
+           [
+             "$.metadata.cluster",
+             "$.metadata.context.cluster",
+             "$.cluster",
+             "$.resource.cluster"
+           ]},
+          {"environment",
```
djwhitt, line 146: :robot:

> Is the replacement behavior of `pick` intentional here? Once any entry resolves, it replaces the `$.resource` fallback. Adding `environment` and the other new paths means this payload will drop `cloud.provider`:
>
> ```elixir
> %{"resource" => %{"environment" => "staging", "cloud" => %{"provider" => "aws"}}}
> ```
>
> If preserving arbitrary resource attributes is intended, a config-only option is to add `{"resource", ["$.resource"]}` to `pick` with `elevate_keys: ["resource"]`; the normalized picks will still win collisions.
>
> Please add a regression test either way—asserting that unpicked resource attributes are deliberately dropped, or that they are preserved alongside newly picked keys.

lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
```diff
@@ -90,6 +90,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
       Field.string("service_name",
         paths: [
           "$.resource.service.name",
+          "$.resource._service_name",
```
djwhitt, line 93: :robot:

> Is it intentional for `$.resource._service_name` to take precedence over the top-level `$.service_name`? Before this change, a payload containing both resolved to the top-level value; after this change, the underscore-namespaced resource value wins.
>
> The existing precedence test covers `$.resource.service.name` over `$.resource._service_name`, but not `_service_name` over the top-level value. Please add a regression test with conflicting underscore-namespaced and top-level values across logs, metrics, and traces to make the intended order explicit.

djwhitt (changes requested):
:robot:

> Changes requested pending confirmation and regression coverage for the two unresolved mapping behaviors:
>
> - Whether new `pick` matches should replace the `$.resource` fallback and drop unpicked resource attributes.
> - Whether `$.resource._service_name` should take precedence over top-level `$.service_name`.
>
> Make the intended behavior explicit and test it across logs, metrics, and traces as applicable.

lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
amokan: Dumb 🤖. The stack handles all of this (_3929 in particular_) 🫠

So either we do it that way or I smash this into a big single PR.

lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
amokan: Test added

djwhitt: :robot:

> Resolved by [#3927](https://github.com/Logflare/logflare/pull/3927), which includes regression coverage confirming that `$.resource._service_name` takes precedence over top-level `$.service_name` across logs, metrics, and traces.

lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
djwhitt: :robot:

> Resolved by [#3929](https://github.com/Logflare/logflare/pull/3929), which adds `pick_mode: :merge` and regression coverage showing that uncurated resource attributes are preserved alongside curated picks across logs, metrics, and traces.

lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
```diff
@@ -265,19 +296,61 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
       Field.flat_map("resource_attributes",
         paths: ["$.resource"],
         pick: [
```
Ziinc, line 298: why not have a shared `@resource_attrs_pick` module constant list?

Ziinc (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
Ziinc: or a function that generates the pick configuration, so that any changes can be inserted into the relevant tuple easily?

djwhitt (approved):

PR #3929, opened 2026-09-04
mapper: add `pick_mode` option to `json`/`flat_map` field config

amokan:
## Overview

Adds a `:pick_mode` option to `json` / `flat_map` mapper fields so a curated `:pick` map and the raw source map can both reach the output. This PR also switches `resource_attributes` over to leverage this feature.

### Key Changes

- New `:pick_mode` field config option: `:replace` (_default, existing either/or behavior_) or `:merge`
- `:merge` unions the `:pick` entries with the remaining values from `:path` / `:paths`, with pick entries winning on key collision
- `:exclude_keys` and `:elevate_keys` continue to apply to the merged result
- `resource_attributes` now uses `:merge` on all three event types, so curated keys and uncurated resource keys both land in the map

### Risks / Considerations

- `:paths` remains a coalesce in both modes — only the first resolving path contributes. Payloads that split resource keys across multiple paths will still only pick up keys from the first one that resolves.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
@@ -126,6 +126,8 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
       ),
       Field.flat_map("resource_attributes",
         paths: ["$.resource"],
+        exclude_keys: ["_project_region", "_service_name"],
+        pick_mode: :merge,
         pick: [
           {"application_id", ["$.app_id", "$.application_id", "$.metadata.app_id"]},
           {"application_name",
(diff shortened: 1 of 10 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
```diff
@@ -126,6 +126,8 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
       ),
       Field.flat_map("resource_attributes",
         paths: ["$.resource"],
+        exclude_keys: ["_project_region", "_service_name"],
+        pick_mode: :merge,
```
djwhitt, line 148: :robot:

> This merge preserves uncurated resource attributes, but makes a trade-off for aliases and collisions: `_project_region` and `_service_name` are normalized and removed, while curated picks overwrite same-named raw resource values. This provides one canonical value per key, but means `resource_attributes` is not a lossless copy of the input when values disagree—for example, `metadata.environment` versus `resource.environment`. O11Y-1601 says all resource keys and values should reach the column; is this normalized, pick-wins behavior the intended interpretation? If so, documenting the trade-off would make the contract explicit.

amokan: This is outlined in the summary, but I'll get something in the moduledocs

amokan: Documented in the moduledocs for the appropriate mapper namespace(s) as well as the config module for ClickHouse.

djwhitt (approved):

PR #3931, opened 2026-09-04
mapper: validate precision, enum values, and filter lengths at compile time

amokan:
## Overview

Moves mapper config validation to compile time. Out-of-range values that previously wrapped or truncated silently at map time are now rejected when the config is compiled.

### Key Changes

- `precision` is range-checked against the values ClickHouse actually accepts, instead of narrowing and producing a column with the wrong scale
- `enum_values` entries are checked against the `Enum8` backing range instead of wrapping to the wrong variant
- String length filters reject negative values, which previously wrapped to a huge `usize` and made the filter silently unsatisfiable
- A `precision` or filter length that is present but not a 64-bit integer (_a bignum, float, or string_) is now a compile error instead of being read as "not set" and silently defaulted
- `Enum8` clamps rather than wraps if an out-of-range value still reaches the map path
- Consolidated the mapper tests into `test/logflare/mapper_test.exs`

### Risks / Considerations

- Errors are confined to `compile_mapping` — the hot data path still never errors, it clamps. A config that was silently producing wrong values will now fail to compile, which is the intended outcome but is a behavior change for any such config.


native/mapper_ex/src/mapping.rs
```diff
@@ -759,6 +776,17 @@ pub fn decode_enum8_data<'a>(env: Env<'a>, field: Term<'a>) -> Result<Enum8Data,
     })
 }
 
+/// Decode a `DateTime64` precision. ClickHouse allows 0-9; anything else would
+/// truncate or wrap when narrowed to `u8` and silently produce a column with
+/// the wrong scale.
+fn decode_precision<'a>(env: Env<'a>, field: Term<'a>) -> Result<u8, String> {
+    match get_int_key(env, field, "precision") {
+        None => Ok(9),
```
djwhitt, line 784: :robot:

> `get_int_key` returns `None` both when the key is absent and when an explicit Elixir integer cannot decode as `i64`. A JSON `precision` or length-filter value of `2^100` therefore compiles: precision silently defaults to `9`, while the filter is omitted. Decode a present term through `Result` and reject type or overflow failures; reserve `None` and defaults for an absent key.

djwhitt (commented):
:robot:

> `Field.datetime64/2` and `Field.array_datetime64/2` still normalize an explicit `precision: false` to `9` via `opts[:precision] || 9` (`field_config.ex:356` and `:386`). Consequently, this invalid supplied value never reaches the new NIF validation, and both configurations compile successfully with precision 9.
>
> Preserve explicitly supplied values—for example with `Keyword.get(opts, :precision, 9)`—so compile-time validation can reject `false`. Add coverage for both constructors.

djwhitt (approved):

PR #3940, opened 2026-09-07
feat: logstash ingestion

FranekKubis:
Closes: [O11Y-2422](https://linear.app/supabase/issue/O11Y-2422/implement-logstash-ingestion-for-the-elk-stack)

Add logstash transport to elastic backend

Adds a third `transport` value to the `:elastic` backend, delivering events to a
Logstash `http` input as ECS-flavoured JSON. `timestamp` and `event_message` are
lifted to `@timestamp` and `message` so a receiving pipeline needs no `date`
filter; remaining fields stay top-level and Logflare metadata is namespaced
under `logflare`. Reuses WebhookAdaptor (HTTP/1.1, optional basic auth, gzip)
via the existing `format_batch` hook.

Also fixes three pre-existing issues surfaced while verifying end to end:

- `ElasticAdaptor.validate_config/1` had no catch-all clause, so an unsupported
  transport raised `CaseClauseError` instead of returning an invalid changeset.
  The existing "rejects unknown transport" test was already failing on this.

- `Logflare.FinchDefaultHttp1` was never started in single-tenant Postgres mode,
  so every WebhookAdaptor-based backend (elastic, loki, webhook) crashed with
  "unknown registry" on delivery.

- The source edit page hardcoded which backend types it would render, hiding
  elastic, loki, clickhouse, s3, axiom, otlp, incidentio and last9 and making
  them impossible to attach to a source. Backend types and labels now derive
  from a single registry in `Logflare.Backends.Backend`, with a compile-time
  check that every adaptor has a label. Side effect: the "New backend" dropdown
  is now ordered alphabetically.

Adds a `logstash` docker-compose service and `test/logstash.conf` for local
verification. Note that outbound requests are SSRF-protected, so a local
Logstash must be reached through a tunnel rather than `localhost`.


lib/logflare/backends/adaptor/elastic_adaptor.ex
```diff
@@ -23,45 +52,216 @@ defmodule Logflare.Backends.Adaptor.ElasticAdaptor do
...
-          %{"Authorization" => "Basic #{basic_auth}"}
-        else
-          %{}
-        end
-    }
+      "logstash" ->
+        %{
+          url: config.url,
+          http: "http1",
+          gzip: Map.get(config, :gzip, true),
```
Ziinc: we should always set this to true and remove it as a configurable.

lib/logflare/backends/adaptor/elastic_adaptor.ex
```diff
@@ -23,45 +52,216 @@ defmodule Logflare.Backends.Adaptor.ElasticAdaptor do
...
-    |> Ecto.Changeset.cast(params, [:username, :password, :url])
+    types = %{
+      transport: :string,
+      # filebeat, logstash
+      url: :string,
+      username: :string,
+      password: :string,
+      # logstash, otlp
+      gzip: :boolean,
+      headers: {:map, :string},
```
Ziinc: We can extend filebeat to handle the headers as well do that this is a common config across all 3

lib/logflare/backends/adaptor/elastic_adaptor.ex
Ziinc: We can remove it from the otlp drain as well, don't see a point in letting users configure it

lib/logflare/backends/adaptor/elastic_adaptor.ex
```diff
@@ -23,45 +52,216 @@ defmodule Logflare.Backends.Adaptor.ElasticAdaptor do
...
+      # otlp
+      endpoint: :string,
+      protocol: :string
```
Ziinc: Protocol can be removed as a config since only one is supported.

lib/logflare/backends/adaptor/elastic_adaptor.ex
```diff
@@ -23,45 +52,216 @@ defmodule Logflare.Backends.Adaptor.ElasticAdaptor do
...
+      transport: :string,
+      # filebeat, logstash
+      url: :string,
+      username: :string,
+      password: :string,
+      # logstash, otlp
+      gzip: :boolean,
+      headers: {:map, :string},
+      # otlp
+      endpoint: :string,
```
Ziinc: Hmm wonder if it's worth converting this to use :url so that it's consistent across all 3...thoughts?

lib/logflare/backends/adaptor/elastic_adaptor.ex
```diff
@@ -78,14 +278,15 @@ defmodule Logflare.Backends.Adaptor.ElasticAdaptor do
...
-
-    changeset
-    |> validate_required([:url])
+  defp redact_headers(headers) do
+    for {k, v} <- headers, into: %{}, do: redact_header(k, v)
   end
 
-  @impl Logflare.Backends.Adaptor
-  def test_connection(_), do: {:error, :not_implemented}
+  defp redact_header(k, v) do
```
Ziinc: There some common header redaction logic across the backends, should move them to a utility function under the HttpBased module

mix.lock
```diff
@@ -152,7 +152,7 @@
...
   "retry": {:hex, :retry, "0.19.0", "aeb326d87f62295d950f41e1255fe6f43280a1b390d36e280b7c9b00601ccbc2", [:mix], [], "hexpm", "85ef376aa60007e7bff565c366310966ec1bd38078765a0e7f20ec8a220d02ca"},
   "rustler": {:hex, :rustler, "0.37.3", "5f4e6634d43b26f0a69834dd1d3ed4e1710b022a053bf4a670220c9540c92602", [:mix], [{:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: false]}], "hexpm", "a6872c6f53dcf00486d1e7f9e046e20e01bf1654bdacc4193016c2e8002b32a2"},
-  "rustler_precompiled": {:hex, :rustler_precompiled, "0.8.2", "5f25cbe220a8fac3e7ad62e6f950fcdca5a5a5f8501835d2823e8c74bf4268d5", [:mix], [{:castore, "~> 0.1 or ~> 1.0", [hex: :castore, repo: "hexpm", optional: false]}, {:rustler, "~> 0.23", [hex: :rustler, repo: "hexpm", optional: true]}], "hexpm", "63d1bd5f8e23096d1ff851839923162096364bac8656a4a3c00d1fff8e83ee0a"},
+  "rustler_precompiled": {:hex, :rustler_precompiled, "0.9.0", "3a052eda09f3d2436364645cc1f13279cf95db310eb0c17b0d8f25484b233aa0", [:mix], [{:rustler, "~> 0.23", [hex: :rustler, repo: "hexpm", optional: true]}], "hexpm", "471d97315bd3bf7b64623418b3693eedd8e47de3d1cb79a0ac8f9da7d770d94c"},
```
Ziinc: No deps changes, should remove this from diff

test/logflare/backends/adaptor/elastic_adaptor_test.exs
```diff
@@ -149,4 +218,114 @@ defmodule Logflare.Backends.Adaptor.ElasticAdaptorTest do
...
+          config: %{
+            transport: "logstash",
+            url: "http://localhost:8080",
+            username: "some user",
+            password: "some password"
+          }
+        )
+
+      start_supervised!({AdaptorSupervisor, {source, backend}})
+      :timer.sleep(500)
```
Ziinc: Should avoid timer usage

Ziinc: Should wrap an assertion with TestUtils retry_assert if there's a race condition and we need to wait for proc to startup

test/logflare_web/live/backends/backends_live_test.exs
```diff
@@ -432,6 +432,67 @@ defmodule LogflareWeb.BackendsLiveTest do
...
+      refute html =~ "Logstash URL"
+
+      html =
+        view
+        |> element("select#elastic-transport")
+        |> render_change(%{backend: %{config: %{transport: "logstash"}}})
+
+      assert html =~ "Logstash URL"
+      refute html =~ "Filebeat URL"
+      refute html =~ "Elastic Supabase Endpoint"
```
Ziinc: The otlp endpoint is bundled with elasticsearch actually, not supabase specific. 
```suggestion
      refute html =~ "Elasticsearch OTLP Endpoint"
```

Ziinc (changes requested):

Ziinc:
Could you also add a demo video of the setup + ingestion? 🙏

lib/logflare/backends/adaptor/elastic_adaptor.ex
FranekKubis: I would rather keep it as is because `endpoint` it is OTLP's own vocabulary (e.g. `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT`) and it is not quite a synonym for `url`.

FranekKubis:
I am adding a rough recording of clicking my way through the app, initializing new backend for logstash and sending logs to the logflare source connected to the logstash backend.
Video posted on linear:
https://linear.app/supabase/issue/O11Y-2422/implement-logstash-ingestion-for-the-elk-stack#comment-b2d875f2

lib/logflare/backends/adaptor/elastic_adaptor.ex
```diff
@@ -23,45 +52,197 @@ defmodule Logflare.Backends.Adaptor.ElasticAdaptor do
...
+
+      {_transport, _args} ->
+        {:error, :not_implemented}
+    end
+  end
+
+  @impl HttpBased.Client
+  def client_opts(%Backend{config: config}) do
+    [
+      url: config.endpoint,
```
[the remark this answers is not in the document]
FranekKubis: added `ssrf: true`

lib/logflare_web/live/backends/components/backend_form.heex
```diff
@@ -199,21 +200,85 @@
...
+          fields =
+            (input_value(f_config, :headers) || %{})
+            |> Stream.concat(Stream.repeatedly(fn -> {"", ""} end))
+            |> Enum.take(headers_number) %>
+          <%= for {{key, val}, i} <- Enum.with_index(fields, 1) do %>
+            <div class="form-group">
+              {label(f_config, "header#{i}_key", "Custom header #{i} - Key")}
+              {text_input(f_config, "header#{i}_key", value: key, class: "form-control")}
+              {label(f_config, "header#{i}_value", "Custom header #{i} - Value")}
+              {text_input(f_config, "header#{i}_value", value: val, class: "form-control")}
```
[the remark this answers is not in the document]
FranekKubis: Refactored into header component.

test/logflare_web/live/source_backends_live_test.exs
```diff
@@ -12,14 +12,37 @@ defmodule LogflareWeb.SourceBackendsLiveTest do
...
+
   test "able to add/remove additional backends", %{conn: conn, user: user, source: source} do
     backend = insert(:postgres_backend, user: user)
 
     {:ok, view, _html} =
       live_isolated(conn, SourceBackendsLive, session: %{"source_id" => source.id})
 
     # create
-    assert render(view) =~ "PostgreSQL"
+    assert render(view) =~ "Postgres"
```
Ziinc, line 45: PostgreSQL is the official name

Ziinc (approved):

PR #3950, opened 2026-09-09
chore: improve ClickHouse insert outcome, drop, and circuit breaker telemetry

amokan:
## Overview

Closes a known insert-layer observability gap. Finch latency/churn metrics were isolable by pool name, but nothing counted insert outcomes, so there was no error rate, no error breakdown, and no count of events actually lost.

The ClickHouse-side alert from `O11Y-2022` only sees requests that reach the server and is blind to pool checkout timeouts, connect timeouts, TLS failures, and circuit-breaker fast-fails. These metrics are emitted client-side, so they cover exactly those.

### Key Changes

- `logflare.clickhouse.insert.result.count` counts one event per batch tagged `backend_id`, `event_type`, `async`, `result`, and `error_class`
- `Ingester.error_class/1` turns a failure reason into a short atom safe to use as a metric tag
- `logflare.clickhouse.circuit_breaker.open.count` registers the breaker's existing `:open` event, which was emitted but never exported
- `logflare.ingest_event_queue.retry_dropped.count` counts events lost outright after an insert failure, tagged `:retries_exhausted` or `:circuit_breaker_open`
- The retry-requeue path emits its own `requeue_queue_unavailable` event instead of borrowing `not_initialized.dropped`, which had two unrelated emitters
- `retriable?/1` treats pool checkout timeouts and TLS alerts as retriable, and reuses `error_class/1` so retries and metrics stay in sync

### Risks

- **Dashboard impact.** Any panel reading `logflare.ingest_event_queue.not_initialized.dropped.count` for a ClickHouse backend loses the requeue-path contribution and needs the new series added. Three live queries in `o11y-team-tools` are affected (_tracked in `O11Y-2477`, to land before this reaches prod_)
- **Retry behavior change.** Pool checkout timeouts and TLS alerts now get one additional attempt (`@max_retries 1`) where they previously failed immediately.
- **Coupling.** Since `retriable?/1` reuses `error_class/1`, changing how errors are classified changes what gets retried. Both are pinned by tests
- `error_class/1` recovers the HTTP status by re-parsing the string `do_insert/5` formats two lines earlier. Contained, but fragile — the structured-error refactor is tracked separately in `O11Y-2478`

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
@@ -20,6 +20,14 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
   @pool_timeout 8_000
   @receive_timeout 15_000
   @too_many_parts_marker "Code: 252."
+  @retriable_error_classes [
+    :pool_timeout,
+    :timeout,
+    :connection_refused,
+    :connection_reset,
+    :connection_closed,
+    :tls_alert
+  ]
 
   @type error_class ::
           :too_many_parts
@@ -175,11 +183,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
   defp retriable?({:ok, %Tesla.Env{status: 429}}), do: true
   defp retriable?({:ok, _env}), do: false
 
-  defp retriable?({:error, reason})
-       when reason in [:timeout, :econnrefused, :econnreset, :closed],
-       do: true
-
-  defp retriable?({:error, _reason}), do: false
+  defp retriable?({:error, reason}), do: error_class(reason) in @retriable_error_classes
 
   @doc """
   Inserts a pre-gzipped RowBinary payload directly into ClickHouse.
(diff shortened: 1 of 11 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -20,6 +20,29 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
   @pool_timeout 8_000
   @receive_timeout 15_000
   @too_many_parts_marker "Code: 252."
+  @retriable_error_classes [
+    :pool_timeout,
```
djwhitt, line 25: :robot:

> This error class is not reachable for the configured ClickHouse pools. Both pools use HTTP/1, and Finch 0.20 raises a `RuntimeError` when `NimblePool.checkout!/4` times out rather than returning `{:error, %Finch.Error{reason: :pool_timeout}}`. The exception bypasses Tesla’s retry middleware and `handle_insert_result/4`.
>
> A saturated one-connection HTTP/1 pool reproduced this behavior: the Tesla call raised and the retry callback was invoked 0 times. Consequently, pool checkout timeouts are still neither retried nor included in the new insert-result metric. Normalize the actual Finch exception and cover it with a real HTTP/1 pool-saturation test; the current mocked return shape does not represent production behavior.

djwhitt (approved):

PR #3952, opened 2026-09-09
chore: return a structured clickhouse insert error instead of a formatted string

amokan:
## Overview

Follow-up to #3950.

`do_insert/5` flattened a non-200 ClickHouse response into `"HTTP #{status}: #{body}"`, and the two consumers that needed the status then parsed it back out of that sentence. Reformatting the message would have silently broken classification, sending the new insert error-rate metric to `:unknown` with no test failure.

Behavior-preserving — same classes for the same failures, same log output, no metric or telemetry changes.

### Key Changes

- `do_insert/5` returns `{:error, {:http, status, body}}` instead of a formatted string
- `error_class/1` matches the status directly, dropping the `"HTTP "` prefix match and `Integer.parse` round trip
- `too_many_parts?/1` accepts the structured shape, keeping its binary clause since `retriable?/1` calls it with the raw response body
- New `Ingester.error_string/1` renders the reason for the log line, wrapping the formatted string in `inspect/1` so the logged `error_string` is the same as before
- Specs move from `{:error, String.t()}` to `{:error, http_error() | term()}`

### Risks / Considerations

- **Stacked on #3950** — _needs a rebase once that PR merges_
- The `inspect/1` in `error_string/1` is required, not stylistic. Without it a multiline ClickHouse response body splits the log entry across lines, and an oversized one writes unbounded — a 9000-byte body goes from 4105 chars to 9010. Both cases are tested
- `Message.failed/2` in `pipeline.ex:485` now carries a tuple rather than a string as the Broadway failure reason. Nothing matches on its shape, and it is only ever inspected
- `retriable?/1` is unaffected: Tesla's `should_retry` sees the raw adapter result, and the tuple is built after retries are exhausted
- Four tests pinned the old string shape and were updated. Two of them (`pipeline_test.exs:1807`, `:1848`) were stubbing the adaptor with a string the real code no longer produces

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
@@ -30,6 +30,8 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
     :tls_alert
   ]
 
+  @type http_error :: {:http, status :: pos_integer(), body :: binary()}
+
   @type error_class ::
           :too_many_parts
           | :http_too_many_requests
@@ -59,7 +61,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
           TypeDetection.event_type(),
           opts :: keyword()
         ) ::
-          :ok | {:error, String.t()}
+          :ok | {:error, http_error() | term()}
   def insert(backend_or_conn_opts, table, log_events, event_type, opts \\ [])
 
   def insert(_backend_or_conn_opts, _table, [], _event_type, _opts), do: :ok
@@ -85,7 +87,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
   end
 
   @spec do_insert(Keyword.t(), String.t(), TypeDetection.event_type(), iodata(), keyword()) ::
-          :ok | {:error, String.t()}
+          :ok | {:error, http_error() | term()}
   defp do_insert(connection_opts, table, event_type, request_body, opts) do
     async? = Keyword.get(opts, :async, false)
     settings = Keyword.delete(opts, :async)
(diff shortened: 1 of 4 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -20,6 +20,31 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
   @pool_timeout 8_000
   @receive_timeout 15_000
   @too_many_parts_marker "Code: 252."
+  @retriable_error_classes [
+    :pool_timeout,
```
djwhitt, line 25: :robot:

> This PR inherits the HTTP/1 pool-checkout timeout issue identified in [the earlier #3950 comment](https://github.com/Logflare/logflare/pull/3950#discussion_r3970527901). The configured ClickHouse pools still raise a `RuntimeError` before Tesla can invoke `retriable?/1` or emit insert-result telemetry.
>
> Carry the fix into this branch when rebasing onto #3950; otherwise this stacked PR remains subject to the same unhandled production path.

lib/logflare/backends/adaptor/clickhouse_adaptor/ingester.ex
```diff
@@ -82,11 +107,50 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.Ingester do
...
+  defp http_error_class(false, 429, _body), do: :http_too_many_requests
+  defp http_error_class(false, status, _body) when status >= 500, do: :http_server_error
+  defp http_error_class(false, status, _body) when status >= 400, do: :http_client_error
+  defp http_error_class(false, _status, _body), do: :http_error
+
+  @doc """
+  Renders an insert failure reason as a log-friendly string.
+  """
+  @spec error_message(term()) :: String.t()
+  def error_message({:http, status, body}) when is_pos_integer(status),
```
djwhitt: :robot:

> **Nit:** This appears to change the logged `error_string` representation. Previously the formatted HTTP error passed through `inspect/1`, which quoted and escaped the value and applied its default printable limit; this branch returns the raw formatted response body.
>
> Is that change intentional? If so, the PR description may need to call it out. If the intent is to preserve the existing log value, consider inspecting or otherwise bounding the formatted string and adding coverage for multiline or large response bodies.

djwhitt (approved):

PR #3953, opened 2026-09-09
feat(repo): configure primary database by URL

djwhitt:
Stacked on #3903. This diff adds `DB_URL` for the primary PostgreSQL database and gives primary and replica URIs the same authentication semantics.

## Configuration

Password authentication:

```text
DB_URL=postgres://logflare:password@database.example.com:5432/logflare?ssl=true&pool_size=10
```

AWS IAM authentication:

```text
DB_URL=postgres://logflare@my-database.cluster-abc.eu-west-1.rds.amazonaws.com:5432/logflare?auth=aws_iam&aws_region=eu-west-1
```

URI values override corresponding individual `DB_*` settings. Settings absent from the URI remain in effect.

## Behavior

- Resolves the primary URL before `Logflare.Repo.init/2` applies authentication, allowing both password and IAM URLs.
- Uses the same Logflare-specific `auth=password|aws_iam` and `aws_region` query parameters for primary and replica URIs, then removes them before the options reach Postgrex.
- Keeps `DB_AUTH` and `DB_AWS_REGION` as defaults when a URL omits those parameters. URL values take precedence.
- Treats a password-bearing URI as explicit password authentication. `auth=password` can use a URI password or inherit `DB_PASSWORD`; `auth=aws_iam` rejects a password in the same URI.
- Preserves Ecto's URL precedence and configured-SSL handling while deriving the socket address family for URL hosts without dropping unrelated socket options.
- Resolves the primary URL before replica pools inherit its omitted settings, while preserving each replica's host and other explicit overrides.
- Redacts URL credentials from parsing and authentication-validation errors.
- Routes the already-supported libcluster Postgrex and Cainophile epgsql clients through the same resolved URL configuration.

## Validation

The combined focused suite passes: **48 tests, 0 failures** across `test/logflare/repo_test.exs`, `test/logflare/repo/aws_iam_test.exs`, and `test/logflare/repo/connection_options_test.exs`.

Formatting and warnings-as-errors compilation pass. `mix lint.all` passes with 32 pre-existing software-design suggestions. Live RDS/Aurora and RDS Proxy authentication remains unvalidated.


lib/logflare/repo/connection_options.ex
```diff
@@ -150,6 +168,173 @@ defmodule Logflare.Repo.ConnectionOptions do
...
+
+          case normalize_url_options(url_options) do
+            {:ok, url_options} ->
+              merge_url_options(config, url_options)
+
+            {:error, reason} ->
+              raise ArgumentError,
+                    "invalid database URL #{inspect(redact_url(url))}: #{reason}"
+          end
+        rescue
```
Baishan: Could have a function level rescue to reduce some of the nesting.

Baishan (approved):

amokan (approved):

PR #3961, opened 2026-09-09
chore: clear the ExSlop DualKeyAccess backlog check

chasers:
Fixes all 15 `ExSlop.Check.Warning.DualKeyAccess` findings, then enables the check by removing it from `ex_slop_backlog` in `config/.credo.exs`. The backlog drops from 10 checks to 9.

The check fires on `||` chains that read one map with both an atom key and a string key. Each site now reads one key type.

## Backend config keeps one shape

`config_encrypted` round-trips through JSON, so it holds string keys after a raw `Repo.get` and atom keys after `typecast_config_string_map_to_atom_map/1`. Both shapes reached `Adaptor.cast_config/2`, so the webhook adaptor read each field twice.

`Backend.validate_config/1` now atomizes the existing config once. This matches the contract `redact_config/1` already had, documented on the `Jason.Encoder` impl. The atomize helper moves out of that impl so both callers share it. The webhook, sentry and postgres adaptors drop their string branches.

## Attrs normalize at the context boundary

`Alerting.create_alert_query/2`, `Alerting.update_alert_query/2`, `Backends.update_backend/2` and `KeyValues.bulk_upsert_key_values/2` take attrs from both API params and internal callers. They stringify the top level keys once through the new `Utils.Map.stringify_top_level_keys/1`, which leaves values untouched. A deep stringify would corrupt struct and payload values. In `update_backend/2` this also collapses the dual `Map.has_key?/2` reads.

## Dead branches

`LogEvent.mapper/3` reads string keys and returns a string-keyed body, so the atom branch in `id/1` never ran. The same holds for the params in `log_event_factory/1`, which stringifies them first.

## Verification

`MIX_ENV=test mix ci` passes. Credo now runs 75 checks, up from 74.

Test files run: backends, alerting, key_values, log_event, logs, utils, the webhook, sentry and postgres adaptors, the backend, rule and key value controllers, and the alerts and rules LiveViews. All pass.

Note for reviewers: `mix ci` fails under `MIX_ENV=dev`, because the reach baseline was written for the test scope and dev adds `priv/tasks`. Run `MIX_ENV=test mix ci` locally, as CI does. That is pre-existing and out of scope here.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -80,7 +80,7 @@ gate to go green - fix the code.
 
 - `mix lint.all` runs `credo --strict` with the **ex_slop** plugin, which adds
   checks for LLM patterns (blanket rescues, narrator comments, anti-idiomatic
-  `Enum`). The `ex_slop_backlog` list in `config/.credo.exs` holds the 10 checks
+  `Enum`). The `ex_slop_backlog` list in `config/.credo.exs` holds the 9 checks
   that have pre-existing findings. They are off so the gate is green on existing
   code. Fixing a backlog check's findings and removing it from that list is a
   welcome change on its own. Never add a check to the list to go green.
--- a/config/.credo.exs
+++ b/config/.credo.exs
@@ -13,7 +13,6 @@ ex_slop_backlog = [
   ExSlop.Check.Refactor.RedundantEnumJoinSeparator,
   ExSlop.Check.Refactor.ReduceMapPut,
   ExSlop.Check.Refactor.WithIdentityElse,
-  ExSlop.Check.Warning.DualKeyAccess,
   ExSlop.Check.Warning.RescueWithoutReraise
 ]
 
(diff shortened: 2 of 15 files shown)


chasers:
Pushed `fa51356de` after a review pass. It fixes two regressions in the first commit.

**1. `Backends.update_backend/2` deleted alert query associations from API calls.** The stringify made `Map.has_key?(attrs, "alert_queries")` true for a raw request body. `Api.BackendController.update/2` passes body params straight through, and the association is `on_replace: :delete`, so `PATCH /api/backends/:token` with `{"alert_queries": []}` deleted every join row for that backend. The function now reads the atom key from the original attrs, which keeps the old behaviour. `test/logflare/backends_test.exs:600` covers it, and the test fails without the fix.

**2. `Backend.validate_config/1` raised on the write path.** The shared atomize helper ran on every changeset, so a stored config key with no live atom raised `ArgumentError` instead of returning a changeset error. It now normalizes the stored config through the adaptor's own `cast_config/1`, the same way `Backends.typecast_config_string_map_to_atom_map/1` already does. No atom creation, and keys no adaptor declares drop out. The `Jason.Encoder` helper goes back to private, so the public function the first commit added is gone.

Verified with `MIX_ENV=test mix ci` and the backends, alerting, key_values, utils, backend controller and adaptor suites. The ClickHouse and syslog suites need containers, so CI covers those.

chasers:

PR #3962, opened 2026-09-09
chore: add connect/disconnect telemetry for ClickHouse read pools

amokan:
## Overview

We have no direct measurement of read pool connection lifecycle. Reconnect storms, recycle behavior, and idle socket death can only be inferred from server-side ClickHouse metrics, which the proxy layer obscures anyway. `DBConnection.TelemetryListener` emits per-connection `[:db_connection, :connected]` and `[:db_connection, :disconnected]` events — the measurement itself rather than an inference from it.

Pairs with the checkout timing metrics already in place (#3896): a disconnect/reconnect rate baseline per pool means a spike at recycle time, or one synchronized across pools, discriminates between the slow-checkout mechanisms directly.

### Key Changes

- `QueryConnectionSup` supervises a named `DBConnection.TelemetryListener` ahead of the pools' `DynamicSupervisor`, so it outlives the pools — which are hard-stopped on config refresh, idle-stopped after 5 minutes, and recycled every ~10 mins
- `build_ch_opts/2` passes `connection_listeners: {[listener], {backend_id, label}}`. The tag is computed at pool start, so dynamically-created labeled pools self-identify
- `logflare.clickhouse.read_pool.connected.count` and `.disconnected.count` tag `backend_id` and `read_cluster`, unpacking the pool tag in `tag_values`. A `keep` guard restricts them to our tag shape, since `[:db_connection, :connected]` is a globally-named event any future `DBConnection` listener would land on
- New `ClickHouseAdaptor.read_cluster_tag/1` maps the legacy unlabeled pool to a stable `"default"`, and is applied to the existing checkout, `query_error`, and `failover` tags as well, so every read pool metric family joins on `read_cluster`
- Two of the new tests run against a real pool end to end, asserting tagged connect and disconnect events rather than just the opts we hand to `Ch.start_link/1`

### Risks / Considerations

- **Dashboard impact.** The legacy pool's `read_cluster` changes from empty to `"(unlabeled)"` on already-shipped read pool metrics.
- **Disconnects are lazy, not instantaneous.** `DBConnection.disconnect_all/2` — what `recycle_pool/1` calls — only stamps the pool; connections drop at their next checkin or idle-scan pass. Recycle-driven disconnects smear across the 60s `recycle_spread` on a busy pool, and an idle pool registers none until traffic touches it. Panel descriptions should say so rather than imply a sawtooth
- The listener is a singleton keyed by name. If it crashes, in-flight monitors are lost and disconnects for existing connections go uncounted until those connections cycle — bounded and self-healing, but it makes the counters lower bounds, not exact
- `Ch.start_link/1` forwards opts untouched to `DBConnection.start_link/2` (_ch 0.5.6, db_connection 2.9.0_), so no adapter change was needed

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor.ex
@@ -47,7 +47,7 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
   @ch_slow_pool_checkout_ms 1_000
   @us_per_hour 3_600 * 1_000_000
   @default_max_event_age_hours 24
-  @default_read_cluster_tag "default"
+  @unlabeled_read_cluster_tag "(unlabeled)"
 
   defdelegate connection_pool_via(arg), to: ConnectionManager
   defdelegate connection_pool_via(arg, label), to: ConnectionManager
@@ -556,11 +556,14 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
 
   @doc """
   Normalizes a read cluster label into a telemetry tag, mapping the legacy
-  unlabeled pool to a stable string.
+  unlabeled pool to `#{@unlabeled_read_cluster_tag}`.
+
+  The value is reserved: `read_only_urls` rejects it as a cluster label, so a
+  labeled pool can never share a tag with the unlabeled one.
   """
   @spec read_cluster_tag(String.t() | nil) :: String.t()
   def read_cluster_tag(label) when is_non_empty_binary(label), do: label
-  def read_cluster_tag(_label), do: @default_read_cluster_tag
+  def read_cluster_tag(_label), do: @unlabeled_read_cluster_tag
 
   @spec default_read_cluster_label(Backend.t()) :: String.t() | nil
   defp default_read_cluster_label(%Backend{config: config}) do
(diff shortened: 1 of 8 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -553,6 +554,14 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
     end
   end
 
+  @doc """
+  Normalizes a read cluster label into a telemetry tag, mapping the legacy
+  unlabeled pool to a stable string.
+  """
+  @spec read_cluster_tag(String.t() | nil) :: String.t()
+  def read_cluster_tag(label) when is_non_empty_binary(label), do: label
+  def read_cluster_tag(_label), do: @default_read_cluster_tag
```
djwhitt: :robot:

> Could we avoid using `"default"` as an unreserved sentinel? `read_only_urls` permits a cluster named `"default"` while leaving `default_read_cluster` unset. In that valid configuration, unlabeled queries use the legacy pool and explicit `"default"` queries use a distinct named pool, but both emit `read_cluster: "default"`, merging their lifecycle, checkout, and error metrics.
>
> One option is adding an explicit discriminator:
>
> ```elixir
> def read_pool_tags(nil),
>   do: %{read_cluster: "default", read_pool_kind: "legacy"}
>
> def read_pool_tags(label),
>   do: %{read_cluster: label, read_pool_kind: "named"}
> ```
>
> Alternatively, namespace the existing tag values:
>
> ```elixir
> def read_cluster_tag(nil), do: "legacy"
> def read_cluster_tag(label), do: "named:" <> label
> ```
>
> If `"default"` is intended to be reserved, config validation should reject that label after checking existing configurations. Please also add a regression test proving the legacy pool and a named `"default"` pool remain distinguishable.

djwhitt (approved):
:robot:

> Approved with one non-blocking telemetry-tagging follow-up.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
amokan: Good call out.

I went with a slightly different variant and use `"(unlabeled)"` in regard to the telemetry tag. But do now ensure no named read cluster can be named that to ensure consistent behavior.

PR #3963, opened 2026-09-09
fix: update clickhouse mapping configuration to ensure `metadata.source` does not get lost

amokan:
The mapping behavior for `log_attributes` (logs), `attributes` (metrics), and `span_attributes` (traces) is configured to elevate keys within `metadata` as well as include all root-level attributes not listed in their respective `exclude_keys` property.

The problem is that a new source coming online with a `metadata.source` attribute would be lost to a root-level `source` property in the existing configuration despite us not leveraging the `$.source` attribute for anything as it already exists as `source_uuid` in ClickHouse.

This PR adds `source` to `exclude_keys` for all three types to ensure the root level `source` attribute is excluded prior to elevating the attributes within `metadata`. Worth noting that `exclude_keys` is applied _before_ `elevate_keys`, which is what makes this work properly.

### DIFF
--- a/lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
+++ b/lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
@@ -24,6 +24,19 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
   This trade-off is deliberate: downstream queries get a stable, predictable key set
   rather than having to reconcile aliases and duplicate keys per row.
 
+  ## Root-level `source` contract
+
+  Every attribute map (`log_attributes`, `attributes`, `span_attributes`) excludes the
+  root-level `source` key. Ingest resolves the target source from `params["source"]`, and
+  `Plug.Parsers` merges the parsed request body into `conn.params`, so a caller may pass
+  the source token in the JSON body rather than the query string. The ingest controller
+  drops only `timestamp` and `id` from that body, so a body-supplied token survives into
+  the event body as a source UUID under `$.source`.
+
+  Because `:exclude_keys` is applied before `:elevate_keys`, dropping the key lets a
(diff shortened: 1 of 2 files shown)


lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
```diff
@@ -40,9 +53,11 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor.MappingDefaults do
...
-  @metric_config_id "00000000-0000-0000-0002-000000000004"
-  @trace_config_id "00000000-0000-0000-0003-000000000005"
+  @log_config_id "00000000-0000-0000-0001-000000000005"
+  @metric_config_id "00000000-0000-0000-0002-000000000005"
+  @trace_config_id "00000000-0000-0000-0003-000000000006"
+
+  @attributes_exclude_keys ["id", "event_message", "source", "timestamp"]
 
   @spec config_id(TypeDetection.event_type()) :: String.t()
   def config_id(:log), do: @log_config_id
```
djwhitt, line 63: :robot:

> `source` is not guaranteed to be the routing UUID. For example, batched events authenticated by an outer `?source=<uuid>` can contain an independent root-level `source` value in each event. `source_uuid` preserves the destination UUID, but not that application value, so this global exclusion silently drops data across logs, metrics, and traces when `metadata.source` is absent.
>
> Remove the routing token at the relevant ingestion boundary, or conditionally discard it only when it matches `source_uuid`. Add a regression test confirming that a legitimate root-level `source` remains mapped.

djwhitt (approved):

lib/logflare/backends/adaptor/clickhouse_adaptor/mapping_defaults.ex
amokan: Agree in general, but probably not worth worrying about now. When we find out we're smashing over some other, unexpected `source` attribute - we'll adjust for that 🤷‍♂️

PR #3965, opened 2026-09-10
feat(helm): allow per-path backends and extra Ingress objects

stigglor:
The chart rendered one Ingress whose every path pointed at the release's own Service on service.port, so nothing could route a path to a preview Service, to a named port, or to a controller annotation, which is how Argo Rollouts splits canary traffic on an ALB. Paths now take optional serviceName and servicePortName, and extraIngresses renders further Ingress objects for routes that cannot share one.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -227,10 +227,31 @@ ingress:
       paths:
         - path: /
           pathType: ImplementationSpecific
+          # Backend for this path. Both empty routes it to the release's own Service on
+          # `service.port`. `serviceName` names another Service in the release's namespace,
+          # `servicePortName` a port on it by name. See README.md#ingress.
+          serviceName: ""
+          servicePortName: ""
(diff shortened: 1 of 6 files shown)


helm/values.yaml
```diff
@@ -227,11 +227,32 @@ ingress:
...
   tls: []
     # - secretName: chart-example-tls
     #   hosts:
     #     - chart-example.local
 
+# Further Ingress objects, each taking the same shape as `ingress` above plus a required
+# `nameSuffix` that keeps its name distinct from `<fullname>`. For routes that cannot share one
+# Ingress, because they belong on different ingress classes or because an annotation applies per
+# Ingress rather than per rule. `ingress.enabled` does not gate these. See README.md#ingress.
+extraIngresses: []
```
Ziinc, line 244: we could make `ingress` an array instead?

PR #3966, opened 2026-09-11
feat: pglogical ddl migrator

Ziinc:
Adds a pglogical ddl migrator that wraps the migration functions and explicitly propagates the ddl statements.

Added a demo script to verify that it is working, tested on local dev.

Part of O11Y-2491

CleanShot 2026-09-11 at 15 47 04@2x

pre-merge checklist:
- [ ] tidy code/docs/scripts to be more concise.

### DIFF
--- a/.github/workflows/pglogical-replica-ci.yml
+++ b/.github/workflows/pglogical-replica-ci.yml
@@ -1,0 +1,107 @@
+name: pglogical Replica CI
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches: ["**"]
+    paths:
+      - ".tool-versions"
+      - "lib/logflare/repo/*"
+      - "lib/logflare/ecto/adapters/pglogical_postgres/**"
+      - "config/runtime.exs"
+      - "config/test.exs"
+      - "docker-compose.pglogical.yml"
+      - "docker/Dockerfile.pglogical"
+      - "test/integration/pglogical_*"
+      - "test/support/pglogical_*"
+      - "test/support/pglogical/**"
+      - ".github/workflows/pglogical-replica-ci.yml"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  pglogical-replica:
+    name: pglogical DDL replication integration tests
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: bitnamilegacy/postgresql:15
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: logflare_test
+          POSTGRESQL_WAL_LEVEL: logical
+        options: >-
+          --health-cmd "pg_isready -d logflare_test -U postgres -p 5432"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 10
+    env:
+      MIX_ENV: test
+      SHELL: /bin/bash
+      PGLOGICAL_PRIMARY_URL: postgresql://postgres:postgres@localhost:5433/logflare_test
+      PGLOGICAL_REPLICA_URL: postgresql://postgres:postgres@localhost:5434/logflare_test
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Read versions from .tool-versions
+        id: v
+        run: |
+          echo "elixir=$(grep '^elixir '  .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
+          echo "erlang=$(grep '^erlang '  .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"
+
+      - name: Set up Elixir
+        uses: erlef/setup-beam@v1
+        with:
+          version-file: .tool-versions
+          version-type: strict
+
+      - name: Restore Mix cache
+        uses: actions/cache@v5
+        with:
+          path: |
+            deps
+            _build
+            priv/native
+          key: mix-v5-${{ runner.os }}-${{ steps.v.outputs.erlang }}-${{ steps.v.outputs.elixir }}-${{ github.job }}-${{ hashFiles('mix.lock', '**/Cargo.lock') }}-${{ hashFiles('mix.exs', 'Cargo.toml', 'config/**', 'lib/**', 'native/**', 'priv/**', 'test/support/**') }}
+          restore-keys: |
+            mix-v5-${{ runner.os }}-${{ steps.v.outputs.erlang }}-${{ steps.v.outputs.elixir }}-${{ github.job }}-${{ hashFiles('mix.lock', '**/Cargo.lock') }}-
+            mix-v5-${{ runner.os }}-${{ steps.v.outputs.erlang }}-${{ steps.v.outputs.elixir }}-${{ github.job }}-
+
+      - name: Install dependencies
+        run: mix do deps.get + deps.compile
+
+      - name: Build and start pglogical primary/replica containers
+        run: docker compose -f docker-compose.pglogical.yml up -d --build
+
+      - name: Wait for pglogical containers to be healthy
+        run: |
+          for port in 5433 5434; do
+            until docker run --rm --network host postgres:15 pg_isready -h localhost -p "$port" -U postgres; do
+              echo "Waiting for Postgres on port $port..."
+              sleep 2
+            done
+          done
+
+      - name: Bootstrap pglogical extension, nodes, replication set, subscription
+        run: mix run test/support/pglogical/bootstrap.exs
+
+      - name: Create and migrate the primary application database
+        run: epmd -daemon; mix do ecto.create --quiet + ecto.migrate --quiet
+
+      - name: Run pglogical DDL replication integration tests
+        run: mix test --only pglogical_replica
+
+      - name: Tear down pglogical containers
+        if: always()
+        run: docker compose -f docker-compose.pglogical.yml down -v
--- a/lib/logflare/ecto/adapters/pglogical_postgres/connection.ex
+++ b/lib/logflare/ecto/adapters/pglogical_postgres/connection.ex
@@ -1,0 +1,74 @@
+defmodule Logflare.Ecto.Adapters.PglogicalPostgres.Connection do
+  @moduledoc """
+  Delegates to `Ecto.Adapters.Postgres.Connection`, wrapping DDL with
+  `pglogical.replicate_ddl_command/2` for replication.
+
+  Replication-set names are validated in `config/runtime.exs` because DDL
+  cannot use bind parameters.
+  """
+  @behaviour Ecto.Adapters.SQL.Connection
+
+  alias Ecto.Adapters.Postgres.Connection, as: PG
+  alias Logflare.Repo.Migrator
+
+  @impl true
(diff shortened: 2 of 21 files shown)


amokan (commented):
🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> Three blocking findings, posted as inline comments: raw `execute/1` SQL being wrapped in `replicate_ddl_command`, `rollback/2` not routed through `Logflare.Repo.Migrator`, and no CI job exercising `priv/repo/migrations/*` through the wrapper (which is what keeps the first one green).

lib/logflare/ecto/adapters/pglogical_postgres/connection.ex
```diff
@@ -0,0 +1,74 @@
...
+  defdelegate table_exists_query(table), to: PG
+
+  @impl true
+  def execute_ddl(command) do
+    replication_sets = Migrator.replication_sets()
+
+    command
+    |> PG.execute_ddl()
+    |> List.wrap()
+    |> Enum.map(&wrap_in_pglogical(&1, replication_sets))
```
amokan, line 54: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Blocking: raw `execute/1` SQL is wrapped here, not just schema DDL.**
>
> Line 52 calls `PG.execute_ddl/1`, which falls through to Ecto's binary catch-all at `deps/ecto_sql/lib/ecto/adapters/postgres/connection.ex:1408` (`execute_ddl(string) when is_binary(string), do: [string]`), so every `Ecto.Migration.execute/1` body reaches `wrap_in_pglogical/2` on this line. What that breaks in migrations already in the repo:
>
> - `priv/repo/migrations/20210729161959_subscribe_to_postgres.exs:34` — `ALTER SYSTEM SET wal_level = 'logical'` (dev/test only, gated at line 33). Inside `replicate_ddl_command`'s function body, Postgres's `PreventInTransactionBlock` guard rejects it, so `mix ecto.migrate` from zero fails outright. That migration sets `@disable_ddl_transaction true` at line 18 to satisfy that same guard. Line 53 (`SELECT pg_drop_replication_slot(...)`, on `down`) hits the same restriction.
> - `20210729161959:42`, `20240708025030:18` and `:21`, `20241204120824:10`, `20251216234322:11`, `20260209080416:11` — `CREATE`/`ALTER PUBLICATION logflare_pub` (name from `config/config.exs:160`) is executed on the *subscriber*. A subscriber without `logflare_pub` fails with `publication "logflare_pub" does not exist`, and pglogical's apply worker retries that transaction indefinitely, so no later DDL arrives either.
> - `20251205181853:5` (`DELETE FROM sources ...`), `20260610224521:8` and `:14` (`UPDATE versions ...`), `20250224140820:5` and `:16` — migration DML is shipped as text and runs on the subscriber; if those tables are in a replication set the row-level changes arrive too and it applies twice.
> - Any future `create index(..., concurrently: true)` — same function-body restriction as `ALTER SYSTEM`, and it would fail only where replication sets are configured. Nothing uses `concurrently` today.
>
> **Fix** — two parts, both needed. First, stop wrapping raw SQL:
>
> ```elixir
> @impl true
> def execute_ddl(command) when is_binary(command), do: PG.execute_ddl(command)
>
> @impl true
> def execute_ddl(command) do
>   # existing body
> end
> ```
>
> That alone silently stops replicating raw SQL that *should* reach the replica: `20190220150309:5` and `20190226225345:5` (`ALTER TABLE ... DROP CONSTRAINT`), `20210810182003:5` and `20220803211705:4` (`... REPLICA IDENTITY FULL`). So second, add an explicit opt-in for those — a `replicated_execute/1` migration helper, or a `@replicate_ddl true` attribute the adapter checks — and update the note at `lib/logflare/ecto/adapters/pglogical_postgres.ex:11-14`, which currently only tells authors to schema-qualify raw SQL, to state that raw SQL isn't replicated unless opted in.
>
> The alternative is keyword-sniffing the statement text for a denylist (`ALTER SYSTEM`, `CONCURRENTLY`, `INSERT`/`UPDATE`/`DELETE`, publication DDL). It replicates more DDL without author effort, but the list needs maintaining forever and gets the default wrong: anything unrecognized still ships to the subscriber.
>
> Not executed against a live pglogical instance — the two function-body failures follow from `PreventInTransactionBlock`, the publication and DML cases from `replicate_ddl_command` executing the command text on the subscriber.

lib/logflare/release.ex
```diff
@@ -7,7 +7,10 @@ defmodule Logflare.Release do
     Application.ensure_all_started(:ssl)
 
     for repo <- repos() do
-      {:ok, _, _} = Ecto.Migrator.with_repo(repo, &Ecto.Migrator.run(&1, :up, all: true))
+      migration_repo = Logflare.Repo.Migrator.migration_repo_for(repo)
+
+      {:ok, _, _} =
+        Ecto.Migrator.with_repo(migration_repo, &Ecto.Migrator.run(&1, :up, all: true))
```
amokan, line 13: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Blocking: `rollback/2` bypasses this routing.**
>
> `migrate/0` now resolves through `Logflare.Repo.Migrator.migration_repo_for/1`, but `rollback/2` at lines 19-21 still passes `repo` straight to `Ecto.Migrator.with_repo/2`. A down-migration therefore drops the object on the primary only; the replica keeps it, and the next up-migration's replicated DDL fails on the subscriber and stalls the apply worker. `mix ecto.rollback` is unrouted too — the only aliases calling `&migrate/1`/`&migrate_quiet/1` are `mix.exs:276`, `:290`, and `:303`.
>
> **Fix:**
>
> ```elixir
> def rollback(repo, version) do
>   migration_repo = Logflare.Repo.Migrator.migration_repo_for(repo)
>
>   {:ok, _, _} =
>     Ecto.Migrator.with_repo(migration_repo, &Ecto.Migrator.run(&1, :down, to: version))
> end
> ```
>
> Add `"ecto.rollback": [&rollback/1]` in `mix.exs` if dev rollback should route as well. The alternative — raising from `rollback/2` when `Migrator.replication_sets() != []` and documenting rollback as unsupported under pglogical — is defensible if you never intend to roll back a replicated schema, but it's more code than routing it.

.github/workflows/pglogical-replica-ci.yml
```diff
@@ -0,0 +1,107 @@
...
+              echo "Waiting for Postgres on port $port..."
+              sleep 2
+            done
+          done
+
+      - name: Bootstrap pglogical extension, nodes, replication set, subscription
+        run: mix run test/support/pglogical/bootstrap.exs
+
+      - name: Create and migrate the primary application database
+        run: epmd -daemon; mix do ecto.create --quiet + ecto.migrate --quiet
```
amokan, line 100: 🤖 AI-generated review feedback (Claude Code), reviewed by @amokan

> **Blocking: no CI job runs the real migration suite through the wrapper.**
>
> This step targets the 5432 service DB, and the job's `env:` (lines 49-53) never sets `LOGFLARE_PGLOGICAL_REPLICATE_DDL_COMMANDS_SETS`. `config/test.exs:43` pins `replication_sets: []`, and the integration test overrides app env at runtime (`test/integration/pglogical_ddl_replication_test.exs:59`) to run only the two toy migrations in `test/support/pglogical_test_migration.ex` and `_alter.ex`. So `priv/repo/migrations/*` never passes through `pglogical.replicate_ddl_command`, which is why the `ALTER SYSTEM` and `ALTER PUBLICATION` failures in `lib/logflare/ecto/adapters/pglogical_postgres/connection.ex` are green here.
>
> **Fix:** add a step that migrates the pglogical primary (5433) from zero with sets enabled:
>
> ```yaml
> - name: Migrate pglogical primary with DDL replication enabled
>   env:
>     LOGFLARE_PGLOGICAL_REPLICATE_DDL_COMMANDS_SETS: my_set
>     POSTGRES_PORT: 5433   # whatever knob points Logflare.Repo at the primary
>   run: mix ecto.migrate
> ```
>
> `config/test.exs:36-37` hardcodes `localhost:5432`, so overriding host/port (env var, or reading `PGLOGICAL_PRIMARY_URL` in `config/test.exs`) is part of this change. An ExUnit test driving `Ecto.Migrator` over the real migrations path would allow richer assertions, but it's slower and duplicates what `mix ecto.migrate` already does.

djwhitt:
@claude Please review this.

.github/workflows/pglogical-replica-ci.yml
```diff
@@ -0,0 +1,117 @@
+name: pglogical Replica CI
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches: ["**"]
+    paths:
```
djwhitt, line 10: :robot:

> **Blocking:** Include `priv/repo/migrations/**` in this path filter. Migration-only PRs are where incompatible or incorrectly classified DDL will usually enter, but they currently skip this integration job; the first run happens only after merge on `main`.
>
> `mix.exs` and `lib/logflare/ecto/adapters/pglogical_postgres.ex` should also trigger this workflow because they control migration routing and wrapping.

mix.exs
```diff
@@ -304,11 +304,28 @@ defmodule Logflare.Mixfile do
...
+      "ecto.rollback": [&rollback/1]
     ]
   end
 
+  # Routes `mix ecto.migrate`/`mix ecto.rollback` through `Logflare.Repo.Migrator`,
+  # so migrations honor LOGFLARE_PGLOGICAL_REPLICATE_DDL_COMMANDS_SETS in dev/test too.
+  defp migrate(args) do
+    Mix.Task.run("app.config")
+    repo = Logflare.Repo.Migrator.migration_repo_for(Logflare.Repo) |> inspect()
+    Mix.Task.run("ecto.migrate", ["-r", repo | args])
```
djwhitt, line 318: :robot:

> **Blocking:** Preserve explicit `-r`/`--repo` arguments instead of unconditionally prepending another repository. `Mix.Ecto.parse_repo/1` collects every occurrence without deduplicating them.
>
> For example, `mix ecto.rollback -r Logflare.Repo` becomes two entries for the same database and can roll back one migration for each entry. Only inject the selected repository when no explicit repository was supplied, or deliberately replace the existing option.

test/integration/pglogical_ddl_replication_test.exs
```diff
@@ -0,0 +1,191 @@
...
+
+  setup %{primary_conn: primary_conn, replica_conn: replica_conn} do
+    initial_migrator_env = Application.get_env(:logflare, Migrator)
+
+    on_exit(fn ->
+      Application.put_env(:logflare, Migrator, initial_migrator_env)
+      Enum.each([Logflare.Repo.Pglogical, PglogicalPrimaryRepo], &stop_repo/1)
+    end)
+
+    reset_public_schema!(primary_conn)
```
djwhitt, line 52: :robot:

> The CI job runs the real migrations on the pglogical primary, but these tests reset both schemas before checking the subscriber. Could we wait for and assert a representative table or column from the real migration suite on the subscriber first? The toy migrations test the replication mechanism, but not the resulting application schema.

.github/workflows/pglogical-replica-ci.yml
djwhitt: :robot:

> This still appears unresolved in the current head: `priv/repo/migrations/**`, `mix.exs`, and `lib/logflare/ecto/adapters/pglogical_postgres.ex` are absent from this workflow’s PR path filter. A migration-only PR therefore skips the pglogical job. Please add these paths.

djwhitt (approved):

PR #3970, opened 2026-09-12
fix: show the top "Load more" button whenever older events can load

chasers:
Stacked on #3969. Merge that first; this PR's base is `fix/search-scroll-to-bottom`.

## Problem

The top **Load more** button showed only when the initial search returned a 101st "sentinel" row inside the current time range (`search_query_executor.ex:274-278`). Once shown, a click pages **without** the time-range filter (`search_operations.ex:352-359`) and extends the URL range — so the button could reach older events outside the range, but only if the range happened to hold more than one page. Visibility followed one rule, behaviour another, and the button came and went with the width of the range.

## Change

- The previous cursor is the **oldest displayed row**, for the initial page and for every previous page. It is `nil` only when the page has no rows, or when a previous page comes back without a sentinel, which means nothing older exists. `EventPagination` is unchanged: a `nil` cursor still hides the button.
- `cursor_condition(:previous, ...)` is strict (`<`, and `id <` at equal timestamps) so the cursor row is not fetched twice. The sentinel-based cursor needed the inclusive form.

## Tests

New LiveView test: a range with 99 events shows the top button; a click loads the three older events from outside the range, patches the URL range, and then hides the button. The existing pagination tests keep passing unchanged.

- `logs_search_lv_test.exs` — 90 tests, 0 failures
- `search_operations_test.exs`, `event_pagination_test.exs` — 36 tests, 0 failures
- `MIX_ENV=test mix ci` — exit 0


chasers:
Second commit: the **bottom** button had the mirror-image bug.

`EventPagination.complete_initial/2` set `next_exhausted?` from the initial page's `has_more?`. The initial page is fetched newest-first, so its sentinel is an *older* event and says nothing about newer ones. With 100 or fewer events in the range the bottom button was hidden from the start, even when the range ended in the past and newer events existed.

The initial page now leaves `next_exhausted?` at `false`. Availability comes from the range end alone (`next_page_available?/4`), and a next page that comes back short still hides the button. `complete_page(:next)` is unchanged — its sentinel really is a newer event.

New test: a 99-event range that ends in the past shows the bottom button; a click loads the newer events, patches the URL range, then hides the button.

Verified: `event_pagination_test.exs` + `logs_search_lv_test.exs` 92/0, `MIX_ENV=test mix ci` exit 0.

stigglor (approved):

Ziinc (approved):

PR #3973, opened 2026-09-12
perf(logs): transform bodies before constructing LogEvent

djwhitt:
## Summary

- Run copy-field, key-value enrichment, and drop-field transforms directly against the body map.
- Construct the final `LogEvent` once instead of allocating a new struct and `{:ok, value}` tuple at every stage.
- Return immediately when all transform configurations are absent or already parsed as empty.
- Specialize common binary-key path lookups while retaining `get_in/2` as a fallback.
- Reuse parent maps when a requested nested drop makes no change.
- Add a reproducible end-to-end `LogEvent.make/2` benchmark.

This is a standalone extraction of #3767 onto current `main`. It does not include #3765's BigQuery payload cleaning or the timestamp changes from #3768 and #3769; the existing stack has not been rewritten.

## Benchmark

Command, run unchanged on the benchmark-only baseline commit `64060f7e` and candidate `94d3b95c`:

```bash
MIX_ENV=dev \
  LF_BENCH_TIME=4 \
  LF_BENCH_WARMUP=2 \
  LF_BENCH_MEMORY_TIME=3 \
  ../bin/x mix run --no-start bench/log_event_body_transforms.exs
```

The table reports the median throughput from three alternating baseline/candidate runs. Memory measurements were identical across each revision's three runs.

| Scenario | Baseline | Candidate | Throughput | Memory |
|---|---:|---:|---:|---:|
| No transforms | 620.59 K ips | 642.91 K ips | **+3.6%** | 4.88 KB → 4.81 KB (**-1.4%**) |
| Three copies and three drops | 466.18 K ips | 474.82 K ips | **+1.9%** | 5.93 KB → 5.41 KB (**-8.8%**) |

Per-run throughput:

- No transforms: baseline 609.96/620.59/629.12 K ips; candidate 642.91/649.93/631.16 K ips.
- Three copies and three drops: baseline 469.32/462.06/466.18 K ips; candidate 473.84/480.41/474.82 K ips.

Environment: Linux, 6 available cores, Elixir 1.19.5, Erlang/OTP 27.3.4.6, JIT enabled. The benchmark exercises full `LogEvent.make/2` calls rather than private helpers. It covers the no-transform path and parsed copy/drop transforms; it intentionally excludes external key-value cache lookups.

## Validation

- `../bin/test test/logflare/log_event_test.exs` — 6 properties, 66 tests, 0 failures
- `MIX_ENV=test ../bin/x mix test.compile`


djwhitt:
Closing. Not worth it.

PR #3976, opened 2026-09-14
fix: keep the "Load more" spinner up until the page query returns

chasers:
**Stack 1/7** — "Load more" pagination fixes.

- **Problem:** the spinner stops in milliseconds. The page query still runs for seconds.
- **Cause:** `phx-click-loading` covers only the event round trip. The query runs async.
- **Fix:** `EventPagination` gets a `:loading` button state.
- `load_events` sets it. The page result and the page error clear it.
- The button stays disabled while it loads. A second click does nothing.


chasers:
Third commit: **scroll to the bottom when a search runs.**

A chart click already scrolled to the newest results via `scroll_to_bottom_on_result?`, but running a search did not, so submitting a query left the viewport wherever it was. `start_search` now sets the same flag.

The search button and the Enter key both route through a single event — `lql_editor_wrapper_hook.js` binds `monaco.KeyCode.Enter` (line 90) and the submit listener (line 55) to the same `submitSearch/0`, which pushes `start_search` — so one change covers both.

Verified: `mix test test/logflare_web/live/search_live/` 119/0, `MIX_ENV=test mix ci` exit 0.

chasers:
Fifth commit — and this one is almost certainly the staging bug.

## The scan had no far side

A page request drops the query's timestamp range so it can page past it, and kept only a **one-sided** partition bound:

```sql
WHERE EXTRACT(DATE FROM t0.timestamp) <= '2026-09-14'
  AND (t0.timestamp < TIMESTAMP_MICROS(?) OR ...)
```

Nothing bounds the other end, so on BigQuery that reads every partition the source has ever had. On a small table it returns; on a production table it does not.

## A page now moves by the window you are looking at

The scan is `[cursor - window, cursor]` going back and `[cursor, cursor + window]` going forward, where `window` is the width of the range currently in view (floor: 60 seconds). The query carries a `TIMESTAMP_MICROS` bound **and** a partition-date bound on both sides, for `:timestamp`, `:pseudo` and Postgres alike.

Two consequences, both wanted:

- **The LQL range grows by that same window on every click**, instead of only when the returned rows happened to fall outside it. That is why the range looked frozen even when rows loaded.
- **An empty window no longer stalls the button.** The cursor shifts by the window rather than clearing, so the next click keeps walking back. `extend_timestamp_range/2` is gone.

## The buttons say what they do

`Load more (-10 minutes)` on top, `Load more (+10 minutes)` on the bottom, humanized from the same window the query uses — seconds, minutes, hours or days, singular or plural.

## Verification

`mix test test/logflare/logs/ test/logflare_web/live/search_live/` — 263 tests, 0 failures, including new SQL-level tests asserting both bounds and the window arithmetic. `MIX_ENV=test mix ci` — exit 0.

amokan (approved):

chasers:

PR #3977, opened 2026-09-14
fix: log a page request that the LiveView drops

chasers:
**Stack 2/7** — "Load more" pagination fixes.

- **Problem:** a dropped page request looks the same as an empty page.
- **Fix:** log a warning when the LiveView drops `load_events`.
- The log shows the intent, the cursor, and the guard state.


amokan (approved):

chasers:

PR #3979, opened 2026-09-14
fix: bound a page request to the window in view

chasers:
**Stack 4/7** — "Load more" pagination fixes.

- **Problem:** a page request scans to the end of the table. On BigQuery that is every partition.
- **Fix:** a page moves by the width of the range in view. The minimum is 60 seconds.
- The query bounds both sides by timestamp and by partition date.
- An empty page moves the cursor by one window. A quiet stretch no longer stalls the button.
- The buttons show the window: "Load more (-10 minutes)".
- The label no longer clips on a long window.


amokan (approved):

chasers:

PR #3980, opened 2026-09-14
fix: let the LiveView drive every scroll

chasers:
**Stack 5/7** — "Load more" pagination fixes.

- **Problem:** a search does not scroll to the newest results. A fresh page load does not scroll either.
- **Cause:** follow-up diffs restore the old scroll anchor and undo the scroll.
- **Fix:** the LiveView pushes `scroll-to-bottom` for every initial page and every tail append.
- The hook holds the scroll until a frame runs it. It skips the anchor restore until then.
- `data-tailing` and `scrollToLatest/0` are gone.


amokan (approved):

chasers:

PR #3981, opened 2026-09-14
fix: write an implied timestamp range into the query on a page request

chasers:
**Stack 6/7** — "Load more" pagination fixes.

- **Problem:** a query with no `t:` filter pages with a fixed 60 second window.
- **Fix:** use the range the chart draws. `c:group_by(t::minute)` means the last 120 minutes.
- The first page request writes that range into the query. The URL then shows what is on screen.


amokan (approved):

chasers:

PR #3982, opened 2026-09-14
fix: stop hiding the load more buttons on a short page

chasers:
**Stack 7/7** — "Load more" pagination fixes.

- **Problem:** a short page hides the "Load more" buttons.
- **Cause:** the rules still read a short page as "no more data". Now it only means a quiet window.
- **Fix:** take the cursor from the returned rows, not from the sentinel row.
- Remove `next_exhausted?` and the helpers that kept it.
- A button now hides in two cases only: the view tails, or it has no cursor.


amokan (approved):

chasers:

PR #3984, opened 2026-09-14
fix: address the review of the load more stack

chasers:
**Stack 8/8** — fixes from a code review of PRs 1–7.

- **Window doubles:** each click moved twice as far as the last one. One window per search now drives the label, the query, the cursor, and the range.
- **Timezone:** an implied range was UTC but was read as local time. The range now uses the search timezone.
- **Open `t:` bound:** a page request replaced `t:>X` with the last 2 hours. It now keeps `X`.
- **Future cursor:** an empty "next" page moved the cursor past now. The cursor now stops at the request time.
- **Stale result:** an old page result could change a new search. The LiveView now drops it.
- **Stuck spinner:** a parse error left "Loading" on. The error path now clears it.
- **Other button:** a click on it killed the running page query. It is now disabled while a page loads.
- **Cleanup:** removes the unused sentinel row, dead clauses, a duplicate alias, and inline comments.
- **Not fixed:** an empty first page still hides both buttons. That needs a design choice.


amokan (approved):

chasers:

PR #3985, opened 2026-09-14
fix: remove the 750ms LiveView latency simulator in dev

chasers:
**Stack 9/9** — "Load more" pagination fixes.

- **Problem:** dev holds back every LiveView message by 750 ms. The "Load more" spinner and every click feel slow.
- **Cause:** the dev root layout calls `liveSocket.enableLatencySim(750)` on each page load.
- **Fix:** remove that script. Same change as #3901.
- **After merge:** the value stays in `sessionStorage`. Run `liveSocket.disableLatencySim()` once in the browser console, or open a new tab.


chasers:
Opened by mistake: this reused the head branch of #3901. The latency simulator removal goes into the stack on its own branch instead.

PR #4002, opened 2026-09-17
feat: poll DBConnection ready count and queue depth for ClickHouse read pools

amokan:
## Overview

The ClickHouse read pool metrics we have are currently all edge-triggered: checkout timing, connect/disconnect counts (#3962), shed checkouts (#3996). None of these show the state of the pool itself.

This PR samples every 30s alongside the existing periodic measurements in an attempt to answer the questions "_during a slow-checkout episode, did ready connections hit zero, how deep did the queue get, and did recovery come by drain or by reconnect?_"

### Key Changes

- `QueryConnectionSup.list_read_pools/0` selects the `CHReadPool` entries out of `BackendRegistry` and returns `{backend_id, label, pool_pid}` for the pools actually running. Managers whose pool idle-stopped are not listed, so no `Backend` lookup per sweep
- `Telemetry.clickhouse_read_pool_metrics/2` joins the existing `telemetry_poller` sweep. `get_connection_metrics/1` is an `:infinity` call, so each pool is polled in its own task via `Task.async_stream` with `on_timeout: :kill_task`, bounded by `@ch_read_pool_poll_timeout` (_5s_). A pool that exited between enumeration and poll is skipped silently; one that does not answer in time is skipped with a warning and a `poll_timeout` count for the backend and read cluster. Any other result shape is logged and skipped, since `telemetry_poller` permanently drops a measurement that raises
- `logflare.clickhouse.read_pool.ready_conn_count` and `.checkout_queue_length` declared as `last_value` gauges tagged `backend_id` and `read_cluster`, joining the other read pool families on the same tags. `logflare.clickhouse.read_pool.poll_timeout` declared as a `sum` on the same tags, so a missing gauge sample can be attributed to a stall rather than an idle-stop
- Tests cover enumeration per label and after refresh, the gauge and counter definitions, one event per active pool, the empty case, the dead-pid race, the timeout path with its `poll_timeout` count, and the unexpected-result fallback

### Risks / Considerations

- **Read the two gauges together, and alongside `connected/disconnected`**. `DBConnection` reports `ready_conn_count` as 0 whenever no connection is immediately available, and `checkout_queue_length` as 0 whenever it has a ready connection. Both zero means no connection is available and no caller is queued. That is the steady state of a saturated pool, but it is also what a pool reports _before_ its first connection is established and while every connection is reconnecting. `connected`/`disconnected` moving at the same time points to connectivity loss rather than saturation. The metric descriptions spell this out
- **Per node, not per cluster.** Every node runs its own pool for a backend, so these are per-instance gauges. Panel by instance, or take a max across nodes; a sum mixes independent pools and means nothing
- **Stale series clean themselves up.** `OtelMetricExporter` buckets `last_value` per export generation, so a pool that idle-stops simply stops appearing rather than reporting its last value forever
- **A missing sample has two causes.** A pool that idle-stopped and a pool that did not answer the poll both leave a gap in `ready_conn_count`/`checkout_queue_length`. The `poll_failure` counter separates them: it increments only when the pool process was alive but the poll produced no sample, and reason says whether it timed out or answered with an unexpected shape. A gap with no matching `poll_failure` is an idle-stop, and a gap with one is worth looking at
- **The timeout bounds the poller, it does not isolate it.** `telemetry_poller` runs every measurement sequentially in one process, so a stuck pool delays the whole sweep by up to 5s per batch of pools (batch size is `System.schedulers_online/0`) and pushes the next tick of every other periodic measurement by the same amount. This is a bounded, one-off delay per stuck pool per tick rather than a hang. If that becomes a problem the sweep can be handed to `Logflare.TaskSupervisors` so the poller returns immediately


lib/telemetry.ex
```diff
@@ -419,6 +425,27 @@ defmodule Logflare.Telemetry do
...
       ),
+      last_value("logflare.clickhouse.read_pool.ready_conn_count",
+        event_name: @ch_read_pool_status_event,
+        measurement: :ready_conn_count,
+        tags: [:backend_id, :read_cluster],
+        description:
+          "Idle, checked-in connections in a ClickHouse read pool on this node, sampled from `DBConnection.get_connection_metrics/1` every #{div(@metrics_interval, 1000)}s. Reads 0 whenever the pool is busy (every connection checked out), so 0 here with 0 `checkout_queue_length` means fully utilized with nobody waiting. A missing sample means the pool was idle-stopped or the poll timed out; `poll_timeout` tells the two apart. Pools are per node; view per instance"
```
djwhitt: :robot:

> `ready_conn_count == 0` with `checkout_queue_length == 0` does not necessarily mean full utilization. `DBConnection.ConnectionPool` also reports this state before any connection is established and while all connections are reconnecting. Please describe it as “no connection is immediately available and no caller is queued,” or add a metric that distinguishes checked-out from connecting/disconnected connections. Otherwise this guidance can misclassify connectivity loss as normal utilization.

lib/telemetry.ex
```diff
@@ -714,12 +741,83 @@ defmodule Logflare.Telemetry do
...
+    :telemetry.execute(@ch_read_pool_poll_timeout_event, %{count: 1}, metadata)
+
+    Logger.warning("ClickHouse read pool did not answer the metrics poll",
+      backend_id: backend_id,
+      read_cluster: metadata.read_cluster,
+      reason: inspect(reason)
+    )
+  end
+
+  defp emit_read_pool_status(other) do
```
djwhitt: :robot:

> This branch suppresses both gauges without emitting `poll_timeout`, so an active pool returning an unexpected shape appears identical to an idle-stopped pool. That contradicts the documented missing-sample distinction. Please emit a bounded poll-failure metric, such as `reason: :unexpected_result`, or weaken the documentation to state that `poll_timeout` identifies only one possible failure mode. Keep the inspected result in logs rather than metric tags.

djwhitt (approved):
:robot:

> Approved. Two non-blocking observability comments identify metric interpretations to tighten before deployment.

lib/telemetry.ex
amokan: changed

lib/telemetry.ex
amokan: I went with a change to the metric name (now `poll_failure`) which now includes the reason `:unexpected_result`.

PR #4004, opened 2026-09-17
feat: replica-aware routing tag for ClickHouse endpoint queries

amokan:
## Overview

ClickHouse Cloud's [replica-aware routing](https://clickhouse.com/docs/products/cloud/features/infrastructure/replica-aware-routing) consistent-hashes the `X-ClickHouse-Replica-Tag` request header onto a replica, so requests carrying the same tag land on the same node and hit a warm cache. Our endpoint queries often carry a project identifier but today spread across all replicas, so per-project cache locality is lost.

This PR lets a ClickHouse backend optionally name one endpoint request parameter whose value is sent as the replica tag.

### Key Changes

- New optional ClickHouse backend config field `replica_routing_param`, the name of the endpoint request parameter to hash on. A blank/empty value disables the feature. Cast, display-allowlisted, and exposed on the backend form and `ClickhouseConfigSchema` alongside the other optional fields
- `validate_config/1` restricts the value to `\w+`, matching what the `@param` parser accepts, so a stray space or a leading `@` is a form error rather than a silently inert setting. Whitespace-only input casts to `nil`
- On the endpoint query path (`{query, declared_params, input_params}` and the `max_limit` variant), `execute_query_with_params/6` looks the configured param up in the request params and adds `x-clickhouse-replica-tag` to the Ch request headers. Raw string and Ecto query paths never set it
- `execute_ch_query/4` now forwards `:headers` from its opts into `do_ch_query_on_label/5`, and the default-cluster failover retry carries the same headers, so a retried query keeps its tag
- Tests cover the header on both endpoint paths, its absence when the field is unset or the param is missing, non-string / oversized / non-printable values, the raw-string path, the failover retry, and config cast and validation. All header tests go through `Mimic.call_original` so the header is actually sent to the local ClickHouse

### Risks / Considerations

- **The header is a best-effort hint, dropped silently** A missing param, a non-string value, a value over 256 bytes, or one containing whitespace or control characters results in an untagged query with no log. _This is deliberate: the feature is an optimization and the check runs on every endpoint query_
- **No-op where routing is unavailable.** ClickHouse ignores the header on services without replica-aware routing enabled and on self-hosted installs, so enabling the field is safe anywhere. It is an Enterprise-tier toggle in ClickHouse Cloud
- **Scaling remaps the ring.** Changing the replica count of a read service redistributes tag-to-replica assignments, so locality resets after a resize. Not workload isolation: other traffic still shares the replica
- **Applies to endpoint-based queries only**


lib/logflare/backends/adaptor/clickhouse_adaptor.ex
```diff
@@ -48,6 +49,10 @@ defmodule Logflare.Backends.Adaptor.ClickHouseAdaptor do
   @us_per_hour 3_600 * 1_000_000
   @default_max_event_age_hours 24
   @unlabeled_read_cluster_tag "(unlabeled)"
+  @replica_tag_header "x-clickhouse-replica-tag"
+  @replica_tag_max_bytes 256
+  @replica_tag_value_pattern ~r/\A[[:graph:]]+\z/
```
djwhitt: :robot:

> `[[:graph:]]` accepts values that Mint cannot encode as HTTP header values. For example, `"café"` passes this regex, but Mint rejects it with `{:invalid_header_value, "x-clickhouse-replica-tag", "café"}`. The adaptor then returns a `:connection_error`, so this best-effort optimization can break an otherwise valid endpoint query instead of omitting the tag.
>
> Please restrict the value to Mint-safe ASCII, such as `~r/\A[\x21-\x7E]+\z/`, or deterministically encode/hash it. Add regression coverage for non-ASCII values and the 256/257-byte boundary.

djwhitt (approved):
:robot:

> Approved. The inline replica-tag validation issue is non-blocking for the expected ASCII identifier inputs.

lib/logflare/backends/adaptor/clickhouse_adaptor.ex
amokan: Fixed in `9b5fd7a`

PR #4006, opened 2026-09-17
fix: stop log-search aggregates crashing on an invalid chart range

chasers:
## Problem

`Logflare.Logs.SearchOperations.intersperse_missing_range_timestamps/4` pipes the result of `Timex.Interval.new/1` straight into `Enum.to_list/1`. `Interval.new/1` returns `{:error, :invalid_until}` when `until <= from` ([timex `interval.ex:168-172`](https://github.com/bitwalker/timex/blob/main/lib/interval/interval.ex)), so `Enum.to_list/1` raises. The aggregates `Task` dies and the user's chart panel returns nothing.

```
** (Protocol.UndefinedError) protocol Enumerable not implemented for Tuple
Got value:  {:error, :invalid_until}
    (elixir 1.19.5) lib/enum.ex:3824: Enum.to_list/1
    (logflare 1.50.12) lib/logflare/logs/search_operations.ex:856:
        Logflare.Logs.SearchOperations.intersperse_missing_range_timestamps/4
    (logflare 1.50.12) lib/logflare/logs/search_operations.ex:826:
        Logflare.Logs.SearchOperations.add_missing_agg_timestamps/1
    (logflare 1.50.12) lib/logflare/logs/logs_search.ex:68: ...search_result_aggregates/1
Function: #Function<1.90654086/0 in Logflare.Logs.SearchQueryExecutor.start_aggs_task/2>
```

## Production impact

**1,933 crashes on 2026-09-17 in a 75-minute window, 11:29–12:43 UTC**, on build 1.50.12 — first on `prod-d` (874), then on `prod-g` (1,059). Every one carried `{:error, :invalid_until}`. Earlier: 38 occurrences across 09-11..09-14 on 1.50.11, and zero on most days. The bug is latent and old; a burst of searches with a degenerate time range turned it into an incident.

## Two ways to reach `until <= from`

`from = DateTimeUtils.truncate(min, chart_period)`, `until = DateTimeUtils.truncate(max, chart_period)`.

1. **Truncation collapse — `from == until`.** `add_missing_agg_timestamps/1` guards on `min == max`, but it compares the *untruncated* values. A range narrower than one chart period (`12:00:10..12:00:40` at `chart_period: :minute`) passes that guard, then truncates both bounds into the same bucket. The existing `from == until` shift only covers `step_period == :seconds`.
2. **Inverted range — `from > until`.** `Helpers.get_min_max_filter_timestamps/2` returns `%{min: lvalue, max: rvalue}` verbatim for the `:range` operator. It is the only clause that skips `min_max_timestamps/1`, so it never sorts its bounds.

Both reproduce against the current dep tree:

```
Interval.new (truncation collapse): {:error, :invalid_until}
OLD CODE RAISES: protocol Enumerable not implemented for Tuple...
Interval.new (inverted range): {:error, :invalid_until}
```

## Fix

Move the interval build into a private `empty_range_aggs/3` and return `[]` when `Interval.new/1` reports an invalid range.

This matches the established intent of the `min == max` early return in `add_missing_agg_timestamps/1`: a zero-width or inverted range means "do not intersperse", not "crash". The chart renders whatever rows the query returned instead of raising.

The guard sits in `intersperse_missing_range_timestamps/4` — the function that raises — so every caller is protected. The in-function `use Timex` goes away with it; `Timex.Interval` is now referenced in full.

## Tests

New `describe "intersperse_missing_range_timestamps/4"` block covering the happy path, both degenerate inputs, the empty-aggregates case, and the existing `:second` special case.

```
mix test test/logflare/logs/search_operations_test.exs \
         test/logflare/logs/search_queries_test.exs test/logflare/lql
2 properties, 676 tests, 0 failures
```

`mix format --check-formatted` passes.

## Deliberately out of scope

- Generalizing the `from == until` shift to every chart period would render one empty bucket instead of none. That changes chart output, so it deserves its own decision.
- Sorting `min`/`max` in the `:range` clause of `Helpers` would change the query itself, not just the chart fill.

Happy to fold either in if you would rather fix the range at the source.


chasers:

PR #4007, opened 2026-09-17
fix: hold the scroll position when older events load above

chasers:
Follows #3986. Reported on staging: clicking "Load more" at the top moves the viewport, and the jump is obvious when a log line wraps to several lines.

## Cause

`#logs-list` is `flex-direction: column-reverse`, so the top button inserts older events at the visual top and pushes every row below them down.

Two things moved the scroll on that load, and neither knew about the other:

1. `updated/0` ran `restoreScrollAnchor/0`, holding the visible row in place.
2. `put_event_page/3` pushed `scroll-to-event`, jumping the oldest new row to the top.

Both ran inside `requestAnimationFrame`, so whichever fired second won. A following diff (the URL patch, the chart, the buttons) ran the anchor restore again. `updated/0` guarded the restore for `scroll-to-bottom` only, so `scroll-to-event` never got that discipline.

Both failure modes scale with the height of the inserted batch, which is why wrapped lines made it visible and single lines did not.

## Fix

- Drop `push_scroll_to_oldest/2` and the `scroll-to-event` push. Nothing else used that event, so the hook handler goes too. The anchor restore is now the only owner of the scroll on a page load.
- Stop falling back to `window.scrollTo(0, this.scrollPosition)`. That offset is stale once rows land above the viewport, and it was wrong by exactly the height of those rows.
- Keep an anchor only when the row carries an id.

## Tests

The "anchor row is gone" test fails against the old fallback and passes against the fix. I verified both directions.

- 226 Elixir tests pass across the 5 affected files
- 70 JS tests pass
- `format`, `lint.all`, `test.slop` (27/29) and `test.structure` pass

Please confirm on staging with a source that has wrapped log lines.


chasers:
## Reproduced on staging, and the fix measured at zero drift

Attached Playwright to a real Chrome over CDP against `test.logflarestaging.com`, source 58, `t:2026-09-17T09:{26..30}:00`, US/Arizona. 100 rows on screen, 100 added per click, tallest row 195px, median 20px.

### As deployed

```
driftPx 2408    anchorTop 459 -> 2867    scrollY 0 -> 225
```

Timeline:

```
t=1151  y=4     scrollBy [0, 2628.5]      <- anchor restore, correct
t=1152  y=2633  scrollIntoView log-...    <- scroll-to-event, wins
t=1177  y=225                             <- lands 2408px off
```

`restoreScrollAnchor/0` computes the right compensation and applies it. One millisecond later `scroll-to-event` runs `scrollIntoView({block: "start"})` on the oldest new row and overwrites it. That is the race this PR removes.

### With `scroll-to-event` suppressed, which is exactly what this PR does

```
driftPx 0    anchorTop 459 -> 459    scrollY 0 -> 2633
```

The row the reader was on does not move.

### Note for reviewers

This does not reproduce against a local dev source. There `document.getElementById(id)` returns null, so the `?.` swallows the call and the jump never happens. You need a source where the oldest row's dom id resolves, which staging has.

assets/js/source_lv_hooks.js
```diff
@@ -22,32 +22,32 @@ hooks.SourceSchemaModalTable = {
...
-        window.scrollBy(0, scrollDelta)
-      } else if (scrollPosition !== undefined) {
-        window.scrollTo(0, scrollPosition)
-      }
+      if (!anchorElement || !this.el.contains(anchorElement)) return
+
+      const scrollDelta = anchorElement.getBoundingClientRect().top - scrollAnchor.top
+
+      if (scrollDelta !== 0) window.scrollBy(0, scrollDelta)
     })
```
amokan: 🤖

> The PR body already notes a following diff, the URL patch, re-runs the anchor restore. Because `beforeUpdate` captures synchronously but the restore waits for a frame, that second capture can read the already-shifted top and its restore cancels the first. Running the restore inline in `updated()` is safe because the DOM is already patched at that point, and `getBoundingClientRect` forces layout.


```suggestion
    if (!scrollAnchor) return

    const anchorElement = document.getElementById(scrollAnchor.id)

    if (!anchorElement || !this.el.contains(anchorElement)) return

    const scrollDelta = anchorElement.getBoundingClientRect().top - scrollAnchor.top

    if (scrollDelta !== 0) window.scrollBy(0, scrollDelta)
```

assets/js/test/source_lv_hooks.test.js
```diff
@@ -93,3 +74,75 @@ describe("SourceLogsSearchList", () => {
...
+
+    return hook;
+  };
+
+  const stubRect = (element, top) => {
+    element.getBoundingClientRect = () => ({ top, bottom: top + 20 });
+  };
+
+  beforeEach(() => {
+    vi.stubGlobal("requestAnimationFrame", (callback) => callback());
```
amokan:
```suggestion
```

_This would be dead if you take the suggested change in `source_lv_hooks.js`_

assets/js/test/source_lv_hooks.test.js
```diff
@@ -93,3 +74,75 @@ describe("SourceLogsSearchList", () => {
...
+
+    stubRect(document.getElementById("log-events-a-1"), 100);
+    hook.captureScrollAnchor();
+
+    document.getElementById("logs-list").innerHTML = "";
+    hook.restoreScrollAnchor();
+
+    expect(scrollBy).not.toHaveBeenCalled();
+    expect(scrollTo).not.toHaveBeenCalled();
+  });
```
amokan, line 146: Test to cover the changes in the `source_lv_hooks.js`

🤖 

> _queues rAF instead of running it inline and asserts scrollBy fired before any frame ran_

```suggestion
  });

  it("restores the anchor before the next frame so a following diff sees the corrected position", () => {
    const hook = mountWithLogList();
    const anchor = document.getElementById("log-events-a-1");
    const rafQueue = [];
    vi.stubGlobal("requestAnimationFrame", (callback) => rafQueue.push(callback));

    stubRect(anchor, 100);
    hook.captureScrollAnchor();

    stubRect(anchor, 420);
    hook.restoreScrollAnchor();

    expect(scrollBy).toHaveBeenCalledWith(0, 320);
    expect(rafQueue).toHaveLength(0);
  });
```

test/logflare_web/live/search_live/logs_search_lv_test.exs
```diff
@@ -2437,8 +2437,7 @@ defmodule LogflareWeb.Source.SearchLVTest do
       expected_ids = events |> Enum.take(102) |> log_event_dom_ids()
       assert visible_log_event_ids(view) == expected_ids
 
-      [oldest_loaded_id | _] = expected_ids
-      assert_push_event(view, "scroll-to-event", %{id: ^oldest_loaded_id})
+      refute_push_event(view, "scroll-to-event", %{})
```
amokan: 🤖 

> Nothing can emit `scroll-to-event` any more, so the assertion can never fail.

```suggestion
```

amokan (changes requested):

assets/js/source_lv_hooks.js
chasers: Taken, in 09a7d8a. The reasoning holds: morphdom finishes before `updated/0` runs and `getBoundingClientRect` forces layout, so the inline read sees the patched DOM. Removing the frame also closes the window where a following diff captured an already shifted top.

Worth noting the staging trace showed the restore computing the right delta even from inside the frame (`scrollBy [0, 2628.5]`). What beat it was `scroll-to-event` one millisecond later, which this PR removes. So the inline change is a correctness hardening rather than the fix itself.

assets/js/test/source_lv_hooks.test.js
chasers: Dropped in 09a7d8a. Right, the stub is dead once the restore is inline.

assets/js/test/source_lv_hooks.test.js
chasers: Added in 09a7d8a, verbatim. `expect(rafQueue).toHaveLength(0)` is the part worth having, since it pins the ordering rather than just the outcome.

test/logflare_web/live/search_live/logs_search_lv_test.exs
chasers: Removed in 09a7d8a.

One thing to flag rather than quietly accept: it was not strictly unfailable. It caught a reintroduction of `push_scroll_to_oldest/2`, which is exactly the regression this PR guards against. I am fine removing it because the e2e test in this PR now covers the behaviour properly, and I have confirmed that test actually runs rather than passing vacuously: the feature suite went from 20 tests to 21 between 6426ee5 and 443b26e, 0 failures.

If the e2e test ever gets skipped or removed, this PR loses its only regression guard.

chasers: Reopening this one: we have put the assertion back in 5580242.

The call is @chasers'. The reasoning:

`refute_push_event(view, "scroll-to-event", %{})` is not unfailable. It fails the moment `push_scroll_to_oldest/2` comes back, and that is precisely the regression this branch exists to prevent. It is a regression guard in the ordinary sense: green against the current tree, red against the change we do not want.

The e2e test covers the same ground more convincingly, through a real browser, and I confirmed it runs rather than passing vacuously (the feature suite went 20 -> 21 between 6426ee5 and 443b26e). But it is a single feature test. If it is ever skipped, retagged or removed, dropping this line leaves the branch with no guard at the unit level.

Cost of keeping it is one line and no runtime. Happy to revisit if you still disagree.

amokan (approved):

PR #4008, opened 2026-09-18
chore: remove the deprecated ClickHouse `read_only_url` backend config option

amokan:
## Overview

The singular `read_only_url` config field predates labeled read clusters (#3693) and has been marked deprecated since that PR landed. Multi-read cluster routing via `read_only_urls` and `default_read_cluster` has been out on prod for weeks, so the old field and its fallback logic can go.

This PR removes the option from the ClickHouse adaptor, the connection manager's host resolution, the backend form, and the OpenAPI schema.

### Key Changes

- `ClickHouseAdaptor` no longer casts, validates, or display-allowlists `read_only_url`. The read-grant check falls straight through to the primary `url` when no labeled clusters exist and a dedicated query user is set
- `ConnectionManager.read_url/2` resolves labeled cluster, then `default_read_cluster`, then primary `url`. The `legacy_read_url/1` step between the last two is deleted
- The unlabeled pool and its `"(unlabeled)"` telemetry tag are unchanged. That pool still serves the primary URL, so the reserved-label validation on `read_only_urls` stays
- The "Read-Only URL (Optional)" input is removed from the backend form. `ReadClusterUrlsComponent` is now the only read cluster UI
- `ClickhouseConfigSchema` drops the property
- Tests: removed the singular cast tests and a fallback describe that only exercised a copied helper. Test-connection and dedicated-query-user tests now configure `read_only_urls`, and "legacy" wording is renamed to "unlabeled"

### Risks / Considerations

- **No data migration for configs that still store `read_only_url`.** Every backend load path, including API create/update responses, re-casts `config_encrypted` through `cast_config/1`, so a stale key is dropped in memory and never reaches the JSON encoder. _The key lingers in the encrypted blob until the backend is next saved, which is harmless_
- **Behaviour change for backends that set `read_only_url` without `read_only_urls`.** Reads move from the old read-only endpoint to the primary `url`. Prod/staging backends were migrated to labeled clusters before this, but any self-hosted install still relying on the singular field needs to move to `read_only_urls` before upgrading
- **OpenAPI schema now documents no read cluster option.** `read_only_urls` and `default_read_cluster` were never added to `ClickhouseConfigSchema`; tracked separately as a follow-up


amokan:
@claude review this

djwhitt (approved):
:robot:

> The OpenAPI risk note is out of date. This diff adds `read_only_urls` and `default_read_cluster` to `ClickhouseConfigSchema`; please update the PR description.

PR #4010, opened 2026-09-18
feat: emit configured pool size alongside the ClickHouse read pool status poll

amokan:
## Overview

The read pool status poll (#4002) emits `ready_conn_count` and `checkout_queue_length`, but nothing emits the pool's size. Without it a dashboard cannot show utilization as a share of the pool, and `ready_conn_count = 0` reads the same on a 32-connection labeled pool and a 50-connection default pool without going back to the backend config.

This PR adds a `pool_size` gauge to the same `pool_status` event. The pool already knows its size when it starts, so it carries that value in its `Registry` entry and the poller reads it back alongside the `DBConnection` counts. No backend cache or database lookup happens in the sweep.

### Key Changes

- `ConnectionManager` registers each read pool under a via tuple whose `Registry` value is `%{pool_size: n}`, using the size `read_pool_size/2` resolved at pool start. `connection_pool_via/2` is unchanged, so every existing lookup and query path still resolves through the two-element via
- `QueryConnectionSup.list_read_pools/0` returns `{backend_id, label, pool_pid, pool_size}`, reading the size from the registry value in the same `Registry.select/2` match spec. This is the running pool's size, not whatever the backend config says now
- `Telemetry.clickhouse_read_pool_metrics/2` emits `pool_size` in the `pool_status` event with `ready_conn_count` and `checkout_queue_length`, so the three measurements share a timestamp. The poll task still wraps only `DBConnection.get_connection_metrics/1`
- `logflare.clickhouse.read_pool.pool_size` declared as a `last_value` tagged `backend_id` and `read_cluster`, joining the other two gauges on the same tags. Pool size is static for the life of a pool, so it is a gauge and not emitted per checkout
- Tests cover the gauge definition, `pool_size` present in every per-pool sample, `read_pool_size: 7` / `labeled_read_pool_size: 4` listing and emitting as 7 for the unlabeled pool and 4 for `"api"` against real pools, and the existing dead-pid, timeout, and unexpected-result paths with the new tuple shape

### Risks / Considerations

- **`(pool_size - ready_conn_count) / pool_size` is an upper bound on utilization, not utilization.** `ready_conn_count` reads 0 for connections still establishing or reconnecting as well as for checked-out ones, so the ratio also climbs during connectivity loss. Read it with `connected`/`disconnected`, as the `pool_size` description states
- **Started size, not current config.** A config change shows up in `pool_size` when the pool restarts, which `refresh_backend/1` does cluster-wide on backend update. Until then the gauge reports the size the pool actually has, so the utilization math cannot go negative against a config the pool has not picked up
- **`list_read_pools/0` changes shape.** Callers that destructure the tuple need the fourth element; `Telemetry` is the only caller today
- **Per node, not per cluster.** Same as the other two gauges: panel by instance, or take a max across nodes


chasers (commented):
**One bug:** [a Postgres error crashes the poller](https://github.com/Logflare/logflare/pull/4010#discussion_r4049955993). Four smaller points follow.

I traced the crash path in the Cachex and `telemetry_poller` source. I did not reproduce it.

Start with [Report the size the pool started with](https://github.com/Logflare/logflare/pull/4010#discussion_r4049956023). That change also removes three other problems:

- [the crash](https://github.com/Logflare/logflare/pull/4010#discussion_r4049955993)
- [the wide catch](https://github.com/Logflare/logflare/pull/4010#discussion_r4049956009)
- [the dropped samples](https://github.com/Logflare/logflare/pull/4010#discussion_r4049956017)

lib/telemetry.ex
```diff
@@ -778,24 +790,44 @@ defmodule Logflare.Telemetry do
...
+    {pool, DBConnection.get_connection_metrics(pool_pid),
+     configured_read_pool_size(backend_id, label)}
   catch
-    :exit, _reason -> {pool, :gone}
+    :exit, _reason -> {pool, :gone, nil}
+  end
+
+  @spec configured_read_pool_size(pos_integer(), String.t() | nil) :: pos_integer() | nil
+  defp configured_read_pool_size(backend_id, label) do
+    case Backends.Cache.get_backend(backend_id) do
```
chasers: **Bug: a Postgres error here crashes the poller.**

`poll_read_pool/1` catches only `:exit`. A raise goes through.

1. The cache entry expires.
2. The next poll reads the row from the replica.
3. The replica is down. Cachex returns `{:error, %Cachex.Error{}}`.
4. `ContextCache.fetch/3` has no clause for that shape. It raises `CaseClauseError`.
5. The task is linked, so the poller dies.

Result: all pools in the sweep lose the sample. No `poll_failure` is counted.

lib/telemetry.ex
```diff
@@ -778,24 +790,44 @@ defmodule Logflare.Telemetry do
...
-          {tuple(), [DBConnection.Pool.connection_metrics()] | :gone}
-  defp poll_read_pool({_backend_id, _label, pool_pid} = pool) do
-    {pool, DBConnection.get_connection_metrics(pool_pid)}
+          {tuple(), [DBConnection.Pool.connection_metrics()] | :gone, pos_integer() | nil}
+  defp poll_read_pool({backend_id, label, pool_pid} = pool) do
+    {pool, DBConnection.get_connection_metrics(pool_pid),
+     configured_read_pool_size(backend_id, label)}
   catch
-    :exit, _reason -> {pool, :gone}
+    :exit, _reason -> {pool, :gone, nil}
```
chasers: **The catch is too wide.**

It now covers the cache lookup too. An exit there reports the pool as `:gone`. No warning is logged. No `poll_failure` is counted.

Fix: wrap only `DBConnection.get_connection_metrics/1`.

lib/telemetry.ex
```diff
@@ -778,24 +790,44 @@ defmodule Logflare.Telemetry do
...
+      %Backend{config: config} -> ConnectionManager.read_pool_size(config, label)
+      nil -> nil
+    end
   end
 
   @spec emit_read_pool_status(term()) :: :ok
-  defp emit_read_pool_status({:ok, {_pool, :gone}}), do: :ok
+  defp emit_read_pool_status({:ok, {_pool, :gone, _pool_size}}), do: :ok
+
+  defp emit_read_pool_status({:ok, {{backend_id, label, _pool_pid}, _metrics, nil}}) do
```
chasers: **This clause has two problems.**

1. It drops good samples. When the lookup returns nil, the pool is still live. A nil can stay in the cache for 20 minutes. Emit `ready_conn_count` and `checkout_queue_length`. Omit only `pool_size`.
2. It matches before the metrics-shape clauses. A bad metrics result gets the tag `:backend_not_found`. The `inspect(result)` log never runs. Check the metrics shape first.

lib/telemetry.ex
```diff
@@ -778,24 +790,44 @@ defmodule Logflare.Telemetry do
...
+  defp poll_read_pool({backend_id, label, pool_pid} = pool) do
+    {pool, DBConnection.get_connection_metrics(pool_pid),
+     configured_read_pool_size(backend_id, label)}
   catch
-    :exit, _reason -> {pool, :gone}
+    :exit, _reason -> {pool, :gone, nil}
+  end
+
+  @spec configured_read_pool_size(pos_integer(), String.t() | nil) :: pos_integer() | nil
+  defp configured_read_pool_size(backend_id, label) do
```
chasers: **Report the size the pool started with.**

The pool reads config only at start. Cached config can change later with no restart. The gauge then shows a size the pool does not have. The utilization math can go negative.

Fix:

1. Store the started size in `ConnectionManager` state or the Registry value.
2. Return it from `QueryConnectionSup.list_read_pools/0`.

This removes the cache and the database from the poller. Three other problems go away:

- [the crash](https://github.com/Logflare/logflare/pull/4010#discussion_r4049955993)
- [the wide catch](https://github.com/Logflare/logflare/pull/4010#discussion_r4049956009)
- [the dropped samples](https://github.com/Logflare/logflare/pull/4010#discussion_r4049956017)

test/logflare/telemetry_test.exs
```diff
@@ -559,14 +561,66 @@ defmodule Logflare.TelemetryTest do
...
+                      %{backend_id: ^backend_id, read_cluster: "api"}}
+    end
+
+    test "counts a pool whose backend no longer exists as a poll failure", %{backend: backend} do
+      {:ok, _} =
+        QueryConnectionSup.start_connection_manager(ConnectionManager.child_spec(backend))
+
+      assert :ok == ConnectionManager.ensure_pool_started(backend)
+      live_pool = ConnectionManager.get_pool_pid(backend)
+      missing_backend_id = backend.id + 100_000
```
chasers: **This test leaks cache state.**

It writes a nil entry for `backend.id + 100_000` into the shared `Backends.Cache`. The sandbox rollback does not clear Cachex. A later test with that id reads nil.

Fix: stub `Backends.Cache.get_backend/1` with Mimic.

chasers (approved):
Approving to unblock

PR #4013, opened 2026-09-18
feat: webhook v2 backend with consolidated ingest

chasers:
Adds the `webhook_v2` backend type. It sends the same HTTP POST batches as `webhook`, but it uses the consolidated ingest path of the ClickHouse backend.

**Stack:** PR 2 of 5. The `webhook` backend does not change.

## Why a new type

Datadog, Elastic, Loki, Axiom, Sentry, and Incident.io use the `webhook` pipeline. A new type keeps them safe. It needs no migration, because `backends.type` is a string.

## v1 compared with v2

| Concern | `webhook` | `webhook_v2` |
| --- | --- | --- |
| Pipelines | One per source per backend | One per backend per node |
| Producer | Full `LogEvent` structs | Pointers, with a cap on in-flight events |
| JSON encode | Whole batch | One event at a time |
| Bad event | Drops the batch | Fails one event |
| Batch size | 250 | `batch_size`, default 1 000, maximum 10 000 |

## How a request is built

1. All sources write to the queue `{:consolidated, backend_id}`.
2. A processor encodes one event body to JSON.
3. The batcher joins the bodies into a JSON array or NDJSON.
4. `WebhookAdaptor.Client` sends the request. It applies gzip and the SSRF protection.

One request can hold events from more than one source. The payload has no source field, the same as v1.

## Also in this PR

A failed request drops its events with `retry_dropped` telemetry. **Retries come in PR 3.**

The backend type label fix for the `not_initialized` telemetry moved to #4018.

## Tests

- New: `webhook_v2_adaptor_test.exs` and `webhook_v2_adaptor/pipeline_test.exs`.
- `mix ci` passes.


chasers:

chasers:
Fixed in `3fbe015`.

- `handle_message/3` now has the `%EncodedEvent{}` clause and a catch-all function clause, mirroring `ClickHouseAdaptor.Pipeline`. Two tests cover both.
- Alias order corrected. It was a rename artifact: the alias used to start with `WebhookV2` and sorted correctly before the rename.

One correction on the credo point: `mix credo --strict` on that file reports no issues and `MIX_ENV=test mix ci` exits 0 both before and after, so this was a convention fix rather than a failing gate.

chasers:
Set the processor count to match the ClickHouse consolidated pipeline in `61d96f0`: `max(schedulers - batcher_concurrency, 6)`, replacing the fixed 2.

Measured with the throughput benchmark from #4027 on 10 schedulers, 96 000 events, 20 ms sink latency:

| sources | before (2 processors) | after (6 processors) |
| --- | --- | --- |
| 1 | 87 671 ev/s | **136 142 ev/s** |
| 4 | 84 553 ev/s | **134 409 ev/s** |
| 16 | 83 257 ev/s | **132 450 ev/s** |

That is ~1.6x across the board — the 2 processors were the limit of the whole pipeline, not the batchers or the queue.

Worth noting the cost, since it is allocated per backend rather than per source: on a 64-scheduler host this is 60 processors + 4 batchers for each active `consolidated_webhook` backend. That is the same bargain the ClickHouse adaptor already makes, and its own comment flags a global cap as the thing to revisit before enabling these broadly.

PR #4014, opened 2026-09-18
feat: webhook v2 retries and circuit breaker

chasers:
Adds one retry for a failed `webhook_v2` batch. The shared circuit breaker limits the retries.

**Stack:** PR 3 of 5. This PR has the requeue logic, which is the part with the most risk.

## Retry rules

| Result | Action |
| --- | --- |
| 2xx | Success |
| 408, 429, 5xx, transport error | Requeue one time |
| Other 4xx | Drop the batch |
| Body does not encode to JSON | Drop one event |

## How it works

- A retry sends the stored `EncodedEvent`. The body is not encoded a second time.
- Each transient failure counts toward `Logflare.Backends.CircuitBreaker` (from PR 1).
- While the breaker is open, the pipeline drops retries. **It still sends new batches.**

## Drop reasons in telemetry

`:retries_exhausted`, `:circuit_breaker_open`, `:queue_unavailable`, `:rejected`.

The reason `:request_failed` from PR 2 is removed.

## Events lost in the queue

An event can wait until the `GenerationJanitor` drops its generation. The processor then finds no event for the pointer.

- Before: the pipeline skipped these pointers without a signal.
- Now: it counts them in `[:logflare, :ingest_event_queue, :missing_ids]`, tagged `backend_type: :webhook_v2`.
- `missing_ids` above 0 means the backlog is older than about 4 minutes. The receiver is too slow for the ingest rate.

## Tests

- Requeue with the same bytes, retries exhausted, breaker open, breaker start.
- Missing events: one `missing_ids` event with the count, and no `retry_dropped` event.
- `mix ci` passes.


chasers:

PR #4016, opened 2026-09-18
feat: webhook v2 sample percentage

chasers:
Adds the `sample_percentage` config field to `webhook_v2`. The backend sends only that percentage of events and drops the rest.

**Stack:** PR 5 of 5.

## Config

- A number greater than `0`, up to `100`. Decimal values such as `12.5` are valid.
- The default is `100`, which sends all events.

## How it works

- The adaptor implements `pre_ingest/3`. A dropped event never enters the queue.
- The sample applies to this backend only. The other backends of the source get every event.
- **The decision is random and independent for each event.** It does not keep the spans of one trace together.

## Telemetry

- Event: `[:logflare, :logs, :ingest_logs, :drop_sampled]` with the drop count.
- Metric: `logflare.logs.ingest_logs.drop_sampled`, tagged `backend_id` and `backend_type`.
- No log line, because a drop is the expected behaviour.

## Also in this PR

The backend form has a **Sample percentage** input. The docs have a "Sampling" section.

## Tests

- Config: default, custom value, out of range.
- `pre_ingest/3`: 100 percent, no field, 50 percent of 10 000 events, empty batch.
- Ingestion: the queue stays empty and the telemetry reports all drops.
- `mix ci` passes.


chasers:

PR #4018, opened 2026-09-18
fix: limit the size of the consolidated startup queue

chasers:
Puts a size limit on the startup queue of a consolidated backend. Today that queue can grow without a limit when the destination is slow.

**Not part of the webhook v2 stack.** It protects ClickHouse today.

## The problem

1. Each producer queue of a consolidated backend holds up to 120 000 events.
2. When all of them are full, `add_to_table/3` sends new events to the startup queue.
3. The startup queue had **no size check**. No `QueueJanitor` runs for a consolidated backend.
4. Only the `GenerationJanitor` limited the backlog, by age (4 to 8 minutes).

Result: a slow destination could hold millions of events in memory.

## The fix

- A new batch drops when the startup queue is at 120 000 events.
- `check_queue_size: false` still skips the check.
- New events drop. Events already in the queue are still sent.

## Why drop, and not HTTP 429

- A 429 must come **before** ingest (`BufferLimiter` does this). At insert time, other backends already have the events. A client retry then makes duplicates.
- A backend that is not a default ingest backend must not stop ingest for the whole source.

## Telemetry

- New event `[:logflare, :ingest_event_queue, :queue_full, :dropped]` with `count`, `backend_type`, `backend_id`.
- New metric `logflare.ingest_event_queue.queue_full.dropped.count`, tagged `backend_type`.
- Fix: the `not_initialized` drop telemetry reported `:clickhouse` for every consolidated backend. It now reports the real type.

## Tests

- At the limit: the batch drops, the queue size stays, the telemetry has the count.
- One below the limit: the batch is kept.
- `check_queue_size: false`: the batch is kept.
- `mix ci` passes.


chasers:

PR #4019, opened 2026-09-18
fix: push back when a consolidated default ingest backend is full

chasers:
Restores HTTP 429 push back for a ClickHouse default ingest backend. It stopped when ClickHouse moved to consolidated ingest (#3089).

## The bug

| Step | What happens |
| --- | --- |
| 1 | `BufferCacheWorker` stores the queue sizes of a consolidated backend under `{:consolidated, backend_id}`. |
| 2 | The 429 check reads `{source_id, backend_id}`. |
| 3 | No code writes that key for a consolidated backend. The check reads 0 events. |
| 4 | `cached_local_pending_buffer_full?/1` returns `false`. The client never gets a 429. |

**Who it affects:** only sources that use ClickHouse as a default ingest backend (both default ingest flags on). BigQuery, Postgres, S3, and Syslog are not affected.

## The fix

- For a default ingest backend with consolidated ingest, the check reads `{:consolidated, backend_id}`.
- The backend is **full** when every producer queue is at 120 000 events.
- The check ignores the startup queue. The 429 starts when the overflow starts.

## What does not change

- A ClickHouse backend **without** the default ingest flag does not push back.
- Sources without `default_ingest_backend_enabled?` check the system default queue only.

## One caveat

All sources of a consolidated backend share its queues. When one busy source fills them, every source with that default ingest backend gets a 429.

## Tests

- The ClickHouse test filled a `{source_id, backend_id}` queue that production never creates. It now uses real consolidated queues. **It fails without this fix.**
- New cases: one producer queue has room, only the startup queue has events, the default ingest flag is off.
- `mix ci` passes.

## Related

#4018 limits the size of the startup queue. With both PRs: 429 first, then a counted drop if the client does not slow down.


chasers:

PR #4020, opened 2026-09-18
feat: restrict signups to a list of email domains

chasers:
Adds an environment variable that restricts new accounts to a list of email domains.

```
LOGFLARE_SIGNUP_ALLOWED_DOMAINS=supabase.com,supabase.io
```

**Not set or empty: all domains are allowed.** Nothing changes for a deployment that does not set it.

## What is checked

Only a **new** identity:

| Path | Code |
| --- | --- |
| New user signs up (OAuth or email link) | `Users.insert_or_update_user/1` |
| New team member accepts an invite | `TeamUsers.insert_or_update_team_user/2` |
| Team member creates an own account | `Users.create_user/1`, `AuthController.create_and_sign_in/2` |

An existing user or team member can still sign in from any domain.

## Why invites are checked too

Without that check, any user can invite an outside address. The restriction then has a hole.

## What is not checked

- The partner API (`Partners.create_user/2`).
- The single-tenant seed user.

These are not signups by a person. `Users.insert_user/1` stays unrestricted for them.

## Match rules

- The domain is the text after the last `@`.
- Exact and case-insensitive. `mail.supabase.com` does **not** match `supabase.com`.
- The list accepts spaces, upper case, and a leading `@`.

## What the person sees

A rejected signup redirects to the login page with the message "New accounts are restricted to approved email domains." It writes no error log.

The email link is still sent to any address. The check runs at sign-in. This keeps the change small and does not reveal which accounts exist.

## Tests

- `SignupDomains`: parse, default, match rules (11 tests).
- `Users`, `TeamUsers`, `AuthController`, dashboard: allowed, rejected, and existing account for each path.
- Checked with the real `config/runtime.exs`: unset, empty, and blank values allow all domains.
- `mix ci` passes.


chasers:
@claude review this one

chasers:

PR #4021, opened 2026-09-19
fix: stop the ClickHouse endpoint round-trip rewriting FINAL as a table alias

amokan:
## Overview

Every ClickHouse endpoint query is parsed to an AST and re-serialized. The pinned [`sqlparser`](https://github.com/apache/datafusion-sqlparser-rs/) 0.54.0 had no grammar for `FROM t FINAL`, so it parsed `FINAL` as a table alias and emitted `FROM t AS FINAL`. ClickHouse accepts that and skips deduplication, returning wrong rows with no error. `ARRAY JOIN` broke the same way but failed loudly at execution.

This PR bumps `sqlparser` 0.54.0 → 0.63.0 one minor version per commit. Each commit compiles and passes the SQL suites on its own, so a regression bisects to a single release. From 0.60.0 the serializer stops adding `AS` to implicit aliases, and 0.62.0 adds real `ARRAY JOIN` grammar.

### Key Changes

- `sqlparser` 0.54.0 → 0.63.0 over nine commits, one per minor version.
- `AstUtils` gains builders for the node shapes the crate now requires (_object-name parts, spanned values, casts_) and a single `extract_parameters/1` replacing two private copies
- The NIF `to_string` returns `{:error, reason}` on a malformed AST instead of panicking
- Round-trip tests for `FINAL` and `ARRAY JOIN` at the parser, `apply_limit`, and `Sql.transform` levels

### Risks / Considerations

- **`FINAL` is still an alias in the AST.** The crate has no `FINAL` grammar; the fix is in the serializer. `FROM t AS x FINAL` remains a parse error, and a sandboxed source referenced as `FROM t FINAL` expands to `(subquery) FINAL`, which ClickHouse rejects loudly
- **BigQuery `@param` now parses as an identifier** (0.55.0). Parameter extraction and the bq→pg translation handle both shapes; endpoint behavior is unchanged
- **Two observable changes.** A `DELETE` inside a CTE is now rejected by the SELECT-only check rather than the tokenizer, and `Parser.to_string/1` can return an error tuple where it previously raised
- **bq→pg drops every `CROSS JOIN`, not only `CROSS JOIN UNNEST`.** Pre-existing; 0.59.0 would have changed it silently. Preserved and pinned with a test. May be worth a follow-up
- **Stale nested `Cargo.lock` removed.** The workspace root lockfile governs the build; CI cache keys only change hash


amokan:
@claude review this

djwhitt (approved):
:robot:

> Please add a test for a quoted BigQuery identifier such as `` `@field` ``. Parameter extraction currently matches every identifier beginning with `@`, without checking `quote_style`; a quoted column name should not become a required parameter.

